PR #88, opened 2019-09-06
qtt agda cleanup

andy-morris:
The only notable part of this is that `Π π S T` is spelt `Π[ π / S ] T` now. Hopefully this will make it a little clearer which numbers are usages and which are bound variables, and also (spoiler!) be more consistent with `[ π / S ]⊗ T` etc when the linear logic stuff is mergeable.

(I made a separate PR for just this because the usage polymorphism also branches from here and I wanted to make it easier to merge both in when they're ready)


andy-morris:
(oops I forgot to rebase first, sorry about that!)

andy-morris:
never mind this, i'll just rebase the LL connectives over the polymorphism stuff locally

PR #92, opened 2019-09-09
Thealmarty/core type checker wip

thealmarty:


src/Juvix/Core/MainLang.hs
```diff
@@ -266,8 +270,22 @@ iType ii g (Free x) =
   case lookup x g of
     Just ann -> return $ snd ann
     Nothing  -> throwError (iTypeErrorMsg ii x)
+--Prim-Const typing rule
 iType _ii _g (Nat _) = return VNats
--- Typing rule not in lang ref atm?
+--App rule, function M applies to N
+iType ii g (App m n) = do
```
cwgoes: either this function should take sigma as an argument, or return it

cwgoes: concluded: it should return sigma

primitive types should have usage w

cwgoes:
@thealmarty Let me know when I should review and/or merge this.

thealmarty:
Reverted  "Add string name to bound variables." commit to separate it out to another branch.

PR #94, opened 2019-09-10
Mariari/unhardcode

mariari:
allows for custom functions, functions like +, -, *, etc.  are not yet removed from being special cased, however once the parser is augmented to accept a parsing order this will be removed. This will be in another MR or this one if it's still up

### DIFF
--- a/src/Juvix/Bohm/Translation.hs
+++ b/src/Juvix/Bohm/Translation.hs
@@ -214,6 +214,7 @@ netToAst net = evalEnvState run (Env 0 net mempty)
                                       rec' aux1 cameFrom newFanMap nodeVarMap
                                     -- Never happens by precondition!
                                     FreeNode → pure Nothing
+                                -- TODO :: unify with through function 19 lines below
                                 fromCircleOrStar con =
                                   let newFanMap = Map.insert i [In con] fanMap
                                       cameFrom = Just (n, Prim)
--- a/src/Juvix/Bohm/Translation.hs
+++ b/src/Juvix/Bohm/Translation.hs
@@ -197,15 +197,16 @@ netToAst net = evalEnvState run (Env 0 net mempty)
               Just port →
                 case port of
                   B.IsAux3 {B._tag3 = tag, B._prim = prim, B._aux2 = aux2, B._aux3 = aux3} →
-                    case tag of
-                      B.IfElse → do
-                        case (prim, aux2, aux3) of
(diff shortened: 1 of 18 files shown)


src/Juvix/Bohm/Translation.hs
```diff
@@ -70,11 +74,53 @@ astToNet bohm = net'
         (Just portInfo, _) → chaseAndCreateFan portInfo
         -- The symbol is Free, but used already, create a sharing node for it
         (Nothing, Just portInfo) → chaseAndCreateFan portInfo
-        -- The symbol is Free, just stash it in a symbol with no rewrite rules
-        (Nothing, Nothing) → do
-          nodeInfo ← (,) <$> (newNode (B.Primar $ B.Symbol s)) <*> pure Prim
-          put @"free" (Map.insert s nodeInfo frees)
-          pure nodeInfo
+        -- Since we now take an environment, we now have to check if this
+        -- unkown symbol is a function or a symbol or what?
```
cwgoes:
```suggestion
        -- unknown symbol is a function or a symbol or what?
```

src/Juvix/Bohm/Type.hs
```diff
@@ -23,9 +24,30 @@ data Bohm
   -- Not valid syntax but for read back of a graph
   | Erase
   -- Not valid syntax but for read back of a graph
-  | Curried (Int → Int) Bohm
-  | CurriedB (Int → Bool) Bohm
+  | Curried3 (Primitive → Primitive → Primitive → Maybe Primitive) Bohm Bohm Bohm
```
cwgoes: Why `Maybe`? Should be a compiler bug if the types mismatch, right?

src/Juvix/Bohm/Type.hs
```diff
@@ -23,9 +24,30 @@ data Bohm
...
+        | Neq  | Lt
+        | Gt   | Ge
+        | Le
+        deriving Show
+
+
+-- | Constructs a Function from a primitive
+-- the final argument is maybe, as if the nodes don't line up
+-- a final type can't be constructed. This is untyped
+-- so type check at a higher level
```
cwgoes: Hmm OK for now

src/Juvix/Bohm/Type.hs
mariari: it would be a compiler bug yes, but how I implement the functions currently is pattern matches on primitive to apply itself, it could throw an error on a mismatch however.

PR #95, opened 2019-09-11
Fixed compile errors and added sources

mariari:
We need to figure out how to get code to properly embed in the PDF, once that is the case, we can add back λ as a replacement for \\.


doc/reference/src/language-reference.bib
```diff
@@ -110,3 +110,21 @@ @misc{plenty-o-nuttin
...
+  chapter={2.4},
+  howpublished={\url{https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node28.html#SECTION00640000000000000000}}
+}
+@misc{cedille,
+  title={Efficient Mendler-Style Lambda-Encodings in Cedille},
+  author={Denis Firsov, Richard Blair, and Aaron Stump},
+  howpublished={\url{https://arxiv.org/abs/1803.02473v1}}
+}
+
+@misc{Scott,
```
cwgoes:
```suggestion
@misc{scott,
```

Let's use uniform lowercase for reference identifiers.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
-- alt-term encoding: $lam x. x + x + x$ ~> $3 * x$ rewrite rule instead of multi-node graph
+- alt-term encoding: $\lambda x. x + x + x$ ~> $3 * x$ rewrite rule instead of multi-node graph
 
 - Over set of underlying opcodes & von-Neumann-ish VM (e.g. x86, LLVM, EVM).
 - Use linear types (e.g. rewrite memory locations in place)
 - Must be semantically equivalent to interaction system encoding after read-back
 - Must preserve any-order reduction (exists some maximum `n` rewrite steps after which graph read-back must have final form regardless of order)
 - Can relax diamond property (strong confluence)? unclear if necessary
 
+### Argument for Correct read-back
```
cwgoes:
```suggestion
### Argument for correctness of read-back
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+
+We wish to show that a read-back of an EAL-expression that has undergone zero or more reductions gives back the same result as normal evaluation up to alpha equality.
```
cwgoes: OK, so this is not only arguing for the correctness of read-back but also the correctness of reduction?

Read-back could be argued separately, e.g. "read-back (encode (term)) = term", but this is fine too.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+
+Before we can prove this result, we must first prove a few lemmas and theorems first as well talk about what fan in's with the same label imply.
+
+The first lemma we wish to prove is the following
+
+\begin{Lemma}{AST→Net has one free port}{AST->Net}
+  Let $A$ = be a valid BOHM ADT.
```
cwgoes:
```suggestion
  Let $A$ be a valid BOHM term.
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  Now Consider the net over $A$, let $N_a$ be this net.
```
cwgoes: What is the "net over A" - do you mean the encoding of A?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  Now let us consider any sub case of $A$, $L = P(n_1 \star \cdots \star n_k)$.
```
cwgoes: What is a sub-case of A - a sub-term?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  The node corresponding to $L$ looks.
+
+  \begin{tikzpicture}
+    \draw (0,0) circle (.8cm);
```
cwgoes: This is a circle. I don't follow.

The sub-graph corresponding to the encoding of a subterm should be a graph with some number of free ports.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \end{tikzpicture}
+
+  Where the labels in $N_a$ are the same as in $L$, with $n_{k+1}$ being an extra port which connects to the ADT above it.
+  \\ \\
+  For an ADT to be realized, we must have all arguments, so all these ports must be connected.
```
cwgoes: Must we? That's not a rule for term (or AST, do you mean AST instead of ADT?) validity in general.

If we are only talking about closed terms, that may be OK, but we should say so explicitly.

I'm not convinced that we do have to reason about only closed terms, I think we could also reason about open ones.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  Now, if $L$ is $A$, then we have one free port, however if it is not, then this free port must not be free but instead be connected to the ADT above it.
```
cwgoes: What does "above" mean?

I think you're getting at "must be connected to the encoding of an outer term".

I agree with this but I do not think it is trivial, we should argue each case of the encode function.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  For free variables which have no $\lambda$ to connect to, a symbol node is created, maintaining the invariant.
```
cwgoes: This seems to conflict with "must have all arguments" above - or are you talking about free variables in a subterm only?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+      \begin{subfigure}[c]{0.1\textwidth}
+        \begin{tikzpicture}
+          \wirecross[]{a}{c}{b}{d}
+        \end{tikzpicture}
+      \end{subfigure}
+    \end{figure}
+
+    We can see that there is only one wire eliminated, the one connecting the main ports of node $f$ and node $g$.
+    \\ \\
+    The rest of the wires are simply rewired to each other, thus not effecting the total number of free ports
```
cwgoes:
```suggestion
    The rest of the wires are simply rewired to each other, thus not altering the total number of free ports.
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+      \end{subfigure}
+      ~
+      \begin{subfigure}[c]{0.1\textwidth}
+        \begin{tikzpicture}
+          \wirecross[]{a}{c}{b}{d}
+        \end{tikzpicture}
+      \end{subfigure}
+    \end{figure}
+
+    We can see that there is only one wire eliminated, the one connecting the main ports of node $f$ and node $g$.
```
cwgoes:
```suggestion
    The wire connecting the main ports of node $f$ and node $g$ is eliminated.
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+
+
+\begin{Theorem}{AST→Net→Eval has one free port}{AST->Net->Eval}
+  Let $A = Net(AST)$.
+  \\ \\
+  By lemma \ref{Lm:AST->Net} $A$ has one free port.
+  \\ \\
+  We must now show $Eval(A)$ does not change the number of free ports.
+  \\ \\
+  We can show this by simply considering how every reduction rule works. For the sake of brevity, we will only consider the three variations of rewrite rules present in the BOHM system WLOG.
```
cwgoes: What is WLOG?

I think you mean that every rule takes one of these forms with `f` and `g` varying, but that's not clear.

I don't think it's true unless you allow the number of auxiliary ports to vary.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+        \end{tikzpicture}
+      \end{subfigure}
+    \end{figure}
+
+    We can see that there is only one wire eliminated, the one connecting the main ports of node $f$ and node $g$.
+    \\ \\
+    The rest of the wires are simply rewired to each other, thus not effecting the total number of free ports
+  \end{tcolorbox}
+
+  \begin{tcolorbox}[title=Case 2 - Isolation, colframe=inner-box-1]
```
cwgoes: "Isolation"?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+      \begin{subfigure}[c]{0.1\textwidth}
+        \begin{tikzpicture}
+          \inode[right = 20 pt]{b}{$⊗$}
+          \inwire[]{b}{b}
+          \draw (-1.5,-1.5) node{a}  .. controls (-1.5,1) and (0,-1.5) .. (0,1) node{c};
+        \end{tikzpicture}
+      \end{subfigure}
+    \end{figure}
+
+    Here we can see that like the first case the wire that connects the main ports are eliminated, and that $a$ and $c$ are rewired.
```
cwgoes: This `f` node has three auxiliary ports, which isn't true for most of the current BOHM rules, and I don't think this pattern matches any of the rules either.

Are you thinking of erasure? Or of beta reduction?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+          \outwire[]{c.out}{b}
+          \outwire[]{d.out}{c}
+          \swire[0]{a/4}{c/1}{}{}
+          \swire[0]{b/1}{d/4}{}{}
+          \swire[0]{a/1}{d/1}{}{}
+          \swire[0]{b/4}{c/4}{}{}
+        \end{tikzpicture}
+      \end{subfigure}
+    \end{figure}
+    Here we see the case for a fan-in node $f_i$ and some arbitrary node $a$. Both nodes get duplicated,
```
cwgoes: I don't think you need to argue this specifically for fan-in nodes, you can just argue that for any rule that takes this form the number of free ports remains constant.

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+    however the number of free ports is preserved, as each node that is created is internally connected to the duplicated nodes.
+    Additionally the four external ports are preserved among the four duplicated nodes.
+  \end{tcolorbox}
+  $\therefore$ since all cases are covered, $Eval(A)$ has only one free port.
```
cwgoes: I'm not convinced that all the cases are covered. If all the BOHM rewrite rules take one of these forms, can we list which rule takes which form?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+\end{Theorem}
+
+We now shall define one more definition before getting to our main theorem.
+
+
+\begin{Definition}{Valid EAL-Net}{Valid EAL-Net}
+  A \textcolor{definitions}{\textbf{Valid EAL-Net}} is a net translated from the EAL subset subset of BOHM and has undergone zero or more reduction steps.
```
cwgoes:
```suggestion
  A \textcolor{definitions}{\textbf{Valid EAL-Net}} is a net translated from the EAL subset of BOHM that has undergone zero or more reduction steps.
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  Now to show this evaluates to the same answer, we must reconstruct the ADT considering all cases.
+  \\ \\
+  For this we must keep two maps as we do this algorithm.
+  \\ \\
+  The first being the variable name from a specific lambda.
+  \\ \\
+  The second map is a map from the fan in number to the current status of our traversal of said fan-in.
+  \\ \\
+  We can now a proof by induction, specifically, we shall consider three cases:
```
cwgoes:
```suggestion
  We can now write a proof by induction, specifically, we shall consider three cases:
```

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+  \\ \\
+  Let $rec$ be this recursive algorithm.
+  \begin{tcolorbox}[title= Case 1 - rest, colframe=inner-box-1]
+    The logic for all other cases is simple. Consider some node $P$ with ports $1 \cdots n$.
+    \\ \\
+    Let port n be the parent ADT port.
+    \\ \\
+    By induction we run this algorithm and ports $1 \cdots n-1$ are $α$ equivalent to normal evaluation.
+    \\ \\
+    Thus we can construct ADT $P$ by $P(rec(n_1) \cdots rec(n_{n-1}))$.
```
cwgoes: I think you keep using ADT where you mean AST

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+
+  3. The rest.
+
+  We will also consider the inductive case first for the following proof.
+  \\ \\
+  Let $rec$ be this recursive algorithm.
+  \begin{tcolorbox}[title= Case 1 - rest, colframe=inner-box-1]
+    The logic for all other cases is simple. Consider some node $P$ with ports $1 \cdots n$.
+    \\ \\
+    Let port n be the parent ADT port.
```
cwgoes: AST?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+    \\ \\
+    By induction we run this algorithm and ports $1 \cdots n-1$ are $α$ equivalent to normal evaluation.
+    \\ \\
+    Thus we can construct ADT $P$ by $P(rec(n_1) \cdots rec(n_{n-1}))$.
+    \\ \\
+    $\therefore$ $P$ is well formed and we get an $α$ equivalent answer.
+  \end{tcolorbox}
+
+  \begin{tcolorbox}[title= Case 2 - $\lambda$ /\ $\mu$, colframe=inner-box-1]
+    The $\lambda$ /\ $\mu$ case is special in that in a Valid EAL-Net the node will be traversed via the primary port via the parent AST and then by the second auxiliary port.
```
cwgoes: What is "we traverse" - the read-back algorithm traverses?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+
+    For this analysis we must consider what port we can enter from, however first we should note that before entering the primary port of a FanIn, that we must have first traversed through either $\star$ or $\circ$ first.
+    \\ \\
+    This is the case because entering a FanIn in this way is known as a FanOut, and denotes the end of sharing.
+    And since ending sharing before starting is not possible from a valid EAL-Net, these cases never happen.
+    \\ \\
+    \begin{tcolorbox}[title= Enter from Prim, colframe=inner-box-2]
+      As shown above, we must have already traversed a FanIn node, namely $\star$ or $\circ$.
+      \\ \\
+      This means the map from the FanIn number $i$ is not empty, and thus we have a history of completing nodes or being in nodes.
```
cwgoes: What do "completing nodes" or "being in nodes" mean?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+        \begin{tikzpicture}
+          \inetcell[]{a}{$f_i$}[5]
+          \draw (-0.29, 0.23) node {$\scalebox{1.9}{$\circ$}$};
+          \draw ( 0.29, 0.29) node {$\scalebox{1.7}{$\star$}$};
+          \cutWire{a.out}{b.north}{}{}
+        \end{tikzpicture}
+    \end{figure}
+
+
+    For this analysis we must consider what port we can enter from, however first we should note that before entering the primary port of a FanIn, that we must have first traversed through either $\star$ or $\circ$ first.
```
cwgoes: This sounds like you are keeping a first-in-first-out stack of ⋆or∘ indicators while traversing, is that right?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+      As shown above, we must have already traversed a FanIn node, namely $\star$ or $\circ$.
+      \\ \\
+      This means the map from the FanIn number $i$ is not empty, and thus we have a history of completing nodes or being in nodes.
+      \\ \\
+      If we have an unfinished $In$, then we pick the most recent $In$ port left to traverse.
+      This is forced as this node is considered a FanOut and thus closes sharing.
+      \\ \\
+      We then mark the node as complete. If we have completed a single node, then we simply choose to leave through the other port.
+      \\ \\
+      Since we are working over valid EAL-Net's which require no oracle, there is no ambiguity in this mechanism. Furthermore, this fact also excludes any other configurations from happening.
```
cwgoes: How does the fan's label come into play here?

doc/reference/src/low-level-execution-model.pdc
```diff
@@ -111,276 +111,542 @@ The oracle-free abstract algorithm for optimal reduction operates on four node t
...
+      \\ \\
+    \end{tcolorbox}
+    \begin{tcolorbox}[title= Enter from Aux, colframe=inner-box-2]
+      We entered the FanIn through an auxiliary port, so we mark in the second map that we are $In$ this port at FanIn node $i$.
+      \\ \\
+      We then leave through the principle port, denoting the beginning of sharing.
+    \end{tcolorbox}
+    These cases handle all possible configurations in which we can enter and the validity of the expressions are handled by induction.
```
cwgoes: Which expressions?

cwgoes:
I think we have to use context semantics for the full argument of correctness.

See [this paper](https://sci-hub.se/https://doi.org/10.1016/j.ic.2010.10.002), I think this is the clearest presentation (it's still quite complex).

Let's discuss this on Monday.

doc/reference/src/low-level-execution-model.pdc
mariari: Correct

doc/reference/src/low-level-execution-model.pdc
mariari: Yeah, as if A is an ADT, then P is a case (hence sub-case) of A. 

Is sub-tem more proper terminology?

doc/reference/src/low-level-execution-model.pdc
mariari: This was the one unfinished graph, there is supposed to be n+1 ports, but alas wasn't sure how to draw that just yet

doc/reference/src/low-level-execution-model.pdc
mariari: This is not about terms being free of closed, simply that to inhabit any ADT, we must give it all arguments. 

An example is
```haskell
type Pair = Pair Bool Bool

t :: Pair
t =  Pair True False

cPair :: Bool -> Pair
cPair = Pair True
````

If were to say partially apply Pair, like in cPair, then it isn't in the set of Pari's but instead in the set of functions which give us a Pair.

For open terms, we just

doc/reference/src/low-level-execution-model.pdc
mariari: above means, the ADT that contains L, if L is not the root expression.

> I agree with this but I do not think it is trivial, we should argue each case of the encode function.

yeah this addition wouldn't be hard, it necessitates that we give an algorithm for encode. As it is, this proof  works under nodes giving back the prots to connect to. This is alluded to in the proof.

doc/reference/src/low-level-execution-model.pdc
mariari: this talks about a free symbol in graph so (lambda f. f x). we have no lambda to connect x to, so we create a node for the symbol x. this preserves the number of free ports

doc/reference/src/low-level-execution-model.pdc
mariari: >What is WLOG?

without loss of generality. It's often used in proofs where one doesn't want to specify every case, as the cases they cover are general and cover all cases by using a specific case

> I think you mean that every rule takes one of these forms with f and g varying, but that's not clear.

Yeah, I could  have been clearer.

> I don't think it's true unless you allow the number of auxiliary ports to vary.

Ideally they would be, but I couldn't figure out how to add ...'s to the diagram, so I left them with 2 arguments

doc/reference/src/low-level-execution-model.pdc
mariari: It isolates the node from the rest of the graph (takes it out of the connected components), and connects an eraser to it

doc/reference/src/low-level-execution-model.pdc
mariari: this case is very similar to what `If` does to `T` or `F`, and is the most general, as it rewrires a wire, and erases another branch.

the WLOG is used here, as all isolating cases act similarly

doc/reference/src/low-level-execution-model.pdc
mariari: I use `fan-in` here, as there are no other nodes which duplicate nodes this way, it's just generic on `fan-in`

doc/reference/src/low-level-execution-model.pdc
mariari: Will do!

doc/reference/src/low-level-execution-model.pdc
mariari: AST/ADT

doc/reference/src/low-level-execution-model.pdc
mariari: Correct, I view them as quite interchangeable, for the purposes of this, but will change if using AST is less clear

doc/reference/src/low-level-execution-model.pdc
mariari: s/traverse/reach the λ node via recursion/

doc/reference/src/low-level-execution-model.pdc
mariari: first-in-last out stack, correct.

doc/reference/src/low-level-execution-model.pdc
mariari: completing means completing the fan-in then fan-out of star or circle. 

In means you have visited the fan-in of star or circle, but not the fan-out

I could have explained this better

doc/reference/src/low-level-execution-model.pdc
mariari: "This means the map from the FanIn number $i$ is not empty, and thus we have a history of completing nodes or being in nodes." 

From above, basically we have the label be a map to the status of going through various ports.

doc/reference/src/low-level-execution-model.pdc
mariari: the read-back expression of running the algorithm

doc/reference/src/low-level-execution-model.pdc
cwgoes: OK, I think "we" is an unclear reference - "the algorithm traverses"?

doc/reference/src/low-level-execution-model.pdc
cwgoes: I thought A was an AST?

doc/reference/src/low-level-execution-model.pdc
cwgoes: Oh you mean the ADT *of BOHM terms*. I think that is evident and you can just mention that free variables receive named nodes.

doc/reference/src/low-level-execution-model.pdc
cwgoes: Ok this can be done later

doc/reference/src/low-level-execution-model.pdc
cwgoes: (do you mean one of the variants of a BOHM term?)

doc/reference/src/low-level-execution-model.pdc
cwgoes: can we say that instead?

doc/reference/src/low-level-execution-model.pdc
cwgoes: right-o

doc/reference/src/low-level-execution-model.pdc
cwgoes: > without loss of generality. It's often used in proofs where one doesn't want to specify every case, as the cases they cover are general and cover all cases by using a specific case

OK we should define it somewhere, it can be earlier.

> Ideally they would be, but I couldn't figure out how to add ...'s to the diagram, so I left them with 2 arguments

Alright, we should mention it, even as a TODO.

doc/reference/src/low-level-execution-model.pdc
cwgoes: We should mention that the number of ports varies, and be clear about what is under the "forall" in these WLOG cases.

doc/reference/src/low-level-execution-model.pdc
cwgoes: I don't think this usage is standard. A BOHM term isn't an ADT - an ADT is the *type*.

Maybe a BOHM term is "an inhabitant of the BOHM ADT", if you like.

doc/reference/src/low-level-execution-model.pdc
cwgoes: can we say that explicitly?

doc/reference/src/low-level-execution-model.pdc
cwgoes: just define the phrases somewhere

doc/reference/src/low-level-execution-model.pdc
cwgoes: I see, we should refer to induction separately in each case where we invoke it for a subterm I think.

doc/reference/src/low-level-execution-model.pdc
cwgoes: that's what I thought, we should be clearer about the separation of context by level though

doc/reference/src/low-level-execution-model.pdc
mariari: I have added todo's now.

The first point still isn't addressed

doc/reference/src/low-level-execution-model.pdc
mariari: Done, it's above where the type is defined.

PR #99, opened 2019-09-17
First-class usages

andy-morris:
- [x] Syntax for usages-as-terms
- [x] Reduction & type rules
- [ ] Make sure subusaging (`↑ π ≾ ω`) is well-behaved
- [ ] Confluence of reduction rules
- [ ] Type safety proof


cwgoes:
Bump :) - do you want to push the state of things & merge this?

andy-morris:
So I added usage terms to what I had already, which was straightforward-ish, except for [multiplying a context by a scalar](https://github.com/cryptiumlabs/juvix/blob/e84e3e55cf62ddef1e879bea9df3bd1e01acb27b/experimental/qtt-agda/Type.agda#L57-L61).

For a multiplication πΓ, where π can contain variables, it might not even be well formed for the whole context! So I'm not sure how to solve this atm. If anyone has any insights please share them. :)

cwgoes:
Maybe we need to define `πΓ` only in the presence of some `Γ_0` under which `π` is well-defined (e.g. variables `v_i` in `π` have `0 v_i : 𝓤`), and multiply the remainder of the context by `π`, e.g.

`(1 x: ℕ, ω y: ℕ, 0 u : 𝓤)  ⊢ u (1 z : ℕ) = (u z : ℕ)`.

All rules involving multiplication of some context `πΓ` will then need to be updated to reflect this "multiplication well-formed" condition.

Did I understand your question correctly / does that make any sense?

andy-morris:
So you're saying all the variables which are not used in π, but bound before one which is, are basically considered out of scope? I think that might work.

Though I wonder if we could get away with just setting them to 0. That would be simpler, I think, and slightly more expressive.

cwgoes:
> So you're saying all the variables which are not used in π, but bound before one which is, are basically considered out of scope? I think that might work.
> 
> Though I wonder if we could get away with just setting them to 0. That would be simpler, I think, and slightly more expressive.

That seems like it could also work, if you think it will let's try it, probably easier to evaluate with the revised typing rules written out explicitly.

andy-morris:
I'm not a fan of how janky [this definition](https://github.com/cryptiumlabs/juvix/blob/8ffc17a1a7d0cf45bad24e0468f362ff5f77c9d6/experimental/qtt-agda/Type.agda#L67-L73) is, but maybe it'll all work out in the end.

andy-morris:
[Look, a polymorphic definition!!](https://github.com/cryptiumlabs/juvix/blob/debb192b2684564c4e0d793f60fce44ba9e50ee0/experimental/qtt-agda/Example.agda#L106-L112)

Also there are [some design considerations](https://github.com/cryptiumlabs/juvix/blob/debb192b2684564c4e0d793f60fce44ba9e50ee0/experimental/qtt-agda/Example.agda#L137-L145) about how smart to make the typechecker when it comes to usage expressions. I think that option 3 is probably better—even though it makes the typechecker more complex, I expect that tedious shuffling like the `auxᵢ` will show up constantly otherwise.

andy-morris:
```agda
module 🙃 where

open import Prelude ; open import QTT ; open import Eval
open import Relation.Binary.Construct.Closure.ReflexiveTransitive

variable n : ℕ ; R S T : Term n ; e f : Elim n

module _ {n} T z s (w : Term n) where
  𝓤-elim-sucω₁ : [ 𝓤-elim T z s w (sucᵘ ωᵘ) ] ⟿ᵗ* substᵗ (substᵗ s _) _
  𝓤-elim-sucω₁ = [ β-𝓤s ] ◅ υ ◅ ε

  𝓤-elim-sucω₂ : [ 𝓤-elim T z s w (sucᵘ ωᵘ) ] ⟿ᵗ* w
  𝓤-elim-sucω₂ = [ 𝓤-elim₅ (sucᵘ-ω) ] ◅ [ β-𝓤ω ] ◅ υ ◅ ε

≡-≋ : S ≡ T → S ≋ᵗ T
≡-≋ refl = Evalᵗ.≋-refl


-- all types are definitionally equal!!!!
S≋T : ∀ (S T : Type n) → S ≋ᵗ T
S≋T S T =
  let open Evalᵗ in
  ≡-≋ sw² ◅◅
  ≋-sym (star-≋ (𝓤-elim-sucω₁ 𝓤 𝓤 (weakᵗ $ weakᵗ S) T)) ◅◅
  star-≋ (𝓤-elim-sucω₂ _ _ _ _)
 where
  sw² : R ≡ substᵗ (substᵗ (weakᵗ (weakᵗ R)) e) f
  sw² {R = R} {e = e} {f = f} rewrite subst-weakᵗ (weakᵗ R) 0 e =
    ≡.sym $ subst-weakᵗ R 0 f
```

Going to have to reformulate this a bit, maybe by having separate types for finite usages and maybe-infinite ones or something

andy-morris:
I suppose this might as well be merged now whenever someone's had time to have a look through it, since it doesn't touch the rest of the codebase.

andy-morris:
(Merging this and opening separate PRs for major proofs, etc, as discussed.)

PR #100, opened 2019-09-18
added proper O(1) compare symbols

mariari:


cwgoes:
Hmm, I'm not sure what I think of this library, [it's rather a hack](https://www.stackage.org/haddock/lts-13.3/symbol-0.2.4/src/Data.Symbol.Unsafe.html#Symbol).

I'm also not convinced it will be more efficient - if symbols aren't known until runtime, we're now manipulating this map every time we parse one.

I think using `Data.Text` (fast unicode bytestring) for symbols is probably fine in the compiler / typechecker; symbol comparisons aren't likely to be the performance bottleneck (much more likely z3 or optimizing transformations which perform complex graph search).

If we do end up needing fast symbol comparisons using `Int` directly in some layer of the core language seems preferable to this implicit map.

mariari:
see #102

PR #105, opened 2019-09-19
The error msgs now more consistently show users the names they are familiar with (as per the parser and lang ref).

thealmarty:
Minimally Tested:
~~~~
*Juvix.Core.MainLang> cType 0 [] (Star 1) (0,VStar 0)
Left "*1 is of type * of a higher universe. But the expected type *0 is * of a equal or lower universe."
*Juvix.Core.MainLang> cType 0 [] (Pi 0 Nats Nats) (0, VStar 0)
Right ()
*Juvix.Core.MainLang> cType 0 [] (Pi 1 Nats Nats) (0, VStar 0)
Right ()
*Juvix.Core.MainLang> cType 0 [] (Pi 1 Nats Nats) (1, VStar 0)
Left "Sigma has to be 0."
*Juvix.Core.MainLang> cType 0 [] (Pi 1 Nats Nats) (0, VStar 1)
Left "Type mismatched. \nNat \n (binder number 0) is of type \n\"*1\" , with 0 usage.\n But the expected type is \"*0\" , with 0 usage."
*Juvix.Core.MainLang> cType 0 [] (Pi 1 Nats Nats) (0, VNats)
Left "The variable type and the result type must be of type * at the same level."
~~~~

### DIFF
--- a/src/Juvix/Core/MainLang.hs
+++ b/src/Juvix/Core/MainLang.hs
@@ -234,7 +234,7 @@ cType _ii _g (Star n) ann = do
         (n < j)
         (throwError $
          show (Star n) ++
-         " is of type * of a higher universe. But the input type " ++
+         " is of type * of a higher universe. But the expected type " ++
          showVal (snd ann) ++ " is * of a equal or lower universe.")
     _ ->
       throwError $ "* n is of type * but " ++ showVal (snd ann) ++ " is not *."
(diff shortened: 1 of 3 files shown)


src/Juvix/Core/MainLang.hs
```diff
@@ -44,7 +44,19 @@ data CTerm
                         -- The abstracted variable's usage is tracked with the Usage(π).
   | Conv ITerm --(CONV) conversion rule. TODO make sure 0Γ ⊢ S≡T
               -- Conv is the constructor that embeds ITerm to CTerm
-  deriving (Show, Eq)
+  deriving (Eq)
+
+instance Show CTerm where
```
cwgoes: Spaces would be nice, probably a bit more readable with them

PR #110, opened 2019-09-23
Even more language reference work

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/90


cwgoes:
I think unification in the elementary affine core will require beta reduction.

Maybe easier to use the untyped (less efficient) inference algorithm for now.

PR #113, opened 2019-09-23
Thealmarty/styraconyx core

thealmarty:
Init core for Styraconyx release.


thealmarty:
Alternatively, we would have a Styraconyx branch? Then I would have these files as MainLang and Parser in the core.

PR #115, opened 2019-09-26
Thealmarty/add operations to nats

thealmarty:
Adding tests is WIP but the nat operations are added in the MainLang and Parser.  The operations use the operations of Haskell's Naturals.

TODOs:
- Still have to investigate why some operations are not parsed properly with the cterm parser. (All operations work fine with the iterm parser, the convITerm parser isn't converting the Ops that aren't the first listed Op in the iterm parser.) 
- Add unit and property tests for the parser. 
- Make this work with cEval.

### DIFF
--- a/app/Interactive.hs
+++ b/app/Interactive.hs
@@ -20,52 +20,51 @@ import qualified Juvix.Nets.Bohm              as Bohm
 
 interactive ∷ Context → Config → IO ()
 interactive ctx _ = do
-  func ← return $ \str -> do
-    return str
+  func <- return $ \str -> do return str
   H.runInputT (settings ctx) (mainLoop func)
 
 settings ∷ Context → H.Settings IO
-settings ctx = H.Settings {
-  H.complete        = H.completeFilename,
-  H.historyFile     = Just (contextHomeDirectory ctx <> "/.jvxi_history"),
-  H.autoAddHistory  = True
-  }
+settings ctx =
+  H.Settings
+    { H.complete = H.completeFilename
+    , H.historyFile = Just (contextHomeDirectory ctx <> "/.jvxi_history")
+    , H.autoAddHistory = True
+    }
 
 mainLoop ∷ (String → IO String) → H.InputT IO ()
 mainLoop func = do
-  input ← H.getInputLine "jvxi >> "
+  input <- H.getInputLine "jvxi >> "
   case input of
-    Nothing → return ()
-    Just i  → do
+    Nothing -> return ()
+    Just i -> do
       case i of
-        (':' : special) →
-          handleSpecial special (mainLoop func)
-        inp → do
+        (':':special) -> handleSpecial special (mainLoop func)
+        inp -> do
           H.outputStrLn =<< liftIO (func inp)
           mainLoop func
 
 handleSpecial ∷ String → H.InputT IO () → H.InputT IO ()
 handleSpecial str cont = do
   case str of
-    "?"    → liftIO (putDoc specialsDoc) >> cont
-    "exit" → return ()
-    "tutorial" → do
+    "?" -> liftIO (putDoc specialsDoc) >> cont
+    "exit" -> return ()
+    "tutorial" -> do
       H.outputStrLn "Interactive tutorial coming soon!"
       cont
-    'c' : 'p' : ' ' : rest -> do
+    'c':'p':' ':rest -> do
       let parsed = Core.parseString Core.cterm rest
       H.outputStrLn $ show parsed
       cont
-    'c' : 't' : ' ' : rest -> do
+    'c':'t':' ':rest -> do
       let parsed = Core.parseString Core.cterm rest
       H.outputStrLn $ show parsed
       case parsed of
         Just cterm -> do
           let eval = Core.cEval cterm []
-          H.outputStrLn $ Core.showVal eval
+          H.outputStrLn $ show eval
         Nothing -> return ()
       cont
-    'c' : 'e' : ' ' : rest -> do
+    'c':'e':' ':rest -> do
       let parsed = Core.parseString Core.cterm rest
       H.outputStrLn $ show parsed
       case parsed of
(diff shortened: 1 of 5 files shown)


app/Interactive.hs
```diff
@@ -20,52 +20,51 @@ import qualified Juvix.Nets.Bohm              as Bohm
 
 interactive ∷ Context → Config → IO ()
```
thealmarty: The arrows weren't changed as far as I can tell, not sure why it's showing these line changes here. I've removed the showVal function to be a show instance so I changed the code here accordingly.

app/Interactive.hs
```diff
@@ -20,52 +20,51 @@ import qualified Juvix.Nets.Bohm              as Bohm
...
       H.outputStrLn $ show parsed
       cont
-    'c' : 't' : ' ' : rest -> do
+    'c':'t':' ':rest -> do
       let parsed = Core.parseString Core.cterm rest
       H.outputStrLn $ show parsed
       case parsed of
         Just cterm -> do
           let eval = Core.cEval cterm []
-          H.outputStrLn $ Core.showVal eval
```
thealmarty: This is the line I changed.

package.yaml
```diff
@@ -125,6 +126,8 @@ tests:
       - juvix
       - tasty
       - tasty-hunit
+      - tasty-quickcheck
+      - QuickCheck
```
thealmarty: I will need to write custom generator for test data generation, so I need to add QuickCheck too.

app/Interactive.hs
```diff
@@ -20,52 +20,51 @@ import qualified Juvix.Nets.Bohm              as Bohm
...
+          H.outputStrLn $ show eval
```
thealmarty: This is the line I changed.

app/Interactive.hs
cwgoes: It looks like they were changed from Unicode to non-Unicode. It's fine.

src/Juvix/Core/MainLang.hs
```diff
@@ -87,25 +108,45 @@ data Value
...
-showVal (VNeutral _n) = "neutral "
-showVal (VNat i) = show i
-
-showFun ∷ (Value → Value) → String
-showFun _f = "\\x.t "
+varX ∷ Value
+varX = VNeutral (NFree (Global "x"))
+
+--(Value -> Value) are equal when given an arbitrary value they return the same value.
+valueToValueEq ∷ (Value → Value) → (Value → Value) → Bool
```
cwgoes: Hmm I don't think this is true in general; what if they only returned the same value for `varX`?

test/MainLangTests.hs
```diff
@@ -0,0 +1,51 @@
...
+instance Arbitrary CTerm where
+  arbitrary = CTerm -}
+natsTypeStar0 ∷ Assertion
+natsTypeStar0 = cType 0 [] Nats (0, VStar 0) @?= Right ()
+
+nat1Inferred ∷ Assertion
+nat1Inferred = iType 0 [] (Nat 1) @?= Right (Omega, VNats)
+
+--function has to be named test_ to be picked up by tasty.
+test_core ∷ Test.Tasty.TestTree
```
cwgoes: Note that you can also use automatic discovery via `tasty-discover` for the individual assertions. This is fine as well.

test/MainLangTests.hs
```diff
@@ -0,0 +1,51 @@
...
+import           Prelude
+
+import           Test.Tasty
+import           Test.Tasty.HUnit
+import           Test.Tasty.QuickCheck
+
+import           Test.QuickCheck
+import           Test.QuickCheck.Gen
+
+--property tests of type checker:
```
cwgoes: Eventually we'll want to test the core evaluator against the LLVM evaluator, when that's ready, which can be done with random core term generation / fuzzing.

test/MainLangTests.hs
```diff
@@ -0,0 +1,51 @@
...
+-- \x.x (any term) evaluates to (any term evaluated)
+{-lamProp :: CTerm -> Env -> Property
+lamProp cterm env = App (cEval (Lam Bound 0) env) cterm == cterm-}
+-- any constant term evaluates to itself
+constProp ∷ CTerm → Env → Bool
+constProp (Star i) env = cEval (Star i) env == VStar i
+constProp Nats env     = cEval Nats env == VNats
+  --constProp _ _          = True --Not testing non-const terms
+
+{-TODO need to combine generators to generate CTerms http://hackage.haskell.org/package/QuickCheck-2.13.2/docs/Test-QuickCheck-Gen.html
```
cwgoes: Some manual testcases for typechecking would be nice as well, e.g. the SKI combinators as in the Agda code.

src/Juvix/Core/MainLang.hs
cwgoes: I think you might want to use `quote0` and check equality of the resulting `CTerm`.

app/Interactive.hs
mariari: It could be the autoformater on save being weird, as the spacing for types and constructors has also changed quite drastically

src/Juvix/Core/MainLang.hs
thealmarty: I can use ```quote0``` to check the equality of the resulting ```CTerm``` like this:

 ```valueToValueEq y1 y2 = quote0 (y1 varX) == quote0 (y2 varX)```

But I still don't have a solution for the concern that they may only return the same value for ```varX```. Any idea? Checking all possible ```Value``` type will result in infinite loops for the types with ```Value -> Value```.

cwgoes: I mean use `quote0` first on the top-level value, e.g.:

```haskell
instance Eq Value where
  x == y = quote0 x == quote0 y
```

Would that work?

thealmarty: I see!  Yes! Thank you!

test/MainLangTests.hs
thealmarty: Will do in another PR.

test/MainLangTests.hs
thealmarty: Will do in another PR.

PR #120, opened 2019-10-02
Build z3 in Makefile

cwgoes:
Build locally & use locally-build version.


cwgoes:
Works, I think, but the binary will need to be run with

```bash
LD_LIBRARY_PATH=.stack-work/prefix/lib juvix interactive
```

which is annoying.

PR #122, opened 2019-10-02
Thealmarty/nat op and parser tests

thealmarty:
Finished:
- parser unit tests for cterm and iterm (included in cterm parser as convITerm).

To-dos:
- nat Operations which works properly in core and App.
- add value parser for natOp
- add value parser for cEval and iEval
- add tests for cType (including unit tests for SKI combinator), iType, cEval, iEval


thealmarty:
I'm sorry the 'file changes' is a huge mess. I don't understand why the #123 merge is showing there? Anyway, I've only changed the Parser and MainLang files. Mainly adding unit tests to cterm parser.

PR #123, opened 2019-10-02
Elementary affine core

cwgoes:


cwgoes:
Merging to avoid possible conflicts, future work to come in a separate PR.

PR #126, opened 2019-10-03
Mariari/semi ring nat andw

mariari:


thealmarty:
I'm still not a big fan of importing libs. This makes the code unnecessarily complex and harder to read. What is bothering you that makes you want the semiring lib and made all these changes?

mariari:
> I'm still not a big fan of importing libs. This makes the code unnecessarily complex and harder to read. What is bothering you that makes you want the semiring lib and made all these changes?

There isn't any more libs being imported that isn't already in the stdlib. Chris has chosen to use Protolude, after using it for a bit I do enjoy it.

Since protolude imports MTL and other annoying things that don't work that well with capability we have elected to have a standard library. 

Due to protolude we get semiring for free.

Also the num instance you have written is unsafe. If you try - it will error, this also causes warnings in the REPL and when we build, we should minamize this. Also conceptually, what we are reasoning over is a semiring, and even in your comment of the type you note this. Thus we should try to be as accurate to the math as possible, and not lie about the type

thealmarty:
> There isn't any additional libs being imported that isn't the standard library. Chris has chosen to use Protolude, after using it for a bit I do enjoy it.
> 
> Since protolude imports MTL and other annoying things that don't work that well with capability we have elected to have a standard library.
> 
> Since we have a standard library, we should use it to be uniform! Protolude was tailored made to solve many issues the normal prelude has.
> 
> Due to using protolude we get semiring for free.
> 
> Also the num instance that is currently there is unsafe. If you try `-` or other unimplemented functions it will error, this also causes warnings in the REPL and when we build, we should minamize this. Also conceptually, what we are reasoning over is a semiring, and even in your comment of the type you note this. Thus for these two reasons we should elect to use SemiRing over Num. I would argue this makes the code more obvious, not more complex. When we have really good types algorithms often become obvious.
> 
> For example when we have a mapping between keys and values, it's more informational to say, `Map Keys Values` over a list of keys and values. The reason for this is that lists in functional languages are overused and thus the meaning of `[(Key,Value)]` is obfuscated. This really bites you when refactoring or changing design. For `bindle` I had to refactor some code from using list to O(1) concat list. It ended up being a huge pain trying to figure out what is the lists that I need to change, and what lists are actually being used as proper lists! This is also why I change all `++` to `<>` ++ is just a specialization of `<>` and when we eventually refactor to remove strings (strings are slow and should rarely be used seriously) with the inclusion of auttoparsec, ++ will just lead to annoying changes that can be avoided earlier.
> 
> I hope this gives you a better insight into the methodology of why these changes are here. Feel free to tell me anything you disagree with this!

Thanks for the detailed explanations! I appreciate this! I was going to bring up if we need protolude but I see I'm outvoted, and I accept this. 

And yes I noticed that ```<>``` is standard in the code base. I'll use my editor to replace ```++``` with ```<>``` in all files. If there's any standard/coding practice that I'm not aware of, please let me know, I can easily replace across all files - you don't need to worry about changing individual instances.

We can merge this after Chris ACK this.

PR #127, opened 2019-10-03
Mariari/mainlang test fix

mariari:
Makes the test compile

### DIFF
--- a/src/Juvix/Core/MainLang.hs
+++ b/src/Juvix/Core/MainLang.hs
@@ -265,12 +265,10 @@ cType ii g (Pa pi varType resultType) ann = undefined
 cType ii g (NPm first second) ann = undefined
 -- (Lam) introduction rule of dependent function type
 -- Lam s should be of dependent function type (Pi pi ty ty').
-cType ii g (Lam s) (sig, VPi pi _ty ty') = do
-  traceShowM s
-  let sVal = cEval s []
+cType ii g (Lam s) (sig, (VPi pi ty ty')) =
   cType
     (ii + 1)
-    ((Local ii, (sig <.> pi, sVal)) : g) -- put s in the context with usage sig*pi
+    ((Local ii, (sig <.> pi, ty)) : g) -- put s in the context with usage sig*pi
     (cSubst 0 (Free (Local ii)) s)      -- x (varType) in context S with sigma*pi usage.
     (sig, ty' (vfree (Local ii)))       -- is of type M (usage sigma) in context T
 cType ii g (Lam s) ann =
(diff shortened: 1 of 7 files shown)


src/Juvix/Core/MainLang.hs
```diff
@@ -265,12 +265,10 @@ cType ii g (Pa pi varType resultType) ann = undefined
...
 -- (Lam) introduction rule of dependent function type
 -- Lam s should be of dependent function type (Pi pi ty ty').
-cType ii g (Lam s) (sig, VPi pi _ty ty') = do
-  traceShowM s
-  let sVal = cEval s []
+cType ii g (Lam s) (sig, (VPi pi ty ty')) =
   cType
     (ii + 1)
-    ((Local ii, (sig <.> pi, sVal)) : g) -- put s in the context with usage sig*pi
+    ((Local ii, (sig <.> pi, ty)) : g) -- put s in the context with usage sig*pi
```
thealmarty: The red (old) code are not my code. I didn't have ```_ty``` there.

src/Juvix/Core/MainLang.hs
```diff
@@ -137,14 +117,14 @@ vfree n = VNeutral (NFree n)
...
---Contexts map variables to their types.
+-- Contexts map variables to their types.
 type Context = [(Name, Annotation)]
 
 --Evaluation
 type Env = [Value]
 
 cEval ∷ CTerm → Env → Value
-cEval (Star i) _d      = VStar i
+ceval (Star i) _d      = VStar i
```
thealmarty: typo.

src/Juvix/Core/MainLang.hs
thealmarty: You are correct. I should have put the ```ty``` (instead of the evaluated ```s```) in the context. This is a quick fix that I will merge in so that the tests pass. In the future we'll all make smaller, separate PR so we can merge them with more flexibility.

cwgoes:
Does this supersede https://github.com/cryptiumlabs/juvix/pull/125? It seems like it does, maybe we should close that one in favor of this.

thealmarty:
> Does this supersede #125? It seems like it does, maybe we should close that PR in favor of this one.

#125 is auto closed by github when I merged #126! How nice! It may make sense to look at #127 again after #131 is merged - the ```MainLang``` test file got overhauled in #131 .

thealmarty:
I looked at the diff and I don't think this PR is relevant anymore. The test fix was done and #126 picked up the rest. @mariari  could you please confirm? Thanks!

mariari:
Yeah It should no longer be relevent, closing now

PR #128, opened 2019-10-04
indented erasure so it doesn't go to character length 165~

mariari:


cwgoes:
... but see https://github.com/cryptiumlabs/juvix/issues/117#issuecomment-538867186.

PR #129, opened 2019-10-06
WIP :: Mariari/llvm

mariari:
Start of LLVM work, a lot needs to be done, once this gets somewhat figured out we can merge and branch off this.


cwgoes:
Might be helpful to draw out the memory layout as a diagram where spacial contiguity reflects spacial contiguity in memory and arrows are pointers etc.

cwgoes:
Is this R4R? (or would reviews be helpful)

src/Juvix/LLVM/Codegen/Block.hs
```diff
@@ -0,0 +1,554 @@
...
+  Symbol →
+  [(Type, Name)] →
+  [BasicBlock] →
+  m Operand
+define retty label argtys body = do
+  addDefn
+    $ GlobalDefinition
+    $ functionDefaults
+      { Global.parameters = params,
+        -- Figure out which is best!
```
cwgoes: Do we have to choose globally or can we choose per-function?

src/Juvix/LLVM/Codegen/Constants.hs
```diff
@@ -0,0 +1,21 @@
...
+import qualified LLVM.AST.Type as Type
+
+portLength ∷ Num p ⇒ p
+portLength = 32
+
+-- | An abbreviation for 'IntegerType' 8
+i4 ∷ Type.Type
+i4 = Type.IntegerType 4
+
+-- TODO ∷ change this based on the machine, C++ pargma?
```
cwgoes: You can use pragmas, but I don't know if we want to tie it to the machine Juvix is compiled on.

src/Juvix/LLVM/Codegen/Graph.hs
```diff
@@ -0,0 +1,217 @@
...
+      )
+    -- TODO ∷ Abstract most of the logic in this function
+    body = do
+      makeFunction "link" args
+      setPort ("node_1", "port_1") ("node_2", "port_2")
+      setPort ("node_2", "port_2") ("node_1", "port_1")
+      _ ← retNull
+      createBlocks
+
+-- preform offsets
```
cwgoes:
```suggestion
-- perform offsets
```

cwgoes:
I suggest focusing on the interaction net operations; native sum/product/etc types are nice but we can add those later, it would be nice to have end-to-end execution working.

src/Juvix/LLVM/Codegen/Block.hs
mariari: Well the choices aren't much.

You can see them [here](http://hackage.haskell.org/package/llvm-hs-pure-9.0.0/docs/LLVM-AST.html#t:Definition).
I believe every function has to be global to the module. Note that we could always make function pointers and use that.

cwgoes: Ah sorry, I should have commented on the line below, I meant the calling convention.

src/Juvix/LLVM/Codegen/Graph.hs
```diff
@@ -0,0 +1,348 @@
...
+      port ← call portType edge (Block.emptyArgs [mainPort])
+      nodePtr ← getElementPtr $
+        Types.Minimal
+          { Types.type' = nodePointer,
+            Types.address' = port,
+            Types.indincies' = Block.constant32List [0, 1]
+          }
+      node ← load nodePointer nodePtr
+      -- run compare function
+      cmp ← undefined
```
cwgoes: This should just depend on the ports, right?

mariari: The function works on a node, the compare should run that the node given back is the same as the current node. I'm thinking I need a reference to the node, and not the node itself to do this function?

From there I can convert PtrToInt.

cwgoes: Ah I see. Yes, I think you will need a reference.

PR #131, opened 2019-10-10
Tidied up tests.

thealmarty:
- Renamed ```MainLangTests.hs``` to ```MainLang.hs``` in ```test/``` to sync with other test files.
- Moved type checker unit tests from ```Erasure.hs``` to ```MainLang.hs``` because the type checkers are from ```MainLang.hs```.
- Rewrote all parser tests to the same style as the other tests.


cwgoes:
@thealmarty I just fixed merge conflicts with develop.

PR #132, opened 2019-10-13
Automatically format using Ormolu

cwgoes:
Closes https://github.com/cryptiumlabs/juvix/issues/117


mariari:
Seems fine, we'll probably want to modify it to fit our needs better in the future though.

PR #134, opened 2019-10-14
Auxiliary triples

A-Manning:


src/Juvix/Nets/Bohm.hs
```diff
@@ -288,43 +326,48 @@ fanInAux2 numFan (numOther, otherLang) level = do
...
-  other1 ← newNode (Primar otherLang)
-  other2 ← newNode (Primar otherLang)
-  let nodeOther1 = RELAuxiliary0 { node       = other1
+fanInAux3 ∷ (InfoNetwork net Lang m)
+          ⇒ Node → (Node, Auxiliary3) → Int → m ()
+fanInAux3 numFan (numOther, otherLang) level = do
+  incGraphSizeStep 3
+  other1 ← newNode (Auxiliary3 otherLang)
+  other2 ← newNode (Auxiliary3 otherLang)
+  other3 ← newNode (Auxiliary3 otherLang)
```
mariari: When fan in meets a node with 3 auxiliary ports, it should only duplicate the node once. Thus this logic needs to be reworked a bit

src/Juvix/Nets/Bohm.hs
```diff
@@ -288,43 +326,48 @@ fanInAux2 numFan (numOther, otherLang) level = do
...
-      nodeFan1   = RELAuxiliary2 { node       = fanIn1
+      nodeFan3   = RELAuxiliary3 { node       = fanIn3
                                  , primary    = ReLink numOther Aux1
                                  , auxiliary1 = Link (Port Aux1 other1)
                                  , auxiliary2 = Link (Port Aux1 other2)
+                                 , auxiliary3 = Link (Port Aux1 other3)
                                  }
-  traverse_ linkAll [nodeOther1, nodeOther2, nodeFan1]
-  deleteRewire [numFan, numOther] [other1, other2, fanIn1]
+  traverse_ linkAll [nodeOther1, nodeOther2, nodeFan1, nodeFan2, nodeFan3]
```
mariari: It seems here you forget about `nodeOther3`,and it's linked in the nodeFan's above

src/Juvix/Nets/Bohm.hs
```diff
@@ -288,43 +326,48 @@ fanInAux2 numFan (numOther, otherLang) level = do
...
-          ⇒ Node → (Node, Auxiliary1) → Int → m ()
-fanInAux1 numFan (numOther, otherLang) level = do
-  incGraphSizeStep 1
-  other1 ← newNode (Auxiliary1 otherLang)
-  other2 ← newNode (Auxiliary1 otherLang)
-  fanIn1 ← newNode (Auxiliary2 (FanIn level))
-  let nodeOther1 = RELAuxiliary1 { node       = other1
-                                 , primary    = ReLink numFan Aux1
-                                 , auxiliary1 = Link (Port Aux1 fanIn1)
+      nodeFan1   = RELAuxiliary3 { node       = fanIn1
```
mariari: Sadly this layer isn't typed, so you are allowed to create a `RELAuxiliary3`, however if you look at FanIn, you had to wrap it in an Auxiliary2. Thus this doesn't work as intended

mariari:
Only things that needs to be done, is adding the rule to the interpreter.

PR #135, opened 2019-10-14
EGtbS stuff

andy-morris:
Having finished studying the paper, filling in the missing proofs, etc, I started actually using it, and... I think it might actually be worse? Even in just defining the syntax. [Look at this mess!!](https://github.com/cryptiumlabs/juvix/blob/6f73613a5f0bd516d4397d389f3eb5629bbd1b7b/experimental/qtt-agda/Term.agda#L77-L79) That used to just be a data constructor.

So I think it might actually be better to stick with de Bruijn indices? Even if the proofs are more complex, at least it'll be easier to tell what's going on. What do you think?


cwgoes:
> So I think it might actually be better to stick with de Bruijn indices? Even if the proofs are more complex, at least it'll be easier to tell what's going on. What do you think?

Would it make sense to use co-de-Bruijin as a sort of IR for easier analysis in the compiler?

I agree that it probably doesn't make sense to use it as any sort of user-facing representation.

If we just intend to use it as an IR (or perhaps not at all) I think it is lower-priority than formalising the theory of first-class usages, so if that's easier to do with standard de Bruijin only let's save EGtbS for later.

andy-morris:
> Would it make sense to use co-de-Bruijin as a sort of IR for easier analysis in the compiler?

Possibly, if it is more efficient to evaluate terms of this form in the typechecker, we could look into using it later. Unless the type checker uses inets as well (I assume it won't, but not sure), in which case I suppose there isn't really anywhere else to put them. 

Anyway, for now I'll switch back to the de Bruijn branch and put first class usages in there. Sorry for running into a dead end like this.

cwgoes:
> > Would it make sense to use co-de-Bruijin as a sort of IR for easier analysis in the compiler?
> 
> Possibly, if it is more efficient to evaluate terms of this form in the typechecker, we could look into using it later. Unless the type checker uses inets as well (I assume it won't, but not sure), in which case I suppose there isn't really anywhere else to put them.
> 
> Anyway, for now I'll switch back to the de Bruijn branch and put first class usages in there. Sorry for running into a dead end like this.

Not at all - worth the investigation!

I would leave the files around somewhere so we can reference them in the future if necessary.

PR #137, opened 2019-10-16
Core reorganisation, tests, erasure work

cwgoes:
This PR makes updates to core per [this task list](https://github.com/cryptiumlabs/juvix/issues/112#issuecomment-542190055) for the next release, in detail:

- Separates out Core into "HR" (human-readable) and "IR" (intermediate representation). The former has text identifiers for variables & the latter has de Bruijin indices. They can be converted between, so the user should only need to interact with the former, and the typechecker just needs to work with one of them (at present, it works with the latter). Test-cases for conversion have been added.
- Parameterizes Core terms (both HR & IR) over the type of primitive types (`𝜅 ∈ 𝐾` from the language reference) and the type of primitive values (`𝑐 ∈ 𝐶` and `𝑓 ∈ 𝐹` from the language reference), so that the architecture can be utilised for different primitive type and primitive value types, which will be the case when we support multiple backends with differing built-in functionality.
- Splits up `MainLang.hs` into `Types.hs` and `Typechecker.hs` so that other modules can import just the types (we should follow this pattern generally).
- Turns the previous parser for Core IR into a parser for Core HR, updates it to match the language reference syntax-wise and parameterise over primitive type & primitive value parsing.
- Renames a few data types to match the language reference.
- Runs `make format` on everything.


cwgoes:
Parsing applications (`f x`) doesn't seem to work; I wonder if this requires special provision.

cwgoes:
- [ ] Fix application parsing
- [ ] Add conversion functions between IR / HR forms
- [ ] Add example set of built-in functions, types
- [ ] Pass through type lookup function to erasure step

cwgoes:
- [ ] Figure out how best to handle the various kinds of names in IR
- [ ] Add more parsing testcases of core HR, including naturals
- [ ] Add primitive functions over naturals
- [ ] Pass through type lookup function to erasure step

cwgoes:
Splitting out the erasure work into a separate PR to ease review somewhat.

cwgoes:
Ready for review. A few outstanding questions:
- I screwed up something in the parser w.r.t parentheses but not sure what.
- Do `IR.Local` variables need to be handled in IR/HR conversion? I don't entirely understand those.

cwgoes:
Parser fixed, thanks @mariari.

src/Juvix/Core.hs
```diff
@@ -1,12 +1,126 @@
...
-import Juvix.Core.SemiRing
+import qualified Juvix.Core.HR as HR
+import qualified Juvix.Core.IR as IR
+import Juvix.Core.Usage
+import Juvix.Core.Utility
+import Juvix.Library
+
+-- contract: no shadowing
+-- TODO - handle this automatically by renaming shadowed vars
+hrToIR ∷ HR.Term primTy primVal → IR.Term primTy primVal
```
mariari: Hmm these changes should probably be added to a `Core/Translate.hs` or something like that

src/Juvix/Core/Erasure.hs
```diff
@@ -3,27 +3,28 @@ module Juvix.Core.Erasure
...
 erase' cterm =
   let (term, env) = exec (erase cterm)
-   in (term, typeAssignment env)
+   in term >>| \term →
+        (term, typeAssignment env)
 
-exec ∷ EnvErasure a → (a, Env)
-exec (EnvEra env) = runState env (Env empty 0 [])
+exec ∷ EnvErasure a → (Either ErasureError a, Env)
+exec (EnvEra env) = runState (runExceptT env) (Env empty 0 [])
```
mariari: Good change, I suspect a lot of our effects that lack `runExceptT` will slowly gain them over time

src/Juvix/Core/HR/Parser.hs
```diff
@@ -0,0 +1,134 @@
...
+baseReservedOpNames ∷ [String]
+baseReservedOpNames =
+  [ "\\", -- lambda
+    "@", -- TODO: remove me, necessary for annotation parsing at the moment
+    ":", -- type & usage annotation
+    "->" -- arrow
+  ]
+
+generateParser ∷ ∀ primTy primVal. Parameterisation primTy primVal → (String → Maybe (Term primTy primVal))
+generateParser parameterisation =
```
mariari: Rather hard to read due to how the code formatter handles this. 

However in the future, we should fix our code formatter to add newlines if it sees ∷ in a let block.

src/Juvix/Core/Parameterisations/Naturals.hs
```diff
@@ -0,0 +1,38 @@
...
+parseVal lexer = Natural . fromIntegral |<< Token.natural lexer
+
+reservedNames ∷ [String]
+reservedNames = ["Nat"]
+
+reservedOpNames ∷ [String]
+reservedOpNames = []
+
+naturals ∷ Parameterisation NatTy NatVal
+naturals = Parameterisation typeOf apply parseTy parseVal reservedNames reservedOpNames
```
mariari: Seems like a fine way of doing parameterization, What else could we parameterize over here?

cwgoes: In the future this will include backend-specific translations, e.g. for LLVM.

src/Juvix/Core.hs
cwgoes: Agreed - moved them there.

PR #146, opened 2019-10-29
updated docker image to properly cover llvm

mariari:
At some point rebuild, I forgot the last rm, so the current image being uploaded is about 34~ GB too big

This is after comitting the `rm -rf` changes to the image

### DIFF
--- a/scripts/llvm-install.sh
+++ b/scripts/llvm-install.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+MIRROR="http://releases.llvm.org"
+
+# Determine architecture
+ARCH=$(dpkg --print-architecture)
+
+# Download
+echo "Downloading llvm-9.0.0.src.tar.xz"
+wget -nv -O "llvm-9.0.0.src.tar.xz"     "${MIRROR}/${LLVM_VERSION}/llvm-9.0.0.src.tar.xz"
+
+# Install
+# echo "Installing ${TARGET}"
+tar xf "llvm-9.0.0.src.tar.xz"
+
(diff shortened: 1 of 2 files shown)


scripts/llvm-install.sh
```diff
@@ -0,0 +1,37 @@
...
+tar xf "llvm-9.0.0.src.tar.xz"
+
+cd ./llvm-9.0.0.src
+
+mkdir build
+cd build
+
+cmake ../
+
+# mkdir ../../bin/opt
```
cwgoes: Should we delete this line? Is it left in as a comment?

scripts/llvm-install.sh
```diff
@@ -0,0 +1,37 @@
...
+
+cmake --build . --target install
+
+cd ../..
+
+
+rm llvm-9.0.0.src.tar.xz
+rm -rf llvm-9.0.0.src
+# Cleanup
+# rm -rf "${DOWNLOAD_FILE}" "${TARGET}/" llvm-install.sh
```
cwgoes: Should we delete this line? Is it left in as a comment?

scripts/llvm-install.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+MIRROR="http://releases.llvm.org"
+
+# Determine architecture
+ARCH=$(dpkg --print-architecture)
+
+# Download
+echo "Downloading llvm-9.0.0.src.tar.xz"
```
cwgoes: Can the version easily be a variable?

mariari: done!

PR #148, opened 2019-10-29
Thealmarty/add doc

thealmarty:
- add comments to code, more clearly referencing the language ref
- fix type checker
- fix formatter errors


doc/reference/src/core-language.pdc
```diff
@@ -228,10 +228,10 @@ Let $S$ be a set of sorts ${i, j, k}$ with a total order.
 #### Introduction rule
 
 \begin{prooftree}
-\AxiomC{$0Γ ⊢ M \overset{0}{:} ∗_i$}
-\AxiomC{$0Γ, x \overset{0}{:} M ⊢ N \overset{0}{:} ∗_i$}
+\AxiomC{$0Γ ⊢ V \overset{0}{:} ∗_i$}
```
thealmarty: M and N are function and var in the app typing rule so I thought V(ariable) and R(esult) may be less confusing?

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -134,57 +176,65 @@ cType p ii g (Lam s) ann =
...
         Right a → return (Omega, VPrimTy a)
         Left f → return (Omega, f)
--- App rule, function M applies to N
+-- App, function M applies to N (Elimination rule of dependent function types)
 iType p ii g (App m n) = do
   mTy ← iType p ii g m -- annotation of M is usage sig and Pi with pi usage.
   case mTy of
     (sig, VPi pi varTy resultTy) → do
-      cType p ii g n (pi, varTy) -- N has to be of type varTy with usage pi
+      cType p ii g n (sig <.> pi, varTy) -- N has to be of type varTy with usage sig*pi
```
thealmarty: As per lang ref, N has to be of type varTy with usage sig*pi, not pi.

doc/reference/src/core-language.pdc
cwgoes: I agree.

src/Juvix/Core/IR/Typechecker.hs
cwgoes: Right, still TBD if we have the constraint of sigma = 1.

src/Juvix/Core/Parameterisations/Naturals.hs
```diff
@@ -6,15 +6,17 @@ import Text.ParserCombinators.Parsec
...
 
+-- c: primitive constant and f: functions
 data NatVal
-  = Natural Natural
-  | Add
-  | Sub
-  | Mul
+  = Natural Natural -- c
+  | Add -- f addition
+  | Sub -- f substraction
```
cwgoes:
```suggestion
  | Sub -- f subtraction
```

PR #151, opened 2019-10-31
Thealmarty/add unit ty

thealmarty:
- Make formatter work on more files.
- Change ```naturals``` to ```nat``` for easier/more consistent naming of primitive types. The parameterisation functions should all be named as the data type without ```Ty```, e.g., NatTy's parameterisation is ```nat```, UnitTy's parameterisation is ```unit```.
- Add unit type as a primitive type.
  - add ```param``` as an input to the test generators(shouldCheck, shouldInfer, shouldEval)
  - add ```shouldParsePrim``` test generator function for testing parsing of primitive types.

May use [this](https://hackage.haskell.org/package/capability-0.2.0.0/docs/Capability-Reader.html) (as suggested by @cwgoes ) to avoid threading the parameterisation through all functions in another commit.

### DIFF
--- a/src/Juvix/Core/HR/Parser.hs
+++ b/src/Juvix/Core/HR/Parser.hs
@@ -1,134 +1,137 @@
+{-# LANGUAGE ExplicitForAll #-}
+
 module Juvix.Core.HR.Parser where
 
-import Data.Functor.Identity
-import Juvix.Core.HR.Types
-import Juvix.Core.Types
-import Juvix.Core.Usage
-import Juvix.Library hiding ((<|>), try)
-import Text.Parsec hiding (try)
-import Text.ParserCombinators.Parsec
-import Text.ParserCombinators.Parsec.Expr
-import Text.ParserCombinators.Parsec.Language hiding (reservedNames, reservedOpNames)
-import qualified Text.ParserCombinators.Parsec.Token as Token
-import Prelude (String)
+import           Data.Functor.Identity
+import           Juvix.Core.HR.Types
+import           Juvix.Core.Types
+import           Juvix.Core.Usage
+import           Juvix.Library                          hiding (try, (<|>))
+import           Prelude                                (String)
+import           Text.Parsec                            hiding (try)
+import           Text.ParserCombinators.Parsec
+import           Text.ParserCombinators.Parsec.Expr
+import           Text.ParserCombinators.Parsec.Language hiding (reservedNames,
+                                                         reservedOpNames)
+import qualified Text.ParserCombinators.Parsec.Token    as Token
 
-baseReservedNames ∷ [String]
+baseReservedNames :: [String]
 baseReservedNames =
-  [ "*", -- sort
-    "[Π]", -- function type
-    "w" -- omega
+  [ "*" -- sort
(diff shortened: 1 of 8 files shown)


src/Juvix/Core/HR/Parser.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE ExplicitForAll #-}
```
thealmarty: To fix formatter error.

cwgoes: Hmm odd, did you try adding it to `package.yaml` under `default-extensions:` ?

cwgoes: Ah I would try uninstalling `stylish-haskell`, I think the new formatter supersedes it anyways.

thealmarty: OK yes I've verified that it is indeed the case. So I removed all the pragmas on top.

src/Juvix/Core/Parameterisations/Unit.hs
```diff
@@ -0,0 +1,50 @@
...
+
+typeOf ∷ ∀ a. ([UnitTy] → a) → UnitVal → Either a UnitTy
+typeOf _ Unit = Right TUnit
+
+apply ∷ UnitVal → UnitVal → Maybe UnitVal
+apply _ _ = Nothing
+
+parseTy ∷ Token.GenTokenParser String () Identity → Parser UnitTy
+parseTy lexer = do
+  Token.reserved lexer "TUnit"
```
cwgoes: `()` is the common shorthand, it's not a big deal though.

src/Juvix/Core/Parameterisations/Unit.hs
```diff
@@ -0,0 +1,50 @@
...
+  pure TUnit
+
+parseVal ∷ Token.GenTokenParser String () Identity → Parser UnitVal
+parseVal lexer = do
+  Token.reserved lexer "Unit"
```
cwgoes: `()` is the common shorthand, it's not a big deal though.

test/CoreTypechecker.hs
```diff
@@ -1,142 +1,176 @@
+{-# LANGUAGE ExplicitForAll #-}
```
cwgoes: I don't think you need this here either.

src/Juvix/Core/Parameterisations/Unit.hs
thealmarty: They can't both be ```()``` so I've changed the type one to be "Unit" and the value to be "()". But the value one's test doesn't pass atm. Feel free to fix it and merge or I can fix it tomorrow.

test/CoreTypechecker.hs
thealmarty: Sorry I missed this. Fixed.

src/Juvix/Core/Parameterisations/Unit.hs
mariari: Fixed it, was a very minor change in the order in which things are parsed.

See #154 for the exact details

PR #173, opened 2019-11-12
WIP: Work to complete the LLVM graph operations

mariari:


src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -29,46 +33,58 @@ link ∷
...
+    HasState "count" Word m,
+    HasState "currentBlock" Name.Name m,
+    HasState "moduleDefinitions" [AST.Definition] m,
+    HasState "names" Names m,
+    HasState "symtab" (Map.HashMap Symbol Operand.Operand) m,
+    HasState "typTab" TypeTable m,
+    HasState "varTab" VariantToType m
+  ) ⇒
+  m Operand.Operand
+isBothPrimary = Block.defineFunction Type.i1 "is_both_primary" args $
```
cwgoes: What is this function doing? Checking if two ports are a primary pair?

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -29,46 +33,58 @@ link ∷
...
+  do
+    -- TODO ∷ should this call be abstracted somewhere?!
+    -- Why should Ι allocate for every port?!
+    mainPort ← allocaNumPortsStatic False (Operand.ConstantOperand (C.Int 32 0))
```
cwgoes: Does this just declare a local variable?

src/Juvix/Backends/LLVM/Net/EAC.hs
```diff
@@ -0,0 +1,57 @@
...
+-- Reduction rules
+--------------------------------------------------------------------------------
+
+-- TODO ∷ Maybe add metadata at some point?
+
+-- These functions work off the nodeType signature not Types.eac
+
+-- mimic rules from the interpreter
+-- This rule applies to Application ↔ Lambda
+anihilateRewireAux = Codegen.defineFunction Type.void "anihilate_rewire_aux" args $
```
cwgoes:
```suggestion
annihilateRewireAux = Codegen.defineFunction Type.void "annihilate_rewire_aux" args $
```

src/Juvix/Backends/LLVM/Net/EAC.hs
```diff
@@ -0,0 +1,57 @@
...
+
+--------------------------------------------------------------------------------
+-- Reduction rules
+--------------------------------------------------------------------------------
+
+-- TODO ∷ Maybe add metadata at some point?
+
+-- These functions work off the nodeType signature not Types.eac
+
+-- mimic rules from the interpreter
```
cwgoes: I wonder if we can split these out neatly into constructor/duplicator rules. If not, it's OK, but it'd be nice.

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -29,46 +33,58 @@ link ∷
...
+    -- TODO ∷ Make sure findEdge is in the environment
+    edge ← Block.externf "find_edge"
+    nodePtr ← Block.externf "node_ptr"
+    node ← load nodeType nodePtr
```
cwgoes: Are we loading the node data here? Do we need to do that to read the port address?

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -29,46 +33,58 @@ link ∷
...
+    port ← call portType edge (Block.emptyArgs [node, mainPort])
+    otherNodePtr ← getElementPtr $
+      Types.Minimal
+        { Types.type' = nodePointer,
+          Types.address' = port,
+          Types.indincies' = Block.constant32List [0, 1]
+        }
+    -- convert ptrs to ints
```
cwgoes: Hopefully this is a zero-cost cast under the hood.

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: This checks if two ports are connected via their primary pair

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: This allocs the main port, which is the 0th index of the port array

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: If the pointer is the same size as the int here, it's a 0 cost cast!

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: we need to load the node, as find_edge requires a nodeType and not a nodePointer

src/Juvix/Backends/LLVM/Net/EAC.hs
mariari: the rules from the original interpreter?

cwgoes: Not exactly, just sharing common logic where applicable.

cwgoes:
Merged, follow-up later (ref https://github.com/cryptiumlabs/juvix/issues/189)

PR #180, opened 2019-11-16
Erasure contd.

cwgoes:
Closes https://github.com/cryptiumlabs/juvix/issues/160

This seems to work but it is a bit messy. We'll need to either fuse the core typechecker & erasure in order to avoid doing duplicate typechecking work here, or pass along type information for subterms somehow.

### DIFF
--- a/src/Juvix/Core/Erasure/Algorithm.hs
+++ b/src/Juvix/Core/Erasure/Algorithm.hs
@@ -64,7 +64,7 @@ eraseTerm parameterisation term usage ty =
         body ← eraseTerm parameterisation body bodyUsage retTyBody
         -- If argument is not used, just return the erased body.
         -- Otherwise, if argument is used, return a lambda function.
-        pure (if usage <> argUsage == Core.SNat 0 then body else Erased.Lam name body)
+        pure (if usage <.> argUsage == Core.SNat 0 then body else Erased.Lam name body)
       Core.Elim elim →
         case elim of
           Core.Var n → pure (Erased.Var n)
--- a/src/Juvix/Core/Erasure/Algorithm.hs
+++ b/src/Juvix/Core/Erasure/Algorithm.hs
@@ -6,12 +6,21 @@ where
 import qualified Juvix.Core.Erased as Erased
 import Juvix.Core.Erasure.Types
 import qualified Juvix.Core.HR.Types as Core
+import qualified Juvix.Core.IR as IR
+import Juvix.Core.Translate (hrToIR, irToHR)
 import qualified Juvix.Core.Types as Core
 import qualified Juvix.Core.Usage as Core
 import Juvix.Library hiding (empty)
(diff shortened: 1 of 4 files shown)


src/Juvix/Core/Erasure/Types.hs
```diff
@@ -1,32 +1,38 @@
...
   = Env
       { typeAssignment ∷ Erased.TypeAssignment primTy,
+        context ∷ IR.Context primTy primVal,
         nextName ∷ Int,
         nameStack ∷ [Int]
       }
   deriving (Show, Eq, Generic)
 
-newtype EnvErasure primTy a = EnvEra (ExceptT ErasureError (State (Env primTy)) a)
+newtype EnvErasure primTy primVal a = EnvEra (ExceptT ErasureError (State (Env primTy primVal)) a)
```
mariari: we should probably eventually write a type alias for the main type here so we don't repeat it for every deriving via.

(Template haskell may be worth looking into for this)

src/Juvix/Core/Erasure/Algorithm.hs
```diff
@@ -40,28 +54,35 @@ eraseTerm parameterisation term usage ty =
...
-            x ← eraseTerm parameterisation x undefined undefined
-            pure (Erased.App f x)
+            let IR.Elim fIR = hrToIR (Core.Elim f)
+            context ← get @"context"
+            case IR.iType0 parameterisation context fIR of
+              Left err → throw @"erasureError" (InternalError (show err <> " while attempting to erase " <> show f))
+              Right (fUsage, fTy) → do
+                let fty@(Core.Pi argUsage fArgTy _) = irToHR (IR.quote0 fTy)
+                f ← eraseTerm parameterisation (Core.Elim f) fUsage fty
+                if argUsage == Core.SNat 0
```
mariari: changes seem fine, however I'm curious about this if statement

cwgoes: See the language reference, page 30:

2019-11-18-133420_459x77_scrot

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -175,7 +175,7 @@ cType p ii g (Lam s) ann =
         ((Local ii, (sig <.> pi, ty)) : g) -- put s in the context with usage sig*pi
         (cSubst 0 (Free (Local ii)) s) -- x (varType) in context S with sigma*pi usage.
         (sig, ty' (vfree (Local ii))) -- is of type M (usage sigma) in context T
-    _ → throwError $ show (snd ann) <> " is not a function type but should be."
+    _ → throwError $ show (snd ann) <> " is not a function type but should be - while checking " <> show (Lam s)
```
thealmarty: Awesome!

cwgoes:
Follow-up tasks to https://github.com/cryptiumlabs/juvix/issues/190

PR #183, opened 2019-11-18
Interaction net updates

cwgoes:
Closes https://github.com/cryptiumlabs/juvix/issues/158

### DIFF
--- a/src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
+++ b/src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
@@ -53,6 +53,7 @@ data Primar primVal
   | Fals
   | IntLit Int
   | Symbol Symbol
+  | PrimVal primVal
   deriving (Show)
 
 data ProperPort primVal
--- a/src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
+++ b/src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
@@ -27,7 +27,7 @@ data Auxiliary3 primVal
   | Curried3 (Primitive → Primitive → Primitive → Maybe Primitive)
   deriving (Show)
 
-data Auxiliary2 primval
+data Auxiliary2 primVal
   = Or
   | And
   | Cons
@@ -35,6 +35,7 @@ data Auxiliary2 primval
   | Lambda
   | App
   | FanIn Int
+  | PrimCurried2 (primVal → primVal → Maybe primVal)
   | Curried2 (Primitive → Primitive → Maybe Primitive)
   deriving (Show)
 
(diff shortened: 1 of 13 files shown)


cwgoes:
@mariari I'll add some more testcases, but want to quickly check if this is the right approach. We should probably have the interpreter mirror the LLVM backend as closely as possible (ideally we can even share a lot of the intermediate construction code).

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
```diff
@@ -563,9 +596,25 @@ curry2 (nodeF, numNode) (p, numPrim) = do
...
+  incGraphSizeStep (- 1)
+  curr ← newNode (Auxiliary1 $ PrimCurried1 $ nodeF p)
+  let currNode = RELAuxiliary1
+        { node = curr,
+          primary = ReLink numNode Aux2,
+          auxiliary1 = ReLink numNode Aux1
+        }
+  linkAll currNode
+  deleteRewire [numNode, numPrim] [curr]
+
```
mariari: What is the difference between this and curry2?

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
```diff
@@ -587,8 +636,28 @@ curry1 (nodeF, numNode) (p, numPrim) =
...
+      incGraphSizeStep (- 1)
+      curr ← newNode (Primar node)
+      let currNode = RELAuxiliary0
+            { node = curr,
+              primary = ReLink numNode Aux1
+            }
+      linkAll currNode
+      deleteRewire [numNode, numPrim] [curr]
+      pure $ True
+
```
mariari: Same for this vs curry1

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
cwgoes: They should be de-duplicated, we'll get rid of `Primitive`.

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
cwgoes: Ditto on the answer.

cwgoes:
Follow-up issue: https://github.com/cryptiumlabs/juvix/issues/184

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
cwgoes: Noted in https://github.com/cryptiumlabs/juvix/issues/184.

src/Juvix/Interpreter/InteractionNet/Nets/Default.hs
cwgoes: Noted in https://github.com/cryptiumlabs/juvix/issues/184.

Issue #189, opened 2019-11-18
LLVM-related tasks


cwgoes:
- [ ] Try converting interpreted inet graph representation to LLVM net @cwgoes 
- [x] Figure out `malloc` / `free` (must be linked, probably) @cwgoes 
- [x] Figure out how to JIT the code @cwgoes 
    - [x] Issue with kaleidoscope: only works with `main`
    - [x] Also we want to actually output the IR sometimes (but the JIT can do this!)
    - [x] Problems with optimisations (removes `externf` usage?)
- [x] Finish rewrite rules @mariari 
- [x] Write evaluation algorithm (start with the simple sequential one from Asperti) @mariari 
- [ ] Write read-back algorithm (ideally we can share code with interpreter...) @mariari 
       Could convert to graph (list of nodes, ports) in LLVM then read-back in Haskell
- [x] Write the environment for code generation @mariari 
- [ ] Is there execution metering? Probably a future research topic. e.g. [this](https://github.com/jwhitham/x86determiniser)
- [ ] Find the x86 assembly for the critical path functions and *optimise* it
- [x] Test it! @mariari @cwgoes

cwgoes:
Various inspirations:
- [Helium](https://github.com/Helium4Haskell/helium)
- [LHC](https://github.com/Lemmih/lhc)

cwgoes:
In the future we might want to use [OrcJIT](https://hackage.haskell.org/package/llvm-hs-9.0.1/docs/LLVM-OrcJIT.html) (very configurable) but the simple one should be fine for now.

cwgoes:
I thought a bit more about converting from Haskell graphs to LLVM. I think we do *not* want to embed the structure of the graph itself into the generated code in any way - the generated code should just be all the reduction logic plus bespoke encodings, in a way that can easily be saved as a sort of "config" to generate the same LLVM code later.

Once we have generated that LLVM code and JIT-ed it into a module, we want some FFI API between the module and Haskell to pass data back and forth. This FFI will need approximately the following calls:

- `createNet : [Node] -> IO ()`
- `readNet : IO [Node]`
- `saveState : IO StateHandle`
- `loadState : StateHandle -> IO ()`
- `appendToNet : [Node] -> IO ()`
- `reduceUntilComplete : IO ReductionStats`

[...]

PR #193, opened 2019-11-19
LLVM updates

cwgoes:
Ref #189 

- [x] Adds JIT functionality for LLVM modules
- [x] Adds external linkage for `malloc` / `free`
- [x] Tests JIT, including of a function, and `malloc` / `free` usage


cwgoes:
Hmm @mariari does this pass for you locally? It does for me. Wonder why CI fails.

mariari:
The tests compile for me, and even all the tests pass, except for the michelson which relies on the OPTIMIZATION c++ pragma

src/Juvix/Backends/LLVM/JIT/Execution.hs
```diff
@@ -0,0 +1,65 @@
...
+passes config = defaultCuratedPassSetSpec {optLevel = convOptLevel (configOptimisationLevel config)}
+
+convOptLevel ∷ OptimisationLevel → Maybe Word
+convOptLevel None = Nothing
+convOptLevel O0 = pure 0
+convOptLevel O1 = pure 1
+convOptLevel O2 = pure 2
+convOptLevel O3 = pure 3
+
+-- Note: in order to allow this to return functions, a green thread is forked to retain module
```
mariari: return functions?

As in, return newely defined functions that are used at the top level?

test/Backends/LLVM.hs
```diff
@@ -0,0 +1,127 @@
...
+import qualified Test.Tasty as T
+import qualified Test.Tasty.HUnit as T
+
+test_malloc_free_jit ∷ T.TestTree
+test_malloc_free_jit = T.testCase "malloc free module should jit" $ do
+  (fn ∷ Word32 → IO Word32) ← jit (Config None) mallocFreeModule "test"
+  res ← fn 7
+  43 T.@=? res
+
+mallocFreeModule ∷ LLVM.AST.Module
```
mariari: Should probably test the functions that generate this Module as well

test/Backends/LLVM.hs
```diff
@@ -0,0 +1,127 @@
...
+                            ),
+                        callingConvention = CC.C,
+                        returnAttributes = [],
+                        arguments = [(LocalReference voidStarTy (UnName 1), [])],
+                        functionAttributes = [],
+                        metadata = []
+                      }
+                  ]
+                  ( Do $ Ret (Just (ConstantOperand (C.Int 32 43))) []
+                  )
```
mariari: Would be nice to load the result somewhere, and return it in the final Ret call. Would be nice to show off since malloc gives us backa  Void Pointer.

src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -273,6 +275,58 @@ terminator trm = do
...
+
+defineMalloc ∷
+  ∀ m.
+  ( HasState "moduleDefinitions" [Definition] m,
+    HasState "symtab" (HashMap Symbol Operand) m
+  ) ⇒
+  m ()
+defineMalloc = do
+  let name = "malloc"
+  op ← external voidStarTy name [(size_t, "size")]
```
mariari: How much can we change the size of the pointer?

mariari:
It seems that on subsequent builds, it just raises all my machine's cores to 100%. Then it spams the terminal with `error 11: full-pipe` if i run it some more times after this.

src/Juvix/Backends/LLVM/JIT/Execution.hs
cwgoes: Return Haskell functions which can then be passed around and used anywhere, yes.

test/Backends/LLVM.hs
cwgoes: Yeah - is there a way to run the `Codegen` monad?

src/Juvix/Backends/LLVM/Codegen/Block.hs
cwgoes: I think it depends on the library we're linking against (that provides `malloc` / `free`)

cwgoes: (for now we can just worry about 64-bit systems)

cwgoes:
I really haven't had success debugging; I just disabled the automatic tests for now (they can still be run manually). Maybe there are some implementation details of the LLVM bindings which do not permit parallel execution (which the test framework does).

cwgoes:
Follow-up testing issue: https://github.com/cryptiumlabs/juvix/issues/208

Issue #191, opened 2019-11-18
Deal with omega / natural usage comparisions


cwgoes:
Right now we use an `Eq` instance, this is a bit confusing and might lead to mistakes.

I think comparing on an ordering might be better, and we should review where exactly in the typechecker we are doing this.

thealmarty:
> Right now we use an `Eq` instance, this is a bit confusing and might lead to mistakes.
> 
> I think comparing on an ordering might be better, and we should review where exactly in the typechecker we are doing this.

As per the language reference, currently usages are exact, so `Eq` instance is required. Do we want to change the spec to *less than or equal to* usages? We can avoid using `Eq` instance then. That is:

0 usage implies no computational usage.
1 usage implies 1 or 0 computational usage.
n usage implies n or less computational uage.
w usage implies any number of usage (unchanged from current spec)

thealmarty:
Actually that still wouldn't solve the issue that checking `== SNat 0` (see if a variable was zero-usage) returns true for Omega, because Omega is defined as any number of usage.

cwgoes:
> Do we want to change the spec to less than or equal to usages?

I think we want exact usages, without weakening (especially with the upcoming first-class usage type theory upgrade) - but there will be some less-than-or-equal check somewhere for `w`, insofar as we are counting usages (since `w` just means 'any number of usages', and we can't ever use a variable `w` times).

Where do we use the `==` instance in the typechecker? It looks like:

[line 134](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Typechecker.hs#L134)
[line 147](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Typechecker.hs#L134)
[line 163](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Typechecker.hs#L163)
[line 183](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Typechecker.hs#L183)

In the first three cases, I think we do want to be checking that the usage is actually `0` (so the current logic is wrong - e.g. `* 0 : w * 1` is not a valid judgement, but it would be accepted). In the last case, equality might be OK, or we might allow `(x : w T) : 3 T` to be admitted, but I don't think the type theory does at the moment, so let's leave it out unless we decide that we need it later (note: that's *not* weakening, weakening would be to admit `(x : 3 T) : 2 T`, which I think we don't want.

[...]

PR #196, opened 2019-11-20
Thealmarty/fix usage eq instance

thealmarty:
- new `Eq` instance as per #191 tests fail still to be investigated. 
- change the Erasure code to use equality instead of case comparison - @cwgoes could you please do the check you were doing again to see check that the behaviour of this new `Eq` instance is desired.


thealmarty:
Closing as per discussion in #191

PR #198, opened 2019-11-21
Continuation of the previous LLVM Work

mariari:
Works to complete #189.

### DIFF
--- a/src/Juvix/Backends/LLVM/Codegen/Graph.hs
+++ b/src/Juvix/Backends/LLVM/Codegen/Graph.hs
@@ -65,7 +65,7 @@ isBothPrimary = Block.defineFunction Type.i1 "is_both_primary" args $
   do
     -- TODO ∷ should this call be abstracted somewhere?!
     -- Why should Ι allocate for every port?!
-    mainPort ← allocaNumPortsStatic False (Operand.ConstantOperand (C.Int 32 0))
+    mainPort ← mainPort
     -- TODO ∷ Make sure findEdge is in the environment
     edge ← Block.externf "find_edge"
     nodePtr ← Block.externf "node_ptr"
@@ -197,7 +197,6 @@ allocaGenH mPortData type' = do
   pure ports
   where
     len = length mPortData
-
 
 allocaPortsH,
   allocaDataH ∷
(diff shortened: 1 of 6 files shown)


cwgoes:
JIT & `malloc` / `free` are up (with tests) in https://github.com/cryptiumlabs/juvix/pull/193

What is immediately blocking next - the graph representation / conversion?

mariari:
> at is immediately blocking next - the graph representation / conversion?

I believe so, once we have that, we can try hooking all these up and start some basic graph testing.

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -61,28 +61,33 @@ isBothPrimary ∷
...
   do
     -- TODO ∷ should this call be abstracted somewhere?!
     -- Why should Ι allocate for every port?!
-    mainPort ← allocaNumPortsStatic False (Operand.ConstantOperand (C.Int 32 0))
+    mainPort ← mainPort
     -- TODO ∷ Make sure findEdge is in the environment
     edge ← Block.externf "find_edge"
     nodePtr ← Block.externf "node_ptr"
     node ← load nodeType nodePtr
     port ← call portType edge (Block.emptyArgs [node, mainPort])
```
cwgoes: Are we reading into `mainPort` here (pointer receiver)?

It seems like we're allocating an (empty?) port then using it as a parameter, but maybe I'm misreading.

src/Juvix/Backends/LLVM/Net/EAC.hs
```diff
@@ -1,33 +1,193 @@
...
+    lamCase ← Codegen.addBlock "switch.lam"
+    eraCase ← Codegen.addBlock "switch.era"
+    dupCase ← Codegen.addBlock "switch.dup"
+    defCase ← Codegen.addBlock "switch.default"
+    extCase ← Codegen.addBlock "switch.exit"
+    car ← Types.loadCar eacList
+    cdr ← Types.loadCdr eacList
+    tag ← tagOf car
+    _term ←
+      Codegen.switch
```
cwgoes: I wonder if we can do a clever thing where we add the tags with an offset (shift) and have only one case switch - not sure if it would really be faster though.

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: mainPort is just an alias for the allocaNumPortStatic call.

So it's just a constantOperand, that allocates the pointer if it needs to be a pointer (it doesn't here)

src/Juvix/Backends/LLVM/Net/EAC.hs
mariari: that could be faster, though I'm usnure how LLVM optimizes that

cwgoes: yeah let's wait on that sort of thing for now

PR #200, opened 2019-11-21
Thealmarty/add allowsUsageOf function

### LINKED ISSUE #191: Deal with omega / natural usage comparisions

thealmarty:
Closes #191:

- removed old custom `Eq` instance for usage.
- added a function is called `allowsUsageOf` for checking usage compatibility.
- an `Eq` instance of usage is *derived* for comparison of `Term` and `Value` or any other exact comparisons e.g., when checking whether usage is zero.
- fixed type checker and all tests pass
- [this](https://github.com/cryptiumlabs/juvix/pull/185/commits/68905877256b9954ddd67e4b4c032ff99f09dc53) commit can be safely reverted (if desired)?

### DIFF
--- a/scripts/precommit.sh
+++ b/scripts/precommit.sh
@@ -7,6 +7,6 @@ make format
 # disabled for now
 # make lint
 
-#cd scripts && ./run-generate.lisp
+cd scripts && sbcl --script "./run-generate.lisp"
 
 exit 0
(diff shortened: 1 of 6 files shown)


cwgoes:
> this commit can be safely reverted (if desired)?

Yup, done (just pushed to this branch).

PR #203, opened 2019-11-22
Thealmarty/add more tests

thealmarty:
Add `depIdentity` test.

Currently (at [this](https://github.com/cryptiumlabs/juvix/pull/203/commits/ec40d2dd0b258cb6ba2111f13f631665c61d06f0) commit), two tests fail:

- `test_depIdentity`'s fail messages circle between `Local 0` and `Local 1`, but I think `Local 0` is correct.
- `test_k_app_I` fails in the way the expected and got type are the same.


thealmarty:
OK I fixed `test_depIdentity`!

thealmarty:
Added tests: 
- `test_depIdentity`
- `test_k_app_I`
- `test_kFun_app_1`

test/CoreTypechecker.hs
```diff
@@ -64,19 +64,21 @@ depIdentity =
...
-      (IR.VStar 0)
-      ( IR.vapp
-          All.all
-          (IR.VLam (IR.VPrimTy))
+      (IR.VStar 0) -- first input is of type type
+      ( const
+          ( IR.VPi
+              (SNat 1)
+              (IR.VNeutral (IR.NFree (IR.Local 0))) -- second input is of type of the first input
+              (const (IR.VNeutral (IR.NFree (IR.Local 0)))) -- the function's return is of the type that's the same as the second input
```
cwgoes: I understand the code but not the comment. This is just the type we bound in the first input, right?

test/CoreTypechecker.hs
```diff
@@ -263,6 +294,9 @@ kAppICompTy =
 -- ((Nat -> Nat -> Nat) -> (Nat -> Nat -> Nat) -> (Nat -> Nat -> Nat)) ->
 -- ((Nat -> Nat -> Nat) -> Nat -> Nat -> Nat)
 -- (Nat -> Nat -> Nat)
+
+-- S (K11) = (K 1) (1 1) doesn't work because one cannot properly annotate the second input of K (1 1).
```
cwgoes: `S` in general might require first-class usages (luckily, we're working on that!)

However, it should be possible to write *specialised* versions of `S` for certain usages (but any type) right now.

thealmarty: Woohoo first-class usages!

I'm not sure how to annotate 

1 applied to 1

but I have to annotate that because I have to annotate the inputs to `K` and `(1 1)` is the second input to `K`. Any idea?

test/CoreTypechecker.hs
thealmarty: Yes, it is of the type of the first input. But it's not the first input, because the first input is the type's type, i.e., `* 0`. 
 
The dependent identity is 

`\t.\x.x:t` 

the first input is `t`, the type of `x`, the second input is `x`, of type `t`. The return is `x`, of type `t`.

So the return type is the second input. 

Did I answer your question?

test/CoreTypechecker.hs
cwgoes: Ah I see - in that case, the first `1` is `((a -> a) -> (a -> a)) -> (a -> a) -> (a -> a)` and the second is `(a -> a) -> a -> a` (like usual) - the `a` in the first is just `a -> a`.

test/CoreTypechecker.hs
cwgoes: Maybe I'm just mixing up the indices, but I'm still confused.

Per my counting:

`\t.\x:t.x:t` - `t` is the first input (a type, with value `t` and type `* 0` itself), `x` is the second input (with value `x` and type `t`), and the return *value* is `x`, which is of type `t`, which is the first input.

Pi should bind the *value* of the first input - which is the type `t` - not the *type* of the first input `* 0` (we don't care about that, I don't think it should be in scope at all).

thealmarty: > Maybe I'm just mixing up the indices, but I'm still confused.
> 

The bound variables are in de bruijn indices, so in line 76, the index is 0 because `t` is the input bounded immediately before. In line 77, the input immediately before is `x:t` so again the index is 0.

> Per my counting:
> 
> `\t.\x:t.x:t` - `t` is the first input (a type, with value `t` and type `* 0` itself), `x` is the second input (with value `x` and type `t`), and the return _value_ is `x`, which is of type `t`, which (`t`) is the value of the first input.
> 
Yes that's the idea.

> Pi should bind the _value_ of the first input - which is the type `t` - not the _type_ of the first input `* 0` (we don't care about that, I don't think it should be in scope at all).

This is the annotation for the dependent identity. So these are all types.

Am I making sense?

test/CoreTypechecker.hs
thealmarty: Yes, I could try Church-encoded numbers. I can also test `S K I 1`.

cwgoes: Ah sorry, I thought you were trying Church-encoded numbers - `1 1` with naturals doesn't make too much sense - `S K I` would be good to test.

test/CoreTypechecker.hs
cwgoes: They are all types, yes, but whenever we type-check a dependent (pi) type, de-Bruijin index 0 should be the *value* of the function argument, right?

Then this pi type seems like it's returning a value of type `x`, which is not what we want - we want to return a value of type `t`, which should be bound at de-Bruijin index 1, since it was the argument to the outer lambda.

In non-de-Bruijin indicies, the dependent identity should be:
`\t : * 0 -> \x : t -> (x : t)` with type `pi (t : * 0) . pi (x : t) . t`.
That term should translate to de-Bruijin as `\.\. (0 : 1)`, right?

Do you see my confusion?

Perhaps it would be helpful to list out the indices when type-checking different sub-terms.

thealmarty: > In non-de-Bruijin indicies, the dependent identity should be:
> `\t : * 0 -> \x : t -> (x : t)` with type `pi (t : * 0) . pi (x : t) . t`.
> That term should translate to de-Bruijin as `\.\. (0 : 1)`, right? That's what you have on line 55. So why is the type on line 77 de-Bruijin index 0 (`x`) instead of de-Bruijin index 1 (`t`)?
> Do you see my confusion?

I see. Line 62 is the annotation for line 61. It says that the type of the output (`x`) is the first input.

For the annotation on lines 67-77, the type checker knows that for the identity function, the output's type has to be the same as the second input. (Hence on line 77, it's index 0) But that's fine because the type of `x` is bounded to the first input's value (on line 76).

Did that answer your question?

thealmarty: It seems the type checker as is annotates such that the return value takes the input as is, not the input's value. I.e., on L77, it takes in the second input, `Nat`, although on L76, it takes in the first input's value.

Will add the following tests to experiment further: 

\t1.\t2.\(x : t1).x
\t1.\t2.\(x : t2).x
\t.\(x:t).\(y:t).x
and \t.\(x:t).\(y:t).y

Will also look into other languages to see if this is the correct/desired type checker behaviour.

thealmarty: I have a better way of explaining why it is correct that the return is the type of `x`, not `x`:

In the annotation all we ever take in is the type of `x`, i.e., `t`. We do not take in the value of `x`. The second inputs is annotated as the first input (i.e., the type of `x`, `t`). And the return type is bound to the second input's value, in this case `t`.

cwgoes: The annotation doesn't contain the value of `x`, yes - it doesn't contain the value of `t` either - but it should express that the identity function should take *some* t as the first input, an x of type *t* as the second input, and return something of type *t*; it can refer to the values in the abstract (then the typechecker will fill them in when we check the annotation against an actual term). The only thing the annotation can refer to are the values of previous inputs (even though it doesn't know what they are yet) - this is how the value of *t* is referred to when specifying the type of the argument `x` - since values, not types, are substituted in when performing application:

2019-11-27-080659_761x268_scrot


The value of the first input, `t`, is the type of the second input and the type of the return value. We don't want to refer to "the type of x" anywhere, annotations shouldn't bind types of anything (when we check the annotation, we should check that the type of the body - `(x : t)` - is `t`, which it should be).

cwgoes: So, this does pass:

```haskell
depIdentityCompTy ∷ ∀ primTy primVal. IR.Annotation primTy primVal
depIdentityCompTy =
  ( SNat 0,
    IR.VPi
      (SNat 0)
      (IR.VStar 0) -- first input is of type type
      (\t ->
          ( IR.VPi
              (SNat 1)
              (IR.VNeutral (IR.NFree (IR.Local 0))) -- second input is of type of the first input
              (const t)
          )   
      )   
  )
```

Why do we have both the lambda arguments and the bound indices? It seems like only one should be necessary.

cwgoes: I think @andy-morris looked at this and also thought an index was off, but I'll let him explain his analysis.

thealmarty: > 
> ```haskell
> depIdentityCompTy ∷ ∀ primTy primVal. IR.Annotation primTy primVal
> depIdentityCompTy =
>   ( SNat 0,
>     IR.VPi
>       (SNat 0)
>       (IR.VStar 0) -- first input is of type type
>       (\t ->
>           ( IR.VPi
>               (SNat 1)
>               t
>               (const t)
>           )   
>       )   
>   )
> ```
> 
> And this fails, which I would also expect:
> 
> ```haskell
> depIdentityCompTy ∷ ∀ primTy primVal. IR.Annotation primTy primVal
> depIdentityCompTy =
>   ( SNat 0,
>     IR.VPi
>       (SNat 0)
>       (IR.VStar 0) -- first input is of type type
>       (\t ->
>           ( IR.VPi
>               (SNat 1)
>               t
>               (\x -> x)
>           )   
>       )   
>   )
> ```
> 

In general I don't think we should mix Haskell's `\x` function into my type checker DSL. The above inputs would never be passed in in practice because the type checker wouldn't generate such annotations.

test/CoreTypechecker.hs
```diff
@@ -263,48 +295,63 @@ kAppICompTy =
...
-                            (IR.Bound 2)
-                            () -- Annotation of x
+
+-- SKI1 = K 1 (I 1)
+-- the first input K takes inputs 1 and 1 so has type signature of Nat -> Nat -> Nat
+-- the second input I has type signature Nat -> Nat
+-- the third input is Nat
+-- type signature of this S is (Nat -> Nat -> Nat) -> (Nat -> Nat) -> Nat -> Nat
+ski1 ∷ NatTerm -- ∀ primTy primVal. IR.Term primTy primVal -- S = \x.\y.\z. (xz) (yz)
+ski1 =
```
andy-morris: It seems to me that `ski1` corresponds to just the `S`, rather than the whole `S K I 1` expression? It looks like the AST for

```
λxyz. (1 ((1 x : 1 ℕ → 0 ℕ → ℕ) z) : 0 ℕ → ℕ)
      ((1 y : 1 ℕ → ℕ) z)
```

i.e. `S`, but `test_ski1` below is asserting that this has the type `ℕ` (the type of the whole `S K I 1`).

test/CoreTypechecker.hs
```diff
@@ -204,14 +206,44 @@ kApp1 =
...
+kFunApp1CompTy =
+  ( SNat 1,
+    IR.VPi
+      (SNat 0)
+      (IR.VPi (SNat 1) (IR.VPrimTy Nat) (const (IR.VPrimTy Nat)))
+      (const (IR.VPrimTy Nat))
+  )
 
 --K: (Nat -> Nat) -> Nat -> (Nat -> Nat) I:Nat -> Nat type checks to Nat -> (Nat -> Nat)
 kAppI ∷ NatElim
```
andy-morris: This looks good, but I was wondering if you think it's worth also adding another test where the argument (L265-271) is replaced with just an unannotated `identity`? Because I think the typechecker should be able to push in the function argument's type and still work.

test/CoreTypechecker.hs
cwgoes: > In general I don't think we should mix Haskell's \x function into my type checker DSL. The above inputs would never be passed in in practice because the type checker wouldn't generate such annotations.

Hmm, are you sure? HOAS is used in value evaluation, e.g. [here](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Evaluator.hs#L27).

andy-morris: > I think @andy-morris looked at this and also thought an index was off, but I'll let him explain his analysis.

I'm not sure this little scribble will be that enlightening, but yeah, I'm pretty sure that the type should end in index `1` rather than `0`, since it has to skip over the binder for `x` itself.

(I thought I'd already posted this but it doesn't seem to have fully gone through, sorry about that)

test/CoreTypechecker.hs
thealmarty: Good idea! Added. It did pass.

thealmarty:
> One last thing that was there before this PR so I don't think github will let me attach it to the line:
> 
> On line 60, `depIdentity` has a usage of `0`; should this be `1` to match `depIdentityCompTy` below?
> 
> Anyway, the rest looks good to me.

When L60 is changed to `SNat 1`, the test does not pass. But I think this is correct behaviour? This `SNat` is annotating the usage of the type (first input), I think. Lmk if you disagree. We definitely want to make sure the type checker has the desired behaviour.

test/CoreTypechecker.hs
thealmarty: Thank you! You are right! I've been trying to annotate `s` but I think I got the levels wrong somehow. See [here](https://gist.github.com/thealmarty/f652a74ead0c84c4b8c58b7bcf54a8c6). Would like to borrow your and @cwgoes 's eyes - my eyes are kinda done - sorry I don't know how to make this easier to read. Recommendations welcome!

test/CoreTypechecker.hs
thealmarty: Nice! Type theory in color! I don't think I understand - but it'll be easier to discuss after the logging is finished - we'll do that then. See #223

thealmarty: > Hmm, are you sure? HOAS is used in value evaluation, e.g. [here](https://github.com/cryptiumlabs/juvix/blob/develop/src/Juvix/Core/IR/Evaluator.hs#L27).
> 
> I would be fine with eliminating it, but I think we should do so uniformly if we do.

You're right, `VLam` and `VPi` take functions. Perhaps the annotations should be written in that form actually.

cwgoes:
> > One last thing that was there before this PR so I don't think github will let me attach it to the line:
> > On line 60, `depIdentity` has a usage of `0`; should this be `1` to match `depIdentityCompTy` below?
> > Anyway, the rest looks good to me.
> 
> When L60 is changed to `SNat 1`, the test does not pass. But I think this is correct behaviour? This `SNat` is annotating the usage of the type (first input), I think. Lmk if you disagree. We definitely want to make sure the type checker has the desired behaviour.

Hmm, isn't L60 the annotation on the body of the function, i.e. `x : 1 t`? I think it should be `1` as well, since the type of the whole expression is `0 t -> 1 (x : t) -> 1 t`, right?

thealmarty:
> 
> Hmm, isn't L60 the annotation on the body of the function, i.e. `x : 1 t`? I think it should be `1` as well, since the type of the whole expression is `0 t -> 1 (x : t) -> 1 t`. Hopefully logs will help debug.

To clarify, the `depIdentity` *definition* (**L56-L65**) says:
         
`\t.\x.x 0: t` 

Or, `x` is of type `t` and `t` is not used in the computation of `x`.

The `depIdentity` *annotation* says very close to what you think the type of the expression should be:
`0 t -> 1 (x : t) -> t` (I think the output type shouldn't have usage? And thus it's making sense for this to pass.) I've commented on the usage lines below, or **L71** for `t` and **L75** for `x` respectively:

```haskell
depIdentityCompTy =
  ( SNat 0, 
    IR.VPi
      (SNat 0) -- usage annotation for t
      (IR.VStar 0) -- first input is of type type
      ( const
          ( IR.VPi
              (SNat 1) -- usage annotation for x
              (IR.VNeutral (IR.NFree (IR.Local 0))) -- second input is of type of the first input
              (const (IR.VNeutral (IR.NFree (IR.Local 0)))) -- output is of type of x
```

thealmarty:
I shall add comments to all the tests explaining what the definitions and annotations are.

cwgoes:
> To clarify, the `depIdentity` _definition_ (**L56-L65**) says:
> 
> `\t.\x.x 0: t`
> 
> Or, `x` is of type `t` and `t` is not used in the computation of `x`.

`x : 0 t` doesn't mean that `t` is not used in the computation of `x`, it means that `x` is of type `t` in the zero-usage fragment, i.e. `x` isn't computationally used. We want the dependent identity to take exactly `x : 1 t` as an argument and return it unmodified, so it should still have `1` usage.

I think the first component of `depIdentityCompTy` should also be `SNat 1`, we want `1` usage of the function, that may be related  - i.e.:

```haskell
depIdentityCompTy =
  ( SNat 1, 
    IR.VPi
       (...)
```

thealmarty:
> 
> `x : 0 t` doesn't mean that `t` is not used in the computation of `x`, it means that `x` is of type `t` in the zero-usage fragment, i.e. `x` isn't computationally used. We want the dependent identity to take exactly `x : 1 t` as an argument and return it unmodified, so it should still have `1` usage.
> 
> I think the first component of `depIdentityCompTy` should also be `SNat 1`, we want `1` usage of the function, that may be related - i.e.:
> 
> ```haskell
> depIdentityCompTy =
>   ( SNat 1, 
>     IR.VPi
>        (...)
> ```

I changed both as you suggested and it did pass.

thealmarty:
Sorry I don't know why I was thinking it was annotating `t`...that's right, it was annotating `x`. And I think it's good that the type checker correctly type checked (I think?) when the `depIdentity` function isn't used/annotated to zero and `x` isn't used/annotated to zero. 

I'm glad you two are looking at the tests and make sure they make sense as I requested! I'm adding more comments to make it more clear. I appreciate further questions/comments - and we'll check further with the logging as well.

thealmarty:
I have added more comments to explain more clearly what the tests are doing. 

I commented out failed tests as I would like to merge this if there is no further comments.

I will add contemplation tests in a separate PR.

PR #205, opened 2019-11-25
Notes and Comments on Frontend syntax

mariari:


doc/Frontend/ideas.org
```diff
@@ -0,0 +1,58 @@
...
+    f 2 : A
+
+    f : A ~2
+    f : ~2 A
+  #+END_SRC
+  + The first idea was rejected because it is ambiguous in some
+    situations
+  + I also don't fully like the second, as how does this work with
+    named arguments? It seems like arbitrary extra syntax
+  + Thus I would like suggestions on what this should look like.
```
cwgoes: What about just `f : 2 A`? I think it will be unambiguous if we always require a usage annotation.

cwgoes: also, there will be variables, e.g. `f : k a`, `f : (n * k + 3) a`, etc.

doc/Frontend/ideas.org
```diff
@@ -0,0 +1,58 @@
...
+  + Note :: the last arrow requires dependent useages
+
+- With sugar it would look something like this
+  #+BEGIN_SRC idris
+    f : (x : Int) -o (y : Int) -> C -(x+y)-> D
+  #+END_SRC
+  + We can see that -o stands for 1 useage, and -> stands for
+    unlimited useages
+  + Note :: In the unicode extension → can replace ->
+  + Note :: In the unicode extension  ⊸ can replace -o
```
cwgoes: I like the sugar options - what's the default though? Why not the simple:

`f : 2 (x : 3 Int) -> Double`, `f : w (x : 2 Int -> y : 1 () -> IO ())`

doc/Frontend/ideas.org
mariari: That would be fine, if we always require an annotation, but wouldn't always requiring an annotation before all arrows be noise? (I suspect many devs would want some default they can change)

doc/Frontend/ideas.org
mariari: I'm assuming most would either pick -o or -> as the default.

The issue with this, is that there is no way to alias this approach, meaning we always have to write explicit numbers, when an arrow signifier would be better.

Though I will say, this allows toplevel functions
```idris
f : 2 Int
f = 3
```

Or something along these lines

doc/Frontend/ideas.org
cwgoes: Possibly... I think the syntax should be pretty explicit by default though, and sugar can be added over time (or enabled by developers) once they're familiar with the underlying desugared behaviour.

doc/Frontend/ideas.org
cwgoes: > The issue with this, is that there is no way to alias this approach, meaning we always have to write explicit numbers, when an arrow signifier would be better.

Hmm why - couldn't we still alias the arrows? e.g. `f : 2 (x -> Double)` (if `f` is linear in `x`, but we need `2` of `f`)? 

Honestly I think I prefer the explicit usages though, it's a few more characters but easier to read since the usages are always between the variable binder and the type.

cwgoes:
(anyways, I think we can merge this as notes)

mariari:
I will write a note in the file about disagreements and which syntaxs we can take and the pros and cons of each approach.

After that part is done I will merge it

doc/Frontend/ideas.org
mariari: > > The issue with this, is that there is no way to alias this approach, meaning we always have to write explicit numbers, when an arrow signifier would be better.
> 
> Hmm why - couldn't we still alias the arrows? e.g. `f : 2 (x -> Double)` (if `f` is linear in `x`, but we need `2` of `f`)?
> 

My issue is that imagine, if we didn't have `Double` but instead `m A`. would this not be amigious with usage m?

cwgoes:
I will add my thoughts later.

Issue #209, opened 2019-11-27
Typechecker "debug mode"


cwgoes:
So that we can better understand the typechecking process for a term, and ensure that all steps match up with the type theory specified in the language reference (and that we understand where precisely they are being applied), I suggest that we implement a "debug mode" in the typechecker where we write a log entry each time a rule from the language reference is applied, and what variables were in scope at the time. This will probably require a writer monad of some sort. Then, if the user wants to examine the typing process, they can instruct Juvix to print out all these lots.

This will also be helpful for error message generation later on, I expect.

thealmarty:
I was thinking the same. I'll get this going before writing more tests because this will help us with testing too.

cwgoes:
Are there remaining tasks here? Separating logs into a datatype, perhaps?

thealmarty:
> Are there remaining tasks here? Separating logs into a datatype, perhaps?

Yes. I'd like to do a bit more debugging first, then I'm ready to tidy it all up into a datatype.

cwgoes:
@andy-morris We already have a basic debug mode, let's preserve it when we update the typechecker for first-class usages. I think we can go ahead & close this issue though.

Issue #190, opened 2019-11-18
Clean up the erasure, general type/usage pass-through compilation stack


cwgoes:
Follow-up to https://github.com/cryptiumlabs/juvix/pull/180

- Avoid calling typechecker in erasure
- Figure out best way to pass through info on types, usages of subterms

cwgoes:
- [ ] Type/usage annotated version of core (for erasure)
- [ ] Type/usage annotated version of erased core (for backends)

cwgoes:
I started altering the typechecker to return annotated core but stopped midway - wanted to consult with @thealmarty first - it's pretty messy with the HOAS values if `typeTerm` / `typeElim` return annotations instead of values. Maybe there's a cleaner way.

thealmarty:
I agree that it's pretty messy. I propose the following options:

1. don't worry about the type checker being called again in `Erasure` - it's overkill to do so much code repetition to avoid running the type checker more than once. It adds other inefficiency and the type checker isn't heavy.

2. type check first then erase in the same function. So we'd have a function similar to [this](https://github.com/cryptiumlabs/juvix/blob/e2f88db44a5efc348746b7c5cf9fde812d7cec2a/src/Juvix/Core/Pipeline.hs#L52) but altered such that the type checker results will pass through without repetition.

Option 1 is easy cause it was already done. Were there other downsides to it? Option 2 should be doable. Thoughts?

cwgoes:
It's not so much the performance that concerns me with (1) but rather that it makes the erasure logic more complex (e.g. it has to track the environment to call the type-checker).

In *either* case we need to pass type & usage information down to the various backends, since they will need it, so I don't think we can avoid some changes to the type-checker to return sub-type information. Given that I am sceptical that (2) is really helpful - I don't think we can combine type-checking, erasure, and backend compilation all in one function without way too much complexity.

cwgoes:
A few things we could do to simplify:
- Move value application (`vapp`, `cEval`, `iEval`) to a separate file, since it is independent from typechecking (used by typechecking, but it doesn't use the typechecker itself)
- Either switch away from HOAS (I think this might be easiest) or at least use only HOAS and not HOAS *and* index-bound variables (ref https://github.com/cryptiumlabs/juvix/pull/203#discussion_r351128599)
- Put the error strings in a `Show` typeclass instances (#174 does this already)

cwgoes:
[...]

PR #210, opened 2019-11-27
Separate out evaluator

thealmarty:
- separate out evaluator from the type checker. (as per #190 [comment](https://github.com/cryptiumlabs/juvix/issues/190#issuecomment-558981355))
- add type checker log as per #209


thealmarty:
Reverted the WIP of adding logging to type checker. It will be in another PR so this PR can merge first.

PR #212, opened 2019-11-27
Add logging to type checker

thealmarty:
Follow up on #209 .

### DIFF
--- a/src/Juvix/Core/IR/Typechecker.hs
+++ b/src/Juvix/Core/IR/Typechecker.hs
@@ -1,101 +1,11 @@
 module Juvix.Core.IR.Typechecker where
 
-import Control.Lens ((^?), ix)
+import Juvix.Core.IR.Evaluator
 import Juvix.Core.IR.Types
 import Juvix.Core.Types
 import Juvix.Core.Usage
 import Juvix.Library hiding (show)
 import Prelude (lookup)
-
--- evaluation of checkable terms
-evalTerm ∷
-  ∀ primTy primVal m.
-  ( HasThrow "typecheckError" (TypecheckError primTy primVal m) m,
-    Show primTy,
-    Show primVal
-  ) ⇒
-  Parameterisation primTy primVal →
-  Term primTy primVal →
-  Env primTy primVal m →
-  m (Value primTy primVal m)
-evalTerm _ (Star i) _d = pure (VStar i)
-evalTerm _ (PrimTy p) _d = pure (VPrimTy p)
-evalTerm param (Pi pi ty ty') d = do
-  ty ← evalTerm param ty d
-  pure (VPi pi ty (\x → evalTerm param ty' (x : d)))
-evalTerm param (Lam e) d = pure (VLam (\x → evalTerm param e (x : d)))
-evalTerm param (Elim ii) d = evalElim param ii d
-
-toInt ∷ Natural → Int
-toInt = fromInteger . toInteger
-
--- evaluation of inferable terms
-evalElim ∷
-  ∀ primTy primVal m.
-  ( HasThrow "typecheckError" (TypecheckError primTy primVal m) m,
-    Show primTy,
-    Show primVal
-  ) ⇒
-  Parameterisation primTy primVal →
-  Elim primTy primVal →
-  Env primTy primVal m →
-  m (Value primTy primVal m)
-evalElim _ (Free x) _d = pure (vfree x)
-evalElim _ (Prim p) _d = pure (VPrim p)
-evalElim param (App elim term) d = do
-  elim ← evalElim param elim d
-  term ← evalTerm param term d
-  vapp param elim term
-evalElim param (Ann _pi term _type) d = evalTerm param term d
-evalElim _ (Bound ii) d =
-  fromMaybe (throw @"typecheckError" (UnboundIndex ii)) (pure |<< (d ^? ix (toInt ii)))
-
--- value application function
-vapp ∷
-  ∀ primTy primVal m.
-  ( HasThrow "typecheckError" (TypecheckError primTy primVal m) m,
-    Show primTy,
-    Show primVal
-  ) ⇒
-  Parameterisation primTy primVal →
-  Value primTy primVal m →
-  Value primTy primVal m →
-  m (Value primTy primVal m)
-vapp _ (VLam f) v = f v
-vapp _ (VNeutral n) v = pure (VNeutral (NApp n v))
-vapp param (VPrim x) (VPrim y) =
-  case Juvix.Core.Types.apply param x y of
-    Just v → pure (VPrim v)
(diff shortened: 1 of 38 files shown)


thealmarty:
After changing some things, got [this](https://gist.github.com/thealmarty/89b625d1c39f30bd0f89f8fdf431fa0a) (more detailed) error. Notice that

~~~~
type Context primTy primVal m = [(Name, Annotation primTy primVal m)]

type Annotation primTy primVal m = (Usage, Value primTy primVal m)

evalTerm ∷
  ∀ primTy primVal m.
  ( HasThrow "typecheckError" (TypecheckError primTy primVal m) m,
    Show primTy,
    Show primVal
  ) ⇒
  Parameterisation primTy primVal →
  Term primTy primVal →
  Env primTy primVal m →
  m (Value primTy primVal m)
~~~~
So, output of `evalTerm` is `m (Value primTy primVal m)`

so I need to take evalTerm's output out of the m for it to be `(Value primTy primVal m)`

@cwgoes `evalTerm` never outputs errors, should I take it out of the monad altogether? I'm aware that it requires changes of all related functions but it seems more elegant to not put it in a monad when there's no error to throw. A few other functions never output errors - maybe they should be cleaned up as well?

cwgoes:
> @cwgoes `evalTerm` never outputs errors, should I take it out of the monad altogether? I'm aware that it requires changes of all related functions but it seems more elegant to not put it in a monad when there's no error to throw. A few other functions never output errors - maybe they should be cleaned up as well?

`evalTerm` calls `evalElim`, which does throw errors. If there are cases where functions neither throw errors nor call other functions which do, we should indeed remove the monad.

thealmarty:
> `evalTerm` calls `evalElim`, which does throw errors. If there are cases where functions neither throw errors nor call other functions which do, we should indeed remove the monad.

Indeed. Will leave them alone.

As you suggested, will use `Capability.Writer` instead so it's easier to combine.

cwgoes:
Do you think the log output should be strings, or a datatype (like the `Error` is now)?

I think a datatype is nice since we can write the `Show` instance elsewhere and more easily parse logs with code in the future should we ever need to do that (e.g. for error messages, perhaps).

thealmarty:
> Do you think the log output should be strings, or a datatype (like the `Error` is now)?
> 
> I think a datatype is nice since we can write the `Show` instance elsewhere and more easily parse logs with code in the future should we ever need to do that (e.g. for error messages, perhaps).

That's a good idea - I'm writing a lot of the logs in function forms for that purpose anyway.  Although they are of type `String` now. I will change them to a log datatype.

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -23,67 +74,216 @@ typeTerm ∷
...
       typeTerm
-        p -- checks resultType is of type Star i
+        p -- param
         (ii + 1)
-        ((Local ii, (pi, ty)) : g)
-        (substTerm 0 (Free (Local ii)) resultType)
-        ann
-    _ →
-      throw @"typecheckError" UniverseLevelMustMatch
+        ((Local ii, (SNat 0, ty)) : g) -- add x of type V, with zero usage to the context
```
thealmarty: According to p.22 of the lang ref, x should be added to the context with 0 usage.

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -23,67 +74,216 @@ typeTerm ∷
...
+        else do
+          logOutput $ passed <> typechecked x ann
+          return ()
+-- Lam (introduction rule of dependent function type), requires Pi (formation rule of dependent function type)
+typeTerm p ii g t@(Lam m) ann = do
+  typeTermIntroLog t ann
+  logOutput
+    "patterned matched to be a Lam term. Checking that input annotation has sigma usage and is dependent function type (Pi). "
   case ann of
     (sig, VPi pi ty ty') → do
```
thealmarty: `sig` should be 0 or 1 according to the current lang ref, but currently that is not checked.

thealmarty:
Follow ups:
- we may want to check that `sigma` is 0 or 1 as in lang ref.
- should I output the logs along with the test outputs or else where? 
- please look through the added logs and confirm that they are correct.

thealmarty:
Added sigma have to be 0 or 1 to the Lam and App rule.
The test of adding `Nat 1` and `Nat 2` infers to `(omega, Nat)` fails after adding the enforcement of sigma to the `App` rule:

~~~~
AddNatural 1Natural 2 should infer to type (w,Right Nat):                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 FAIL
    test/CoreTypechecker.hs:407:
    expected: Left Sigma must be 0 or 1.
     but got: Right (w,Right Nat)
~~~~
According to the p.22 of lang ref, the function (M) has to be of usage `sigma`:

![image](https://user-images.githubusercontent.com/39599143/70344849-8600d900-180f-11ea-97b7-9fc381ce1d03.png)

However, in the type checker for `Prim-Fn` (see [here](https://github.com/cryptiumlabs/juvix/blob/0e09997d96d3774f7ff3b647927d1acb3b59c04e/src/Juvix/Core/IR/Typechecker.hs#L355)), we have primitive constants and functions annotated with `omega` usage.

L355 also contradicts the `Prim-Fn` formation and introduction rule


![image](https://user-images.githubusercontent.com/39599143/70344346-57cec980-180e-11ea-869f-688b0d9ec79c.png)

which says that `f`'s annotation is `(sigma, Pi)`.

I shall rewrite the type checker for `Prim-Fn` and add the elimination rule `Prim-App`:

![image](https://user-images.githubusercontent.com/39599143/70344407-8ba9ef00-180e-11ea-8f46-ed777f7752d8.png)

cwgoes:
Primitive constants should be allowed to have `w` usage. I'm less sure about applications of primitive functions. I think we should try first only primitive constants in `w` usage - I don't think we depend on using the results of applying primitive functions multiple times anywhere in the test-cases.

cwgoes:
@thealmarty If you like, you could split the sigma = 0 or 1 check to a separate PR so we can merge the logging code and dig into questions on #203 using it.

cwgoes:
I'll update the constant / function / application rules in a separate PR.

thealmarty:
> @thealmarty If you like, you could split the sigma = 0 or 1 check to a separate PR so we can merge the logging code and dig into questions on #203 using it.

Done. @cwgoes and @andy-morris please review. If there's no problem this can be merged.

thealmarty:
> utACK
> 
> Do we print the logs when a test-case fails?

Oh right, sorry not yet. I'll look into that. In a separate PR.

PR #213, opened 2019-11-27
Michelson backend follow-up

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/168


cwgoes:
Apparently operations can't be serialised, so they can't be in (serialised) closures...

cwgoes:
... for now we will just prohibit lambdas in closures, so we can store them as typed tuples.

cwgoes:
Needs more cleanup but basic closures work & I think this IR is right.

PR #216, opened 2019-12-02
DSL layer work over LLVM

mariari:


mariari:
This PR will have a followup issue regarding refactoring the changes I've propogated in the last two PRs.

Passing around an argument like this is rather poor form, and can be done much better.

PR #233, opened 2019-12-10
Started work on freeing a node

mariari:


doc/Code/Juvix.org
```diff
@@ -58,6 +58,39 @@ This module provides the default environment to run LLVM code in
...
 Operations necessary to update nodes
+- Allocation layout as described by =mallocNodeH=
+  + layout :
+    [portSize | PortArray[portLocation | NodePtr] | DataArray[Data]]
+  | Part         | Alloca Or Malloc  |
+  |--------------+-------------------|
+  | portSize     | Malloc            |
+  | PortArray    | Malloc            |
+  | DataArray    | Malloc            |
+  | PortLocation | Previously Allocd |
```
cwgoes: Can we document where?

doc/Code/Juvix.org
```diff
@@ -99,6 +132,22 @@ Provides a mechanism for defining Sum types
...
+  + layout :
+    {eac | NodePtr* [portSize | PortArray[portLocation | NodePtr] | DataArray[Data]]}
+    * Similar to the one in Graph, however it also has the eac tag
+   | Part         | Alloca Or Malloc   |
+   |--------------+--------------------|
+   | eac          | Malloc             |
+   | portSize     | Malloc             |
+   | PortArray    | Malloc             |
+   | DataArray    | Malloc             |
+   | PortLocation | Not Allocated Here |
```
cwgoes: Can we document where (ditto for `NodePtr`, `Data`)?

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -259,7 +295,31 @@ defineRewire nodePtrType = Block.defineFunction Type.void "rewire" args $
         (numPorts, "port_two")
       ]
 
-delNode = undefined
+deAllocateNode ∷ Define m ⇒ Operand.Operand → Type.Type → m Operand.Operand
+deAllocateNode nodePtr nodePtrType = do
+  node ← load (nodeType nodePtrType) nodePtr
```
cwgoes: So, reading this I am sorta wondering why these aren't contiguous

If they were contiguous (even if they had separate pointers) we could free them with one `free`, one would suppose.

We can worry about this later though

cwgoes:
If we combine all the docs, we should end up with a table which has a bidirectional mapping between every structure stored in memory and the associated `malloc` / `free` calls for that structure

doc/Code/Juvix.org
mariari: Will add tomorrow!

src/Juvix/Backends/LLVM/Codegen/Graph.hs
mariari: I'd say let's make a TODO of this, I actually thought of this myself after writing it.

However, I decided to can it for the time being, as I'm not the most confident in the free and malloc behaviors when interacting with LLVM code.

src/Juvix/Backends/LLVM/Codegen/Graph.hs
```diff
@@ -259,7 +295,31 @@ defineRewire nodePtrType = Block.defineFunction Type.void "rewire" args $
         (numPorts, "port_two")
       ]
```
cwgoes:
```suggestion
-- TODO: Should these be contiguous? If they were contiguous (even if they had separate pointers) we could free them with one free, one would suppose.
```

src/Juvix/Backends/LLVM/Codegen/Graph.hs
cwgoes: Added a TODO

Issue #223, opened 2019-12-03
Make sure type checker behaves as desired


thealmarty:
See discussion on #203 regarding `depIdentity`'s annotation.

After #212  is done this can be investigated further.

Also, after #212  we should examine all the logs of the tests and make sure they all make sense.

thealmarty:
The type checker currently doesn't check that sigma is 0 or 1 but the lang ref says sigma is 0 or 1 (see P.20) - I remember we had doubts about that. 

But I should either stick with the lang ref and check sigma is 0 or 1 or change the lang ref and not check that. Please advise @andy-morris and @cwgoes

cwgoes:
> The type checker currently doesn't check that `sigma` is 0 or 1 but the lang ref says `sigma` is 0 or 1 (see P.20) - I remember we had doubts about that.
> 
> But I should either stick with the lang ref and make sure `sigma` is 0 or 1 or change the lang ref and leave the code as is. Please advise @andy-morris and @cwgoes

For now, in the current version, we should check that `sigma` is 0 or 1 in judgements.

thealmarty:
I think the type checker is doing fine in terms of [these cases]( https://gist.github.com/thealmarty/c92f0dc76fec6a86abfa8d1aaacf924b) (see conclusion in the last paragraph).

thealmarty:
I document the type checker behaviour [here](https://gist.github.com/thealmarty/c92f0dc76fec6a86abfa8d1aaacf924b) regarding how it type checks `depIdentity`. I think the type checker behaves as per the typing rule. @cwgoes and @andy-morris you can see what you think of the behaviour and let me know if you have any questions/comments.

thealmarty:
> > The type checker currently doesn't check that `sigma` is 0 or 1 but the lang ref says `sigma` is 0 or 1 (see P.20) - I remember we had doubts about that.
> > But I should either stick with the lang ref and make sure `sigma` is 0 or 1 or change the lang ref and leave the code as is. Please advise @andy-morris and @cwgoes
> 
> For now, in the current version, we should check that `sigma` is 0 or 1 in judgements.

Typing rules to be updated to properly constrain `sigma` to 0 or 1. WIP #237

thealmarty:
[...]

PR #237, opened 2019-12-10
Thealmarty/refine typechecker

thealmarty:
Follow up on #223


thealmarty:
Outdated. Will follow up after TTG/first class usage implementation.

PR #238, opened 2019-12-11
Various language reference updates

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/204

- Updates primitive typing rules, `App` rule, removes `Prim-App` rule.
- Adds example terms with annotations which should typecheck.

Doesn't finish that issue, there will be further follow-up PRs.


doc/reference/src/core-language.pdc
```diff
@@ -187,9 +187,19 @@ Term judgements have the form:
 Γ ⊢ M \overset{σ}{:} S
 \end{align}
 
-where $σ \in {0,1}$. A judgement with $σ = 0$ constructs a term with no computational content, while a judgement with $σ = 1$ constructs a term which will be computed with. 
+where $σ \in {0,1}$, except for primitive constants.
```
thealmarty: Primitive constants are annotated with `gamma`, not `sigma` now right? So `sigma = {0,1}` without exceptions?

cwgoes: Yes; the judgement syntax is otherwise the same, but σ is always 0 or 1.

Removed this.

doc/reference/src/core-language.pdc
```diff
@@ -692,9 +694,62 @@ To-do: Lay out syntax-directed typechecker following McBride's paper.
 
 ## Examples
 
+### Dependent identity function
```
thealmarty: This section can be removed as it is the same as the I combinator below?

doc/reference/src/core-language.pdc
```diff
@@ -692,9 +694,62 @@ To-do: Lay out syntax-directed typechecker following McBride's paper.
...
+
+The dependent K ("constant") combinator can be typed as: $⊢ λt1.λt2.λx.λy.x \overset{1}{:} (t1 \overset{0}{:} ∗_i) → (t2 \overset{0}{:} ∗_i) → (x \overset{1}{:} t1) → (y \overset{0}{:} t2) → t1$.
+
+This should also typecheck if the $x$ and $y$ argument usages are replaced with $w$ (instead of $1$ and $0$).
+
+#### I combinator
+
+The dependent I ("identity") combinator can be typed as: $⊢ λt.λx.(x \overset{1}{:} t) \overset{1}{:} (t \overset{0}{:} ∗_i) → (x \overset{1}{:} t) → t$.
+
+This should also typecheck if the $x$ argument usage is replaced with $w$ (instead of $1$).
```
thealmarty: [It did type checked with `Omega`](https://github.com/cryptiumlabs/juvix/pull/260/commits/35b5dc1364d1f3c408673704608ffab5175b474a)! Woohoo!

thealmarty:
The examples make sense to me. I'll revisit the McBride paper and review the typing rule changes after the Isohypsibius release.

doc/reference/src/core-language.pdc
cwgoes: Aye, good catch, removed.

cwgoes:
> The examples make sense to me. I'll revisit the McBride paper and review the typing rule changes after the Isohypsibius release.

:+1: - I'll merge for now, and we can review again after the release (next one may add first-class usages anyways, still tbd).

PR #240, opened 2019-12-11
Add custom NatVal show instance.

thealmarty:
The test outputs are easier to read now I think.

### DIFF
--- a/src/Juvix/Core/Parameterisations/Naturals.hs
+++ b/src/Juvix/Core/Parameterisations/Naturals.hs
@@ -29,10 +29,10 @@ data NatVal
   deriving (Eq)
 
 instance Show NatVal where
-  show (Natural x) = " Nat " <> Text.Show.show x <> " "
-  show Add = " + "
-  show Sub = " - "
-  show Mul = " * "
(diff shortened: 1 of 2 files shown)


src/Juvix/Core/Parameterisations/Naturals.hs
```diff
@@ -25,7 +26,14 @@ data NatVal
   | Sub -- f subtraction
   | Mul -- f multiplication
   | Curried NatVal Natural
-  deriving (Show, Eq)
+  deriving (Eq)
+
+instance Show NatVal where
+  show (Natural x) = " Nat " <> Text.Show.show x <> " "
```
thealmarty: Not sure if the last space here is overkill...

mariari: I believe it is so.

we get results like
```haskell
"(+ 3 4)" should parse as Elim (App (App (Prim  + ) (Elim (Prim  Nat 3 ))) (Elim (Prim  Nat 4 ))):
```

We should end up with
```haskell
"(+ 3 4)" should parse as Elim (App (App (Prim +) (Elim (Prim Nat 3 ))) (Elim (Prim Nat 4))):

PR #242, opened 2019-12-12
Tried to make lines fit below 100 characters long

mariari:
Also I cleaned up some trailing whitespace

This also fixes the code formatters issue with parsing `Typechcker`

### DIFF
--- a/src/Juvix/Core/IR/Typechecker.hs
+++ b/src/Juvix/Core/IR/Typechecker.hs
@@ -267,14 +267,14 @@ typeTerm p ii g t@(Elim e) ann = do
   if not (fst ann' `allowsUsageOf` fst ann)
     then
       ( do
-          logOutput "Usages not compatible. "
+          logOutput "Usages not compatible."
           throw @"typecheckError" (UsageNotCompatible ann' ann)
       )
     else
       if (annt /= annt')
         then
           ( do
-              logOutput "The types are not the same. "
+              logOutput "The types are not the same."
               throw @"typecheckError" (TypeMismatch ii (Elim e) ann ann')
           )
         else return ()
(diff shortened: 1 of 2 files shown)


src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -73,12 +73,17 @@ typeTerm ∷
   Term primTy primVal →
   Annotation primTy primVal m →
   m ()
-
--- * (Universe formation rule)
+-- ★ (Universe formation rule)
```
thealmarty: I noticed the error. Glad it's fixed!

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -73,12 +73,17 @@ typeTerm ∷
...
 
 typeTerm _ _ii _g t@(Star i) ann = do
   typeTermIntroLog t ann
-  logOutput "patterned matched to be a * term. Type checker applies the universe formation rule. Checking that Sigma is zero. "
+  logOutput
```
thealmarty: This is fine but the spacing at the end were intentional. I think having a white space between sentences are easier to read.

mariari: Hmm, adding it here, seems rather error prone, as one can forget it. Could this be added to the logOutput function itself?

thealmarty: That may be a good idea. Can be tweaked further...I expect I'll have to change the logs along the way.

mariari:
> Maybe revert the last commit? Doesn't matter much because I expect I'll have to edit the logging for readability anyway when I read the logs.

I'll probably just add them back, as it's shifted between a few commits.

I think the tricky thing here, is making it so the first log doesn't have an extra space, as if we add it to `logOutput` I suspect this will have to be programmed around

src/Juvix/Core/IR/Typechecker.hs
thealmarty: I was going to take them out as a new datatype as Chris suggested. But after we've debugged more.

thealmarty:
> > Maybe revert the last commit? Doesn't matter much because I expect I'll have to edit the logging for readability anyway when I read the logs.
> 
> I'll probably just add them back, as it's shifted between a few commits.
> 
> I think the tricky thing here, is making it so the first log doesn't have an extra space, as if we add it to `logOutput` I suspect this will have to be programmed around

Oh no I don't want you to spend so much time on it! We can merge as is. Again, I'll change it anyway.

PR #244, opened 2019-12-12
Thealmarty/print test logs

thealmarty:
This PR aims to display logs when tests fail.

I think it might be worth it to move away from `tasty-discover` so I can customize the test output easier.
As we add more tests we'd want more customization on how to display them, display with logs etc. It shouldn't be too bad to put the test groups in `defaultMain` and call them, after the tests are organized into `TestGroup`.

Regardless of our decisions on staying with `tasty-discover` or not, I'll start organizing `TestTree`'s into groups in a separate PR.


cwgoes:
I think moving away from `tasty-discover` is fine; explicit organisation would be nice anyways, and we will want to customise the output for some tests I expect.

thealmarty:
The backend LLVM tests are removed from tests atm because otherwise the [CI tests don't pass](https://circleci.com/gh/cryptiumlabs/juvix/1805?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) although those tests do pass on my local test runs.

thealmarty:
I'm blocked here. I don't know why the `HideSuccesses` option isn't working. I was hoping to hide the passed tests so that I can add back the failed tests and debug the type checker easier. The failed test logs can be read from the CI outputs still and we can debug that way for now.

cwgoes:
I tried `stack test --fast --test-arguments --hide-successes`; that didn't work either.

cwgoes:
Ah, found it:

```bash
stack test --fast --test-arguments '--hide-successes --ansi-tricks false'
```

Probably you can also do this in code, I think the CLI is a nice way though

thealmarty:
> Ah, found it:
> 
> ```shell
> stack test --fast --test-arguments '--hide-successes --ansi-tricks false'
> ```
> 
> Probably you can also do this in code, I think the CLI is a nice way though

The commit above was me trying to do this in code, but it doesn't seem to work. I'll remove that code and we can use the CLI for that. Thank you!

cwgoes:
> The commit above was me trying to do this in code, but it doesn't seem to work. I'll remove that code and we can use the CLI for that. Thank you!

It's the `--ansi-tricks false` part that did it.

thealmarty:
I think as we add more tests doing it in the code would be better. But we can change that later.

PR #251, opened 2019-12-14
LLVM graph API

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/189#issuecomment-563434186

- Alter the JIT interface (in `Backends/LLVM/JIT/Execution.hs`) so that we can JIT a module then retrieve multiple functions by name which might have different types.


cwgoes:
Merging this so we have the better JIT function. Follow-up PR for the API coming soon (tm).

PR #252, opened 2019-12-14
Isohypsibius language reference updates contd.

cwgoes:
Closes https://github.com/cryptiumlabs/juvix/issues/204

- [x] Figure out if we should just special-case `Eq` / `Refl` or do something more Agda-like
- [x] Clean up desiderata & ingredients


cwgoes:
We will punt `Eq / Refl` to the next release, I think, so this can be reviewed.

cwgoes:
The content can be reviewed.

Still having troubles with the latest `pandoc` which should be fixed before this is merged.

cwgoes:
Thanks for rebuilding, *grr* Arch Linux problems *grr*.

PR #257, opened 2019-12-16
Compiling of the LLVM code to actual LLVM

mariari:
There are a few things that we have to do before we achieve this goal

1. [x] Execution of Codegen

2. [ ] Moving Define from being partly in Codegen, to being out
     - This probably involves just running the body in the call site.

3. [ ] Having runLLVM up, and properly passing code to the JIT


src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -200,8 +227,8 @@ externf ∷ Externf m ⇒ Name → m Operand
 externf name = getvar (nameToSymbol name)
 
 nameToSymbol ∷ Name → Symbol
-nameToSymbol (UnName n) = (intern (show n))
-nameToSymbol (Name n) = (intern (show n))
+nameToSymbol (UnName n) = (intern (filter (/= '\"') (show n)))
```
cwgoes: Why no quotes? Are they disallowed? Maybe we should check this when `Name`s are created.

PR #260, opened 2019-12-17
Thealmarty/improve log

thealmarty:
Follow up on #223 - adding more logs where tests fail.


thealmarty:
Some of the log intros aren't working atm, e.g., [here](https://github.com/cryptiumlabs/juvix/blob/7bf6d20a044f93f1f2c8e81b0500dd5b2f6e84fa/src/Juvix/Core/IR/Typechecker.hs#L265). To be fixed.

thealmarty:
Logs are a lot more informative now. They are quite long. Eventually need to take it out of the `TestTree`s so that I can 
- `putStr` and have newlines properly
-  use pretty printer like [pretty-simple](https://hackage.haskell.org/package/pretty-simple) so we can have colours etc

thealmarty:
Took out the `depIdentity` test that correctly fails. See explanations [here](https://gist.github.com/thealmarty/657a4484f104f801b50c64bd4924eae0).

thealmarty:
Added `depK` test as per #238 's example. Test passed.

thealmarty:
`s` test does not pass atm and I think there is a bug in the type checker that makes it not able to type check variables that are used more than once. In this case, the third input of `s` (`z`). Typing rule to be fixed as per #238 .

thealmarty:
Improved logs and added more passed tests for the Isohypsibius release!

TODO: 
- improve the 'input' vs 'term' annotation wording, atm it's a bit confusing.
- fix typing rules as per #238 and make sure the dependent S combinator example there pass.

Issue #239, opened 2019-12-11
Formalisation todo list


andy-morris:
- [x] Add first-class usages (#99)
- [ ] Make sure subusaging is well-behaved
- [ ] Type safety proof
- [ ] Confluence of reduction
- [ ] Termination of reduction on well-typed terms

cwgoes:
- [ ] Formalise self-types for induction over lambda-encoded data types
 (blocked on language reference explaining this much more clearly)
- [ ] Formalise linear logic types (low-priority for now)
- [ ] Formalise parameterisation over primitive types / primitive values (minor)

cwgoes:
@andy-morris I was discussing a bit with someone and he mentioned that it might be nice to also check that evaluation-with-all-nonzero-usages-as-omega should be equivalent to evaluation with specific usages - that seems like it would be relatively easy to prove? Would be nice to isolate any potential mistakes in our usage accounting from actual computational behaviour.

PR #261, opened 2019-12-18
Add statements of properties we want

andy-morris:
Adds modules in the `Metatheory` subdirectory with the type signatures of the things in the first comment of #239. This also required reshuffling what was a term and what was an elimination to [cleanly define what a value is](https://github.com/cryptiumlabs/juvix/blob/0c38bddb2626bce0ab151b82a4c658c82dca79d9/experimental/qtt-agda/Metatheory/Value.agda#L24-L69).


cwgoes:
Let's merge this as-is for now, and continue later when we pick up the formalisation work again.

PR #270, opened 2019-12-22
Michelson backend follow-up

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/168

- [x] Use `APPLY`
- [x] Ensure lambda functions in closures work properly
- [x] Ensure (to-be) dependently-typed primitives work properly
- [ ] Hook up the interpreter
- [ ] Update the CLI / REPL


cwgoes:
Blocked on operations not being capturable in closures.

I suppose we have to special-case primitive application or something somehow.

cwgoes:
Going to merge this bit since it does fix lambda captures.

PR #278, opened 2019-12-28
Adding the Lisp code org generation 

mariari:


cwgoes:
Are we generating any comments for Lisp code yet?

mariari:
> Are we generating any comments for Lisp code yet?

Yes, it is now finished. Just had to implement the last few functions

PR #280, opened 2019-12-30
Added HasReader Instances and removed some HasState.

mariari:
This PR also cleans up some long lines


mariari:
If there are any other instances, feel free to commit and change them.

PR #284, opened 2020-01-01
Start frontend language documentation

cwgoes:
Ref https://github.com/cryptiumlabs/juvix/issues/273.

Covers in detail:
- [ ] Basic syntax
- [ ] Implicit arguments
- [ ] Datatype definitions
- [ ] Pattern matching & case expressions
- [ ] Totality checking
- [ ] Various pipelines for different backends (incl. visualisation)

Outlines in brief:
- [ ] Typeclasses
- [x] Do-notation
- [x] Module system
- [x] Automatic proof search

It may be easier to review [the rendered version](https://github.com/cryptiumlabs/juvix/blob/cwgoes/start-frontend-language-description/doc/reference/language-reference.pdf).


cwgoes:
Datatype-wise, for frontend language 1.0 I think we'll just stick with inductive families, but keep in mind the goal to eventually be able to compile them to lambda-encoded datatypes that typecheck with only dependent intersections.

doc/reference/src/frontend/syntax.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+elaborates to
+
+```
+zipWith : (a : 0 Type) -> (b : 0 Type) -> (c : 0 Type) -> (n : 0 Nat) -> n (a -> b -> c) -> Vect n a -> Vect n b -> Vect n c
+zipWith (a : 0 Type) (b : 0 Type) (c : 0 Type) (Z : 0 Nat)   _ Nil        Nil       = Nil
+zipWith (a : 0 Type) (b : 0 Type) (c : 0 Type) (S k : 0 Nat) f (x :: xs)  (y :: ys) = f x y :: zipWith a b c k f xs ys
+
+zipAdd : (a : 0 Type) -> (n : 0 Nat) -> (d : 0 (Num a)) -> Vect n a -> Vect n a -> Vect n a
```
andy-morris: Doesn't `d` need to have usage `n`? (or `ω` if that can't be inferred)

cwgoes: Yes, probably, `d` is a typeclass & I don't know exactly how we'll set usages there. `n` for now.

PR #297, opened 2020-01-06
Graph API work contd.

cwgoes:
`test_eval_jit` in `test/Backends/LLVM.hs` should suffice for print testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ tmp/
 
 # Agda temporary
 *.agdai
+
+# LLVM temporary
+*.ll
+*.s
--- a/src/Juvix/Backends/LLVM/Net/API.hs
+++ b/src/Juvix/Backends/LLVM/Net/API.hs
@@ -58,13 +58,13 @@ opaqueNetType = Type.PointerType eacListPointer (Addr.AddrSpace 0)
 -- reduceUntilComplete :: Ptr Net -> IO ()
 -- test :: IO ()
 
-defineTest :: Codegen.Define m => m Operand.Operand
+defineTest ∷ Codegen.Define m ⇒ m Operand.Operand
 defineTest =
   Codegen.defineFunction Type.void "test" [] $ do
-    create_net <- Codegen.externf "create_net"
-    append_to_net <- Codegen.externf "append_to_net"
+    create_net ← Codegen.externf "create_net"
+    append_to_net ← Codegen.externf "append_to_net"
     reduce_until_complete ← Codegen.externf "reduce_until_complete"
-    ptr <- Codegen.call opaqueNetType create_net (Codegen.emptyArgs [])
+    ptr ← Codegen.call opaqueNetType create_net (Codegen.emptyArgs [])
     Codegen.callVoid append_to_net (Codegen.emptyArgs [ptr, Operand.ConstantOperand (C.Null nodePointer), Operand.ConstantOperand (C.Int 64 0)])
     Codegen.callVoid reduce_until_complete (Codegen.emptyArgs [ptr])
     Codegen.retNull
@@ -108,8 +108,8 @@ defineAppendToNet =
     eac_list ← Types.cons appNode (Operand.ConstantOperand (C.Null Types.eacLPointer))
     Codegen.store netPtr eac_list
     Codegen.retNull
-    where
-      args = [(opaqueNetType, "net"), (nodePointer, "nodes"), (int32, "node_count")]
+  where
+    args = [(opaqueNetType, "net"), (nodePointer, "nodes"), (int32, "node_count")]
 
 defineAppendToNet' ∷ (Codegen.Define m, Codegen.MallocNode m) ⇒ m Operand.Operand
 defineAppendToNet' =
@@ -159,8 +159,8 @@ defineAppendToNet' =
     Codegen.setBlock forExit2
     -- TODO: Set eac list pointer?
     Codegen.retNull
-    where
-      args = [(opaqueNetType, "net"), (nodePointer, "nodes"), (int32, "node_count")]
+  where
+    args = [(opaqueNetType, "net"), (nodePointer, "nodes"), (int32, "node_count")]
 
 defineReduceUntilComplete ∷ Codegen.Define m ⇒ m Operand.Operand
 defineReduceUntilComplete =
--- a/src/Juvix/Backends/LLVM/Net/API.hs
+++ b/src/Juvix/Backends/LLVM/Net/API.hs
@@ -65,7 +65,7 @@ defineTest =
     append_to_net ← Codegen.externf "append_to_net"
     reduce_until_complete ← Codegen.externf "reduce_until_complete"
     ptr ← Codegen.call opaqueNetType create_net (Codegen.emptyArgs [])
-    Codegen.callVoid append_to_net (Codegen.emptyArgs [ptr, Operand.ConstantOperand (C.Null nodePointer), Operand.ConstantOperand (C.Int 64 0)])
+    Codegen.callVoid append_to_net (Codegen.emptyArgs [ptr, Operand.ConstantOperand (C.Null nodePointer), Operand.ConstantOperand (C.Int Codegen.addressSpace 0)])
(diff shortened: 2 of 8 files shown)


cwgoes:
Presently, when JIT-ing the initial module:

```bash
juvix-test: EncodeException "A type definition requires a structure type but got: IntegerType {typeBits = 64}"
```

cwgoes:
The problem is `numPorts`, which is defined as an integer if the address space is large enough.

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,59 @@ defineAppendToNet =
...
+    addr ← Codegen.getElementPtr (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind, Operand.ConstantOperand (C.Int 32 0)]})
+    Codegen.store addr kind
+    next ← Codegen.add int32 ind (Operand.ConstantOperand (C.Int 32 1))
+    Codegen.store counter next
+    cond ← Codegen.icmp IntPred.EQ node_count next
+    Codegen.cbr cond forLoop forExit
+    -- Exit case: next loop.
+    Codegen.setBlock forExit
+    Codegen.retNull
+    Codegen.store counter (Operand.ConstantOperand (C.Int 32 0))
```
mariari: Why store 0 on the counter?

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,59 @@ defineAppendToNet =
...
+    Codegen.br forLoop2
+    -- Second loop: link nodes.
+    Codegen.setBlock forLoop2
+    ind ← Codegen.load int32 counter
```
mariari: Since we store 0 on the counter before coming here, this should always be 0. Thus indexs here are on 0?

.gitignore
```diff
@@ -23,3 +23,7 @@ tmp/
 
 # Agda temporary
 *.agdai
+
+# LLVM temporary
+*.ll
```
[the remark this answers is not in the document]
cwgoes: Yeah, but for now it's convenient.

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,59 @@ defineAppendToNet =
...
+    forExit2 ← Codegen.addBlock "for.exit.2"
+    -- Create a counter to track position
+    counter ← Codegen.alloca int32
+    Codegen.store counter (Operand.ConstantOperand (C.Int 32 0))
+    Codegen.br forLoop
+    -- Loop case: convert node, increment counter.
+    Codegen.setBlock forLoop
+    ind ← Codegen.load int32 counter
+    -- Load node at index `ind`.
+    node ← Codegen.loadElementPtr (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind]})
```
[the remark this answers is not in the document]
cwgoes: Aye, updated.

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,59 @@ defineAppendToNet =
...
+    -- Create the in-memory node.
+    kind ← EAC.mallocApp -- TODO: Switch on node kind.
+      -- Write the address to the list.
+    addr ← Codegen.getElementPtr (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind, Operand.ConstantOperand (C.Int 32 0)]})
```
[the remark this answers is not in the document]
cwgoes: Aye, updated.

src/Juvix/Backends/LLVM/Net/API.hs
cwgoes: It's being re-used for a second loop.

src/Juvix/Backends/LLVM/Net/API.hs
cwgoes: It's incremented later (line 155) - the blocks are not in execution order (confusingly I suppose)

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,59 @@ defineAppendToNet =
...
+    Codegen.setBlock forExit
+    Codegen.retNull
+    Codegen.store counter (Operand.ConstantOperand (C.Int 32 0))
+    Codegen.br forLoop2
+    -- Second loop: link nodes.
+    Codegen.setBlock forLoop2
+    ind ← Codegen.load int32 counter
+    -- Load node at index `ind`.
+    node ← Codegen.loadElementPtr (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind]})
+    ptr ← Codegen.loadElementPtr (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind, Operand.ConstantOperand (C.Int 32 0)]})
```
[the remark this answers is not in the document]
cwgoes: Aye, updated.

src/Juvix/Backends/LLVM/Translation.hs
```diff
@@ -97,6 +88,26 @@ evalErasedCoreInLLVM parameterisation term = do
...
+nodeToIR ∷ ∀ k a primVal c (dataTy ∷ k). (INIR.Address, (a, Lang primVal, c)) → INIR.Node dataTy
+nodeToIR (ind, (_, l, edges)) =
+  INIR.Node
+    { INIR.nodeAddress = ind,
+      INIR.nodeKind = case l of
+        Primar Erase → 0
+        Auxiliary2 Lambda → 1
+        Auxiliary2 App → 2
+        Auxiliary2 (FanIn i) → i
+      --INIR.nodePorts = map (\(_, toNode, toPort) → INIR.Port toNode (portTypeToIndex toPort)) edges
```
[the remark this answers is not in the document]
cwgoes: It is, to-do for a future PR.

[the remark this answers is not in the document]
cwgoes: (still need to determine the exact memory model to pass through to LLVM)

src/Juvix/Backends/LLVM/Net/API.hs
mariari: Well the reason why I found this odd, is because you jump to the wrong loop, see below

src/Juvix/Backends/LLVM/Net/API.hs
```diff
@@ -96,56 +108,63 @@ defineAppendToNet =
...
+    node ← Codegen.loadElementPtr
+      (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind]})
+    ptr ← Codegen.loadElementPtr
+      (Codegen.Minimal {Codegen.address' = nodes, Codegen.type' = nodePointer, Codegen.indincies' = [ind, Operand.ConstantOperand (C.Int 32 0)]})
+    -- TODO: Link things, lookup node pointers.
+    -- Alter parameter?
+    next ← Codegen.add int32 ind (Operand.ConstantOperand (C.Int 32 1))
+    Codegen.store counter next
+    cond ← Codegen.icmp IntPred.EQ node_count next
+    Codegen.cbr cond forLoop forExit
```
mariari: Here you probably what forLoop2 and forExit2.

No?

src/Juvix/Backends/LLVM/Net/API.hs
mariari: Well, the reason it seems odd, is that when we hit this point, we only ever add 1 to it before never touching this part again. See comments below

src/Juvix/Backends/LLVM/Net/API.hs
mariari: `    Codegen.cbr cond forLoop2 forExit2`

this is probably what you wanted, with this, your increment would be correct

PR #299, opened 2020-01-08
Trees that grow

andy-morris:


andy-morris:
One thing I'm wondering about is what to do with the `Show` instances. At the moment `Show` only exists for terms without extensions but it'd also be possible to insert them in somehow.

I'd actually like to suggest having derived `Show` instances and a separate pretty-printing class (or set of functions, or whatever) that uses the concrete syntax of core.

src/Juvix/Core/IR/Types.hs
```diff
@@ -1,58 +1,140 @@
+{-# LANGUAGE ConstraintKinds, EmptyCase, TypeFamilies, UndecidableInstances #-}
```
cwgoes: Can we avoid `UndecidableInstances` or is it required for this pattern?

cwgoes:
Would be nice to see an example of substantive annotations & combining two different annotations.

> One thing I'm wondering about is what to do with the `Show` instances. At the moment `Show` only exists for terms without extensions but it'd also be possible to insert them in somehow.

Can we derive/define Show instances for the terms which use Show instances for the extensions if they are available?

> I'd actually like to suggest having derived `Show` instances and a separate pretty-printing class (or set of functions, or whatever) that uses the concrete syntax of core.

I agree with this - we will want nice things later (e.g. colors) that we don't want in show.

src/Juvix/Core/IR/Types.hs
andy-morris: The error message specifically cites nested constraints—e.g. the context for `Eq Term' _ _ _` expands out to:

```haskell
  ( Eq primTy, Eq primVal,
    ( Eq (XStar ext primTy primVal),
      Eq (XPrimTy ext primTy primVal),
      -- etc
    ),
    ( Eq (XBound ext primTy primVal), {- etc -} )
  )
```

...[which is not allowed in instance contexts (or superclasses)](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-ConstraintKinds). It'd probably be possible not to use `UndecidableInstances`, at the cost of more duplication, but I'd like to avoid that unless you feel particularly strongly.

Since this particular restriction only applies to classes & instances, I think this is the only module that will need `UndecidableInstances` enabled.

cwgoes: I see - it's not the end of the world - agree that avoiding duplication is also important..

cwgoes:
By the way, I think this merits an [architectural choice record](https://github.com/cryptiumlabs/juvix/tree/develop/doc/Architecture).

andy-morris:
Turns out the thing I said about `type HRAnn = (HR, Ann)` doesn't work, since the annotations in `HRAnn.Term` should themselves be `HRAnn.Term`s (rather than `IRAnn.Term`s)... oops!

src/Juvix/Core/IR/Types.hs
andy-morris: btw, since "extension-polymorphic" functions also involve classes, IR.Evaluator needs UIs as well now.

src/Juvix/Core/HR/Types.hs
```diff
@@ -1,21 +1,56 @@
...
+
+  -- (disable these constructors)
+  type XFree HR primTy primVal = Void
+  type XBound HR primTy primVal = Void
+  type ElimX HR primTy primVal = Symbol
+
+
+type Term = IR.Term' HR
+
+pattern Star :: Natural -> Term primTy primVal
```
cwgoes: Thanks for writing these. It would be cool if there were a way to *derive* them, do you know if that is possible?

(without Template Haskell, it isn't worth Template Haskell)

src/Juvix/Core/HRAnn/Erasure.hs
```diff
@@ -1,21 +1,20 @@
...
 
+import Juvix.Library
 import qualified Juvix.Core.HR.Types as HR
 import Juvix.Core.HRAnn.Types
+import Juvix.Core.IR.TransformExt
+
+
+hrForgetter :: ExtTransformTE HRAnn HR.HR primTy primVal
+hrForgetter =
+  ExtTransformTE
```
cwgoes: We should name the fields, since it's easy to mix up the order.

src/Juvix/Core/IR/Evaluator.hs
```diff
@@ -9,48 +11,86 @@ import Juvix.Core.IR.Types
 import Juvix.Core.Types
 import Juvix.Library hiding (show)
 
+-- TODO: extend functions to annotated terms:
```
cwgoes: What is required to do this? It looks like it's already under a `∀ ext`? I guess `Value` discards the annotations though.

src/Juvix/Core/IR/Types.hs
```diff
@@ -1,58 +1,139 @@
...
+
+deriving instance
+  (Eq primTy, Eq primVal, TEAll Eq ext primTy primVal) =>
+  Eq (Term' ext primTy primVal)
+
+deriving instance
+  (Show primTy, Show primVal, TEAll Show ext primTy primVal) =>
+  Show (Term' ext primTy primVal)
 
+-- FIXME add pretty-printer
```
cwgoes: Yes. We should track this separately. https://github.com/cryptiumlabs/juvix/issues/308.

src/Juvix/Core/IR/Types.hs
```diff
@@ -66,7 +147,8 @@ data Name
 data Value primTy primVal m
```
cwgoes: Do we want to either add extensions to values (so we can pass them through evaluation), or alter evaluation so it just returns a suitably-restricted `Term` (though we couldn't typecheck this)? We will want the ability to evaluate but keep annotations (maybe with some transform).

src/Juvix/Core/HR/Types.hs
mariari: I think we can do these via, `deriving via`, as they seem to fall directly out of the type itself?

Though I think even that approach may be a bit of a pain.

src/Juvix/Core/IR/Extension.hs
```diff
@@ -0,0 +1,97 @@
...
+where
+
+import Juvix.Library
+import Data.Kind (Type, Constraint)
+
+
+-- | Marker type for having no extensions
+data NoExt
+
+-- | Extensions for terms and eliminations.
```
mariari: Probably best to put this documentation also at the top show it shows up in the documentation generation?

src/Juvix/Core/IR/TransformExt.hs
```diff
@@ -0,0 +1,115 @@
...
+                                      <*> etfApp fs e
+extTransformEF fs (Ann' π s t e) = Ann' π <$> extTransformTF fs s
+                                          <*> extTransformTF fs t
+                                          <*> etfAnn fs e
+extTransformEF fs (ElimX e) = ElimX <$> etfElimX fs e
+
+extTransformE :: ExtTransformTE ext1 ext2 primTy primVal
+              -> Elim' ext1 primTy primVal -> Elim' ext2 primTy primVal
+extTransformE fs t = runIdentity $ extTransformEF fs t
+
```
mariari: These all seem to fall out from the ADT itself.

Could we somehow do a deriving via (not sure if it can get the fields this way!). To basically auto generate this.

Would be useful if we use this technique in more places.

src/Juvix/Core/IR/TransformExt.hs
```diff
@@ -0,0 +1,115 @@
...
+
+forgetter :: (TermX ext primTy primVal ~ Void,
+              ElimX ext primTy primVal ~ Void)
+          => ExtTransformTE ext NoExt primTy primVal
+forgetter =
+  ExtTransformTE (const ()) (const ()) (const ()) (const ())
+                 (const ()) (const ()) (const ()) (const ())
+                 (const ()) (const ()) absurd     absurd
```
mariari: Why absurd?

Also it seems there are as many const () or abusrd here as there are forms?

Something else that should be generated with this?

src/Juvix/Core/HR/Types.hs
andy-morris: I thought `deriving via` could only produce instances? As far as I know the only way to generate other kinds of declarations is with TH, unfortunately.

src/Juvix/Core/IR/Evaluator.hs
andy-morris: oops, sorry, that's a comment reminding me to add the `∀ ext` that I forgot to remove.

src/Juvix/Core/HR/Types.hs
```diff
@@ -1,21 +1,56 @@
...
-  = Var Symbol --variable
-  | Prim primVal --primitive constant
-  | App (Elim primTy primVal) (Term primTy primVal) --application
-  | Ann Usage (Term primTy primVal) (Term primTy primVal) --type & usage annotation
-  deriving (Show, Eq, Generic)
+
+data HR
+
+instance IR.TEExt HR where
+  -- FIXME: shouldn't XPi also be a name?
```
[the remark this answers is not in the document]
andy-morris: That's right. I thought that `HR` was the AST coming straight out of the parser, so `Pi` would need to have a name attached the same way as `Lam`.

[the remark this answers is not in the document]
cwgoes: Ah, yes, it probably will.

src/Juvix/Core/IR/Types.hs
andy-morris: I think we probably will at some point, yes. My reasoning for stripping the extensions in `eval*` was that I thought they would most likely be annotations (e.g. types) that we may want to recalculate to make sure the evaluator works how we expect.

src/Juvix/Core/IR/TransformExt.hs
andy-morris: The last two are `absurd` because those two cases must be `Void`—if we're stripping all extensions then extra constructors can't be present because there's nothing to map them to.

An alternative is to give those two fields a type like `TermX ext1 primTy primVal -> f (Term ext2 primTy primVal)` or something, so that specific transformations can replace them with something reasonable if possible.

Anyway, yes, you're right, a lot of this code is completely systematic based on the datatype, but I'm not aware of any way to automatically generate it without template haskell.

andy-morris:
@mariari made a good point that the `complete` pragmas could easily get out of date when we extend `Term` and `Elim`, so it might be better to just remove them. This does mean that you'd have to write e.g. `TermX x` instead of `Var x` in `HR.Term`, but that's probably a price worth paying.

src/Juvix/Core/HRAnn/Erasure.hs
andy-morris: Done, and also for `TransformExt.forgetter`.

cwgoes:
Merging; future upgrade to https://github.com/cryptiumlabs/extensible expected - https://github.com/cryptiumlabs/juvix/issues/325.

PR #305, opened 2020-01-14
Push/Enter and Constant Propagation for Michelson

mariari:


mariari:
This commit will also change the stack representation from an alist where the position is the index into the list into something similar to the following

```ocaml

type Store = Constant Type Constant
           | OnStack Type int

type env = {map : Map Symbol Store; size : Int}

position sym env = size env - lookupI env sym
```

This allows us to save time on compilation, along with having a better invariant on the structure.

Instead of having the length of the list be the invariant, meaning we can't skip constants on the list!, we have it be this size of the stack, and can check against that.

mariari:
This commit will also add a size parameter to the stack, that way we don't have to do length checks all the time.

Constant values will be stored on this stack (inefficient to lookup!), however they won't count towards the length of the stack, and will be noted as a skip when looking up offsets!

src/Juvix/Backends/Michelson/Compilation/Checks.hs
```diff
@@ -4,47 +4,51 @@ module Juvix.Backends.Michelson.Compilation.Checks where
...
-                    ( mconcat
-                        [ "stack mismatch while compiling ",
-                          show term,
-                          " - end stack: ",
-                          show end,
-                          ", lifted stack: ",
-                          show endType
-                        ]
+  case maybeInstr of
+    Left _ → pure maybeInstr
```
cwgoes: Note that we can still check the stack type here, by simulating what the lambda would be.

src/Juvix/Backends/Michelson/Compilation/Datatypes.hs
```diff
@@ -26,18 +27,35 @@ unpack ∷
 unpack (Type ty _) binds =
   case ty of
     Tbool → do
-      modify @"stack" (drop 1)
+      modify @"stack" (VStack.drop (1 ∷ Int))
```
cwgoes: Annotation should no longer be necessary, we constrained the type

src/Juvix/Backends/Michelson/Compilation/Datatypes.hs
```diff
@@ -26,18 +27,35 @@ unpack ∷
...
       return (SeqEx [])
     TPair _ _ fT sT →
       case binds of
         [Just fst, Just snd] → do
-          modify @"stack" (appendDrop [(VarE fst, fT), (VarE snd, sT)])
+          modify @"stack"
+            ( VStack.appendDrop
+                ( VStack.fromList
+                    [ (VStack.varE fst Nothing, fT),
```
cwgoes: Is it worth a helper function for `VarE x Nothing` ?

src/Juvix/Backends/Michelson/Compilation/Lambda.hs
```diff
@@ -0,0 +1,174 @@
...
+    Left (LamPartial ops captures args body _) → do
+      -- TODO: Actually compile the lambda to a closure.
+      -- We should never need to do this elsewhere.
+      -- ergo, if we do not return a lambda, we should never
+      -- compile to a lambda, except for the michelson built-ins
+      -- which take lambdas
+      -- todo: formalise this a bit
+      -- todo: deal with michelson builtins which take lambdas
+      -- they will have to use this function or something
+      undefined
```
cwgoes: Ah here we need to call `funcToLambda`

src/Juvix/Backends/Michelson/Compilation/Lambda.hs
```diff
@@ -0,0 +1,174 @@
...
+      )
+      captures
+  -- Step 2: Figure out what the stack will be in the body of the function.
+  current@(VStack.T currentStack _) ← get @"stack"
+  let realValuesGen xs =
+        length
+          $ filter (VStack.inT . fst)
+          $ take (length xs) currentStack
+
+      realValues = realValuesGen captures
```
cwgoes: I wonder if we should alter the formatter, I don't really like the newlines here

src/Juvix/Backends/Michelson/Compilation/Lambda.hs
```diff
@@ -0,0 +1,174 @@
...
+
+      numVarsInClosure = (length (captures <> args))
+
+  extraArgsWithTypes ← zip args . drop (length args) <$> typesFromPi lamTy
```
cwgoes: Why are we dropping `(length args)`, that seems wrong now (maybe this was old logic)

src/Juvix/Backends/Michelson/Compilation/Term.hs
```diff
@@ -7,167 +7,276 @@ import Juvix.Backends.Michelson.Compilation.Prim
...
+    J.App _ _ → do
+      -- TODO: figure out how to do stack check here
+      let (lam, args) = argsFromApps ann
+      case lam of
+        (J.LamM captures arguments body, _usage, lamTy) → do
+          insts ← evaluateAndPushArgs arguments lamTy args paramTy
+          recurseApplication (captures, arguments, body) lamTy args insts paramTy
+        ann@(J.Prim prim, _, primTy) → do
+          -- Treat the primitive as a function with n arguments, the body will eventually be inlined (or packed in a lambda).
+          let arguments = replicate (arity prim) "_"
```
cwgoes: I think we decided we needed to reason about ordering here & whether it will be correct (or if not, how to fix it)

src/Juvix/Backends/Michelson/Compilation/Term.hs
```diff
@@ -7,167 +7,276 @@ import Juvix.Backends.Michelson.Compilation.Prim
...
+              ( M.SeqEx
+                  ( insts
+                      <> [ f,
+                           M.PrimEx
+                             (M.DIP (replicate argsL (M.PrimEx M.DROP)))
+                         ]
+                  )
+              )
+        Left (LamPartial ops captures remArgs body ty) → do
+          -- Fully applied case with a returned lambda: return the instructions followed by the lambda, defer evaluation of the body.
```
cwgoes: Here is where we might need to worry about deleting the arguments, since we evaluate them (`insts`) but don't necessarily drop them anywhere - I wonder if we need to track this on `LamPartial` (how many arguments to drop after inlining the body)

src/Juvix/Backends/Michelson/Compilation/Term.hs
```diff
@@ -7,167 +7,276 @@ import Juvix.Backends.Michelson.Compilation.Prim
...
+          -- Will this type be correct if we compile this to a lambda?
+          pure (Left (LamPartial (insts <> ops) captures remArgs body ty))
+    LT → do
+      -- Underapplied case, we don't yet have enough arguments to inline.
+      -- Here we are turning formerly bound arguments into captures.
+      let (evaluatedArgs, remainingArgs) = splitAt argsL lamArguments
+      lamTy ← dropNArgs lamTy argsL
+      pure (Left (LamPartial insts (captures <> evaluatedArgs) remainingArgs body lamTy))
```
cwgoes: Here I don't think we need to worry about dropping the arguments since the # of args will be correct later (is that right?)

src/Juvix/Backends/Michelson/Compilation/Term.hs
```diff
@@ -7,167 +7,276 @@ import Juvix.Backends.Michelson.Compilation.Prim
...
+    GT → do
+      -- Overapplied case, we must figure out what the body is and recurse.
+      let (args, extraApplied) = splitAt lamArgsL args
+      eitherFuncOp ← termToInstr body paramTy
+      case eitherFuncOp of
+        Right _ → failWith "invalid case"
+        Left (LamPartial ops parCaptures parArgs parBody parTy) → do
+          -- todo: are these captures correct
+          -- Will this type be correct if we compile this to a lambda?
+          recurseApplication (parCaptures, parArgs, parBody) parTy extraApplied (insts <> ops) paramTy
```
cwgoes: Will we correctly drop the initially evaluated arguments (`insts`) here (later on)? I'm not sure that we will

src/Juvix/Backends/Michelson/Compilation/Term.hs
```diff
@@ -7,167 +7,276 @@ import Juvix.Backends.Michelson.Compilation.Prim
...
+            modify @"stack" (cons (VStack.varE name (Just v), typeV))
+        pure (argEval : instrs)
+    )
+    []
+    namedArgs
+-- TODO ∷ find a monadic version of mapAccumR
 
-changesTop ∷ Stack → Stack → Bool
-changesTop post pre = drop 1 post == pre
+-- argTypeFromLam ∷
```
cwgoes: Can we just delete this commented-out function?

src/Juvix/Backends/Michelson/Compilation/VirtualStack.hs
```diff
@@ -0,0 +1,208 @@
...
+  deriving (Show, Eq, Generic)
+
+--------------------------------------------------------------------------------
+-- T operation functions
+--------------------------------------------------------------------------------
+
+ins ∷ (Elem, Untyped.Type) → (Int → Int) → T → T
+ins v f (T stack' size) = T (v : stack') (f size)
+
+-- | 'inT' determines if the given
```
cwgoes: determines if the given what?

src/Juvix/Backends/Michelson/Compilation/VirtualStack.hs
```diff
@@ -0,0 +1,208 @@
+-- |
+-- - Serves as a virtual stack over Michelson
+-- - This stack has a few properties
+--   + The values on this stack may or may not be on the real
+--     stack. However for convention this should be largely ignored,
+--     except when you wish to do an operation like pair
+--     * This can be fix in the future
```
cwgoes:
```suggestion
--     * This can be fixed in the future
```

test/Backends/Michelson.hs
```diff
@@ -34,10 +34,10 @@ backendMichelson ∷ T.TestTree
 backendMichelson =
   T.testGroup
     "Backend Michelson"
-    [ --identityFn,
-      --identityApp,
-      --identityApp2,
-      --identityExpr,
+    [ identityFn,
```
cwgoes: May want to comment these out again for now

src/Juvix/Backends/Michelson/Compilation/Lambda.hs
mariari: it seems we need to add paramTy also

src/Juvix/Backends/Michelson/Compilation/Term.hs
mariari: I think we drop as we are storing the first n arguments in the closure we are building up, so what is left to feed the lambda is the remaining args.

src/Juvix/Backends/Michelson/Compilation/Term.hs
mariari: I don't see where we drop any of this, correct

src/Juvix/Backends/Michelson/Compilation/Datatypes.hs
mariari: Done

mariari:
Currently tracing the `termToInstr=

We get something akin to this

```ocaml
let f a = (+) a

f 2 3
```

1. [Name a push 2, push 3] -- from <- insts
2. split into [2] [3]
3. recurse (+) a
4. [ Name (a, _1) push 2, Name _2 Push 3] -- from <- insts
...

We end up with [Push 2, Push 3, +]. Which is almost correct, namely we need to swap the first n arguments + needs.

Also we need to consider what happens in a partial application case on a primitive function, I think due to inconsistent naming that it would not end too well.

src/Juvix/Backends/Michelson/Compilation/Lambda.hs
cwgoes: indeed we do

cwgoes:
> Also we need to consider what happens in a partial application case on a primitive function, I think due to inconsistent naming that it would not end too well.

We could try giving the primitive functions named arguments, then reordering them on the stack in the order of the named arguments (e.g. alphabetical with `_0`, _1`, etc.) prior to inlining the primitive operation, I think that would work.

cwgoes:
Also, names `_0`, `_1`, etc. will need to be uniquely namespaced for each function to deal correctly w/partial application.

PR #309, opened 2020-01-23
Made the compilation file, and outlined our current strategy and why the vstack

mariari:
renderered version can be seen [here](https://github.com/cryptiumlabs/juvix/blob/mariari/compilation-documentation/doc/Architecture/Compilation.org)

### DIFF
--- a/doc/Architecture/Compilation.org
+++ b/doc/Architecture/Compilation.org
@@ -1,0 +1,62 @@
+* Compilation
+** Michelson
+Michelson is an interpreted stack based language on-top of the Tezos
+chain.
+
+Since we are compiling to Michelson, some interesting considerations
+must be made
+*** Push/Enter vs Eval/Apply
+Although the drawbacks of Push/Enter compared to Eval/Apply are well
+known [fn:1] [fn:2], for an interpretive stack based languages, the
+caveats in that statement go away.
+1. _Eval/Apply is easier to compile to a normal C like language_
+   - Michelson is a stack based language, this point does not apply
+2. _Push/Enter spread complexity around when trying to make a fast compiler_
+   - Michelson is an interpreted language, and in particular, creating
+     lambdas (closures) for over compilation is quite costly. What
+     would be cheaper for this language would be to simply push the
+     arguments onto the stack for the new function to consume, and
+     inline said function.
+*** Compilation Strategy
+Currently annotated Core is translated directly into Michelson
+
+This is less than ideal for more than a few reasons, and will be
+changed at a current release.
+
+However until that time, we will document what is *currently* being
+done and how to improve it
+
+One last note is that names are assumed to not be overlapping, as this
+happens higher up in the compilation process
+**** Virtual Stack
+Currently in compiling Michelson we have what we labeled the =Virtual
+Stack= (vstack), This stack could all be in the real stack or none of it could
+be. What could be stored on the vstack is all the various constants,
+as these would waste computation if passed around normally.
+
+This vstack was designed so that, one could easily reason about the
+stack without having to worry about “If I dup this, would it really
+dup it, or just dup what is on top of the *real* stack?”
+
+Thus primitives have to be aware of this virtual stack.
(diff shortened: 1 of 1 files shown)


doc/Architecture/Compilation.org
```diff
@@ -0,0 +1,62 @@
+* Compilation
+** Michelson
+Michelson is an interpreted stack based language on-top of the Tezos
```
cwgoes:
```suggestion
Michelson is an interpreted stack based language which runs on the Tezos
```

Also, perhaps link to it?

doc/Architecture/Compilation.org
```diff
@@ -0,0 +1,62 @@
...
+must be made
+*** Push/Enter vs Eval/Apply
+Although the drawbacks of Push/Enter compared to Eval/Apply are well
+known [fn:1] [fn:2], for an interpretive stack based languages, the
+caveats in that statement go away.
+1. _Eval/Apply is easier to compile to a normal C like language_
+   - Michelson is a stack based language, this point does not apply
+2. _Push/Enter spread complexity around when trying to make a fast compiler_
+   - Michelson is an interpreted language, and in particular, creating
+     lambdas (closures) for over compilation is quite costly. What
```
cwgoes: Do you mean overevaluation?

doc/Architecture/Compilation.org
```diff
@@ -0,0 +1,62 @@
...
+     would be cheaper for this language would be to simply push the
+     arguments onto the stack for the new function to consume, and
+     inline said function.
+*** Compilation Strategy
+Currently annotated Core is translated directly into Michelson
+
+This is less than ideal for more than a few reasons, and will be
+changed at a current release.
```
cwgoes: Changed in a future release?

doc/Architecture/Compilation.org
```diff
@@ -0,0 +1,62 @@
...
+**** Virtual Stack
+Currently in compiling Michelson we have what we labeled the =Virtual
+Stack= (vstack), This stack could all be in the real stack or none of it could
+be. What could be stored on the vstack is all the various constants,
+as these would waste computation if passed around normally.
+
+This vstack was designed so that, one could easily reason about the
+stack without having to worry about “If I dup this, would it really
+dup it, or just dup what is on top of the *real* stack?”
+
```
cwgoes: Is there a particular case where constant propagation would work with a virtual stack but not with an earlier (pre-stack) inlining pass? Can we talk about that case (or cases)?

doc/Architecture/Compilation.org
```diff
@@ -0,0 +1,62 @@
...
+This is less than ideal for more than a few reasons, and will be
+changed at a current release.
+
+However until that time, we will document what is *currently* being
+done and how to improve it
+
+One last note is that names are assumed to not be overlapping, as this
+happens higher up in the compilation process
+**** Virtual Stack
```
cwgoes: Separately, we should talk about how we track the (real) Michelson stack while compiling (currently in order to map names to stack positions & look up types of things)

doc/Architecture/Compilation.org
mariari: yeah you are correct, changing it now!

doc/Architecture/Compilation.org
mariari: Fixed and thank you

doc/Architecture/Compilation.org
mariari: Sure I can add a bit. But basically it boils down to maybe do some smart compilation and then trying to interpret 

```factor
: add +
```

We may compile this ahead of a time, and not be sure of constants. So if we do this with a vstack, this is fine, and we can do some smart compilation without doing extra analysis.

doc/Architecture/Compilation.org
mariari: Sure will do

doc/Architecture/Compilation.org
mariari: Done!

PR #311, opened 2020-01-27
Move org generation code to separate repository

mariari:
Please everyone try to download ros and run 
```bash
ros install cryptiumlabs/org-generaiton
```

If this works for everyone, then we can update the contributing document and merge this


cwgoes:
> Installation works for me. (I don't get an `org-generation` executable yet but it sounds like that's still in progress?)

I do get a script in `~/.roswell/bin/org-generation`.

andy-morris:
Oh sorry me too, I misread the roswell instructions and was looking in `~/.local/bin`

mariari:
Sorry my bad, I should have clarified where the binary would be

PR #312, opened 2020-01-28
Starting the Language Design Choices

mariari:


doc/Language/Proof-Search.org
```diff
@@ -0,0 +1,47 @@
+* Proof Searching of Juvix
+- Data Structures in Juvix should provide complimentary proofs that
+  are relevant to the users.
+** The Argument for Implicit Refinements Over Indexed Types For Structures
```
cwgoes: This looks like a bit about what the standard library should provide? I think we might want a file specifically for the stdlib.

doc/Language/Interactive.org
```diff
@@ -0,0 +1,3 @@
+* Interaction of the Juvix Programming Language
+In many respects Juvix should
```
cwgoes: should what?

What a cliffhanger!

doc/Language/Proof-Search.org
mariari: We probably should have one, however I think this should stay here.

doc/Language/Interactive.org
mariari: That is for you to fill in, and for me to find out!

PR #314, opened 2020-02-01
Writing Token Contract with F*

mariari:
Currently this contract is not complete, but it should have enough form to understand the approach

### DIFF
--- a/experimental/examples/token/Token.fst
+++ b/experimental/examples/token/Token.fst
@@ -47,7 +47,7 @@ unopteq type storage = {
   accounts : a : accounts { total_supply = add_account_values a };
 }
 
-val emtpy_storage : storage
+val empty_storage : storage
 let empty_storage = {
   total_supply = 0;
   accounts = Map.empty;
@@ -66,17 +66,6 @@ let has_n accounts add to_transfer =
   match Map.select add accounts with
   | Some n -> to_transfer <= n
   | None   -> false
-
-// useless lemma, thought i needed it, but I just did to_transfer >= n.
-// in the function above by accident.
-val has_more : sto : storage
-             -> add : address
-             -> num : nat
-             -> Lemma
-               (requires (has_n sto.accounts add num))
-               (ensures  (num <= Some?.v (Map.select add sto.accounts)))
-let has_more sto add num = ()
-
 
 val account_sub : acc : accounts
                 -> add : address
--- a/experimental/examples/token/Token.fst
+++ b/experimental/examples/token/Token.fst
@@ -62,18 +62,18 @@ unopteq type token = {
 }
 
 unopteq type tx_transfer = {
-    from_account     : address;
-    to_account       : address;
-    transfer_ammount : nat
-  }
+  from_account    : address;
+  to_account      : address;
+  transfer_amount : nat
+}
 
 unopteq type tx_mint = {
-  mint_ammount    : nat;
+  mint_amount     : nat;
   mint_to_account : address
 }
 
 unopteq type tx_burn = {
-  burn_ammount      : nat;
+  burn_amount       : nat;
   burn_from_account : address
 }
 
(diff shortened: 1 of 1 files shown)


experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,157 @@
...
+let add_account_values accounts =
+ MapProp.fold (fun _key (v : nat) (acc : nat) -> v + acc) accounts 0
+
+
+unopteq type storage = {
+  total_supply : nat;
+  accounts : a : accounts { total_supply = add_account_values a };
+}
+
+val emtpy_storage : storage
```
cwgoes:
```suggestion
val empty_storage : storage
```

experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,157 @@
...
+}
+
+
+val has_n : accounts -> address -> nat -> bool
+let has_n accounts add to_transfer =
+  match Map.select add accounts with
+  | Some n -> to_transfer <= n
+  | None   -> false
+
+// useless lemma, thought i needed it, but I just did to_transfer >= n.
```
cwgoes: Shall we delete this then?

experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,157 @@
...
+val transfer_add : acc : accounts
+                 -> add : address
+                 -> num : nat
+                 -> Lemma
+                  (ensures ( add_account_values acc + num
+                           == add_account_values (account_add acc add num)))
+
+let transfer_add acc add num =
+  match Map.select add acc with
+  | Some balance -> admit ()
```
cwgoes: Looks like this needs a proof?

experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,157 @@
...
+                 -> num : nat
+                 -> Lemma
+                  (requires (has_n acc add num))
+                  (ensures (  add_account_values acc - num
+                           == add_account_values (account_sub acc add num)))
+let transfer_sub acc add num =
+  match Map.select add acc with
+  | Some balance ->
+    let remaining : nat = balance - num in
+    admit ()
```
cwgoes: Looks like this needs a proof?

cwgoes:
Let's also add another two transaction types, `mint` and `burn`, which *do* change the total supply, but only when authorised by a particular owner (another field in the token contract storage).

experimental/examples/token/Token.fst
mariari: Yeah, will this commit

experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,276 @@
...
+        accounts     = account_sub token.storage.accounts burn_from_account burn_amount
+      }}
+
+type transaction_error =
+  | Not_enough_funds
+  | Not_same_account
+  | Not_owner_token
+  | Not_enough_tokens
+
+val execute_transaction : token -> tx -> c_or transaction_error token
```
cwgoes: We should prove properties of this top-level function

experimental/examples/token/Token.fst
```diff
@@ -0,0 +1,276 @@
...
+let execute_transaction token tx =
+  match tx.tx_data with
+  | Transfer _ ->
+    if valid_transfer token tx
```
cwgoes: This reads storage once more than necessary

cwgoes: or perhaps leave this out of F* for now but take notes on the general problem

experimental/examples/token/Token.fst
mariari: That is done with the newest push

experimental/examples/token/Token.fst
mariari: We should probbaly leave it out, as it's not too exciting without say a reader monad (F* has them I think, just lacking effect polymorphism is a pain!)

A future PR will be had adding it to the documents.

Which folder should we use?

cwgoes: I think this is relevant to a standard library or examples section of language design choices - https://github.com/cryptiumlabs/juvix/pull/312.

PR #317, opened 2020-02-06
Attempts to get apply for Michelson correct

mariari:
- Currently this PR just includes removing cylic dependencies, so the parameterisation file can use functions in say `Term.hs`


mariari:
Since this mostly just shuffles around defintions I will merge this now after fixing hte merge issues

PR #322, opened 2020-02-12
Higher Level DSL for Michelson

mariari:


mariari:
Notes on Future implementation of this DSL


I think with partial application we will increment a number for things, so in
```ocaml
let f a = (+) {a}
in f c d
```

we get
```haskell
names <- traverse assignNamesAddToVStack args -- in this case a
pure Curr { fun = add, args, names, left = (2 - length names) }
``` 
maybe something like this?

add does what it should, we must also note the arguments and names this will take, so we can look it up properly and do swaps
 for a real computation say
(+) (* 2 b)
what we would do really do  is `2 * b` in the Instructions.

But when add gets it, it is a name, say "_1" which will look it up, and move/dup it to the front of the stack if it isn't already there

when we go back up, we see this, curry, and see if we have enoguh args, if so great then we name d for it, then have + lookup the names and the order will be protected by how this all works

cwgoes:
Generally this structure makes sense to me, we just have to be careful that our Haskell functions are always semantically equivalent to the Michelson ones they're simulating at compile time (namely, the functions must return the same output as their Michelson equivalents would for the entire domain).

src/Juvix/Backends/Michelson/Compilation/Prim.hs
```diff
@@ -28,34 +28,35 @@ primToInstr prim ty =
...
-      case const of
-        M.ValueNil → do
-          let ErasedAnn.PrimTy (Env.PrimTy t@(M.Type (M.TList elemTy) _)) = ty
-          modify @"stack" (VStack.cons (VStack.Val VStack.FuncResultE, t))
-          pure (M.PrimEx (M.NIL "" "" elemTy))
-        _ → do
-          let ErasedAnn.PrimTy (Env.PrimTy t) = ty
-          modify @"stack" (VStack.cons (VStack.Val VStack.FuncResultE, t))
-          pure (M.PrimEx (M.PUSH "" t const))
+  undefined
```
cwgoes: Still waiting on modifying this function?

src/Juvix/Backends/Michelson/DSL/Environment.hs
```diff
@@ -0,0 +1,128 @@
...
+  deriving (Show)
+
+newtype Fun = Fun (∀ m. Reduction m ⇒ [Types.NewTerm] → m Expanded)
+
+unFun ∷ Reduction m ⇒ Fun → [Types.NewTerm] → m Expanded
+unFun (Fun f) = f
+
+-- TODO ∷ have usage information stored with the args?
+-- May be fine without, as the terms we take ourselves should know their usage?
+data Curr
```
cwgoes: As in `curried`? `current`?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,402 @@
...
+              let constructed = Env.C
+                    { Env.fun = f,
+                      Env.argsLeft = names,
+                      Env.left = fromIntegral lPrim,
+                      Env.captures = Set.empty,
+                      Env.ty = ty,
+                      Env.argsApplied = []
+                    }
+              apply constructed args []
+            -- this should never happen, due to type checking??
```
cwgoes: Correct, we should never overapply primitives, throwing an error is the right behaviour.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,402 @@
...
+              Env.ty = ty,
+              Env.argsApplied = []
+            }
+       in apply constructed args []
+    Ann.Var _ → do
+      v ← inst form
+      case v of
+        Env.Curr c → apply c args []
+        -- these two cases would only be valid if we expanded into a Michelson
+        -- lambda, however would we ever do that?
```
cwgoes: As written, no, however we might have *variables* which are Michelson lambdas (e.g. passed as function arguments), do we need to handle that case here? Seems like we might.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,402 @@
...
+-- TODO ∷ drop extra things from the vstack, mainly
+-- 1. the function we move to the front
+-- 2. the expanded lambda we create
+-- Note that it isn't vital, as these aren't stored in the real Michelson stack
+-- Other things considered:
+-- We don't need to drop the arguments we eval and name, as they should be eaten
+-- by the functions they call with the appropriate usages
+apply ∷ Env.Reduction m ⇒ Env.Curr → [Types.NewTerm] → [Symbol] → m Env.Expanded
+apply closure args remainingArgs = do
+  let total_length = fromIntegral (length args + length remainingArgs)
```
cwgoes: Odd snake_case but OK

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,402 @@
...
+              Env.captures = foldr Set.insert (Env.captures closure) captured,
+              Env.ty = Env.ty closure,
+              Env.argsApplied = [],
+              Env.fun = Env.Fun $ \args →
+                Env.unFun
+                  (Env.fun closure)
+                  (args <> fmap makeVar (reverse captured))
+            }
+
+      consVal con (Env.ty closure)
```
cwgoes: Is it the type of the closure? Haven't we already applied some arguments?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,402 @@
...
+          evalArgsAndName
+          expanded ← app
+          recur expanded remainingArgs
+        LT → do
+          let notRemaining = length args - fromIntegral (Env.left closure)
+              newRemaining = drop notRemaining remainingArgs
+          evalArgsAndName
+          expanded ← app
+          recur expanded newRemaining
+        -- only case where we can't name all the args according to this schema!
```
cwgoes: Some more comments with examples would help with legibility here

src/Juvix/Backends/Michelson/Compilation/Prim.hs
mariari: This code will likely just be erased, once everything is setup

src/Juvix/Backends/Michelson/DSL/Environment.hs
mariari: Curried

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: We do handle this, if we are applying functions, we do not want a Michelson lambda. When a lambda is an argument to a primtive we promote it. The only other time a Michelson Lambda gets made is if we don't have enough arguments at the end of all these applications

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Yeah didn't realize I did that

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Yeah the type should be updated. Mind adding a pr to change it?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Sure, I'll add some examples

mariari:
> Structure generally looks correct, see comments, also:
> 
> * Maybe a silly question, but where in `apply` do we accumulate the actual Michelson instructions required to evaluate the arguments (which we will eventually sequence) (the equivalent to `ops` in the old `LamPartial`)?

This happens in the `inst` call, were it evals the form and names the arguments. Notice we eval the arguments right to left in apply, and then for primitve functions, we append the name of the var we just made to the application list

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Added examples, see last commit

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Note I don't have the logic for promote for lambdas online, I was hoping to delegate this to you, which would be using the DSL to cleanup and replace the existing one!

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: I have since updated the type, So we dispatch to the follwoing


```haskell
          con = Env.C
            { ...
              Env.ty = eatType (fromIntegral totalLength) (Env.ty closure),
            }
      consVal (Env.Curr con) (Env.ty con)
      pure (Env.Curr con)

eatType ∷ Natural → Types.Type → Types.Type
eatType 0 t = t
eatType x (Ann.Pi _ _ a) = eatType (pred x) a
eatType _ _ = error "Only eat parts of a Pi types, not any other type!"
```
Does this look okay to you?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,513 @@
...
+  abs,
+  isNat,
+  or ∷
+    Env.Reduction m ⇒ Types.Type → [Types.NewTerm] → m Env.Expanded
+add = intGen Instructions.add (+)
+mul = intGen Instructions.mul (*)
+sub = intGen Instructions.sub (-)
+and = onBoolGen Instructions.and (&&)
+xor = onBoolGen Instructions.xor (/=)
+or = onBoolGen Instructions.or (||)
```
mariari: For some of these, they are more polymorphic than what I've written here. They also operate on bytes and mutez...

Though mutez isn't in `V.Value'`?

Can these types not be computed at compile time?

src/Juvix/Backends/Michelson/DSL/Environment.hs
cwgoes:
```suggestion
data Curried
```

cwgoes: Alright, it might be clearer to name it `data Curried` then.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: `V.Value'` has an `Int`, which can be `Int`, `Nat`, or `Mutez` in Michelson. Morley has all of those in `Type`.

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -23,7 +23,7 @@ import Prelude (String)
 -- TODO: Add rest of primitive values.
 -- TODO: Add dependent functions for pair, fst, snd, etc.
 typeOf ∷ PrimVal → NonEmpty PrimTy
-typeOf (PrimConst v) = PrimTy (M.Type (constType v) "") :| []
+typeOf (Constant v) = PrimTy (M.Type (constType v) "") :| []
```
cwgoes: Given the `V.Value` `Int` / `Nat` / `Mutez` behaviour above; we may need to refactor this in a follow-up - https://github.com/cryptiumlabs/juvix/issues/326.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,513 @@
...
+apply closure args remainingArgs = do
+  let totalLength = fromIntegral (length args + length remainingArgs)
+  case totalLength `compare` Env.left closure of
+    EQ → do
+      -- usage and type doesn't matter here!
+      evalArgsAndName
+      app
+    LT → do
+      evalArgsAndName
+      -- make new closure, and apply args to the fun in the closure
```
cwgoes: Odd phrasing, you mean evaluate the arguments which have been applied so far and then return a closure which will take the remaining arguments then inline the function, correct?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -10,4 +10,513 @@
...
+          recur expanded newRemaining
+        -- In this case, we lack enough names for arguments since | args | > left
+        -- So we need to see how many arguments are left, this is `| args | - left`
+        -- we then need to reserve that many, eval right to left, then append the args
+        -- left to the remaining args, since they too need to be applied
+        GT → do
+          let unNamed = drop (fromIntegral $ Env.left closure) args
+          moreReserved ← reserveNames (fromIntegral (length unNamed))
+          traverseName (zip moreReserved unNamed)
+          traverseName (zip (Env.argsLeft closure) args)
```
cwgoes: Do we need to name `remainingArgs` on the stack here as well or not?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: Yes, that looks correct.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: Thanks!

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: Hmm, I don't quite follow - if lambdas are passed in as arguments, they will be equivalent (in core / the higher-level language) to native functions, so we will be *applying* them to things - do you mean that we'll promote them to this `Env.Curr` form? I'm not sure that we can, and in any case we will need logic to actually call `EXEC`, which we wouldn't do with `Env.Curr`, right?

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: `reaminingArgs` are already named, and so don't need any more names until we apply it

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Well maybe inline, it depends if we have enough arguments by the top level whether we can inline or not, but yes correct.

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Well if we pass them as an argument, to say map, then they should be `promoted` from `Env.Curr` to a real Michelson Lambda.

However if we are just passing around a function in the sense of
```ocaml
let f g = g 3
```
then, we would compile g as a `Env.Curr`, and then in f we will look up g (which is an `Env.Curr`) and then go through the apply logic.

if this is the top level and g takes two arguments, then we would `promote` this into a Michelson lambda.

As per calling `EXEC`, we have all the instructions in the Environment, so you could temporarly empty it, run the capture logic you need to, then create a Michelson lambda.


`Env.Curr` Does not deal with this as this is not `Env.Curr's` responsability

src/Juvix/Backends/Michelson/DSL/Environment.hs
mariari: Done

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: OK

src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: Right, I am wondering what we would do in the case of a lambda passed as an argument to a function in our contract (so one that we do not compile, but that we do evaluate).

mariari: It would compile correctly, as it'd dispatch to the application case on var, which looks it up from the env, and sees if we are LT GT or EQ in respect to how many arguments are left.

Thus if it has EQ number of args, it'd inline completely. There may be some slowdown if it's 

let f a = (+) a, since we have to move names a bit too much (I have this exact case written on paper) in the shuffles, but otherwise be optimal

cwgoes: Right, but in that case when do we use `EXEC`? `EXEC` isn't required for (normal) inlined functions which we're compiling (and aren't being provided as arguments), but it *is* required for lambda functions passed as arguments.

mariari: Ahh that's what you meant. This happens when we promote the top of the VStack. when we see a function like map, it'll put the top two arguments on the stack and promote them, meaning our lambda will turn into a real compiled lambda.

This is the logic I hope you to write (as it's the promotoing our curried -> Michelson Lambda)

cwgoes: This is for the case of external parameters, not functions we compile - we should change [this](https://github.com/cryptiumlabs/juvix/blob/07ec900f4ece3e57d8b4eed5de66fcf8a3eb91c2/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs#L178).

cwgoes: We can expect external Michelson functions to either be curried or use an explicit tuple type.

cwgoes:
- [ ] Figure out how to deal with lambdas in storage (`EXEC`)
- [ ] Clean up old code

cwgoes:
Definitely in the PR
- [ ] Figure out how to deal with lambdas in storage (`EXEC`)
- [ ] Finish parameterisation, including apply
- [ ] Alter test-cases to compile, then interpret, using the Michelson interpreter
- [ ] Run the test-cases and ensure that they work
- [ ] Find any todos / undefineds remaining & clean them up

Maybe in this PR, or alternatively later
- [ ] Deal with usages
- [ ] Deal with datatypes & pattern-matching, a bit blocked on frontend hookup

PR #323, opened 2020-02-13
Thealmarty/token proof idris

thealmarty:
(Please view file instead of the diff. The diff is more messy.)
Note that this isn't proof because the functions do checks each time. 
- This could be a good way for users to get safety without having to write proofs.
- The entrypoints/functions with safety will have higher gas cost because each operation performs more checks. But it solves the problem of non-separability of contract and proofs in the F* contract.)
- This would be good for invariants that are really important because it checks the specific case - writing proofs has risks of capturing the wrong facts!
- This method would be handy for really difficult proofs - again still getting the safety guarantee without having to prove anything. 
- we can write lib/functions to assist this.


thealmarty:
Test results: (not sure why outputs via provenAction are a bit different)
```haskell
*Token_noTransferFrom> createAccount "asdf" 500 initStorage
Right (MkStorage (M 1 (Branch2 (Leaf "asdf" 500) "asdf" (Leaf "qwer" 500))) 1 1000 "Cool" "C" "qwer") : Either Error Storage

*Token_noTransferFrom> provenAction (createAccount "asdf" 500) initStorage
Right (MkStorage (insert "asdf"
                         (plus (case lookup "asdf" (M 0 (Leaf "qwer" 1000)) of   Nothing => 0 Just balance => balance) 500)
                         (insert "qwer" (minus (case lookup "qwer" (M 0 (Leaf "qwer" 1000)) of   Nothing => 0 Just balance => balance) 500) (M 0 (Leaf "qwer" 1000))))
                 1
                 1000
                 "Cool"
                 "C"
                 "qwer") : Either Error Storage

*Token_noTransferFrom> provenAction (performTransfer "qwer" "asdf" 500) initStorage
Right (MkStorage (insert "asdf"
                         (plus (case lookup "asdf" (M 0 (Leaf "qwer" 1000)) of   Nothing => 0 Just balance => balance) 500)
                         (insert "qwer" (minus (case lookup "qwer" (M 0 (Leaf "qwer" 1000)) of   Nothing => 0 Just balance => balance) 500) (M 0 (Leaf "qwer" 1000))))
                 1
                 1000
                 "Cool"
                 "C"
                 "qwer") : Either Error Storage

*Token_noTransferFrom> performTransfer "qwer" "asdf" 500 initStorage
Right (MkStorage (M 1 (Branch2 (Leaf "asdf" 500) "asdf" (Leaf "qwer" 500))) 1 1000 "Cool" "C" "qwer") : Either Error Storage          
```

mariari:
Hey Marty, thanks for writing this up.

I have notes

> The entrypoints/functions with safety will have higher gas cost because each operation performs more checks. But it solves the problem of non-separability of contract and proofs in the F* contract.

Well The proof in the F* contract is more along the lines of the curry howard correspondence, where inhabiting the type is the proof of the existance of it.

So for example, having having 3 is a proof of a Natural, and giving a program with the type `Int → Int` is a proof of the type Int to Int.

So in
```fstar
unopteq type storage = {
  total_supply : nat;
  accounts : a : accounts { total_supply = add_account_values a };
}
```
the depdent/refinement types just add on an invariant, we could however write this

```fstar
unopteq type storage = {
  total_supply : nat;
  accounts     : accounts;
}

type storage_secure =
  s : storage { s.total_supply = add_account_values s.accounts }

```

which would separate out the inhabitance proof, so a normal storage could have a balance that is out of whack.

However, since total_supply is an optimization, we should always make sure our facts don't go out of whack! so the storage type would have to have comments around it in all functions saying the invariant to be careful


> This would be good for invariants that are really important because it checks the specific case - writing proofs has risks of capturing the wrong facts!

Well if we know what to check for via some boolean checks like this, then you know what you want to prove!

> This method would be handy for really difficult proofs - again still getting the safety guarantee without having to prove anything.

I'm quite interested, mind elaborating on this?

experimental/examples/token/Token_noTransferFrom.idr
```diff
@@ -49,42 +41,44 @@ getAccount address accounts = case lookup address accounts of
...
+||| provenTotalSupplyAction checks that totalSupply is conserved from running the input function.
+||| @fn the input function (e.g., performTransfer) and all its input arguments except the storage
+||| @storage the storage input to fn
+provenAction : (fn : (Storage -> Either Error Storage)) -> (storage : Storage) -> Either Error Storage
+provenAction fn storage =
+ let result = fn storage in
+ case result of
+   Left _ => result
+   Right newStorage =>
+     if invariants storage newStorage then result
```
cwgoes: This code will check the invariants at run-time, but we rather want to check them at compile-time (where possible) - try constructing a new `Storage` datatype which is indexed over the total supply, much like vectors are indexed over their lengths (see [here](http://docs.idris-lang.org/en/latest/tutorial/typesfuns.html#vectors)), and also enforces in the type that the sum of account balances is equal to the total supply index.

cwgoes:
> The entrypoints/functions with safety will have higher gas cost because each operation performs more checks. But it solves the problem of non-separability of contract and proofs in the F* contract.

Conceivably one could have an indexed version of a datatype & a non-indexed version, and somehow write the same functions over them (but only receive proofs for the indexed version), this seems like it will require a lot of typeclass-fu though.

thealmarty:
> Well The proof in the F* contract is more along the lines of the curry howard correspondence, where inhabiting the type is the proof of the existance of it.
> 
> So for example, having having 3 is a proof of a Natural, and giving a program with the type `Int → Int` is a proof of the type Int to Int.
> 
> So in
> 
> ```fstar
> unopteq type storage = {
>   total_supply : nat;
>   accounts : a : accounts { total_supply = add_account_values a };
> }
> ```
> 
> the depdent/refinement types just add on an invariant, we could however write this
> 
> ```fstar
> unopteq type storage = {
>   total_supply : nat;
>   accounts     : accounts;
> }
> 
> type storage_secure =
>   s : storage { s.total_supply = add_account_values s.accounts }
> ```
> 
> which would separate out the inhabitance proof, so a normal storage could have a balance that is out of whack.
> 
> However, since total_supply is an optimization, we should always make sure our facts don't go out of whack! so the storage type would have to have comments around it in all functions saying the invariant to be careful

I see. There's probably a way to do it in the F* contract to abstract out the functions like in the Idris one? What I like about this is that we can provide a library for users to import and they can "turn on" safety checks if they please. And if they don't care about the proof then they don't turn it on and can get the min gas cost. Does this make sense? 

If you don't mind, I'd like to see the equivalent of this in F*, with the proofs separated and the functions abstracted. Because that would be strictly superior to the Idris one, since the checks will be at compile time. (As Chris suggested I've try using dependent type to turn the checks to compile time in Idris too.)

> Well if we know what to check for via some boolean checks like this, then you know what you want to prove!

For this case yes - I think for a lot of cases too. What I mean here is that it isn't abstract at all like this.
> 
> > This method would be handy for really difficult proofs - again still getting the safety guarantee without having to prove anything.
> 
> I'm quite interested, mind elaborating on this?

So sometimes it could be really difficult to prove some facts for all x. But what I have here checks the specific run instance - if it runs, it's proven for that instance. No need to prove for all x.

experimental/examples/token/Token_noTransferFrom.idr
thealmarty: Will do.

thealmarty:
> Unfortunately it is just not likely to be possible in general to check invariants (even if you're willing to pay a bit of extra gas), because some of them will be so expensive (such as the sum of account balances being equal to total supply, which requires iterating over all accounts to check).
> 
> Conceivably one could have an indexed version of a datatype & a non-indexed version, and somehow write the same functions over them (but only receive proofs for the indexed version), this seems like it will require a lot of typeclass-fu though.

It's true, one would not want to check a certain subset of invariants with this because of gas cost. I think we can think of this as a way to complement the actual proof system. It's not a replacement. I think it's very little cost to add a library like this and I've shown possible use cases for it in this contract, yes?

mariari:
> > Well The proof in the F* contract is more along the lines of the curry howard correspondence, where inhabiting the type is the proof of the existance of it.
> > So for example, having having 3 is a proof of a Natural, and giving a program with the type `Int → Int` is a proof of the type Int to Int.
> > So in
> > ```fstar
> > unopteq type storage = {
> >   total_supply : nat;
> >   accounts : a : accounts { total_supply = add_account_values a };
> > }
> > ```
> > 
> > 
> > the depdent/refinement types just add on an invariant, we could however write this
> > ```fstar
> > unopteq type storage = {
> >   total_supply : nat;
> >   accounts     : accounts;
> > }
> > 
> > type storage_secure =
> >   s : storage { s.total_supply = add_account_values s.accounts }
> > ```
> > 
> > 
> > which would separate out the inhabitance proof, so a normal storage could have a balance that is out of whack.
> > However, since total_supply is an optimization, we should always make sure our facts don't go out of whack! so the storage type would have to have comments around it in all functions saying the invariant to be careful
> 
> I see. There's probably a way to do it in the F* contract to abstract out the functions like in the Idris one? What I like about this is that we can provide a library for users to import and they can "turn on" safety checks if they please.
> 


Well the issue with this, is that there is no point to really remove the said properties. They don't really make writing functions on it harder. In fact it would make it easier, as the programmer's hand is forced to obey the laws.

A really good example of this to play is from [Yaron Minsky's effective ML talk](https://blog.janestreet.com/effective-ml-video/) In which he gives the following example

```ocaml
type connection_state =
| Connecting
| Connected
| Disconnected

type connection_info = {
  state: connection_state;
  server: Inet_addr.t;
  last_ping_time: Time.t option;
  last_ping_id: int option;
  session_id: string option;
  when_initiated: Time.t option;
  when_disconnected: Time.t option;
}
```

When writing functions for this type it gets very confusing because one could have a weird mix of all these maybes. However if one thinks about the form a bit at the data type construction time and remove all the maybes one ends up with

```ocaml
type connecting = { when_initiated: Time.t; }
type connected = { last_ping : (Time.t * int) option;
                   session_id: string; }
type disconnected = { when_disconnected: Time.t; }

type connection_state =
| Connecting of connecting
| Connected of connected
| Disconnected of disconnected

type connection_info = {
  state : connection_state;
  server: Inet_addr.t;
}

```
in which now writing functions on it is obvious.

I'd conjecture writing a refinmenet type like with  `storage_secure` is much the same. As I know, that no matter what my functions do, as long as I have the `storage_secure` type, my invariants is upheld (much like not having `when_disconnected` and `when_initiated` at the same time)!


> And if they don't care about the proof then they don't turn it on and can get the min gas cost. Does this make sense?

The proof doesn't cost anything! In fact having these refinement types should make the program faster and allocate less, since I don't have to allocate an either if the transfer works.

Instead if the properties hold up before hand, we can avoid uneeded allocations and checks! Thus we can abuse this to make programs simpler (like replace a fold with a map, as we don't have to worry about erroring half way through!).

> If you don't mind, I'd like to see the equivalent of this in F*, with the proofs separated and the functions abstracted. 

For the functions themselves, there isn't really much to abstract, There is a bit of repeat code in the type signatures itself, though the main length of the code comes in from avoiding long lines, and newlining on every `->` in a function type for readability (this accounts for over 10% of the code length!). 

Well since the type itself is the proof, we don't really need to write many proofs!

The issue is that once you start using `storage` instead of `storage_secure` in your types everywhere, you now have an explicit proof burden that the user and you must fill.

For example in

```idris
performTransfer from dest tokens storage =
  let fromBalance = getAccount from (accounts storage)
      destBalance = getAccount dest (accounts storage) in
        case lte tokens fromBalance of
             False => Left NotEnoughBalance
             True => ...
``` 

How do I know if the lte is correct here? Maybe the order got swapped (I accidently did this in my F* code, but the compiler yelled at me!), and so when we hit the `True` branch, we could be doing something like {from_acc = (4 - 6 = 0); to_account = (n + 6)}, and thus the total_supply is the same but it has changed in the map.

Thus the programmer has to take time out of their day to explicitly prove properties they get practically free by just having the correct invariant. It is also important to note that putting these invariants on the structure itself is fairly simple and is something someone writing the library would naturally just assume in a non dependently typed language!  Thus, nothing is really gained here by removing the refinement. 

Sometimes for harder invariants one doesn't use a refinement type on the structure itself, but on the functions itself, or even accept a failure inside the function returning some kind of either result.

> Because that would be strictly superior to the Idris one, since the checks will be at compile time. (As Chris suggested I've try using dependent type to turn the checks to compile time in Idris too.)



> 
> > Well if we know what to check for via some boolean checks like this, then you know what you want to prove!
> 
> For this case yes - I think for a lot of cases too. What I mean here is that it isn't abstract at all like this.


Hmm, I'm not sure what you mean. All the runtime checks boil down to boolean checking at the end of the day. You can simply promote those checks right before the function call, much like I did for the transfer/burn/mint function

```fstar
    if valid_transfer token tx
    then Right (transfer token tx)

val valid_transfer : token -> tx -> bool
let valid_transfer token tx =
  match tx.tx_data with
  | Transfer {from_account; transfer_amount} ->
    has_n token.storage.accounts from_account transfer_amount
    && tx.tx_authroized_account = from_account
  | Mint _ | Burn _ ->
    false

val transfer : tok : token -> tx : tx { valid_transfer tok tx } -> token
let transfer token transaction =
```

This also gives you the benefit, in that if you check for a tighter invariant, you can run this without having to recheck safely!

> 
> > > This method would be handy for really difficult proofs - again still getting the safety guarantee without having to prove anything.
> > 
> > 
> > I'm quite interested, mind elaborating on this?
> 
> So sometimes it could be really difficult to prove some facts for all x. But what I have here checks the specific run instance - if it runs, it's proven for that instance. No need to prove for all x.

Ahh, but that is precisely a refinement type is!

for a function like 
```f*
val sub_pos : n : nat -> m : nat { m >= n } -> nat
let sub_pos n m = m - n
```
if we ask if m > n at run time (or we know it via other facts!) then we can run this. Thus you just need to prove for this instance to get it!

thealmarty:
re @mariari You've got some good arguments here for refinement types, it'd be good to document these somewhere else. 

What I've demonstrated here is a different approach, a clean way to add safety.

PR #328, opened 2020-02-21
Cleanup of the Codebase

mariari:
This PR aims to get most lines under 80 characters long, and cleanup various patterns in the codebase.

There are still some files to cleanup but progress is being made

Other changes that have yet to be had

1. [ ] Change Tuples along param to a proper data structure with names


mariari:
The latest commit changes things like

```haskell
Nat.NatTy -- --> Nat.Ty
nat -- ---> Nat.t

```
Since we always import qualified, no need to be repetitive

mariari:
Last commit changed the config type to T, and qualified the import of config

mariari:
This could probably be merged now, what needs to be done is

1) [ ] change the param tuples into a proper type

2) [ ] format tests to be around 80 characters long

PR #335, opened 2020-02-24
Language BNF

mariari:
This PR starts the Language BNF.

This is needed before a parser can be made

### DIFF
--- a/doc/Frontend/ideas.org
+++ b/doc/Frontend/ideas.org
@@ -84,6 +84,15 @@ syntax.
              natural to anyone who has been using the language but
              with functions only. There is no bifurcation of syntax,
              where modules look like this special thing (ala OCaml).
+
+- Point redacted :: it seems the first syntax does not require special
+                    casing.
+                    #+BEGIN_SRC haskell
+                    <name> =
+                      foo a = 3
+                    #+END_SRC
+                    in the above example, it's unambiguous what <name>
+                    is, as expression ≠ top level!
 ** Like Functions
 The like functions syntax would look something like this
 #+BEGIN_SRC idris
@@ -112,10 +121,6 @@ The like functions syntax would look something like this
   2. Modules look like functions and thus familiar
      - Lower barrier of entry and confusion and forgetting of syntax
 - _Cons_
-  1. We *may* have to reserve capitals for modules and type
-     constructors alone
-     - This *may* mean type names will have to be lower case like in
-       F^{★} or OCaml
   2. Module signature is special cased
   3. We may need an end for the =Module= keyword
      - This depends on how indent sensitive we wish to be
@@ -143,7 +148,6 @@ The Haskell/Idris like syntax would look like
 - The functor and signature syntax could use some refinement
 - _Pros_
   1. Looks Familiar to existing ML solutions
-  2. We have the freedom of using capital names for type names
 - _Cons_
   1. requires a lot of special casing
      - Higher barrier to entry, and forgetting of syntax
--- a/doc/Frontend/ideas.org
+++ b/doc/Frontend/ideas.org
@@ -1,7 +1,4 @@
-
-
 * Usage Syntax
-
 ** Inside Arrows (Jeremy)
 *** Arrows
 **** Initial Idea
@@ -78,7 +75,81 @@
   + cons
     * A few extra characters for every signature
     * How does sugar get added from this system? (Jeremy)
+* Module Syntax
+Currently there are two roads we can go down in regards to module
+syntax.
 
+- mariari :: I personally prefer the "like functions" syntax as it
+             seems writing modules are encouraged, and feel very
+             natural to anyone who has been using the language but
+             with functions only. There is no bifurcation of syntax,
+             where modules look like this special thing (ala OCaml).
+** Like Functions
+The like functions syntax would look something like this
+#+BEGIN_SRC idris
+  Foo : Module
+    type a
+    foo : a -> a
+  Foo =
+    type a = int
+
+    foo : int -> int
+    foo c = c + 2
+
+  type sig = Module
+    type a
+    foo : a -> a
+    val : int -> a
+
+  Foo : sig -> Module
+  Foo A =
+    type t = A.t
+    val x = A.(foo (int x))
+#+END_SRC
+- _Pros_
+  1. Typing Modules is lightweight
+     - Thus encourages module use!
+  2. Modules look like functions and thus familiar
+     - Lower barrier of entry and confusion and forgetting of syntax
+- _Cons_
+  1. We *may* have to reserve capitals for modules and type
+     constructors alone
+     - This *may* mean type names will have to be lower case like in
+       F^{★} or OCaml
+  2. Module signature is special cased
+  3. We may need an end for the =Module= keyword
+     - This depends on how indent sensitive we wish to be
+** Haskell/Idris Like
+The Haskell/Idris like syntax would look like
+#+BEGIN_SRC idris
+  module Foo sig
+    type a
+    foo : a -> a
+  module Foo where
+    type a = int
+
+    foo : int -> int
+    foo c = c + 2
+
+  module t sig
+    type a
+    foo : a -> a
+    val : int -> a
+
+  module Foo (A : t) where
+    type t = A.t
+    val x = A.(foo (int x))
+#+END_SRC
+- The functor and signature syntax could use some refinement
+- _Pros_
+  1. Looks Familiar to existing ML solutions
+  2. We have the freedom of using capital names for type names
+- _Cons_
+  1. requires a lot of special casing
+     - Higher barrier to entry, and forgetting of syntax
+  2. A lot more special casing in the implementation
+  3. Heavier syntax
+     - This means module usage is more discouraged
 * Additive conjunction
 ** Initial ideas (Jeremy)
 - The literature uses `&` for this, however to be less confusing for the user I suggest this
@@ -92,3 +163,4 @@
   + Note :: _1 is a placeholder for a better name of accessing the
             first variant without being able to inspect on the second
   + Note :: in the unicode extension ∨ can replace `\/`
+* If/Cond Syntax
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -13,8 +13,9 @@
 
   ;;;; Functions ======================================================
 
-  function ::= <name> <args>+ = <expression>
-             | <name> <args>+ <cond-logic>*
+  ;; let until indent sensitivity is gone
+  function ::= let <name> <args>+ = <expression>
+             | let <name> <args>+ <cond-logic>*
 
   ;; Match-logic could just be a name, and so is thus fine
   ;; # is for implicit
@@ -22,12 +23,13 @@
   args ::= <match-logic>
          | #<match-logic>
 
+  ;; sig only here while indent sensitivity is not here
   ;; if we only have usage in arrows, we should allow a way to have
   ;; usage information of variables as well!
-  signature ::= <name>         :                 <arrow-type>
-              | <name>         : <type-name> =\> <arrow-type>
-              | <name> <usage> :                 <arrow-type>
-              | <name> <usage> : <type-name> =\> <arrow-type>
+  signature ::= sig <name>         :                 <arrow-type>
+              | sig <name>         : <type-name> =\> <arrow-type>
+              | sig <name> <usage> :                 <arrow-type>
+              | sig <name> <usage> : <type-name> =\> <arrow-type>
 
   ;;;; Types ==========================================================
 
@@ -109,10 +111,30 @@
   ;; they can take any value, however for examples, we will encourage the use
   ;; of taking (and thus parameterizing) modules
 
-  module ::= <name> <args>+ = <top-level>
-           | <name> <args>+ <cond-top>*
+  ;; let and end are there until indent sensitivity is in
+  module ::= let <name> <args>+ = <top-level>* end
+           | let <name> <args>+ <cond-top>* end
 
-  cond-top ::= \| <expression> = <expression>
+
+  ;; what should we allow in a module signature?
+  ;; begin and end are only here while we aren't indent sensitive
+
+
+
+  module-signature ::= <module-signature-helper> Module <sigs-and-types>+ end
+
+  sigs-and-types ::= <sig>
+                   | <module-signature>
+                   | <type>
+
+  module-signature-helper
+    ::= sig <name>         :                 <arrow-type>?
+      | sig <name>         : <type-name> =\> <arrow-type>?
+      | sig <name> <usage> :                 <arrow-type>?
+      | sig <name> <usage> : <type-name> =\> <arrow-type>?
+
+
+  cond-top ::= \| <expression> = <top-level>*
   ;;;; Types Classes ==================================================
 
   ;; Need end if we are indent sensitive!
@@ -135,14 +157,20 @@
                | <where>
                | <string>
                | <number>
-               ; TODO
+               | <lambda>
+               | <tuple>
+               | <parens>
+               | <application>
                | <symbol>
                | <prefix>
                | <infix>
+               ; This is useful for having nested do's or matchs
+               | <block>
                | <do>
-               | <tuple>
                | <comments>
-               | <parens>
+               ; TODO
+               | <record-update>
+               | <record-creation>
 
   usage ::= <expression>
 
@@ -150,6 +178,36 @@
 
   module-lookup ::= <module-name>.<name>
 
+  application ::= <name> <expressions>*
+
+  lambda ::= \\ <match-logic>* -\> <expression>
+
+  symbol ::= <name>
+
+  ;; useful for match, and nested do's!
+  block ::= begin <expression> end
+
+  infix ::= `<name>`
+
+  do ::= <do-body>*
+
+  do-body ::= <exprsesion> \; <expression>
+
+  prefix ::= (<infix-symbol>)
+
+  tuple ::= ( <command-tuple>* )
+
+  commad-tuple ::= <exprsesion> , <expression>
+
+  parens ::= ( <expression> )
+
+  comments ::= -- <any-text-not-new-line> \n
+             | <comments-rec>
+
+  comments-rec ::= <multi-comments>
+                 | {- <comments-rec> -}
+
+  multi-comments ::= {- <any-text-not-{-> -}
 
   ;;; Matching ===================================
 
@@ -234,19 +292,22 @@
   ;; ; is comment in bnf
   comma            ::= ,
   semi             ::= \;
-  name             ::= <ascii-non-reserved>
-  non-capital-name ::= <ascii-no-capital>
-  capital-name     ::= <ascii-capital>
+  name             ::= <utf8-non-reserved>
+  non-capital-name ::= <utf8-no-capital>
+  capital-name     ::= <utf8-capital>
   ;; may want m e or Map.t int string?
   type-name   ::= <name> <others-names>+
   other-names ::= <name>
                 | <universe-expression>
                 | ( <arrow-type> )
-  module-name      ::= <name> ; enforce capital names?
+
+  infix-symbol ::= <utf-infix-symbol>
+
+  module-name ::= <name> ; enforce capital names?
 
   constructor-name ::= <capital-name-and-symbols>
 
-  ascii-non-reserved ::=
-  ascii-no-capital   ::=
-  ascii-capital      ::=
+  utf8-non-reserved ::=
+  utf8-no-capital   ::=
+  utf8-capital      ::=
 #+END_SRC
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -56,8 +56,8 @@
   sum ::= \| <name>
         | \| <name> <product>
 
-  product ::= { <name-type-comma>+ <name-type>* } <type-name>
-            | { <name-type-comma>* }              <type-name>
+  product ::= { <name-type-comma>+ <name-type>* } -\> <type-name>
+            | { <name-type-comma>* }              -\> <type-name>
             | { <name-type-comma>+ <name-type>* }
             | { <name-type-comma>* }
             | <arrow-type>
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -69,6 +69,7 @@
   ;; should we move ()'s somewhere else?
   ;; # is implicit argument
   arrow-type ::= <type-refine>
+               | <name>  : <type-refine>
                | <name>  : <type-refine> <arrow> <arrow-type>
                | #<name> : <type-refine> <arrow> <arrow-type>
                | <type-refine>           <arrow> <arrow-type>
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -233,9 +233,10 @@
   non-capital-name ::= <ascii-no-capital>
   capital-name     ::= <ascii-capital>
   ;; may want m e or Map.t int string?
-  type-name        ::= <name>
-                     | <name> <type-name>
-                     | <universe-expression> <type-name>
+  type-name   ::= <name> <others-names>+
+  other-names ::= <name>
+                | <universe-expression>
+                | ( <arrow-type> )
   module-name      ::= <name> ; enforce capital names?
 
   constructor-name ::= <capital-name-and-symbols>
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -1,0 +1,14 @@
+
+* Syntax
+#+BEGIN_SRC bnf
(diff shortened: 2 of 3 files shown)


doc/Frontend/ideas.org
```diff
@@ -78,7 +75,81 @@
   + cons
     * A few extra characters for every signature
     * How does sugar get added from this system? (Jeremy)
+* Module Syntax
+Currently there are two roads we can go down in regards to module
+syntax.
+
+- mariari :: I personally prefer the "like functions" syntax as it
```
cwgoes: I generally agree, I actually wonder if there is a strong reason to distinguish "modules" and "records" at all - is there?

doc/Frontend/syntax.org
```diff
@@ -0,0 +1,126 @@
...
+         | <alias-declaration>
+
+  ; Should we let there be a refinement in adt type declaration?
+  data-declaration  ::= type <non-capital-name> : <arrow-type> = <adt>
+                      | type <non-capital-name>                = <adt>
+
+  type-declaration  ::= type <non-capital-name> = <name> <type-refine>
+  alias-declaration ::= type <non-capital-name> = <type-refine>
+
+  adt ::= <sum>*
```
cwgoes: Hmm this will look like inductive families, at least in the top-level frontend, not like Haskell ADTs - see the bottom of p.17 in the language reference.

doc/Frontend/syntax.org
```diff
@@ -0,0 +1,126 @@
...
+
+  usage ::= <expression>
+
+
+  record-access ::= <name>.<name>
+
+  module-lookup ::= <module-name>.<name>
+
+  ; we should remove either if or cond!?
+  if   ::= if   <cond-logic>*
```
cwgoes: I'm not sure if we even want to special-case this for now. Ideally we can have some "syntax" pragma like Idris.

doc/Frontend/ideas.org
mariari: The only reason is for syntactic beauty.

A record would have { } before it and ,/;'s to delineate separation.

Also records, have to be pre-declared, whereas a module is a bit more free form

doc/Frontend/syntax.org
mariari: These are alias declarations, not data declarations see,

```ebnf
  data-declaration  ::= type <non-capital-name> : <arrow-type> = <adt>
                      | type <non-capital-name>                = <adt>


  product ::= { <name-type-semi>+ <name-type>* }
            | { <name-type-semi>* }
            | <arrow-type>


  arrow-type ::= <type-refine>
               | <name> : <type-refine> <arrow> <arrow-type>
               | <type-refine>          <arrow> <arrow-type>
               | ( <arrow-type> )
```


namely the <arrow-type> part.

Though I should add

```ebnf
  product ::= { <name-type-semi>+ <name-type>* } <name-type>
            | { <name-type-semi>* } <name-type>
            | { <name-type-semi>+ <name-type>* }
            | { <name-type-semi>* }
            | <arrow-type>
```

This amounts to inductive family syntax.

doc/Frontend/syntax.org
mariari: Well we need some kind of if semantics.


notably

```ocaml
if b
then a
else c
```

is more special cased than

```haskell
if | b = a
   | else = c
```

Because the latter form already exists in the form of

```haskell
f b =
  | foo b = a
  | otherwise = c
```

doc/Frontend/ideas.org
cwgoes: Aye, makes sense.

doc/Frontend/ideas.org
```diff
@@ -78,7 +75,81 @@
...
+             with functions only. There is no bifurcation of syntax,
+             where modules look like this special thing (ala OCaml).
+** Like Functions
+The like functions syntax would look something like this
+#+BEGIN_SRC idris
+  Foo : Module
+    type a
+    foo : a -> a
+  Foo =
+    type a = int
```
cwgoes:
```suggestion
    a : Type
    a = int
```

doc/Frontend/ideas.org
```diff
@@ -78,7 +75,81 @@
...
+- mariari :: I personally prefer the "like functions" syntax as it
+             seems writing modules are encouraged, and feel very
+             natural to anyone who has been using the language but
+             with functions only. There is no bifurcation of syntax,
+             where modules look like this special thing (ala OCaml).
+** Like Functions
+The like functions syntax would look something like this
+#+BEGIN_SRC idris
+  Foo : Module
+    type a
```
cwgoes:
```suggestion
    a : Type
```

doc/Frontend/ideas.org
```diff
@@ -78,7 +75,81 @@
...
+    foo : a -> a
+  Foo =
+    type a = int
+
+    foo : int -> int
+    foo c = c + 2
+
+  type sig = Module
+    type a
```
cwgoes:
```suggestion
    a : Type
```

doc/Frontend/ideas.org
```diff
@@ -78,7 +75,81 @@
...
+    foo : a -> a
+    val : int -> a
+
+  Foo : sig -> Module
+  Foo A =
+    type t = A.t
```
cwgoes:
```suggestion
    t : Type
    t = A.t
```

doc/Frontend/syntax.org
cwgoes: Right right, guard syntax seems reasonable to me.

cwgoes: One note is that we'll want nice exhaustiveness checking (this might need to use proofs in the context, it's not possible in general of course).

cwgoes:
We should write down what implicit variables will look like.

I think @cyberglot's point on user-friendliness / familiarity is important to consider; if upper-case types helps with UX we should consider it for that reason alone.

doc/Frontend/ideas.org
cwgoes: Maybe we should make it more different, if it can't be unified with records.

mariari:
> We should write down what implicit variables will look like.
> 
> I think @cyberglot's point on user-friendliness / familiarity is important to consider; if upper-case types helps with UX we should consider it for that reason alone.

After thinking about it some more, In either case, they can all live in the same name-space.

It turns out to not be ambiguous in the former case, as consider the following

```haskell
<name> = 
  foo = 3
```

In this,  regardless of the case, the parser can tell we have a top level value, and determine that <name> must be a module.

In all other regards to whether types can be capital or not, is not related to the syntax of the modules themselves.

Really it's quite interesting, the real question now becomes. Do we have the case of the word be by convention only (I'm in favor of this), or strictly enforced in the parsing? For the former we'd have a lexer tell the user, don't create a type with lowercase and what not.

mariari:
Everything should be done now, except for named arguments

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
+Map = Data.Map
```
cwgoes: We are assuming inference, here, yes?

(so there could also be a signature `Map : Type`)

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+
+-- common module interfaces for data structures in Common?
+MonoidMeasure : Common.Measure
+MonoidMeasure =
+  type T = Nat
+
+  -- (<>) is normal semi-group  type class function
```
cwgoes: Are we assuming that the semi-group typeclass is automatically in scope? Should we `import Prelude` ?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+-- need OrdMap, as modules that one goes to can't be modules?
+Accounts = Map.OrdMap Data.String MonoidMeasure
+
+-- we send in the string module, as it has a total order
+-- function named compare, that is also part of a general type class
+Token =
+  type Address  = s : String.T {String.length s == 36}
+
+  type Storage = {
+    total-supply : Nat.T;
```
cwgoes: Are we going to allow `-` in identifiers? Will we be very permissive (like Agda) or more restrictive?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+    accounts     : Accounts.T { Accounts.measure-value == total-supply }
+  }
+
+  empty-storage : Storage
+  empty-storage = {
+    total-supply = 0;
```
cwgoes: Why `;` instead of `,` on records? I think we might want to reserve `;` for possible later use as monadic sequence.

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+    accounts     = Accounts.empty;
+  }
+
+  type T = {
```
cwgoes: Is this a record without a constructor? Seems like we might want a named constructor.

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+
+-- need OrdMap, as modules that one goes to can't be modules?
+Accounts = Map.OrdMap Data.String MonoidMeasure
+
+-- we send in the string module, as it has a total order
+-- function named compare, that is also part of a general type class
+Token =
+  type Address  = s : String.T {String.length s == 36}
+
+  type Storage = {
```
cwgoes: Is this a record without a named constructor? Is `Storage` the named constructor?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+    storage = Storage;
+    version = Nat.T;
+    name    = String.T;
+    symbol  = Char.T;
+    owner   = Address;
+  }
+
+
+Transaction =
+  type Transfer = {
```
cwgoes: Same comments / questions on records here.

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+           -> num : Nat.T {has-n acc add num}
+           -> Accounts.T
+account-sub accounts add number =
+  case Accounts.select accounts add of
+  | Just balance ->
+     Accounts.put accounts ~key:add ~val:(balance - number)
+
+account-add : Accounts.T -> Token.Address -> Nat.T -> Accounts.T
+account-add accounts add number =
+  Accounts.update accounts ~f:(number +) ~key:add
```
cwgoes: `number +` is sectioned operators?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+transfer-stor : stor   : Token.Storage
+              -> ~from : Token.Address
+              -> ~to   : Token.Address
+              -> num   : Nat.T {has-n stor.accounts from num}
+              -> Token.Storage
+transfer-stor stor ~from:add_from ~to:add_to num =
+  let new-acc = account-add (account-sub stor.accounts add_from) add-to num in
+  -- This is valid, because the monoid and measure of account track this
+  { total-supply = stor.total-supply
+  ; accounts     = new-acc
```
cwgoes: Same question on `;` usage

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+
+transfer-stor : stor   : Token.Storage
+              -> ~from : Token.Address
+              -> ~to   : Token.Address
+              -> num   : Nat.T {has-n stor.accounts from num}
+              -> Token.Storage
+transfer-stor stor ~from:add_from ~to:add_to num =
+  let new-acc = account-add (account-sub stor.accounts add_from) add-to num in
+  -- This is valid, because the monoid and measure of account track this
+  { total-supply = stor.total-supply
```
cwgoes: I like this record construction syntax and `.` for field access. Do we have a record update syntax?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+      token.owner == tx-tx-authorized-account
+    | Transaction.Burn _ | Transaction.Transfer _ ->
+      false
+
+  transfer token tx =
+    case tx.data of
+    | Transaction.Transfer {from-account; amount} ->
+      has-n token.storage.accounts from-account amount
+      && tx.authroized-account == from-account
+    | Transaction.Burn _ | Transaction.Mint _ ->
```
cwgoes: Is there a reason this match can't just be `_` ?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+  burn     : T
+  mint     : T
+Validation =
+  type T = Token.T -> Transaction.T -> Bool
+
+  mint token tx =
+    case tx.data of
+    | Transaction.Mint ->
+      token.owner == tx-tx-authorized-account
+    | Transaction.Burn _ | Transaction.Transfer _ ->
```
cwgoes: Is there a reason this match can't just be `_` ?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+      && tx.authroized-account == from-account
+    | Transaction.Burn _ | Transaction.Mint _ ->
+      false
+
+  Burn token tx =
+    case tx.data of
+    | Transaction.Burn {burn-from-account; burn-ammount} ->
+      has_n token.storage.accounts burn-from-account burn-amount
+      && tx.authroized-account == burn-from-account
+    | Transaction.Mint _ | Transaction.Transfer _ ->
```
cwgoes: Is there a reason this match can't just be `_` ?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+
+Operation =
+
+  type TokenTransaction f =
+    tok : Token.T -> tx : tx { f tok tx } -> Token.T
+
+  transfer : TokenTransaction Validation.transfer
+  transfer token transaction =
+    case transaction.data of
+    | Transfer {from-account; to-account; amount} ->
```
cwgoes: I generally like this record pattern match syntax modulo the question about `;`. Can it be used to match only a subset of fields (and only scope those)? That would be nice.

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+      { token
```
cwgoes: Is this the record update syntax? How would it work for updating more than one field?

experimental/juvix/contract.jvx
mariari: Well Map here, isn't of type Type, it's of type Module!

experimental/juvix/contract.jvx
mariari: I assume we have implicit prelude?

experimental/juvix/contract.jvx
mariari: I think we should be restrictive, but allow `-` as I personally find it more readable than camlCase or snake_case

experimental/juvix/contract.jvx
mariari: I think `;` is fine here. This doesn't ruin the use of `;` for monadic sequence, as these are very different contexts!

Though we should probably discuss this more in depth, as per pros and cons

experimental/juvix/contract.jvx
mariari: I would probably argue, that we shouldn't have one.

This is because if we add a constructor then, we cut off the ability to abstract. Imagine an example like this


```ocaml

type Foo = {
  i : int;
  b : string;
}

type Data1 = 
  | Foo Foo
  | B Int  

type Blah = 
  | Blah Foo
  | C String

f { i ; b } = to_string i ++ b

data1 = function
  | Foo foo = f foo
  | B i = to_string i

blah = function
  | Blah foo = f foo
  | c string = string
```

experimental/juvix/contract.jvx
mariari: `Storage` is a record type, it has no named consturctor, just constructed with {}. This allows for better abstractions in ADTs.

experimental/juvix/contract.jvx
mariari: Yes much like how Haskell works

experimental/juvix/contract.jvx
mariari: I just have thought of where, though we can have some kind of functional lens of `:=` along with another for updates

experimental/juvix/contract.jvx
mariari: We want exhaustion. Imagine if we updated the variant wtih another operation, maybe it is a kind of a subset of a mint. It's always important to be exhaustive and  try to never use _

experimental/juvix/contract.jvx
mariari: Same reason as above

experimental/juvix/contract.jvx
mariari: Yes, you should be able to, this is less of a syntax question, and more of what we want.

We should have flags to warn/error on that behavior though. Shops like jane street do that for safety when updating record

experimental/juvix/contract.jvx
mariari: with another
`<field> = <expression>`

experimental/juvix/contract.jvx
cwgoes: The questions are:
- How do we deal with two records with the same fields but distinct types?
- Do we want row polymorphism?
- Can we get row polymorphism without type theory alterations?

mariari: Here is an example from OCaml
```ocaml

utop # type foo = {a : int};;
type foo = { a : int; }

utop # type bah = {a : int};;
type bah = { a : int; }

utop # {a = 3};;
- : bah = {a = 3}

utop # let foo : foo = {a = 3};;
val foo : foo = {a = 3}

utop # let bah : bah = {a = 3};;
val bah : bah = {a = 3}
utop # foo == bah;;
Error: This expression has type bah but an expression was expected of type foo

utop # bah = { a = 3 };;
- : bool = true


utop # foo = { a = 3 };;
- : bool = true

```

It seems to allow the creation of records with the same fields.

It infers the last one made (a sane choice!)

It then disallows them being the same.

However thanks to nice type inference, if we make one on the spot it is the correct type.

experimental/juvix/contract.jvx
cwgoes: Ah right - still, should there be a signature?

experimental/juvix/contract.jvx
cwgoes: Hmm alright.

experimental/juvix/contract.jvx
cwgoes: Hmm, I don't have an opinion on the case question but I also don't think we should be too restrictive, except as prohibited by the parser, we can always add conventions later.

experimental/juvix/contract.jvx
cwgoes: Why not just `,` like Haskell?

I think `;` is confusing since it is very different than monadic sequence.

experimental/juvix/contract.jvx
cwgoes: Right, that looks pretty reasonable. Ocaml doesn't have row polymorphism, though, should we consider that as well? What would it require?

experimental/juvix/contract.jvx
cwgoes: OK, sure.

experimental/juvix/contract.jvx
cwgoes: Hmm, I wonder if there's a way to do it without introducing `where` as a keyword.

experimental/juvix/contract.jvx
cwgoes: Hmm, I wonder if there's a way to do it without a keyword / that would also work well inline.

experimental/juvix/contract.jvx
mariari: well in OCaml and F*, they use `;` for records. It kind of reminds me of C structs.

Though, since we are using , for tuples and for lists, I will change it to `,`

experimental/juvix/contract.jvx
mariari: If there are two words in a row

```ocaml
{ token
  storage = 
}
```
we could probably assume record updating IFF, token is a record, though I'm not sure if this has any conflicts

cwgoes:
- [ ] Add the example contract as a test for the parser (eventually)

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
+let Map = Data.Map
+
+-- common module interfaces for data structures in Common?
+sig MonoidMeasure : Common.Measure
```
cwgoes: What is this declaring, exactly? A type alias?

experimental/juvix/contract.jvx
```diff
@@ -0,0 +1,198 @@
...
+let MonoidMeasure =
+  type T = Nat
+
+  -- (<>) is normal semi-group type class function
+  let (<>)    = (<>)
+  let measure = id
+end
+
+-- need OrdMap, as modules that one goes to can't be modules?
```
cwgoes: I don't understand this comment, can you explicate?

experimental/juvix/contract.jvx
cwgoes: I think I am in agreement on the strategy here but this should be in a design doc (this comment thread, basically).

experimental/juvix/contract.jvx
cwgoes: Right, I think that's probably too implicit though, `with` is fine for now.

experimental/juvix/contract.jvx
mariari: the Map is a module.

MonoidMeasure, is a module, so I'm stating the module type.

experimental/juvix/contract.jvx
mariari: Map can't be a functor as it's a common path. So we need a function inside of it, let's say `OrdMap`. That generates the module we wish to use

experimental/juvix/contract.jvx
cwgoes: Gotcha, makes sense.

experimental/juvix/contract.jvx
cwgoes: Right, OK

experimental/juvix/contract.jvx
mariari: > Right, that looks pretty reasonable. Ocaml doesn't have row polymorphism, though, should we consider that as well? What would it require?

Acutally OCaml does!

Here is a passage from Real World Ocaml.

"Objects have some advantages over records: they don’t require type definitions, and their support for row polymorphism makes them more flexible. However, the heavy syntax and additional runtime cost means that objects are rarely used in place of records"

cwgoes: Interesting. What is the additional runtime cost? That sounds problematic, we definitely want to avoid additional runtime costs.

mariari: The additional runtime costs aren't due to row polymorphism, it's the runtime cost of the object system (dynamic calls).

We use simulated row polymorphism all the time in Juvix, via capabilities

cwgoes: Right right, that's fine then.

cyberglot:
> Sync with @cyberglot on proof syntax & include some proofs in the example contract(s)

@mariari what's the status of this in terms of plans/implementation?

mariari:
> > Sync with @cyberglot on proof syntax & include some proofs in the example contract(s)
> 
> @mariari what's the status of this in terms of plans/implementation?

Mainly Coq style tactics, within normal functions.

Often I suspect, for bodies we'll just place `smt ()`  for many proofs.

For signatures I suspect we'll have something like F*, where we have an effect and an expression for the proof statement.

cyberglot:
> Mainly Coq style tactics, within normal functions.
> 
> Often I suspect, for bodies we'll just place `smt ()` for many proofs.
> 
> For signatures I suspect we'll have something like F*, where we have an effect and an expression for the proof statement.

I think we should go agda-style (functions and pattern matching) for now, since the Proof UX project is kind of on-hold. What do you think, @cwgoes ?

cwgoes:
> > Mainly Coq style tactics, within normal functions.
> > Often I suspect, for bodies we'll just place `smt ()` for many proofs.
> > For signatures I suspect we'll have something like F*, where we have an effect and an expression for the proof statement.
> 
> I think we should go agda-style (functions and pattern matching) for now, since the Proof UX project is kind of on-hold. What do you think, @cwgoes ?

Let's go with that for now.

mariari:
> > Mainly Coq style tactics, within normal functions.
> > Often I suspect, for bodies we'll just place `smt ()` for many proofs.
> > For signatures I suspect we'll have something like F*, where we have an effect and an expression for the proof statement.
> 
> I think we should go agda-style (functions and pattern matching) for now, since the Proof UX project is kind of on-hold. What do you think, @cwgoes ?

Well by Coq style tactics, I do mean functions and pattern matching, just tactic like functions over it.

I really dislike how pattern matching in the Coq tactic system is via -'s, so it's not clear

PR #336, opened 2020-02-24
Add style guide.

thealmarty:
Outstanding TODOs:
- Make sure formatter is in sync with the style guide (regarding imports,
  indentation, etc)
- Make sure default HLint setting is desired/in sync with the style guide. 
- Add comments with Haddock syntax for future Haddock doc generation.
- Add a link to this style guide in the [handbook](https://github.com/cryptiumlabs/handbook/tree/master/subteams/juvix). (please @cwgoes ?)
- Clean up errors so that `-Werror` can be turned on (as in the style guide). Everyone please open a PR and turn on `-Werror` and fix the errors relevant to you. After all errors are fixed we can turn on `-Werror` in develop.


doc/STYLEGUIDE.md
```diff
@@ -0,0 +1,336 @@
...
+Always use explicit import lists or `qualified` imports for standard
+and third party libraries. The *Prelude* is an exception.
+
+Comments
+--------
+
+### Punctuation
+
+Write proper sentences; start with a capital letter and use proper
+punctuation. Make sure there is no typos.
```
cwgoes:
```suggestion
punctuation. Make sure there are no typos.
```

doc/STYLEGUIDE.md
```diff
@@ -0,0 +1,336 @@
...
+### Modules
+
+Use singular when naming modules e.g. use `Data.Map` and
+`Data.ByteString.Internal` instead of `Data.Maps` and
+`Data.ByteString.Internals`.
+
+Dealing with laziness
+---------------------
+
+By default, use strict data types and lazy functions.
```
cwgoes: Hmm, do we necessarily want this? I think we can wait on this for now until we actually are thinking about the performance of the compiler, and then we can decide accordingly (e.g. maybe `-XStrict` makes sense)

doc/STYLEGUIDE.md
```diff
@@ -0,0 +1,336 @@
...
+### Blank Lines
+
+One blank line between top-level definitions.  No blank lines between
+type signatures and function definitions.  Add one blank line between
+functions in a type class instance declaration if the function bodies
+are large.  Use your judgement.
+
+### Whitespace
+
+Surround binary operators with a single space on either side.  Use
```
cwgoes: Can we add an example?

cwgoes:
> Add a link to this style guide in the handbook. (please @cwgoes ?)

Done in https://github.com/cryptiumlabs/handbook/commit/a460d80816645ebfd7dff205833e26f7e8869d27.

> Clean up errors so that -Werror can be turned on (as in the style guide). Everyone please open a PR and turn on -Werror and fix the errors relevant to you. After all errors are fixed we can turn on -Werror in develop.

We can try this but I wonder if it's a bit premature, maybe before a `1.0` release? I find that `-Werror` adds a bit of overhead in actually writing code.

mariari:
> > Add a link to this style guide in the handbook. (please @cwgoes ?)
> 
> Done in [cryptiumlabs/handbook@a460d80](https://github.com/cryptiumlabs/handbook/commit/a460d80816645ebfd7dff205833e26f7e8869d27).
> 
> > Clean up errors so that -Werror can be turned on (as in the style guide). Everyone please open a PR and turn on -Werror and fix the errors relevant to you. After all errors are fixed we can turn on -Werror in develop.
> 
> We can try this but I wonder if it's a bit premature, maybe before a `1.0` release? I find that `-Werror` adds a bit of overhead in actually writing code.

I would second this, I vastly prefer `-Wall` (or really mass warnings) over mass erroring.

This is mainly due to the Haskell REPL being subpar when it comes to reloading code. Thus anything that actually freezes my repl from giving me useful results will really slow me down.

I'd prefer for our build system to fail or for us to stop pushing on the server side as it takes the warnings as errors, but as far as development goes, keeping it to warnings.


I will likely open a PR to this style guide, as to more motivation as to why we should always qualify our imports, and how our code formatter effects proper haskell formation style.

thealmarty:
> I will likely open a PR to this style guide, as to more motivation as to why we should always qualify our imports, and how our code formatter effects proper haskell formation style.

This is not a doc for explanations, it's a guide, so I don't think that's necessary. We should keep it concise.

thealmarty:
Changes as per comments
- removed laziness section
- added whitespace example
- removed `Werror`
I'm not sure what to do with the "warnings" section because it seems people want to leave the warnings there. Should I remove the section altogether?

cwgoes:
Hmm, if we do want to qualify our imports it seems like that should be in here, don't you think?

cwgoes:
> I'm not sure what to do with the "warnings" section because it seems people want to leave the warnings there. Should I remove the section altogether?

Let's just remove it for now.

doc/STYLEGUIDE.md
```diff
@@ -0,0 +1,271 @@
...
+your better judgement for the insertion of spaces around arithmetic
+operators but always be consistent about whitespace on either side of
+a binary operator.  Don't insert a space after a lambda. For example:
+
+```haskell
+plus_4 n = n + 4  -- whitespace on either side of `+`
+
+(\x -> x + 4)  -- no space after the lambda 
+
```
andy-morris:
````suggestion
```

````

thealmarty: Thanks for catching that!

andy-morris:
Should there be something about auxiliary definitions? e.g. whether to use `let` or `where`, and how we're going to indent them.

thealmarty:
> Should there be something about auxiliary definitions? e.g. whether to use `let` or `where`, and how we're going to indent them.

Sure! I think that's a good idea! Feel free to add a section!

thealmarty:
> Hmm, if we do want to qualify our imports it seems like that should be in here, don't you think?

If you want to add motivations, maybe add as a footnote? That way it's optional to read and the main guide is still concise.

cwgoes:
@mariari Did you want to add a bit on import qualification?

mariari:
> @mariari Did you want to add a bit on import qualification?

Sure, will do now

cwgoes:
What do we need to do to update the formatter ([here](https://github.com/cryptiumlabs/ormolu)) to match this guide?

mariari:
In terms of what our current formatter does not follow:

1. ,'s and ->'s are at the end of lines instead of the start
  - This reads rather awkwardly. There is an odd exception for infix functions however.

2. Nothing is aligned
  - Normally comments should always be aligned
  - Normally data grouped together should be align, but ormulo forces them to not be aligned properly

3. Data Declarations
  - Namely Records aren't properly aligned, and ,'s are at the end, not the start of lines

4. I'd like to add Paren usage.
  - This is not in the style guide, but our ending )'s should hug each other and be close to lisp form, relying on indentation for meaning, having a trailing end of )'s and ]'s makes the tests very very noisy

thealmarty:
As per meeting, I will look into [Brittany](https://github.com/lspitzner/brittany/) more. It'll either be Brittany or Ormolu.

thealmarty:
- updated style guide. 
  - added link to Tweag's style guide.
  - added why we chose Ormolu.
  - I left the parts of our style guide where we diverge from them or where we have added more details.
- updated README. I added the version number there so we can all be sure we are in sync. We'll have to update that when we update to a new version.

PR #340, opened 2020-02-28
Parser for the BNF

mariari:


mariari:
Currently this parser is not indent sensitive.

This is due to attoparsec lacking Monad Transformers. 

However this is easily fixable! Since I have prepared for a time where we have to be, we just need to modify the SN variant of functions (namely `spaceLiner`) to behave like this
```haskell
spaceLiner p = do
  spacer p
  peek <- peek
  if | newLine peek =  do
      let spacies = takeWhile space
      indentToStart <- get @"indentLevel"
      case  identToStart `compare` spacies of
        EQ -> fail "expression start of new line" <?> spaceLiner
        GT -> pure p
        LT -> fail "Expression start of new line" <?> spaceLiner
    | otherwise = pure p
```

in the mean time the following changes are made to the parser bnf that is **not reflected in the BNF**

let prefixs functions/modules.
val prefixes signatures.

cwgoes:
I think it's fine to make some compromises (e.g. non-indent-sensitivity & extra keywords) right now to speed up writing the parser, but we should keep clear documentation of the difference between the current syntax & the syntax we eventually want.

cwgoes:
Another note is that we can support `let` / `val` syntax in the REPL if that's convenient.

mariari:
Do syntax is currently bugged, gets stuck in a loop, will bring in 

https://hackage.haskell.org/package/attoparsec-expr

to handle this for me.

test/Frontend.hs
```diff
@@ -0,0 +1,164 @@
+module Frontend where
```
cwgoes: Eventually we should turn these into Tasty-style tests and add them to `test/Main.hs` so `stack test` works.

src/Juvix/Frontend/Parser.hs
```diff
@@ -0,0 +1,807 @@
...
+  Expr.buildExpressionParser tableExp (spaceLiner (expressionGen' p))
+
+expression' ∷ Parser Types.Expression
+expression' = expressionGen (fail "")
+
+expression ∷ Parser Types.Expression
+expression = expressionGen (Types.Do <$> do')
+
+usage ∷ Parser Types.Expression
+usage = string "u#" *> expression
```
cwgoes: Do we really need this prefix to all usage expressions?

src/Juvix/Frontend/Lexer.hs
```diff
@@ -0,0 +1,92 @@
...
+-- Hopefully this is fast!
+validStartSymbol' ∷ Integral a ⇒ a → Bool
+validStartSymbol' = Unicode.isAlpha . wordToChr
+
+-- Unicode.isUpper 'İ' = True!
+validUpperSymbol ∷ Integral a ⇒ a → Bool
+validUpperSymbol = Unicode.isUpper . wordToChr
+
+dash ∷ Word8
+dash = 45
```
cwgoes: Can we use `Char.chrFromChar` or something like that so these are easier to read?

src/Juvix/Frontend/Parser.hs
mariari: I don't use this at all, this can be removed. Arrows and signatures already have it properly built in

src/Juvix/Frontend/Lexer.hs
mariari: Does chrFromChar give me back a word8?

test/Frontend.hs
mariari: I agree

https://github.com/typedefs/typedefs/blob/master/src/Typedefs/Test/ParseTests.idr#L25

Seems fairly good form wise. I'll see what i can do over time.

Need to derive equality for the syntax

src/Juvix/Frontend/Parser.hs
cwgoes: Nice, let's remove it then.

src/Juvix/Frontend/Lexer.hs
cwgoes: Ah, that was a hypothetical function, but presumably some such function exists.

mariari: Sure I can rig something with ord and chr, I'll likely, put inlining pragmas with each, just to make sure they inline

mariari: Done!

```haskell
charToWord8 ∷ Char → Word8
charToWord8 = fromIntegral . ord
{-# INLINE charToWord8 #-}
```

is all it took

PR #343, opened 2020-03-14
Integrate extensible-data

andy-morris:
There are still some things left over that are just boilerplate, which future versions of e-d could generate (like the whole of the `TransformExt` module, probably), but the most obvious & repetitive stuff is gone at least.


cwgoes:
(looks like `setup-dependencies` is failing on CI though, we should investigate that)

stack.yaml.lock
```diff
@@ -382,6 +382,20 @@ packages:
       sha256: 05ebc44a9cc5ecda047aa0179020615941ae5b698586854d121b4219ec3cc114
   original:
     url: https://github.com/adjoint-io/poly/archive/b75b853025ea75ba64f62f39d4524832f0117e36.tar.gz
+- completed:
```
thealmarty: This .lock file should not be in the commit I think?

cwgoes: I think it's helpful, it makes our builds more reproducible

PR #344, opened 2020-03-16
Add non-fungible token contract example in Idris.

thealmarty:
In this PR:
- Added NFT contract. All mandatory functions of the ERC 721 standard are there. `Mint` function, which is optional, is added as well. 
- Error type moved to FakeLib.
- Minimally tested. See [test log](https://gist.github.com/thealmarty/f6d7bf62163f4cade457553c2c4baf70). Further tests should be done with more interesting storage's.

Things to consider:
- in the standard, the list of operators is a map of `Address` and `Bool`. Does
  it make sense to just have a `list`/`vect`? Addresses on the list are equivalent
  to `True` in the Address Bool map.
- in the standard, there can at most be one approved address. Does it make sense
  to allow more than one address? This is like the list of operators in
  `Account` except that the list is linked to a specific token instead of a
  specific address.
- should I check in `setApprovalForAll` that the owner address is not in the
  operator list? (Owner always have powers equivalent to an operator.)

### DIFF
--- a/experimental/Examples/NFT.idr
+++ b/experimental/Examples/NFT.idr
@@ -17,7 +17,7 @@ record Account where
 record Token where
   constructor MkToken
   tokenOwner : Address
-  approved : List Address -- approved address(es) (TODO fig out how to use Vect)
+  approved : Maybe Address -- approved address (there can at most be one address)
 
 ||| The storage has type Storage which is a record with accounts and tokens.
 record Storage where
@@ -69,13 +69,13 @@ mint token dest storage =
           case currentCaller == owner storage of
             False => Left FailedToAuthenticate
             True =>
-              let newAcc = insert dest (record {ownedTokens = (1 + getAccountBal dest acc)} destAcc) acc
-                  newToken = insert token (MkToken dest Nil) (tokens storage) in
+              let newAcc = insert dest (record {ownedTokens $= (+ 1)} destAcc) acc
+                  newToken = insert token (MkToken dest Nothing) (tokens storage) in
                     Right
                       (record
                         {accounts = newAcc,
                          tokens = newToken,
-                         totalSup = (totalSup storage) + 1
+                         totalSup $= (+ 1)
                         } storage
                       )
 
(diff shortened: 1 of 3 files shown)


experimental/Examples/NFT.idr
```diff
@@ -0,0 +1,230 @@
...
+(from : Address) -> (dest : Address) -> (token : TokenId) -> Either Error Storage
+transfer from dest token =
+  case ownerOf token of
+    Left e => Left e
+    Right add =>
+      case add == from of
+        False => Left NotOwnedByFromAddress
+        True =>
+          case currentCaller == from ||
+               currentCaller == maybe (owner storage) id (approvedAdd token) ||
```
cwgoes: Why are we using the global `storage` here? Shouldn't we be using `storage` as a parameter somehow?

experimental/Examples/NFT.idr
```diff
@@ -0,0 +1,230 @@
...
+               isApprovedForAll add currentCaller of
+            False => Left FailedToAuthenticate
+            True =>
+              let newAcc = modifyAccBal from (minus 1) (modifyAccBal dest (+ 1) currentAcc)
+                  newToken = insert token (MkToken dest Nothing) currentTok in
+                Right
+                  (record
+                    {accounts = newAcc,
+                    tokens = newToken} storage
```
cwgoes: Ditto on `storage` as a parameter

experimental/Examples/NFT.idr
```diff
@@ -0,0 +1,230 @@
...
+
+||| approve [standard function] set the approved address for an NFT.
+||| When a transfer executes, the approved
+||| address for that NFT (if any) is reset to none.
+total approve : Address -> TokenId -> Either Error Storage
+approve address token =
+  case getApproved token of
+    Left e => Left e
+    Right approvedAdd =>
+      let isApproved = currentCaller == (maybe (owner storage) id approvedAdd)
```
cwgoes: Why are we using the global `storage` here?

experimental/Examples/NFT.idr
```diff
@@ -0,0 +1,230 @@
...
+                (record
+                   {tokens =
+                     insert
+                     token
+                     (MkToken
+                       owner
+                       (Just address)
+                     )
+                     currentTok
+                   } storage
```
cwgoes: Why are we using the global `storage` here?

experimental/Examples/NFT.idr
```diff
@@ -0,0 +1,230 @@
...
+                      )
+
+total initStorage : Storage
+initStorage =
+  case mint 1 "qwer" emptyStorage of
+    Left _ => emptyStorage
+    Right s => s
+
+total storage : Storage
+storage =
```
cwgoes: This global seems to be used in a lot of places where it seems like it should be a parameter, can we just get rid of it?

cwgoes:
> In the standard, the list of operators is a map of Address and Bool. Does
it make sense to just have a list/vect? Addresses on the list are equivalent
to True in the Address Bool map.

I think this is because it's O(1) to lookup in the map vs O(n) to scan a list.

> In the standard, there can at most be one approved address. Does it make sense
to allow more than one address? This is like the list of operators in
Account except that the list is linked to a specific token instead of a
specific address.

Maybe, one is fine for now though I think.

> Should I check in setApprovalForAll that the owner address is not in the
operator list? (Owner always have powers equivalent to an operator.)

Sure, that would be a good sanity check.

experimental/Examples/NFT.idr
thealmarty: This is following the standard that doesn't seem to take `storage` as an input for those functions. The global `storage` here is a fake current storage. I can change it all to taking a `storage` though.

experimental/Examples/NFT.idr
thealmarty: Yes, I can get rid of it.

experimental/Examples/NFT.idr
cwgoes: Yeah, I think all the functions should take a `storage` parameter

It's probably just an implicit contract `storage` in the Ethereum standard, it's the equivalent to a parameter in our case

thealmarty:
> > In the standard, the list of operators is a map of Address and Bool. Does
> > it make sense to just have a list/vect? Addresses on the list are equivalent
> > to True in the Address Bool map.
> 
> I think this is because it's O(1) to lookup in the map vs O(n) to scan a list.
I see. I didn't think the list of address should be long enough for that to matter, and it's making the storage larger with the `Bool`. But I'll stick with the standard for now.

thealmarty:
Changes done as per comments:
- checked in setApprovalForAll that the owner address is not in the operator list
- removed global `storage` and add `storage` as an input to all functions.

thealmarty:
> ACK modulo test log
> 
> You can write tests in Idris inline and have the typechecker check them, if you like, e.g.:
> 
> ```idris
> test_1 : 2 + 2 = 4
> test_1 = Refl
> ```

Good idea. I think I will do that in a separate file!

mariari:
> > > In the standard, the list of operators is a map of Address and Bool. Does
> > > it make sense to just have a list/vect? Addresses on the list are equivalent
> > > to True in the Address Bool map.
> > 
> > 
> > I think this is because it's O(1) to lookup in the map vs O(n) to scan a list.
> 
> I see. I didn't think the list of address should be long enough for that to matter, and it's making the storage larger with the `Bool`. But I'll stick with the standard for now.

if you just need to do an inclusion test, then a set is what you want.

This way you get O(log_32(n)) access to the contents (O(1) in an imperative set), without storing booleans

thealmarty:
[Test log](https://gist.github.com/thealmarty/f6d7bf62163f4cade457553c2c4baf70) updated.

PR #345, opened 2020-03-17
Added Qualified Imports, and some format changes

mariari:


doc/STYLEGUIDE.md
```diff
@@ -157,8 +160,49 @@ Put a blank line between each group of imports.  The imports in each
 group should be sorted alphabetically, by module name.
 
 Always use explicit import lists or `qualified` imports for standard
-and third party libraries. The *Prelude* is an exception.
+and third party libraries.  The *Juvix Standard Library* is an exception.
 
+Qualified imports help two fold:
```
thealmarty: I thought we are putting the motivation in a footnote?

mariari: Well there are no other footnotes in the document, and the document isn't expected to grow too much, so it'll be odd to add it as such here

thealmarty: hmm...I prefer otherwise because people tend to not digest as well when there's a lot of info. I want people to follow the guide well. But it's up to you. No worries. :)

cwgoes: Let's leave it here for now, but we can separate out footnotes if the guide continues to grow

doc/STYLEGUIDE.md
```diff
@@ -157,8 +160,49 @@ Put a blank line between each group of imports.  The imports in each
...
+
+1. easily determine what code code comes from where and refactoring dependencies
+2. Enforce consistency between different Haskell Modules.  For example:
```
cwgoes:
```suggestion
2. Enforce consistency between different Haskell modules. For example:
```

PR #349, opened 2020-03-18
Removing the various tuples in Core

mariari:


mariari:
So far I've changed two tuples, into record types, however there are many more tuples to squash.

There are hacks in this, which are noted in comments. These are due to not being able to easily derive `Eq` and `Show`.

I will probably also qualify more imports while I'm at it.

PR #354, opened 2020-03-26
Switching formatter to Ormolu version 0.0.3.1

thealmarty:
Following up on #336 .

These formatting changes are applied using ormolu version 0.0.3.1. There are no errors with the appropriate extensions turned on. (Extensions that we use and are not in the `--ghc-opt` list are turned on by default by Ormolu) 

I recommend that we use ormolu straight without configuration, for the following reasons:

- easy maintenance of the formatter. 
  - You can install ormolu with `stack install` and any updates can be retrieved by re-running `stack install`. 
  - If we use our own fork, we'll have to maintain it ourselves for the parts we change. Note that our code base is getting larger and there may be bugs that we will notice down the road. We will need to periodically merge from `master` to get bug fixes and there may be merge conflicts. 
- more standard formatting. We'll have to adopt a style guide that matches ormolu but it'll be the ormolu standard which other devs who use ormolu are familiar with.
 
So long as the format is consistent, I think we should minimise the time we spend on maintaining these tools so we can focus on developing.


mariari:
Overall looks fine,

src/Juvix/Backends/LLVM/Net/EAC/Types.hs
```diff
@@ -98,16 +100,17 @@ checkNull xs = do
...
-      }
-  cdr ←
+  newList <- Codegen.malloc (Codegen.nodePointerSize + Codegen.nodePointerSize) eacLPointer
+  car <-
+    Codegen.getElementPtr $
+      Codegen.Minimal
+        { Codegen.type' = Codegen.pointerOf eacPointer,
+          Codegen.address' = newList,
+          Codegen.indincies' = Codegen.constant32List [0, 0]
+        }
```
mariari: Example of 2.

src/Juvix/Backends/LLVM/Net/Environment.hs
```diff
@@ -27,16 +27,20 @@ initialModule = do
...
   modify @"varTab"
-    ( Map.insert "numPorts_small" Codegen.S
-        { Codegen.sum' = "numPorts",
-          Codegen.offset = 0,
-          Codegen.tagSize' = 1
-        }
-        . Map.insert "numPorts_large" Codegen.S
+    ( Map.insert
+        "numPorts_small"
+        Codegen.S
```
mariari: Example of  2 again.

src/Juvix/Backends/LLVM/Codegen/Types.hs
```diff
@@ -350,54 +349,55 @@ numPorts
...
-  }
-
-portPointer ∷ Type
+portType :: Type -> Type
+portType nodePtr =
+  StructureType
+    { isPacked = True,
+      elementTypes =
+        [ nodePtr, -- the pointer to the other node
+          numPortsNameRef -- the offset from the base of the node where the port is
```
mariari: Example of 1

src/Juvix/Backends/LLVM/Codegen/Types.hs
```diff
@@ -350,54 +349,55 @@ numPorts
...
+portTypeName :: IsString p => p
 portTypeName = "graph_port"
 
-portType ∷ Type → Type
-portType nodePtr = StructureType
-  { isPacked = True,
-    elementTypes =
-      [ nodePtr, -- the pointer to the other node
-        numPortsNameRef -- the offset from the base of the node where the port is
-      ]
```
mariari: Example of 1

Makefile
```diff
@@ -28,7 +28,7 @@ lint:
 	stack exec -- hlint app src test
 
 format:
-	find . -path ./.stack-work -prune -o -path ./archived -prune -o -type f \( -name "*.hs" ! -name "ReplDoc.hs" \) -exec ormolu --mode inplace {} \;
+	find . -path ./.stack-work -prune -o -path ./archived -prune -o -type f -name "*.hs" -exec ormolu --mode inplace {} --ghc-opt -XTypeApplications --ghc-opt -XUnicodeSyntax --ghc-opt -XPatternSynonyms \;
```
andy-morris: It seems fine at the moment but do you think it would be worth adding `--ghc-opt -XTemplateHaskell` here as well?

thealmarty: According the Tweag that is turned on by default. If we need -XTemplateHaskellQuotes then we'll have to add that there I think. See `ormolu --manual-exts`. Does the relevant code look fine to you?

andy-morris: If I'm reading ormolu's code right, `TemplateHaskellQuotes` is enabled by accident, since it's implied by `TemplateHaskell`, and AFAICT ormolu doesn't add `-XNo…` flags for the things that aren't enabled.

Maybe we should explicitly add it, since this seems unintentional and might be changed in future?

thealmarty: OK, thanks for point that out. I did notice that it changed a few files' formatting before when I turned on that flag. See if you like the changes in #358.

PR #356, opened 2020-03-27
Annotated terms in typechecker

andy-morris:


andy-morris:
@mariari @cwgoes @thealmarty I think this is ready for review whenever you have time.

I haven't merged it into `develop` yet, but I think that will probably make the diffs unreadable. So I was thinking I could maybe do all the conflict resolution etc while you're reviewing this, and then push the merge commit just before merging, if that sounds ok?

andy-morris:
turns out rebasing didn't mess up the diff after all, so here's the actual thing

src/Juvix/Core/IR/Evaluator.hs
```diff
@@ -6,177 +6,389 @@
...
+
+
+class HasWeak a where
+  weak' :: Natural -> a -> a
+
+weak :: HasWeak a => a -> a
+weak = weak' 0
+
+instance HasWeak ()   where weak' _ () = ()
+instance HasWeak Void where weak' _ v = absurd v
```
mariari, line 23: Very interesting. I should learn to do this more myself

src/Juvix/Core/IR/Evaluator.hs
```diff
@@ -6,177 +6,389 @@
...
+
+type AllWeak ext primTy primVal =
+  (IR.TermAll HasWeak ext primTy primVal,
+   IR.ElimAll HasWeak ext primTy primVal)
```
mariari, line 27: Very cleaver, trick noted.

src/Juvix/Core/IR/Evaluator.hs
```diff
@@ -6,177 +6,389 @@
...
-  ( IR.TermAll (CanSubst ext primTy primVal) ext primTy primVal,
-    IR.ElimAll (CanSubst ext primTy primVal) ext primTy primVal
-  ) =>
-  Natural ->
-  IR.Elim' ext primTy primVal ->
-  IR.Term' ext primTy primVal ->
-  IR.Term' ext primTy primVal
+
+
+class HasWeak a where
```
cwgoes, line 16: Should we put this in `Juvix.Library` perhaps?

src/Juvix/Core/IR/Typechecker/Types.hs
```diff
@@ -0,0 +1,182 @@
...
+throwTC = throw @"typecheckError"
+
+
+data T
+
+do
+  let typed = Ext.Ann $ \primTy primVal -> [t|Annotation $primTy $primVal|]
+  extT <- IR.extendTerm "Term" [] [t|T|] $
+    IR.defaultExtTerm
+      { IR.typeStar   = typed
```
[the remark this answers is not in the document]
andy-morris: `HasThrowTC'` is an abbreviation for a `HasThrow "typecheckError"` constraint, so it's just applying `throw` to the type argument to disambiguate.

The independent `do` is a TH splice. They can be arbitrary expressions, not just function calls, and I needed to share the `typed` variable between `extendTerm` and `extendElim`. Do you think it would it be more obvious that it's a splice if it was `$(do ...)` instead?

src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -1,511 +1,221 @@
...
-        logOutput $ passed <> typechecked x ann
-        pure ()
-  where
-    checking = logOutput "Checking that input annotation is of type *0. "
+typeTerm _ ii ctx tm@(IR.PrimTy pty) ann@(Annotation σ ty) = do
+  tellLogs [TermIntro ctx tm ann, CheckingPrimTy, CheckingSigmaZero]
+  if σ == Usage.SNat 0 then
+    tellLog SigmaIsZero
+  else
+    throwLog UsageMustBeZero
```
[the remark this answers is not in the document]
andy-morris: Oh, yeah, I replaced most of the `SNat 1`/`SNat 0` but I must have missed this one, sorry

src/Juvix/Core/IR/Typechecker/Log.hs
```diff
@@ -0,0 +1,190 @@
...
+      (IR.Term primTy primVal)  -- ^ argument
+  | AppInferredAs Usage.T (IR.Value primTy primVal)
+
+  | InferringAnn
+  | CheckingAnnIsType (IR.Term primTy primVal)
+  | CheckingAnnTerm (IR.Term primTy primVal) Usage.T (IR.Term primTy primVal)
+  deriving (Eq, Show)
+
+describe :: (Show primTy, Show primVal) => Log primTy primVal -> Text
+describe (TermIntro ctx tm (Annotation σ ty)) = mconcat
```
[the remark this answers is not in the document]
andy-morris: Tbh I think the final form of this code will be using a pretty printing library to produce legible output instead of the derived `Show` form (same for `TypecheckError`). I just left the messages more or less as they were for now, to come back to later

src/Juvix/Core/IR/Evaluator.hs
andy-morris: You mean the `HasWeak` class? Yeah, maybe, since it's not specific to core terms (unlike `subst` which has that `Elim` argument).

PR #359, opened 2020-03-28
Arithmetic Circuits MVP

cyberglot:
### MVP 0
  - [x] Field elements
    + [x] addition
    + [x] multiplication
    + [x] equality check operator
  - [x] Simple circuit that checks if a number (private witness) is a square root of the public data

### MVP 1
   - [x] Field elements: addition, multiplication, subtraction, division, negation, comparison
   - [x] Integers: addition, multiplication, subtraction, division, negation, comparison
      + [x] compiled to operations on field elements
   - [x] Prove knowledge of a solution to a quadratic or cubic equation
        + `ax^2 + bx = -c`
        + Term: `\x . \w . (a * (w^2) + b * w) - x == 0`
        + All terms are two-argument functions, the first arg is the public input (of some type), the second arg is the private witness (of some type), the type of the body is a boolean

### MVP 2
   - [x] Conditionals
   - [x] Booleans: AND and OR
   - [x] Prove disjoint knowledge of either a solution to one quadratic or a solution to another (but not which)
      + [x]  `\x . \w . (w ** 2 == x) || (w ** 3 == (x + 4))`
      + [x] `\x . \w . (w ** 3 == x) && (w ** 4 == (x - 2))`
  - [ ] Implement [MiMC](https://byt3bit.github.io/primesym/mimc/)
  - [ ] Prove knowledge of preimage of a hash
     + [ ] `\x . \w . hash_mimc(w) == x`

### MVP 3
 - [x] Generate and verify ZKP proofs
 - [x] Integrate ZKP proofs into the tests


src/Juvix/Backends/ArithmeticCircuit/Parameterisation/Booleans.hs
```diff
@@ -0,0 +1,85 @@
...
+import Juvix.Library hiding ((<|>))
+import Text.ParserCombinators.Parsec
+import qualified Text.ParserCombinators.Parsec.Token as Token
+import qualified Text.Show as Show
+import Prelude (String)
+import Circuit.Arithmetic (Wire(..))
+import Circuit.Expr
+import Circuit.Lang
+
+-- k: primitive type: naturals
```
cwgoes:
```suggestion
-- k: primitive type: booleans
```

src/Juvix/Backends/ArithmeticCircuit/Parameterisation/FieldElements.hs
```diff
@@ -0,0 +1,64 @@
...
+typeOf (Val _) = Ty :| []
+typeOf (Curried _ _) = Ty :| [Ty]
+typeOf Add = Ty :| [Ty, Ty]
+typeOf Mul = Ty :| [Ty, Ty]
+
+apply ∷ Val → Val → Maybe Val
+apply Add (Val x) = pure (Curried Add x)
+apply Mul (Val x) = pure (Curried Mul x)
+apply (Curried Add x) (Val y) = pure (Val (add x y))
+apply (Curried Mul x) (Val y) = pure (Val (add x y))
```
cwgoes:
```suggestion
apply (Curried Mul x) (Val y) = pure (Val (mul x y))
```

src/Juvix/Backends/ArithmeticCircuit/Parameterisation/Integers.hs
```diff
@@ -0,0 +1,75 @@
+module Juvix.Backends.ArithmeticCircuit.Parameterisation.Integers where
```
cwgoes: (I think integers are lower-priority for now)

cyberglot: Integers are not really integrated to `Parametrisation.hs` but I did code it so I decided to just leave it there for when its time arrives.

src/Juvix/Backends/ArithmeticCircuit/Parameterisation/FieldElements.hs
cyberglot: Fixed.

src/Juvix/Backends/ArithmeticCircuit/Parameterisation/Booleans.hs
cyberglot: Fixed.

cyberglot:
> we'll need to add `==` on field elements
Done.

Working on the circuit now.

test/Backends/ArithmeticCircuit.hs
```diff
@@ -28,10 +29,34 @@ equalTerm = J.Lam "x" (J.Var "x")
...
-equalCircuit =
-  C.ArithCircuit
-    { C.weights = C.GateWeights {C.wL = [], C.wR = [], C.wO = []},
-      C.commitmentWeights = [],
-      C.cs = []
-    }
+power :: Term
+power = J.LamM "x" (J.LamM "w" (Base.eq (Base.exp (J.Var "w") (Base.c 2)) (J.Var "x")))
+
+quadratic :: Parameterisation.Integer a b => a → b → Term
```
cwgoes, line 35: Are these tested with proofs somewhere (we should do so)?

test/Backends/ArithmeticCircuit.hs
```diff
@@ -28,10 +29,34 @@ equalTerm = J.Lam "x" (J.Var "x")
...
 equalType = J.Pi Omega (J.PrimTy ()) (J.PrimTy ())
 
-equalCircuit :: C.ArithCircuit Fr
-equalCircuit =
-  C.ArithCircuit
-    { C.weights = C.GateWeights {C.wL = [], C.wR = [], C.wO = []},
-      C.commitmentWeights = [],
-      C.cs = []
-    }
+power :: Term
```
cwgoes, line 32: This is proving knowledge of a square-root, right?

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,57 @@
...
+  = NotYetImplemented
+  | SomethingWentWrongSorry
+
+data Memory x where
+  Mem :: Parameterisation.FieldElement x => [x] -> Memory x
+
+eval :: Mem x -> Term -> (Mem x, Expr.Expr Circuit.Wire Parameterisation.F a)
+eval mem CoreErased.AnnTerm{ CoreErased.term = CoreErased.Term (CoreErased.Prim term) } = (mem, translateTerm term)
+eval mem CoreErased.AnnTerm{ CoreErased.term = CoreErased.Term (CoreErased.Var var) } = (mem, Expr.EVar v)
+eval (Mem ms) CoreErased.AnnTerm{ CoreErased.term = CoreErased.Term CoreErased.LamM{ CoreErased.arguments = arguments }}
```
cwgoes, line 37: Should we deal with this? I think we will run into this case when compiling `λx.λw. w^2 == x`, for example. Maybe you want to specifically match out the first two arguments, and inline everything else?

test/Backends/ArithmeticCircuit.hs
cyberglot: It is, sorry!! My arithmetic is broken, I just guide myself by algebraic laws and hope for the best.

test/Backends/ArithmeticCircuit.hs
cyberglot: They're been used with `zkp` now, update to follow.

src/Juvix/Backends/ArithmeticCircuit/Compilation.hs
```diff
@@ -1,31 +1,16 @@
...
-      C.cs = []
-    }
-{-
- - find all output values
- - follow wire back to gate
- - if add/mul: add add/mul constraint of input values & output or intermediate value
- - if input: stop
- - otherwise: continue
- -}
+runCirc = Expr.execCircuitBuilder . Expr.compile
```
cwgoes, line 16: Do we use this function? Looks like we could use it above.

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+  = Add
+  | Mul
+  | Sub
+  | Exp
+  | Eq
+  | And
+  | Or
+  deriving (Show, Eq)
+
+data Op = Neg
```
cwgoes, line 34: `UnaryOp`, I think

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+  deriving (Show, Eq)
+
+type Term = CoreErased.AnnTerm () PrimVal
```
cwgoes, line 37: This needs to be `CoreErased.AnnTerm PrimTy PrimVal`, for some `PrimTy`, which needs to be defined

We should be able to construct a `Core.Parameterisation` record (example - https://github.com/metastatedev/juvix/blob/develop/src/Juvix/Backends/Michelson/Parameterisation.hs#L139)

It's fine not to worry about the parsing parts for now, but we will/do need type-checking

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+                            mem' = foldr (\(sy, val) m -> insert sy val m) mem matchedArgParams
+                         in transTerm mem' body
+    _ -> panic $ show SomethingWentWrongSorry
+
+transPrim :: PrimVal -> Memory -> ArithExpression Par.F
+transPrim (Element f) _ = FExp $ Lang.c f
+transPrim (Boolean b) _ = BoolExp $ Circuit.EConstBool b
+transPrim (FEInteger i) _ = FExp $ Lang.c (fromIntegral i)
+-- I miss lenses
+transPrim (BinOp Add prim prim') mem = let (_, FExp prim1) = transTerm mem prim
```
cwgoes, line 100: Do we want to ignore the memory updates here? Won't we lose variable mappings?

Also, I think this would be cleaner in `MonadState Memory`

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+insert :: J.Symbol -> ArithExpression Par.F -> Memory -> Memory
+insert sy x (Mem map_ n) = Mem (Map.insert sy (n, x) map_) (n + 1)
+
+lookup :: J.Symbol -> Memory -> Maybe (Int, ArithExpression Par.F)
+lookup sy (Mem map_ _) =
+  case Map.lookup sy map_ of
+    Just res -> Just res
+    Nothing -> Nothing
+
+transTerm :: Memory
```
cwgoes, line 70: IMO this would be cleaner in `MonadState Memory`

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+                                          (_, FExp prim2) = transTerm mem prim'
+                                          in BoolExp $ Lang.eq prim1 prim2
+transPrim (BinOp And prim prim') mem = let (_, BoolExp prim1) = transTerm mem prim
+                                           (_, BoolExp prim2) = transTerm mem prim'
+                                           in BoolExp $ Lang.and_ prim1 prim2
+transPrim (BinOp Or prim prim') mem = let (_, BoolExp prim1) = transTerm mem prim
+                                          (_, BoolExp prim2) = transTerm mem prim'
+                                          in BoolExp $ Lang.or_ prim1 prim2
+-- implements exponentiation by hand
+transPrim (BinOp Exp prim CoreErased.Ann { CoreErased.term = CoreErased.Prim (FEInteger i)}) mem
```
cwgoes, line 119: Looks like this will fail to pattern-match on non-constant exponentiation?

That's fine, for now, but we should explicitly error

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cwgoes: ... you have a `Ty` defined in the other file which seems correct

cwgoes: ... so basically this needs to be the same type as in `Parameterisation.hs`, since that's what we compile from.

It's fine to use other types "inside" the compilation logic, if convenient

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+    CoreErased.Ann { CoreErased.term = CoreErased.LamM { CoreErased.body =  body
+                                                       , CoreErased.arguments = arguments
+                                                       }
+                   } -> let transTermParams = snd . sequenceA $ map (transTerm mem) params
+                            matchedArgParams = zip arguments transTermParams
+                            mem' = foldr (\(sy, val) m -> insert sy val m) mem matchedArgParams
+                         in transTerm mem' body
+    _ -> panic $ show SomethingWentWrongSorry
+
+transPrim :: PrimVal -> Memory -> ArithExpression Par.F
```
mariari, line 95: It seems this function is non exhaustive

```haskell
    In an equation for ‘transPrim’:
        Patterns not matched:
            (BinOp Exp _ (CoreErased.Ann _ _ (CoreErased.Var _))) _
            (BinOp Exp _ (CoreErased.Ann _ _ (CoreErased.Prim (Element _)))) _
            (BinOp Exp _ (CoreErased.Ann _ _ (CoreErased.Prim (Boolean _)))) _
            (BinOp Exp _ (CoreErased.Ann _ _ (CoreErased.Prim (BinOp _ _ _))))
            _
```

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+                                           in FExp $ Lang.sub prim1 prim2
+transPrim (BinOp Eq prim prim') mem = let (_, FExp prim1) = transTerm mem prim
+                                          (_, FExp prim2) = transTerm mem prim'
+                                          in BoolExp $ Lang.eq prim1 prim2
+transPrim (BinOp And prim prim') mem = let (_, BoolExp prim1) = transTerm mem prim
+                                           (_, BoolExp prim2) = transTerm mem prim'
+                                           in BoolExp $ Lang.and_ prim1 prim2
+transPrim (BinOp Or prim prim') mem = let (_, BoolExp prim1) = transTerm mem prim
+                                          (_, BoolExp prim2) = transTerm mem prim'
+                                          in BoolExp $ Lang.or_ prim1 prim2
```
mariari, line 117: This seems to be doing the applicative state monad by hand, basically


if we promote mem to being a state then we can write
```haskell
transPrim (BinOp Add prim prim') =
  FExp . Lang.add <$> transTerm prim <*> transTerm prim'
```

src/Juvix/Backends/ArithmeticCircuit/Compilation.hs
```diff
@@ -1,31 +1,16 @@
-module Juvix.Backends.ArithmeticCircuit.Compilation where
+module Juvix.Backends.ArithmeticCircuit.Compilation (compile, runCirc, add, mul, sub, neg, eq, exp, int, c, and', or', Term, Type, lambda, var, input, cond) where
```
mariari, line 1: Probably best to put this on multiple lines

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,183 @@
...
+import qualified Juvix.Core.Usage as Usage
+import qualified Juvix.Library as J
+import qualified Juvix.Backends.ArithmeticCircuit.Parameterisation as Par
+import qualified Circuit.Expr as Expr
+import qualified Circuit.Lang as Lang
+import qualified Circuit
+import qualified Data.Map as Map
+import Numeric.Natural()
+
+import Protolude
```
mariari, line 13: Probably best to import `Juvix.Library` over this

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
mariari: Ditto, it'd make the trans functions lcleaer cleaner as well

src/Juvix/Backends/ArithmeticCircuit/Parameterisation.hs
```diff
@@ -1,37 +1,119 @@
...
+  deriving (Show, Eq)
 
-type Type = J.Type PrimTy
+type F = Fr
+type BooleanVal = Booleans.Val (Expr.Expr Arith.Wire F Bool) Bool
+type FEVal = FieldElements.Val (Expr.Expr Arith.Wire F F)
+type FEInteger = FEInteger.Val (Expr.Expr Arith.Wire F Int) Int
 
-type PrimTy = ()
+instance FieldElements.FieldElement (Expr.Expr Arith.Wire) where
```
mariari, line 36: I'm getting warnings over this
```haskell
src/Juvix/Backends/ArithmeticCircuit/Parameterisation.hs:39:10-58: warning: [-Wmissing-methods] …
    • No explicit implementation for
        ‘FieldElements.intExp’
    • In the instance declaration for
        ‘FieldElements.FieldElement (Expr.Expr Arith.Wire)’
```

src/Juvix/Backends/ArithmeticCircuit/Parameterisation.hs
```diff
@@ -1,37 +1,119 @@
...
+feValToAll ∷ FEVal → Val
+feValToAll = FEVal
+
+typeOf ∷ Val → NonEmpty Ty
+typeOf (BoolVal x) =
+  fmap boolTyToAll (Booleans.typeOf x)
+typeOf (FEVal x) =
+  fmap feTyToAll (FieldElements.typeOf x)
+typeOf (Curried _ _) = Ty :| [Ty]
+typeOf Eq = BoolTy Booleans.Ty :| [FETy FieldElements.Ty, FETy FieldElements.Ty]
```
mariari, line 88: It seems IntegerValue is not matched here

```haskell
Patterns not matched: (IntegerVal _)
```

src/Juvix/Backends/ArithmeticCircuit/Proof.hs
```diff
@@ -0,0 +1,43 @@
+module Proof (proof) where
```
mariari, line 1: It seems this file does not really compile?

The project builds, as this isn't added to the yaml file

src/Juvix/Backends/ArithmeticCircuit/Proof.hs
```diff
@@ -0,0 +1,43 @@
...
+import Data.Aeson
+import qualified Data.Curve.Weierstrass.BN254 as G1
+import qualified Data.Curve.Weierstrass.BN254T as G2
+import Data.Field.Galois (Extension, rnd, toP)
+import Data.Pairing.BN254 (BN254, Fr, Pairing (..))
+import qualified Protocol.Groth as Groth
+import Protolude hiding (readFile)
+import QAP (QAP)
+import System.Directory
+import System.FilePath.Posix
```
mariari, line 13: This requires a new dep namely 

```haskell
  - filepath
```

src/Juvix/Backends/ArithmeticCircuit/Proof.hs
```diff
@@ -0,0 +1,43 @@
...
+deriving instance FromJSON G2.PA
+
+deriving instance ToJSON G1.PA
+
+deriving instance ToJSON G2.PA
+
+deriving instance FromJSON (Extension G2.U G1.Fq)
+
+deriving instance ToJSON (Extension G2.U G1.Fq)
+
```
mariari, line 27: Seems these can't be derived

src/Juvix/Backends/ArithmeticCircuit/Parameterisation/FieldElements.hs
```diff
@@ -0,0 +1,68 @@
...
+  Neg :: FieldElement e => Val (e f g)
+  IntExp :: FieldElement e => Val (e f g)
+  Eq :: FieldElement e => Val (e f g)
+  Curried :: FieldElement e => Val (e f g) → e f g → Val (e f g)
+
+typeOf ∷ Val a → NonEmpty Ty
+typeOf (Val _) = Ty :| []
+typeOf (Curried _ _) = Ty :| [Ty]
+typeOf Add = Ty :| [Ty, Ty]
+typeOf Mul = Ty :| [Ty, Ty]
```
mariari, line 43: This seems non exhaustive
```haskell
    Pattern match(es) are non-exhaustive
    In an equation for ‘typeOf’:
        Patterns not matched:
            Neg
            IntExp
            Eq
```

src/Juvix/Backends/ArithmeticCircuit/ZKP.hs
```diff
@@ -0,0 +1,51 @@
...
+import qualified Data.Curve.Weierstrass.BN254 as G1
+import qualified Data.Curve.Weierstrass.BN254T as G2
+import Data.Field.Galois (Extension, rnd, toP)
+import qualified Protocol.Groth as Groth
+import QAP
+import qualified Data.Map as Map
+import Fresh (evalFresh, fresh)
+import Juvix.Backends.ArithmeticCircuit.Compilation as Base
+import Circuit
+import Protolude
```
mariari, line 16: Please replace with `Juvix.Library`

src/Juvix/Backends/ArithmeticCircuit/ZKP.hs
```diff
@@ -0,0 +1,51 @@
+module Juvix.Backends.ArithmeticCircuit.ZKP where
+
+import Circuit.Arithmetic (generateRoots)
+import Crypto.Number.Serialize
+import Data.Aeson
```
mariari, line 5: Seems these imports are unused

src/Juvix/Backends/ArithmeticCircuit/ZKP.hs
```diff
@@ -0,0 +1,51 @@
...
+import Data.Pairing.BN254 (BN254, Fr, Pairing (..))
+import qualified Data.Curve.Weierstrass.BN254 as G1
+import qualified Data.Curve.Weierstrass.BN254T as G2
+import Data.Field.Galois (Extension, rnd, toP)
```
mariari, line 9: Same here, and `toP` is unused

package.yaml
```diff
@@ -43,6 +43,12 @@ dependencies:
   - aeson
   - Cabal
   - extensible-data
+  - arithmetic-circuits
+  - galois-field
+  - crypto-numbers
+  - zkp
+  - pairing
+  - MonadRandom
```
mariari, line 51: We should probably Comment that these came from arithCircuit backend.

Will probably need to merge upstream so all the comments on where what deps come from apply.

package.yaml
```diff
@@ -74,6 +80,8 @@ default-extensions:
   - ConstraintKinds
   - EmptyDataDeriving
   - TemplateHaskell
+  - NamedFieldPuns
+  - TupleSections
```
mariari, line 84: Should these be in the defaults?

I like both of these (and have used them in places), and would be fine for adding it, but we should probably decide.

stack.yaml
```diff
@@ -38,6 +38,11 @@ extra-deps:
 - userid-0.1.3.4@sha256:607890fb4363c1d63368a2e96920d3671955feae815d5e2622e35717badad6f6,1172
 - boomerang-1.4.5.7@sha256:67358e1ed44db2da2bf5ba8b5ccde4e3cd19adbc286c5308dba448f89fb03c4a,1979
 - web-routes-th-0.22.6.5@sha256:bdbb13f1c2307ca8b27c45c6cfd03e5c9b014c096e9b661a672625d46e0f6bb2,1561
+- arithmetic-circuits-0.2.0@sha256:8c1d1b2aa22cb1df41c810d4303ef44db172de454cd9be0e2647c9345ddaa48a,5774
+- galois-fft-0.1.0@sha256:6a1e1e255c236b521379435ba528bbb5bc2a159c3ebb7fb960090cd5cf950b20,2409
+- bulletproofs-1.1.0@sha256:eebdf3963372e40675fd501f3f3cc6f553d3925965ce38a112998a321cc7f6f2,5395
+- github: adjoint-io/zkp
+  commit: 2004a19bb097f34a23cb01d6927965b23c9f0f12
```
mariari, line 45: Same commenting deps here

src/Juvix/Backends/ArithmeticCircuit/Compilation.hs
cyberglot: I use them in testcases to compile to `ArithCircuits`, should I move it to tests?

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cyberglot: I think it should be caught by typechecking? You can only pass `Int` to the second parameter anyway?

src/Juvix/Backends/ArithmeticCircuit/Proof.hs
cyberglot: Oh damn, I thought I had deleted it. I replaced it with `ZKP.hs`.

package.yaml
cyberglot: They won't be here for the long term anyway, but I agree that we should comment about it so we know if/when we need to remove them.

package.yaml
cyberglot: They are there just to make my life easier and not write too much. Regardless, I think they relatively are harmless.

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cwgoes: Yeah, it should be, but we should still fail gracefully

src/Juvix/Backends/ArithmeticCircuit/Compilation.hs
cwgoes: If we only use it in test-cases, let's move it to the tests file, yeah

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cyberglot: True, major oversight on my part, thanks for pointing that out. I had a more complicated function before that that was made less bad by `let .. in` and I didn't think of rewriting it as applicative later.

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cyberglot: I haven't do any typechecking for it yet. Glad to add it now 🙂

package.yaml
cyberglot: Removed them all.

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cyberglot: The problem I had was that I cannot make `ArithExpression` a `Functor`, that's why I need pattern matching :( Problem wasn't solved by the `MonadState` unfortunatelly.

test/Main.hs
```diff
@@ -26,9 +26,9 @@ backendTests :: T.TestTree
 backendTests =
   T.testGroup
     "Backend tests"
-    [ ArithmeticCircuit.backendCircuit,
-      LLVM.backendLLVM,
-      Michelson.backendMichelson
+    [ ArithmeticCircuit.backendCircuit
+    -- LLVM.backendLLVM,
```
cwgoes, line 30: Why are these commented now?

test/Main.hs
```diff
@@ -38,11 +38,11 @@ allCheckedTests :: T.TestTree
...
-    [ coreTests,
-      backendTests,
-      frontEndTests,
-      EAC2.eac2Tests,
-      Erasure.erasureTests
+    [ --coreTests,
+      backendTests
+      -- frontEndTests,
+      -- EAC2.eac2Tests,
+      -- Erasure.erasureTests
```
cwgoes, line 45: Ditto, why are these commented now?

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
```diff
@@ -1,19 +1,48 @@
...
+  | Eq
+  | And
+  | Or
+  deriving (Show, Eq)
 
-type WireType = ()
+data UnaryOp = Neg
+  deriving (Show, Eq)
+
+type Term = CoreErased.AnnTerm () PrimVal
```
cwgoes, line 32: Is there a way to convert an `AnnTerm Ty Val` (from `Parameterisation.hs`)  to this type? We'll need that, since the type defined by the parameterisation is what the core type-checker will use.

test/Main.hs
cyberglot: Omg, this wasn't supposed to be committed D:

src/Juvix/Backends/ArithmeticCircuit/Compilation/Types.hs
cyberglot: Probably? I just need a `* -> * -> *` datatype to implement the interface in Parameterisation.

cyberglot:
> I don't understand why the formatter changes so many other files? Are we using different versions, perhaps?
I'm running this version:
```~ λ ormolu --version
ormolu 0.1.0.0 UNKNOWN UNKNOWN
using ghc-lib-parser 8.10.1.20200523```

Is it the correct one?

cwgoes:
> > I don't understand why the formatter changes so many other files? Are we using different versions, perhaps?
> 
> I'm running this version:
> 
> ```
> ormolu 0.1.0.0 UNKNOWN UNKNOWN
> using ghc-lib-parser 8.10.1.20200523
> ```
> 
> Is it the correct one?

Ah, it looks like you have the latest version & the rest of us probably need to upgrade, that's fine then.

cwgoes:
I still think we'll need a bit of glue to hook this all up, but that can wait for the pipeline work.

PR #366, opened 2020-03-31
Updating the Parser to properly parse the example contract

mariari:


cwgoes:
Looking good - can we actually just set up the tests to load example contracts from files (instead of inlining them as strings)? I'm sure we'll want to do that eventually to keep the tests manageable.

mariari:
> Looking good - can we actually just set up the tests to load example contracts from files (instead of inlining them as strings)? I'm sure we'll want to do that eventually to keep the tests manageable.

Yes, though I wasn't sure where the file would live, so I made a macro to copy line by line.

It's easy enough to read.

cwgoes:
We can just make a `test/contracts` directory or the like.

cwgoes:
- [ ] Also treat terms & types in the same syntactic fashion

Issue #373, opened 2020-04-21
Michelson: Fix usage and stack shuffling,


mariari:
Currently our backend for Michelson when considering usages has an issue.

consider the following examples

```ocaml
let f y x z = (\a -> a * x) x

f 2 3 4

let g y x z = x * x

g 2 3 4  
```

In both the g and f case, we can deduce that x has two usages while y and z has 0 usages.

This is great, however under our current model, we do the following
```
Executing g:

-- assume non constnat
Instructions: push 2, push 3, push 4. 
VStack: [(z,0), (x,2), (y,1)]

-- See lookup X
Instructions: SeqEx [PrimEx (DIG 1),PrimEx (DUP @),PrimEx (DUG 2)]
VStack [(x,1),(z,0), (x,1), (y,1)]

-- See lookup X
Instructions: DIG 0
VStack [(x,1),(z,0), (x,1), (y,1)]

-- Mult
```

The result is similar for f.

This happens because we move `n-1` usages of the variable (in this case `x`) to the front, leaving 1 usage behind. This has adverse behavior when we have to use the same x in the same application twice or more.

f shows us a more pathological case, as if we desugar the execution, we don't have `x` being called twice in the lambda.

I believe there are two ways to fix this issue.

Both these ways, have a new notion in the `VStack`, known as `save`.

What `save` does, is that it marks 1 usage as saved, and thus effectively to move it to the front, means to dup it (as its' been saved in it's position).

Thus here are two solutions

1) The easy solution, just implement save, and do no changes to how usages are carried

    An example of what this may look like is the following
   ```
   -- First Lookup X from Example above
   Instructions: SeqEx [PrimEx (DIG 1),PrimEx (DUP @),PrimEx (DUG 2)]
    VStack [(x,1, SAVE),(z,0), (x,1), (y,1)]
   -- Second Lookup X from Example
   Instructions: PrimEx (DIG 2)
    VStack [(x,1, SAVE), (x,1, SAVE), (z,0), (y,1)]
    ````

    This would also work under renaming with minimal modification.
    If a variable is renamed, and it is saved, we would have to look for all aliases as well
    Our current model does not do this, but could easily be done.

2) The Hard solution. Just move all n usages up, When a variable is about to be consumed with extra usages left, copy it before the execution spot.

    This solution should be more efficient, as instead of `DIG 2` above, we could instead just do a `dup`, and even avoid the dig dug.

[...]

PR #374, opened 2020-04-21
Minor Stack Fix

mariari:
Currently this will implement Strategy 1 in #373 

Strategy 2 can safely be done once we have a normalizing Core up.

### DIFF
--- a/src/Juvix/Backends/Michelson/Compilation/VirtualStack.hs
+++ b/src/Juvix/Backends/Michelson/Compilation/VirtualStack.hs
@@ -214,11 +214,39 @@ lookupType n (T stack' _) = go stack'
     go ((_, _) : xs) = go xs
     go [] = Nothing
 
+constToInstr :: T.Type -> V.Value' Instr.ExpandedOp -> Instr.ExpandedOp
 constToInstr ty c =
   case c of
     V.ValueNil ->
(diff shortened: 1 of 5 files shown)


src/Juvix/Backends/Michelson/Compilation/VirtualStack.hs
```diff
@@ -41,22 +41,44 @@ data T lamType
       }
   deriving (Show, Eq)
 
+--------------------------------------------------------------------------------
+-- Usage Information
+--------------------------------------------------------------------------------
+
+type Saved = Bool
```
cwgoes, line 48: Hmm can we `newtype` this?

mariari: I'd rather not, that way I can use it in boolean expressions.

```haskell
isUsageFree :: Usage -> Bool
isUsageFree (Usage x saved)
  | x == mempty
      || x == one && saved =
    False
  | otherwise = True
``

for example

cwgoes: Ah, OK

mariari:
> ACK modulo tests
> 
> However, I think we should use method 2 in the future.

I agree, sadly we can't do that safely until normalization is in.

PR #381, opened 2020-04-25
Add datatype stuff to the type checker

andy-morris:
This doesn't build yet, but I'm pushing it early so @thealmarty can take a look at the datatype representation and let me know if it's no good

edit: oops also it's on top of #380, sorry! I can untangle it if necessary. But the actual datatype stuff is https://github.com/cryptiumlabs/juvix/blob/25a0f64364905af9e929f5b969a1ba4e9d565241/src/Juvix/Core/IR/Typechecker/Env.hs#L19-L52

### DIFF
--- a/src/Juvix/Core/IR/Typechecker/Env.hs
+++ b/src/Juvix/Core/IR/Typechecker/Env.hs
@@ -22,6 +22,7 @@ type Globals primTy primVal =
 data Global primTy primVal
   = GDatatype (Datatype primTy primVal)
   | GDataCon  (DataCon  primTy primVal)
+  | GFunction (Function primTy primVal)
   deriving (Show, Eq, Generic)
 
 data Datatype primTy primVal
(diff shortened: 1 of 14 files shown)


cwgoes:
I merged #380, so the diff should be clean now.

andy-morris:
This can be merged if it looks ok for everyone else. The evaluator doesn't reduce functions yet, because that's a bit of a big job that shouldn't hold up everything else, so I'll put it in a separate PR when it's done

src/Juvix/Core/IR/Typechecker/Env.hs
```diff
@@ -1,11 +1,25 @@
...
+      }
+  deriving (Show, Eq, Generic)
+
+type Globals primTy primVal =
+  HashMap IR.GlobalName (Global primTy primVal)
+
+data Global primTy primVal
+  = GDatatype (IR.Datatype primTy primVal)
+  | GDataCon (IR.DataCon primTy primVal)
+  | GFunction (IR.Function primTy primVal)
```
cwgoes, line 22: Is the idea that top-level functions will end up as `GFunction` ?

andy-morris: Yeah, exactly, the intention is that a top level function (or a lifted one) is an element in the typechecker state of type `IR.Function`, and a datatype is one `Datatype` and zero or more `DataCon`s.

A datatype also has a list of all its constructors because this will presumably be necessary for the coverage checker etc

PR #388, opened 2020-05-05
Dummy Contract Env And Interpreter

mariari:
Gets the Interpreter for Michelson up. To a state where we can easily use it.


mariari:
> Semantic test-cases for compiled contracts (using the interpreter)

this part is now in and being done!

src/Juvix/Backends/Michelson/Compilation/Types.hs
```diff
@@ -52,10 +53,21 @@ data CompilationLog
   | OptimisedByMorley SomeInstr SomeInstr
   deriving (Generic, Show)
 
+data EmptyInstr where
```
cwgoes, line 56: Should we clean these up?

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -37,6 +44,52 @@ constType v =
...
+              let newTerm =
+                    Run.applyPrimOnArgs (Prim.toAnn fun) (Prim.toAnn <$> args)
+                  -- TODO ∷ do something with the logs!?
+                  (compd, _log) = Compilation.compileExpr newTerm
+               in case compd >>= Interpreter.dummyInterpret of
+                    Right x ->
+                      Constant x
+                        |> Prim.Return
+                        |> Right
+                    -- TODO :: promote this error
```
cwgoes, line 84: I believe that this failure indicates a mistake in the typechecker, which should have been run already by this stage

test/Backends/Michelson.hs
```diff
@@ -19,24 +20,44 @@ import Prelude (show)
...
+runExpr term =
+  fst (compileExpr term)
+
+runContractWrap :: Term -> Type -> Either DSL.CompError (Contract' ExpandedOp)
+runContractWrap term ty =
+  runContract (Ann zero newTy (J.LamM [] ["gen%%%"] term)) newTy
+  where
+    newTy = J.Pi zero (primTy unitPair) ty
+
+shouldCompExp :: AnnTerm PrimTy NewPrim -> M.Value -> T.TestTree
```
cwgoes, line 37: Confusing name, but OK

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -37,6 +44,52 @@ constType v =
...
+    (Prim.Return PrimTy PrimVal)
+applyProper fun args =
+  case Prim.term fun of
+    Constant _i ->
+      case length args of
+        0 ->
+          Right (Prim.Return (Prim.term fun))
+        _x ->
+          Left (CoreTypes.TypecheckerError "Applied a constant to argument")
+    Inst instruction ->
```
cwgoes, line 61: Somewhere we need to deal with dependency, namely, polymorphic functions in Michelson will be dependent in Juvix, e.g. `ADD`:

```idris
add : (t : Type) -> t -> t -> t
```

— so we need to deal with the first application to `t` (e.g. `Int` or `Tez`) somewhere, how do you think we should do that?

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
 dependencies:
   - base >= 4.11 && < 5
   - protolude
+  - directory
+  - lens
+  - Cabal
+  - extensible-data
+  - capability >= 0.2.0.0
+  ###########################
+  # Depedency for OS
```
cwgoes, line 25:
```suggestion
  # Dependency for OS
```

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
...
+  ###########################
+  - turtle
+  - time
+  ###########################
+  # Depedency for Arithmetic
```
cwgoes, line 30:
```suggestion
  # Dependency for Arithmetic
```

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
...
+  ###########################
+  - bulletproofs
+  - elliptic-curve
+  - sonic
+  ###########################
+  # Depedency for Structs
```
cwgoes, line 36:
```suggestion
  # Dependency for Structs
```

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
...
   - containers
   - unordered-containers
   - hashable
   - fgl
-  - lens
+  - fgl-visualize
+  - bytestring
+  - enummapset
+  ###########################
+  # Depedency for Parsers
```
cwgoes, line 48:
```suggestion
  # Dependency for Parsers
```

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
...
   - attoparsec
   - attoparsec-expr
-  - capability >= 0.2.0.0
-  - fgl-visualize
-  - enummapset
-  - turtle
-  - directory
+  - aeson
+  ###########################
+  # Depedency for LLVM
```
cwgoes, line 55:
```suggestion
  # Dependency for LLVM
```

package.yaml
```diff
@@ -16,33 +16,54 @@ extra-source-files:
...
+  ###########################
   - llvm-hs
   - llvm-hs-pure
   - llvm-hs-pretty
-  - bytestring
-  - time
+  ###########################
+  # Depedency for Morley
```
cwgoes, line 61:
```suggestion
  # Dependency for Morley
```

src/Juvix/Backends/Michelson/Parameterisation.hs
mariari: I'd still like to reserve the errors of the compiler itself, and so an Either over Maybe is nicer?

src/Juvix/Backends/Michelson/Compilation/Types.hs
mariari: Yes I'll remove them

cwgoes:
Per discussion, we need to add monomorphic version of all primitives, e.g. `add_int_int`, `add_nat_nat`.

test/Backends/Michelson.hs
mariari: Changed to interpretExpression

src/Juvix/Backends/Michelson/Parameterisation.hs
mariari: This is now promoted, however I changed the 

Juvix.Core.Types.TermAssignment

PR #396, opened 2020-05-18
Added Context structure to be used in the various compiler phases

mariari:


src/Juvix/Core/Common/Context.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE DeriveTraversable #-}
```
cwgoes, line 1: Are we using this extension?

mariari: to get the following deriving
```haskell
  deriving (Show, Traversable)
```

It seems we can't derive Traversable via `via` for whatever reason.

If we ever do something with efffects on this structure, then we'll need traverse

cwgoes: Ah yes, misread.

PR #400, opened 2020-05-22
Record Fix and Parser Slowness fix

mariari:
Test via

```bash
stack test

 time .stack-work/dist/x86_64-linux-tinfo6/Cabal-3.0.1.0/build/juvix-test/juvix-test
```

the last binary may change locations based on your system.


mariari:
Some minor testing

```org
- let foo = 3 in foo
  | number | Time                          |
  |--------+-------------------------------|
  |      1 | real 0m0.042s                 |
  |      5 | real 0m0.052s                 |
  |     10 | real 0m0.272s  user 0m1.480s  |
  |     20 | real 0m47.748s user 3m28.042s |
  + many expression
- a
  | number | Time                        |
  |--------+-----------------------------|
  |    160 | real 0m0.059s user 0m0.021s |
  |    320 | real 0m0.083s user 0m0.232s |
  |    640 | real 0m0.079s user 0m0.127s |
  |   1280 | real 0m0.062s user 0m0.121s |
  + many' argsSN
  + turns out to not be that different between calls
```

cwgoes:
Can we add these sorts of benchmarks to CI, or at least a script in the repository, perhaps?

mariari:
> Can we add these sorts of benchmarks to CI, or at least a script in the repository, perhaps?

Sure, we'll add a berncharmking lib.


Also 

# Finding the issue.

Consider

```ocaml
# 3 + let foo = 3 + let foo = 3 + let foo = 3 in foo in foo in foo;;
- : int = 12
```
is valid in ocaml, and thus in our syntax 

```ocaml
let foo = 3
let foo = 3
let foo = 3
```

thus the parser tries to parse application of 3 on the rest and cascade it, upon fidning it can't, it back tracks then does it all over agaoin!


The way out is doing much of what ocaml does


```ocaml
# (+) 3 let foo = 3 in foo;;
Error: Syntax error

# (+) 3 (let foo = 3 in foo);;
- : int = 6
```

basically if the form is ambi if it expects more things, then don't allow it in application!

cwgoes:
I see - that seems quite reasonable.

mariari:
The problem is fixed, just need to revert back the tests to the old tests

mariari:
Going to merge now, and make a new pr which utilizes stack bench

PR #403, opened 2020-05-28
Starting type translation and stage2 pass

mariari:


src/Juvix/FrontendDesugar/RemoveDo/Transform.hs
```diff
@@ -0,0 +1,285 @@
...
+
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Juvix.FrontendDesugar.RemoveDo.Types as New
+import qualified Juvix.FrontendDesugar.RemovePunned.Types as Old
+import Juvix.Library
+
+-- The actual transform we are doing
+
+transformDo :: Old.Do -> New.Expression
+transformDo (Old.Do'' body) = foldr f (transformLast last) body
```
cwgoes, line 11: ... we need to also solve the problem of lifting application over an error monad

cwgoes: ... and a state monad

cwgoes: maybe we should use `MonadIdentity` ( =/ )

src/Juvix/Frontend/Types/Base.hs
```diff
@@ -0,0 +1,443 @@
...
+      deriving (Show, Generic, NFData)
+
+    data Product
+      = Record !Record
+      | Arrow Expression
+      | ADTLike [Expression]
+      deriving (Show, Generic, NFData)
+
+    data Record
+      = Record''
```
andy-morris, line 135: I wonder if we should maybe come up with a different naming scheme for these things than variously having a single/double `'`?

Maybe like `RecordInfo` or something, but I'm not that fussed about what exactly if someone has a better suggestion.

mariari: Yes  I would like to do that at some point

PR #407, opened 2020-06-12
Start datatypes to Michelson

cwgoes:
Closes https://github.com/metastatedev/juvix/issues/405

Work in progress, need to clean up the erasure pipeline first.

I think erasure needs to happen before datatype conversion since e.g. datatypes with 0-usage members will change.

Current ordering:

- IR + datatypes (type-checked)
- Erasure step
- Erased core + datatypes
- Data-type compilation step
- Erased core simplified
- Auxiliary conversion step
- Erased core annotated (with multi-arg lambdas as Michelson uses)

Tasks:

- [x] Switch `Juvix.Core.Erased` to use extensible (extending `Juvix.Core.IR.Types.Base`)
- [x] Switch to `HRAnn` in erasure, passing through to annotated `ErasedAnn` after conversion
- [x] Write conversion function from erased to `ErasedAnn`, which the Michelson backend seems to use at the moment
- [x] Confirm structure, write test-cases
- [x] Update erasure algorithm to pass through datatypes, result in some sort of erased + datatypes
- [ ] Write pass to transform erased + datatypes into erased for Michelson
   - [ ] Simplify complex data-types & pattern matches (per https://github.com/metastatedev/juvix/issues/405)
   - [ ] Search for data constructors by name, turn them into Michelson instructions
   - [ ] Nest pattern matches (per https://github.com/metastatedev/juvix/issues/405)
   - [ ] Compile pattern matches (per https://github.com/metastatedev/juvix/issues/405)


src/Juvix/Core/IRAnn/Extend.hs
```diff
@@ -0,0 +1,37 @@
...
+
+import qualified Extensible as Ext
+import qualified Juvix.Core.IR.Types.Base
+import qualified Juvix.Core.IR.Types.Base as IR
+import qualified Juvix.Core.Usage as Usage
+import Juvix.Library
+
+data T
+
+data Annotation primTy primVal
```
cwgoes, line 11: ... this should be abstracted over `T` to avoid later repetition

cwgoes:
Closing in favour of https://github.com/metastatedev/juvix/pull/450.

PR #408, opened 2020-06-16
Finishing Phase-2 Passes

mariari:


src/Juvix/FrontendDesugar/RemoveCond/Extend.hs
```diff
@@ -0,0 +1,13 @@
+module Juvix.FrontendDesugar.RemoveCond.Extend
```
cwgoes, line 1: Ah yes, this is a nice trick

PR #410, opened 2020-06-19
Add some -Werror flags

andy-morris:
It turned out there are already are a couple of nonexhaustive matches so I think I should fix those before merging this

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           no_output_timeout: 1200
       - run:
           name: Build Juvix
-          command: stack build --copy-bins --fast
+          command: stack build --copy-bins --fast --flag juvix:incomplete-error
       - persist_to_workspace:
           root: /root
           paths:
@@ -36,7 +36,7 @@ jobs:
           at: /root
       - run:
           name: Run tests
-          command: stack test --fast --jobs=1 --test-arguments "--hide-successes --ansi-tricks false" +RTS -N4
+          command: stack test --fast --flag juvix:incomplete-error --jobs=1 --test-arguments "--hide-successes --ansi-tricks false" +RTS -N4
           no_output_timeout: 1200
 
 workflows:
--- a/package.yaml
+++ b/package.yaml
@@ -113,6 +113,9 @@ ghc-options:
   - -fno-warn-missing-pattern-synonym-signatures
   - -j6
   - -static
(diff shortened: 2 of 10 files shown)


andy-morris:
I did the ones I could but I don't know enough about Michelson or interaction nets to finish these definitions:

- `Juvix.Interpreter.InteractionNet.interactionNetASTToErasedCore`
- `Juvix.Backends.Michelson.DSL.Instructions.toNumArgs`
- `Juvix.Backends.Michelson.Parameterisation.{typeOf,constType}`

andy-morris:
Changed back to a warning just for those modules for now.

andy-morris:
@mariari @thealmarty would this change be too disruptive to your way of working? If so maybe this could be hidden behind a package flag which is enabled in the CI or something

thealmarty:
> @mariari @thealmarty would this change be too disruptive to your way of working? If so maybe this could be hidden behind a package flag which is enabled in the CI or something

Not at all! I prefer turning on `-Werror` too! I have suggested this before. Thanks for this!

mariari:
> @mariari @thealmarty would this change be too disruptive to your way of working? If so maybe this could be hidden behind a package flag which is enabled in the CI or something

The main issue I have with this, is that when I want to write new code (or refactor existing code), I often want to play with types in the repl or query it for information.

So being forced to first finish said changes (manually doing it in steps from the repl->File before playing around with forms), would likely disrupt how I flow.

With that said, I'd prefer it to happen on `stack build` that way we get the error behavior without it screwing with the haskell REPL.

If we can't do that, then the CI thing would be good

andy-morris:
I thought that might be the case. I moved it into a package flag `incomplete-error`, which the CI uses, and you can use yourself with `stack build --flag juvix:incomplete-error`.

I also noticed that `package.yaml` contained `-j6`. I changed it to just `-j` to base the number of jobs on many cores the current machine has.

andy-morris:
The arithmetic circuit stuff has some missing cases and methods. I demoted the errors to warnings in those modules too, but if it is easy to complete the code it would probably be better to do that if someone that understands that part of the code has the chance.

andy-morris:
Oops! #417

cwgoes:
Can we merge this or do you want to wait on https://github.com/metastatedev/juvix/pull/417?

cwgoes:
(bump)

PR #413, opened 2020-06-19
Add 'let' to core

andy-morris:
More specifically, a `let` binds an *elimination* to a single variable (not an arbitrary pattern), and the body is a term.

### DIFF
--- a/src/Juvix/Core/HR/Types.hs
+++ b/src/Juvix/Core/HR/Types.hs
@@ -12,7 +12,9 @@ IR.extendTerm "Term" [] [t|T|] $
       { IR.nameLam = "Lam0",
         IR.typeLam = Just [[t|Symbol|]],
         IR.namePi = "Pi0",
-        IR.typePi = Just [[t|Symbol|]]
+        IR.typePi = Just [[t|Symbol|]],
+        IR.nameLet = "Let0",
+        IR.typeLet = Just [[t|Symbol|]]
       }
 
 -- TODO allow extendTerm to reorder fields?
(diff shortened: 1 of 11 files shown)


andy-morris:
I'm going to add some tests and then merge this if that works for everyone else.

andy-morris:
I've just realised: is having `let` in base Core even something we want?

My first impression was "yes, of course we do, we have let in the surface syntax and we will need it later for ANF". But the thing is the `let`s in the input program and those in ANF are basically unrelated.

- In some pass before vanilla Core we will have lifted all the source `let` bindings to the top level (we will need to lift recursive and/or matching ones, so we may as well do all of them).
- ANF needs to introduce a binding for *everything* other than variables and primitive values, so which expressions also had a binding in the source program isn't really relevant.

So I think `let` might actually be better as an extension in the pass that produces ANF.

andy-morris:
After a discussion with @mariari: lifting indiscriminately can inferfere with the optimiser, so actually having a Core `let` and keeping bindings from the source where possible is probably the right thing to do

cwgoes:
Hmm. I'm pretty concerned about having performance depend on whether and/or where the user places `let` in the frontend language, I think we should try to re-calculate such things in the compiler, if possible, since it may not be clear to users what the performance implications are. That can be addressed later, though.

mariari:
> Hmm. I'm pretty concerned about having performance depend on whether and/or where the user places `let` in the frontend language, I think we should try to re-calculate such things in the compiler, if possible, since it may not be clear to users what the performance implications are. That can be addressed later, though.

Well one cited issue with always lambda lifting (it seems for GHC, they found this for after 5), is that register spilling may occur and performance degredation may happen. 

Something programers often want is being able to look at a piece of code and determine how performant it may be. This is often a selling point of languages like OCaml, C, and (I assume) Rust. 

I think this is important to note, in that if we apply a lot of transformations that have some degredation in code that it should be clear. 
  - Note this is impossible, as once inlining happens, we'll start to see odd performance regressions due to code size.
  - But overall, more of these may hurt the user being able to see how to effective optimize code

Also besides that `let` will be optimized rather heavily due to an anf transformation that will utilize the existing lets

andy-morris:
Is this ok with everyone to merge?

PR #415, opened 2020-06-22
ran the formatter

mariari:


mariari:
This pr adds `make build-format` and uses 0.0.3.1 ormulo until we change to GHC 8.10

PR #416, opened 2020-06-22
Some Contract Examples

mariari:


cwgoes:
We should have examples which:
- Trigger sum type logic (`either a b`, match on which it is)
- Include a nested match
- Use helper functions

cwgoes:
We should also port [this prior example](https://github.com/metastatedev/juvix/blob/develop/experimental/juvix/contract.jvx) to the examples directory.

cwgoes:
Per suggestion by @mariari 
- Remove strings
- No numbers in variable names for now
- Don't use `val`, it's reserved

cwgoes:
More recent examples have been since merged.

PR #417, opened 2020-06-23
Add 'let' to forgotten places

andy-morris:
So um, like I said, I often lose the 'incomplete patterns' warnings amongst the rest of Stack's output. Sorry.

Anyway, part of this PR is adding `let` to erased core. Unfortunately I have no idea what to do for `Juvix.Core.EAC.ConstraintGen.boxAndTypeConstraint` so I'll need some help with that.


cwgoes:
> Anyway, part of this PR is adding let to erased core. Unfortunately I have no idea what to do for Juvix.Core.EAC.ConstraintGen.boxAndTypeConstraint so I'll need some help with that.

Just error for now, we'll deal with it later.

PR #419, opened 2020-06-24
transform.hs template for Stage 3 frontend context passes

thealmarty:
This PR has a template for `transform.hs` and all its supporting infrastructure. It's done in the folder for the erase type alias pass, but that pass has not finished.


cyberglot:
Change this PR to be into #420 rather than into develop.

thealmarty:
> Change this PR to be into #420 rather than into develop.

I'd rather not. It keeps better track this way.

cyberglot:
> > Change this PR to be into #420 rather than into develop.
> 
> I'd rather not. It keeps better track this way.

What you mean? We will need to get #420 and this one here into one anyway before merging into develop. I cannot see why have multiple PRs that will get develop into a broken state since they're the same feature.

package.yaml
```diff
@@ -165,6 +166,13 @@ library:
     - Juvix.FrontendDesugar.RemoveDo.Transform
     - Juvix.FrontendDesugar.RemoveDo.Extend
     - Juvix.FrontendContextualise.Contextify.Transform
+    - Juvix.FrontendContextualise.EraseTypeAliases.Types
+    - Juvix.FrontendContextualise.EraseTypeAliases.Transform
+    - Juvix.FrontendContextualise.EraseTypeAliases.Extend
+    - Juvix.FrontendContextualise.EraseTypeAliases.Environment
+    # - Juvix.FrontendContextualise.InfixPrecedence.Types
+    # - Juvix.FrontendContextualise.InfixPrecedence.Transform
+    # - Juvix.FrontendContextualise.InfixPrecedence.Extend
```
mariari, line 175: We should probably remove these

src/Juvix/FrontendContextualise/InfixPrecedence/ShuntYard.hs
```diff
@@ -1,91 +0,0 @@
...
-
--- This list should be of length 1 after all is said and done, and an
--- application given by shunt
-combine :: [Application a] -> Application a
-combine (x : _) = x
-combine [] =
-  error "precondition failed: Shunt.combine was given an empty list"
-
-predSymbol :: Precedence -> Symbol
-predSymbol (Pred s _ _) = s
```
mariari, line 91: We should merge with develop so this file comes back

src/Juvix/FrontendContextualise/FrontendContextualise.hs
```diff
@@ -0,0 +1,18 @@
...
+contextualise :: [Initial.TopLevel] -> Env.Context Target.TopLevel
+contextualise initials =
+  Env.forKey
+    (linearContextualisation initials)
+    [ -- all passes in order
+      Infix.transformTopLevel
+    ]
+
+linearContextualisation :: [Initial.TopLevel] -> Env.Context Target.TopLevel
+linearContextualisation = undefined
```
mariari, line 18: We can remove this for now, I'll recreate this once our passes are over

src/Juvix/FrontendContextualise/InfixPrecedence/ShuntYard.hs
thealmarty: I merged with `develop` already but I thought the idea is to remove all stuff related to other passes? I can easily add this file back to this PR now if desired. Lmk.

thealmarty:
> Good job 
> 
> Some minor changes still need to be cleaned up, but we can merge once those are done

Thank you! Just pushed with the requested changes.

Issue #412, opened 2020-06-19
Frontend context passes (phase 3)


cwgoes:
From https://github.com/metastatedev/juvix/issues/394, in greater detail:

The frontend context passes need to start from desugared frontend ([here](https://github.com/metastatedev/juvix/blob/develop/src/Juvix/FrontendDesugar/RemoveDo/Types.hs), which can be transformed into from the frontend syntax using [this logic](https://github.com/metastatedev/juvix/blob/develop/src/Juvix/FrontendDesugar.hs)) and output [core + datatypes](https://github.com/metastatedev/juvix/blob/develop/src/Juvix/Core/IR/Types/Base.hs).

They should start by creating a [context](https://github.com/metastatedev/juvix/blob/develop/src/Juvix/Core/Common/Context.hs) and putting all of the definitions in it for lookup. Then, the following passes need to happen, in approximately this order:

1. Deal with module opens, record opens (context pass) (@cyberglot)
    1. Resolving names
1. De-sugaring records (@cyberglot)
    1. Translate constructors into regular data type constructors
    1. Translate named field accessors into numbered field accessors (potentially)
    1. Keep names around as metadata
    1. Translate to core in the most straightforward way
1. Deal with type aliases or desugar (@thealmarty)
    1. Erase type aliases (translate into whatever the alias points to)
    1. Later on, the type-checker will treat them differently (& we'll keep metadata)
1. Deal with newtype declarations (desugar)
    1. Erase newtype declarations
    1. Later on, the type-checker will treat them differently (& we'll keep metadata)
1. Resolve operators by precedence (@cyberglot)
    1. Self-evident
1. Convert frontend representation of datatypes & pattern-matching into the core representation of datatypes & pattern-matching
    1. Hopefully should be straightforward, ask questions if problems arise
    1. Check with @andy-morris 
    1. Also see whether we really need this

After those passes, we should end up with core + datatypes, we'll need to check that all of the types match up correctly.

cwgoes:
The structure of what this will require is pretty similar to https://github.com/metastatedev/juvix/pull/408/files.

cwgoes:
Additional things we might need to deal with:
- Named types
- Float let-types to the top-level and rename them properly
- Convert constants into backend-specific representation
- Desugar arrow expressions into Pi

thealmarty:
[...]

PR #420, opened 2020-06-24
Context Passes as defined in #412

cyberglot:
- [x] Environment
 - [ ] Initial context transformation (from base types to definitions, ongoing)
 - [ ] Infix Precedence (ongoing)
 - [ ] De-sugaring records
 - [ ] Module and Record Open


cwgoes:
What's the state of this PR / should we close this?

PR #428, opened 2020-07-01
Cleaned up Some parts of Arithmetic Circuit

mariari:
Namely I've

1. Created type alises
2. made sure code become 80 characters per line or less
3. abstracted some patterns
4. used the capability signature instead of the type itself
5. Qualified all imports modulo library and parser
6. Making things rely more on exhaustion than catchalls
    -   I can do this better
7. Replacing unexhaustive matches with undefineds


cyberglot:
LGTM 👍🏽

PR #435, opened 2020-07-13
Core to Michelson (sans datatypes)

cwgoes:
Split out from the work in https://github.com/metastatedev/juvix/pull/407 for ease of review & integration.

This PR does the following:
- Hooks up the pipeline so that core can be taken all the way to Michelson
- Adds a conversion function from `Erased` types to `ErasedAnn` types
- Adds an annotation for the bind in `let` so that this can be passed down the pipeline
- Fixes a bug in erasure where `Pi` types with zero-usage arguments were not properly erased
- Adds some simple test-cases of type-checking, erasing, and compiling HR terms to Michelson

This does _not_ yet add datatype compilation, that will be a separate PR.

### DIFF
--- a/src/Juvix/Core/ErasedAnn/Conversion.hs
+++ b/src/Juvix/Core/ErasedAnn/Conversion.hs
@@ -1,0 +1,34 @@
+module Juvix.Core.ErasedAnn.Conversion where
+
+import Juvix.Core.ErasedAnn.Types
+import qualified Juvix.Core.Erasure.Types as E
+import qualified Juvix.Core.Usage as Usage
+import Juvix.Library hiding (Type)
+
+convertTerm :: forall primTy primVal m. (Monad m) => E.Term primTy primVal -> Usage.T -> m (AnnTerm primTy primVal)
+convertTerm term usage = do
+  let ty = E.getType term
+  ty' <- convertType ty
+  case term of
+    E.Var sym _ -> pure (Ann usage ty' (Var sym))
+    E.Prim p _ -> pure (Ann usage ty' (Prim p))
+    E.Lam sym body _ -> do
+      -- TODO: Is this the right usage?
+      -- TODO: Deal with arguments, captures.
+      body <- convertTerm body usage
+      pure (Ann usage ty' (LamM [] [sym] body))
+    E.App f a _ -> do
+      f <- convertTerm f usage
+      a <- convertTerm a usage
+      -- TODO: Deal with multi-arg application.
+      pure (Ann usage ty' (AppM f [a]))
+
+convertType :: forall primTy primVal m. (Monad m) => E.Type primTy -> m (Type primTy primVal)
+convertType ty =
+  case ty of
+    E.SymT s -> pure (SymT s)
+    E.PrimTy p -> pure (PrimTy p)
+    E.Pi u a r -> do
+      a <- convertType a
+      r <- convertType r
+      pure (Pi u a r)
--- a/src/Juvix/Core/ErasedAnn/Conversion.hs
+++ b/src/Juvix/Core/ErasedAnn/Conversion.hs
@@ -1,5 +1,6 @@
 module Juvix.Core.ErasedAnn.Conversion where
 
+import Data.List ((\\))
 import qualified Juvix.Core.Erased as Erased
 import Juvix.Core.ErasedAnn.Types
 import qualified Juvix.Core.Erasure.Types as E
@@ -19,7 +20,7 @@ convertTerm term usage = do
     E.Prim p _ -> pure (Ann usage ty' (Prim p))
     E.Let sym bind body (bindTy, _) -> do
       -- Calculate captures.
-      let captures = free (E.Lam sym body undefined)
+      let captures = Erased.free (Erased.Lam sym (E.eraseAnn body))
       -- TODO: Is this the right usage?
       bind <- convertTerm bind usage
       body <- convertTerm body usage
@@ -34,7 +35,7 @@ convertTerm term usage = do
       case body of
         -- Combine nested lambdas into multi-argument function.
         Ann _ _ (LamM cap' arg' body') ->
-          pure (Ann usage ty' (LamM cap' (sym : arg') body'))
+          pure (Ann usage ty' (LamM (cap' \\ [sym]) (sym : arg') body'))
         _ ->
           pure (Ann usage ty' (LamM (free term) [sym] body))
     E.App f a _ -> do
--- a/src/Juvix/Core/ErasedAnn/Conversion.hs
+++ b/src/Juvix/Core/ErasedAnn/Conversion.hs
@@ -1,31 +1,56 @@
 module Juvix.Core.ErasedAnn.Conversion where
 
+import qualified Juvix.Core.Erased as Erased
 import Juvix.Core.ErasedAnn.Types
 import qualified Juvix.Core.Erasure.Types as E
+import qualified Juvix.Core.Types as Types
 import qualified Juvix.Core.Usage as Usage
 import Juvix.Library hiding (Type)
 
-convertTerm :: forall primTy primVal m. (Monad m) => E.Term primTy primVal -> Usage.T -> m (AnnTerm primTy primVal)
+free :: forall primTy primVal. E.Term primTy primVal -> [Symbol]
+free = Erased.free . E.eraseAnn
+
+convertTerm :: forall primTy primVal compErr m. (HasThrow "error" (Types.PipelineError primTy primVal compErr) m) => E.Term primTy primVal -> Usage.T -> m (AnnTerm primTy primVal)
 convertTerm term usage = do
   let ty = E.getType term
   ty' <- convertType ty
   case term of
     E.Var sym _ -> pure (Ann usage ty' (Var sym))
     E.Prim p _ -> pure (Ann usage ty' (Prim p))
+    E.Let sym bind body (bindTy, _) -> do
+      -- Calculate captures.
+      let captures = free (E.Lam sym body undefined)
+      -- TODO: Is this the right usage?
+      bind <- convertTerm bind usage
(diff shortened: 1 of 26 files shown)


src/Juvix/Core/ErasedAnn/Conversion.hs
```diff
@@ -0,0 +1,60 @@
...
+      pure (Ann usage ty' (AppM lam [bind]))
+    E.Lam sym body _ -> do
+      -- TODO: Is this the right usage?
+      body <- convertTerm body usage
+      case body of
+        -- Combine nested lambdas into multi-argument function.
+        Ann _ _ (LamM cap' arg' body') ->
+          pure (Ann usage ty' (LamM (cap' \\ [sym]) (sym : arg') body'))
+        _ ->
+          pure (Ann usage ty' (LamM (free term) [sym] body))
```
mariari, line 40: Yes it would be best to just add `Let` to the Michelson backend instead of converting it. Would be rather simple to set this up in the InstructionsEff

src/Juvix/Core/HR/Extend.hs
```diff
@@ -0,0 +1,24 @@
...
+extTerm =
+  \_primTy _primVal ->
+    IR.defaultExtTerm
+      { IR.nameLam = "Lam0",
+        IR.typeLam = Just [[t|Symbol|]],
+        IR.namePi = "Pi0",
+        IR.typePi = Just [[t|Symbol|]],
+        IR.nameLet = "Let0",
+        IR.typeLet = Just [[t|Symbol|]]
+      }
```
mariari, line 16: The default extension should already handle this for you, no!?

src/Juvix/Core/HRAnn/Extend.hs
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+extElim =
+  \primTy primVal ->
+    IR.defaultExtElim
+      { IR.typeBound = Nothing,
+        IR.typeFree = Nothing,
+        IR.typeElimX = [("Var", [[t|Symbol|]])],
+        IR.nameApp = "App0",
+        IR.typeApp = Just [[t|AppAnnotation $primTy $primVal|]]
```
mariari, line 56: Wouldn't this be handled by the default extension already?

src/Juvix/Core/HRAnn/Extend.hs
```diff
@@ -0,0 +1,57 @@
...
+  \primTy primVal ->
+    IR.defaultExtTerm
+      { IR.nameLam = "Lam0",
+        IR.typeLam = Just [[t|BindAnnotation $primTy $primVal|]],
+        IR.namePi = "Pi0",
+        IR.typePi = Just [[t|Symbol|]],
+        IR.nameLet = "Let0",
+        IR.typeLet = Just [[t|LetAnnotation $primTy $primVal|]],
+        IR.nameElim = "Elim0",
+        IR.typeElim = Just [[t|Annotation $primTy $primVal|]]
```
mariari, line 40: Same goes for this

src/Juvix/Core/Pipeline.hs
```diff
@@ -1,18 +1,54 @@
...
 import Juvix.Library
+import qualified Michelson.Printer as Michelson
+import qualified Michelson.TypeCheck as Michelson
+import qualified Michelson.Untyped as Michelson
+
+coreToMichelson ::
+  ( HasWriter "log" [Types.PipelineLog Michelson.PrimTy Michelson.PrimVal] m,
+    HasReader "parameterisation" (Types.Parameterisation Michelson.PrimTy Michelson.PrimVal) m,
+    HasThrow "error" (Types.PipelineError Michelson.PrimTy Michelson.PrimVal Michelson.CompErr) m,
+    HasReader "globals" (IR.Globals Michelson.PrimTy Michelson.PrimVal) m
```
mariari, line 20: We'll want to alias these, esp since they are quite long

src/Juvix/Core/Pipeline.hs
```diff
@@ -1,18 +1,54 @@
...
+  ) =>
+  HR.Term Michelson.PrimTy Michelson.PrimVal ->
```
mariari, line 22: Even for an alias for this would be nice

src/Juvix/Core/Pipeline.hs
```diff
@@ -1,18 +1,54 @@
...
+coreToMichelsonContract ::
+  ( HasWriter "log" [Types.PipelineLog Michelson.PrimTy Michelson.PrimVal] m,
+    HasReader "parameterisation" (Types.Parameterisation Michelson.PrimTy Michelson.PrimVal) m,
+    HasThrow "error" (Types.PipelineError Michelson.PrimTy Michelson.PrimVal Michelson.CompErr) m,
+    HasReader "globals" (IR.Globals Michelson.PrimTy Michelson.PrimVal) m
+  ) =>
+  HR.Term Michelson.PrimTy Michelson.PrimVal ->
+  Usage.T ->
+  HR.Term Michelson.PrimTy Michelson.PrimVal ->
+  m (Either Michelson.CompErr (Michelson.Contract' Michelson.ExpandedOp, Michelson.SomeContract))
```
mariari, line 42: We could even add another alias to alias out the signautre since it's shared/duplicated

test/Pipeline.hs
```diff
@@ -0,0 +1,196 @@
...
+      HasSource "globals" (IR.Globals primTy primVal),
+      HasSink "globals" (IR.Globals primTy primVal)
+    )
+    via StateField "globals" (EnvExecAlias primTy primVal compErr)
+  deriving
+    (HasReader "globals" (IR.Globals primTy primVal))
+    via ReaderField "globals" (EnvExecAlias primTy primVal compErr)
+  deriving
+    (HasThrow "error" (Core.PipelineError primTy primVal compErr))
+    via MonadError (EnvExecAlias primTy primVal compErr)
```
mariari, line 56: We could probably add this to the src

test/Pipeline.hs
```diff
@@ -0,0 +1,196 @@
...
+shouldCompileToContract ::
+  String ->
+  (HR.Term PrimTy PrimVal, Usage.T, HR.Term PrimTy PrimVal) ->
+  Typed.Globals PrimTy PrimVal ->
+  Text ->
+  T.TestTree
+shouldCompileToContract name (term, usage, ty) globals contract =
+  T.testCase name $ do
+    res <- toMichelsonContract term usage ty globals
+    res T.@=? Right contract
```
mariari, line 90: Abstracting out common values and types would be nice

test/Pipeline.hs
```diff
@@ -0,0 +1,196 @@
...
+
+lamTerm :: HR.Term PrimTy PrimVal
+lamTerm = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App (HR.Prim AddI) (HR.Elim (HR.Var "x"))) (HR.Elim (HR.Var "y")))))
+
+appLam2 :: HR.Term PrimTy PrimVal
+appLam2 = HR.Elim (HR.App (HR.App (HR.Ann (Usage.SNat 1) lamTerm2 lamTy2 0) (HR.Elim (HR.Prim (Constant (M.ValueInt 2))))) (HR.Elim (HR.Prim (Constant (M.ValueInt 3)))))
+
+lamTerm2 :: HR.Term PrimTy PrimVal
+lamTerm2 = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App (HR.Prim AddI) (HR.Elim (HR.Var "x"))) (HR.Elim (HR.Prim (Constant (M.ValueInt 10)))))))
+
```
mariari, line 182: Would be nice if we could eventually get these to be smaller lines using `$` or `|>` to avoid extra lines with just ()'s

src/Juvix/Core/ErasedAnn/Conversion.hs
cwgoes: Sure, let's do this in a follow-up though - https://github.com/metastatedev/juvix/issues/436.

src/Juvix/Core/HR/Extend.hs
cwgoes: I don't quite follow, this is `HR`, it adds names for variables (instead of de Bruijin indices in IR)

src/Juvix/Core/HRAnn/Extend.hs
cwgoes: I don't quite follow, this is `HR`, it adds names for variables (instead of de Bruijin indices in IR)

src/Juvix/Core/HRAnn/Extend.hs
cwgoes: Same response as above

src/Juvix/Core/Pipeline.hs
cwgoes: Aliased

src/Juvix/Core/Pipeline.hs
cwgoes: Sure, although eventually this file shouldn't really be referencing particular backends, this is temporary

src/Juvix/Core/Pipeline.hs
cwgoes: Done

test/Pipeline.hs
cwgoes: I suppose, but mostly I think it makes sense to wait for quasiquotation and then write the tests that way?

test/Pipeline.hs
cwgoes: yeah, but first we should unify the stack at least somewhat - https://github.com/metastatedev/juvix/issues/159 

right now, this is only in `app`, which we should avoid `tests` being dependent upon

test/Pipeline.hs
cwgoes: Sure, done

PR #446, opened 2020-07-22
Removes Infix from the frontend Parser 

mariari:


src/Juvix/FrontendContextualise/InfixPrecedence/Transform.hs
```diff
@@ -0,0 +1,481 @@
+{-# LANGUAGE LiberalTypeSynonyms #-}
+
```
cwgoes, line 2: Can we add a top-level comment to this file with a pseudo-syntax description of what the transform does?

mariari: Sure I will give the changes in the format defined in FrontendDesugar

mariari: See the Extend.hs file for this

PR #447, opened 2020-07-23
Remove Module open

mariari:


src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -0,0 +1,96 @@
...
+  f (NonEmpty (Old.FunctionLike Old.Expression)) Old.Signature Old.Type
+
+type New f =
+  f (NonEmpty (New.FunctionLike New.Expression)) New.Signature New.Type
+
+type WorkingMaps m =
+  ( HasState "old" (Old Context.T) m,
+    HasState "new" (New Context.T) m,
+    HasThrow "error" Error m,
+    HasState "modMap" ModuleMap m
```
thealmarty, line 26: Do you think it's better to alias `HasState "modMap" ModuleMap m` separately so we can use it below?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -0,0 +1,96 @@
...
+qualifyName sym@(s :| _) = do
+  qualifieds <- get @"modMap"
+  case qualifieds Map.!? s of
+    Just preQualified ->
+      pure $ preQualified <> sym
+    Nothing ->
+      pure sym
+
+addModMap ::
+  HasState "modMap" ModuleMap m => Symbol -> NonEmpty Symbol -> m ()
```
thealmarty, line 85: Since we'll qualify `Env` do you think it's better to just name these `add`, `lookup` etc?

mariari: `add` and `lookup` are already in Env through the import.

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: Since we don't use this signature much, it's fine to stay how it is.

src/Juvix/FrontendContextualise/ModuleOpen/Transform.hs
```diff
@@ -0,0 +1,444 @@
...
+  HasState "modMap" Env.ModuleMap m => NonEmpty Symbol -> m (NonEmpty Symbol)
+transformSymbol = Env.qualifyName
+
+transformModuleOpenExpr ::
+  Env.WorkingMaps m => Old.ModuleOpenExpr -> m New.Expression
+transformModuleOpenExpr (Old.OpenExpress modName expr) = do
+  fullQualified <- Env.qualifyName modName
+  let reconstructed = reconstructSymbol fullQualified
+      err = throw @"error" (Env.UnknownModule reconstructed)
+  looked <- Env.lookup (reconstructSymbol fullQualified)
```
thealmarty, line 27: `reconstructed` can be used here in place of `reconstructSymbol fullQualified`?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
thealmarty: Oh right!

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
thealmarty: Alright!

PR #452, opened 2020-07-24
Add example contracts

thealmarty:
- Remove tasty quickcheck. 
- Add eg contract.


test/Frontend.hs
```diff
@@ -92,6 +95,11 @@ removeNoComment =
 -- Parse Many at once
 --------------------------------------------------------------------------------
 
+parsedIdString :: Either String [Juvix.Frontend.Types.TopLevel]
+parsedIdString =
+  Parser.parse
+    (encodeUtf32BE $ unsafePerformIO . readFile $ "/test/Id-Strings.jvx")
```
cwgoes, line 101: Hmm, we should find a way to write these tests that does not use `unsafePerformIO` - Tasty can do IO.

test/examples/Id-Strings.jvx
```diff
@@ -0,0 +1,17 @@
...
+
+let pair = "%pair"
+
+-- Identity contract for strings. 
+
+type storage = unit 
+
+-- Default entry point
+sig default : string -> storage -> storage
+￼let default para storage = Pair [] storage
```
cwgoes, line 13:
```suggestion
￼let default para storage = pair [] storage
```

do we have sugar for `[]` (cc @mariari)? this needs to translate to an empty Michelson list, if we use it like this

cwgoes:
It would be nice to combine this with https://github.com/metastatedev/juvix/pull/416, and we should end up with examples covering the functionality described in https://github.com/metastatedev/juvix/pull/416#issuecomment-648680217, e.g. in total:

- Identity contract
- Contract which swaps the two elements of a tuple (param, storage)
- Contract which uses addition (with integers, for example)
- Contract which matches on a sum type
- Contract which uses a nested match on a sum/product type

test/examples/Id-Strings.jvx
mariari: We don't yet, but I can add it in 30~ minutes if you want it

test/Main.hs
```diff
@@ -64,5 +67,18 @@ translationPasses =
     [ FrontendDesugar.allDesugar
     ]
 
+contractTests :: FilePath -> IO ()
+contractTests file = do
+  parsed <- readFile $ file
+  let parsedIdString = Parser.parse (encodeUtf32BE parsed)
+  case parsedIdString of
+    Left s -> putStrLn s
+    Right _t -> putStrLn $ "parsed " <> file
```
thealmarty, line 76: The output console doesn't print out either of these outputs atm.

thealmarty: Actually nvm. It does when I comment out the other tests and run `make test`. Both files parsed.

test/Main.hs
```diff
@@ -64,5 +67,26 @@ translationPasses =
     [ FrontendDesugar.allDesugar
     ]
 
+contractTests :: FilePath -> IO ()
+contractTests file = do
+  parsed <- readFile file
+  let parsedIdString = (parse (many1 topLevelSN) . removeComments) (encodeUtf32BE parsed)
```
thealmarty, line 73: I have reason to believe it isn't parsing anything atm. 

The most recent commit using `many1` (which applies the action p one or more times) and `parse` instead of `parseOnly` yields 'Fail', with an empty context `[]` and the result of `Left "string"`:

```
juvix> test (suite: juvix-test)
                   
Fail-- standard library functions

let string = "%prim.string"

let pair = "%Michelson.pair"

-- Identity contract for strings. 

type storage = unit 

-- Default entry point
sig default : string -> storage -> storage
��let default para storage = pair [] storage

-- Initialize storage to ().
sig init-storage : storage
let init-storage = ()
[]"string"
```

When `many` is used, it succeeds but it seems it's because it applies action p *zero* times. From the output, you can see the result is an empty list:
```
juvix> test (suite: juvix-test)
                   
Success[]
```

thealmarty:
Output to this commit is 
```
juvix> test (suite: juvix-test)
                   
partial
```
It's an improvement because it's not `Fail`?

thealmarty:
After expanding `Partial`, got this result:

```
juvix> test (suite: juvix-test)
                   
partial
Fail (after partial) []"not enough input"
```

thealmarty:
The test output of this commit is:
```
partial
Success (after partial) [Function' (Func' (Like' {functionLikedName = string, functionLikeArgs = [], functionLikeBody = Body' (Primitive' (Prim' (Michelson :| [string]) ()) ()) (), annLike = ()}) ()) ()]

juvix> Test suite juvix-test passed
``` 
however when parsing the file I still get:
```
juvix> test (suite: juvix-test)

Fail-- standard library functions

let string = %Michelson.string

let pair = %Michelson.pair

let unit = %Param.unit

-- Identity contract for strings. 
let storage = unit

-- Default entry point
sig default : string -> storage -> storage
��let default para storage = pair [] storage

-- Initialize storage to ().
sig init-storage : storage
let init-storage = unit
[]"string"

juvix> Test suite juvix-test passed
```

thealmarty:
See [this log](https://gist.github.com/thealmarty/67c86bced022e8fe02f60649436ebf40). It seems `readFile` or somewhere along the pipeline of the file being translate to `rawContract` messes up the `bytestring` somehow.

test/Main.hs
```diff
@@ -64,5 +66,34 @@ translationPasses =
     [ FrontendDesugar.allDesugar
     ]
 
+contractTests :: FilePath -> IO ()
```
cwgoes, line 69: We should turn this into a `Tasty` test?

thealmarty:
> It would be nice to combine this with #416, and we should end up with examples covering the functionality described in [#416 (comment)](https://github.com/metastatedev/juvix/pull/416#issuecomment-648680217), e.g. in total, for a first set of examples:
> 
> * Identity contract
> * Contract which swaps the two elements of a tuple (param, storage)
> * Contract which uses addition (with integers, for example)
> * Contract which defines a custom datatype for storage
> * Contract which matches on a sum type
> * Contract which uses a nested match on a sum/product type

Done for identity contract. Will be adding contracts similar to #416 and the rest of the list.

test/Frontend.hs
```diff
@@ -89,138 +92,55 @@ removeNoComment =
...
-        <> "  sig exec : Token.T -> Transaction.T -> Either.T Error Token.T \n"
-        <> "  let exec token tx = \n"
-        <> "    case tx.data of \n"
-        <> "    | Transfer _ -> \n"
-        <> "      if | Validation.transfer token tx = Right (transfer token tx) \n"
-        <> "         | else                         = Left NotEnoughFunds \n"
-        <> "    | Mint _ -> \n"
-        <> "      if | Validation.mint token tx = Right (mint token tx) \n"
-        <> "         | else                     = Left NotEnoughFunds \n"
-    )
```
mariari, line 223: This code is still needed

cwgoes: Can we add this as a file instead?

PR #463, opened 2020-08-04
Restored test and formatter

mariari:


mariari:
> Sorry the formatting changes didn't get in there.
> 
> As for the code I think it's best to put that code in a file and run it via the parse test. It's not a big deal to add it back though so feel free.

I will add it to a file later once I make some ingestion functions for ease of use in the codebase

PR #464, opened 2020-08-05
Improve parse file test and add addition contract example

thealmarty:
- Replaces #416's files.
- Abstract file parse tests. 
- Improve readability of test outputs.
- Add `Addition` example. It parses successfully.

### DIFF
--- a/test/examples/Addition.ju
+++ b/test/examples/Addition.ju
@@ -1,0 +1,22 @@
+let nat = %Micheslon.nat
+
+let add = %Michelson.add
+
+let pair = %Michelson.pair
+
+type Storage  = {
+  value  : nat,
+  valu : nat
+}
+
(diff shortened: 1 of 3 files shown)


test/examples/Addition.ju
```diff
@@ -0,0 +1,22 @@
+let nat = %Micheslon.nat
```
cwgoes, line 1:
```suggestion
let nat = %Michelson.nat
```

thealmarty: Huh. Nice catch Chris. @mariari any idea why the typo parsed?

thealmarty: OK I checked the test results. It parses the primitives name as they are without checking if they're valid. 

Later, we can start thinking about matching them at this parse stage and spit out error messages that suggest the user may want something that has similar spelling.

PR #470, opened 2020-08-07
Set up golden tests

thealmarty:
- set up golden tests using `Tasty.Silver`.
- add Id contract golden test.


test/examples/Id-Strings.ju.golden
```diff
@@ -0,0 +1 @@
...ure' (Sig' {signatureName = default, signatureUsage = Nothing, signatureArrowType = Infix' (Inf' {infixLeft = Name' (string :| []) (), infixOp = -> :| [], infixRight = Infix' (Inf' {infixLeft = Name' (storage :| []) (), infixOp = -> :| [], infixRight = Name' (storage :| []) (), annInf = ()}) (), annInf = ()}) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = default, functionLikeArgs = [ConcreteA' (MatchLogic' {matchLogicContents = MatchName' para (), matchLogicNamed = Nothing, annMatchLogic = ()}) (),ConcreteA' (MatchLogic' {matchLogicContents = MatchName' storage (), matchLogicNamed = Nothing, annMatchLogic = ()}) ()], functionLikeBody = Body' (Application' (App' {applicationName = Name' (pair :| []) (), applicationArgs = List' (ListLit' [] ()) () :| [Name' (storage :| []) ()], annApp = ()}) ()) (), annLike = ()}) ()) (),Signature' (Sig' {signatureName = init-storage, signatureUsage = Nothing, signatureArrowType = Name' (storage :| []) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = init-storage, functionLikeArgs = [], functionLikeBody = Body' (Name' (unit :| []) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Can we run a formatter on this, so it's not one big line?

Will that mess up how the golden tests work?

thealmarty:
> tACK, though can we reduce the output when running `make test`? I don't think we need to print the parse results when the test passes, only when it fails.

I'm thinking we can predominately use the golden test but when a contract doesn't pass we can use the stdout to debug because it prints out more output. So the stdout out would only have one file (`test.ju`).

test/examples/Id-Strings.ju.golden
thealmarty: I will do more testings and see if there is a way. I fear that if we change the golden file in anyway the tests would fail. But I'll see what I can do.

mariari: If it just parses it as a haskell string there is a way to trick it, probably the same thing I've done in the FrontendTests (adding \ to the end and start to signal haskell where to begin again

cwgoes:
We should probably only be committing one of the files (whatever one Golden keeps)?

thealmarty:
In this latest commit I use `readMaybe` instead of `read`, which gives a different error. But it's really just the same as the last commit's: `read` has an exception `Prelude.read: no parse`. 

I have verified that at the `readFileMaybe` step it *did* return a `Just ByteString`, so that step isn't the problem. Working on debugging the rest of the pipeline towards `read`.

thealmarty:
Verified in this commit that using derived `Read` and `Show` instances for `Usage` and `Symbol` doesn't fix the error. Probing further.

thealmarty:
Debugging in progress - will be testing more `TopLevel`s and figure out which one's instance may be causing the problem.

mariari:
> Debugging in progress - will be testing more `TopLevel`s and figure out which one's instance may be causing the problem.

you can try parsing expressions, then invoking the read instances back on it modulo the quotes.

thealmarty:
- Custom instances for `Usage` and `Symbol` replaced by derived instances.
- golden test now compares ASTs directly.
- pretty-show function to be finished in another PR - right now it doesn't make anything prettier.
- other frontend tests to be fixed in another PR. They will be changed to comparing the ASTs instead of their strings.

test/examples/Id-Strings.ju.golden
```diff
@@ -0,0 +1 @@
...x' (Inf' {infixLeft = Name' (Sym "string" :| []) (), infixOp = Sym "->" :| [], infixRight = Infix' (Inf' {infixLeft = Name' (Sym "storage" :| []) (), infixOp = Sym "->" :| [], infixRight = Name' (Sym "storage" :| []) (), annInf = ()}) (), annInf = ()}) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = Sym "default", functionLikeArgs = [ConcreteA' (MatchLogic' {matchLogicContents = MatchName' (Sym "para") (), matchLogicNamed = Nothing, annMatchLogic = ()}) (),ConcreteA' (MatchLogic' {matchLogicContents = MatchName' (Sym "storage") (), matchLogicNamed = Nothing, annMatchLogic = ()}) ()], functionLikeBody = Body' (Application' (App' {applicationName = Name' (Sym "pair" :| []) (), applicationArgs = List' (ListLit' [] ()) () :| [Name' (Sym "storage" :| []) ()], annApp = ()}) ()) (), annLike = ()}) ()) (),Signature' (Sig' {signatureName = Sym "init-storage", signatureUsage = Nothing, signatureArrowType = Name' (Sym "storage" :| []) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = Sym "init-storage", functionLikeArgs = [], functionLikeBody = Body' (Name' (Sym "unit" :| []) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Can you run the formatter on all these files

test/examples/Token.ju.golden
```diff
@@ -0,0 +1 @@
...), annCondExpression = ()}]) ()) (), annMatchL = ()} :| [MatchL' {matchLPattern = MatchLogic' {matchLogicContents = MatchCon' (Sym "Mint" :| []) [MatchLogic' {matchLogicContents = MatchName' (Sym "_") (), matchLogicNamed = Nothing, annMatchLogic = ()}] (), matchLogicNamed = Nothing, annMatchLogic = ()}, matchLBody = Cond' (C' (CondExpression' {condLogicPred = Application' (App' {applicationName = Name' (Sym "Validation" :| [Sym "mint"]) (), applicationArgs = Name' (Sym "token" :| []) () :| [Name' (Sym "tx" :| []) ()], annApp = ()}) (), condLogicBody = Application' (App' {applicationName = Name' (Sym "Right" :| []) (), applicationArgs = Parened' (Application' (App' {applicationName = Name' (Sym "mint" :| []) (), applicationArgs = Name' (Sym "token" :| []) () :| [Name' (Sym "tx" :| []) ()], annApp = ()}) ()) () :| [], annApp = ()}) (), annCondExpression = ()} :| [CondExpression' {condLogicPred = Name' (Sym "else" :| []) (), condLogicBody = Application' (App' {applicationName = Name' (Sym "Left" :| []) (), applicationArgs = Name' (Sym "NotEnoughFunds" :| []) () :| [], annApp = ()}) (), annCondExpression = ()}]) ()) (), annMatchL = ()}], annMatch'' = ()}) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Ditto

test/examples/Addition.ju.golden
```diff
@@ -0,0 +1 @@
..."default", signatureUsage = Nothing, signatureArrowType = Infix' (Inf' {infixLeft = Name' (Sym "nat" :| []) (), infixOp = Sym "->" :| [], infixRight = Infix' (Inf' {infixLeft = Name' (Sym "storage" :| []) (), infixOp = Sym "->" :| [], infixRight = Name' (Sym "storage" :| []) (), annInf = ()}) (), annInf = ()}) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = Sym "default", functionLikeArgs = [ConcreteA' (MatchLogic' {matchLogicContents = MatchName' (Sym "addnat") (), matchLogicNamed = Nothing, annMatchLogic = ()}) (),ConcreteA' (MatchLogic' {matchLogicContents = MatchName' (Sym "storage") (), matchLogicNamed = Nothing, annMatchLogic = ()}) ()], functionLikeBody = Body' (Application' (App' {applicationName = Name' (Sym "pair" :| []) (), applicationArgs = Parened' (Application' (App' {applicationName = Name' (Sym "add-storage" :| []) (), applicationArgs = Name' (Sym "storage" :| []) () :| [], annApp = ()}) ()) () :| [Parened' (Application' (App' {applicationName = Name' (Sym "foo" :| []) (), applicationArgs = Name' (Sym "input" :| []) () :| [Name' (Sym "storage" :| []) ()], annApp = ()}) ()) ()], annApp = ()}) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Ditto

test/Main.hs
```diff
@@ -46,13 +46,13 @@ allCheckedTests =
...
-      EAC2.eac2Tests,
-      Erasure.erasureTests,
-      Shunt.allInfixTests
+      -- pipelineTests,
+      -- backendTests,
+      frontEndTests
+      -- translationPasses,
+      -- EAC2.eac2Tests,
+      -- Erasure.erasureTests,
+      -- Shunt.allInfixTests
```
mariari, line 55: We need all our tests back

test/Frontend.hs
```diff
@@ -2,38 +2,43 @@ module Frontend where
...
+      -- matchMoreComplex,
+      -- condTest1,
+      -- record1,
+      -- parens1,
+      -- -- pre-processor tests
+      -- removeNoComment,
+      -- removeNewLineBefore,
+      -- removeSpaceBefore
+      contractFiles
     ]
```
mariari, line 42: we need all these back as well

src/Juvix/Library.hs
```diff
@@ -79,6 +79,7 @@ import Protolude hiding
     reader,
     state,
   )
+import qualified Text.Read as R
```
mariari, line 82: This seems like an unused import?

src/Juvix/Core/Usage.hs
```diff
@@ -13,6 +13,7 @@ module Juvix.Core.Usage
 where
 
 import Juvix.Library hiding (pred, show)
+import qualified Text.Read as R
```
mariari, line 16: Same here

test/examples/Id-Strings.ju.golden
thealmarty: As mentioned in the comment I'll do the pretty show in another PR.

test/Main.hs
thealmarty: I will put them back once I update the test to compare the ASTs instead of strings.

src/Juvix/Library.hs
thealmarty: Yes, will remove them.

test/examples/Id-Strings.ju.golden
mariari: Well would still be nice since this pr uses the read instance, and since the originals were split up per line for readability reasons, we can do even better formatting with the read instance here!

cwgoes:
Looks like `make test` is still failing?

thealmarty:
> Looks like `make test` is still failing?

Yeah still debugging. It can't find the golden files now. :( Should get it solved by end of today I hope. Another batch of tests will be added back in too.

thealmarty:
Tests updated and cleaned up. R4R. Sorry about the large PR and diffs - ignore changes that are not related to tests.

test/examples/Id-Strings.ju.golden
thealmarty: I'll format the golden files in another PR. Just really want this to merge first. All tests pass now.

test/Frontend/Parser.hs
```diff
@@ -91,209 +114,15 @@ removeNoComment =
...
 --------------------------------------------------------------------------------
 -- Parse Many at once
 --------------------------------------------------------------------------------
 
-contractTest =
-  parseOnly
-    (many Parser.topLevelSN)
-    ( ""
-        <> "mod Token = "
-        <> "  let Address = s : String.T {String.length s == 36} \n"
```
mariari, line 167: Where did this get moved to again?

test/Frontend/Parser.hs
```diff
@@ -308,43 +137,105 @@ many1FunctionsParser =
...
+            )
+            ()
+        )
+        (),
+      Function'
+        ( Func'
+            ( Like'
+                { functionLikedName = "bah",
+                  functionLikeArgs = [],
+                  functionLikeBody = Body' (Application' (App' {applicationName = Name' ("foo" :| []) (), applicationArgs = Constant' (Number' (Integer'' 1 ()) ()) () :| [Constant' (Number' (Integer'' 2 ()) ()) (), Constant' (Number' (Integer'' 3 ()) ()) ()], annApp = ()}) ()) (),
```
mariari, line 207: This needs to be newlined... it's 281

test/Frontend/Parser.hs
```diff
@@ -308,43 +137,105 @@ many1FunctionsParser =
...
+            )
+            ()
+        )
+        (),
+      Function'
+        ( Func'
+            ( Like'
+                { functionLikedName = "nah",
+                  functionLikeArgs = [],
+                  functionLikeBody = Guard' (C' (CondExpression' {condLogicPred = Infix' (Inf' {infixLeft = Name' ("bah" :| []) (), infixOp = "==" :| [], infixRight = Constant' (Number' (Integer'' 5 ()) ()) (), annInf = ()}) (), condLogicBody = Constant' (Number' (Integer'' 7 ()) ()) (), annCondExpression = ()} :| [CondExpression' {condLogicPred = Name' ("else" :| []) (), condLogicBody = Constant' (Number' (Integer'' 11 ()) ()) (), annCondExpression = ()}]) ()) (),
```
mariari, line 219: Same here

test/Frontend/Parser.hs
```diff
@@ -308,43 +137,105 @@ many1FunctionsParser =
...
...nctionLikeArgs = [], functionLikeBody = Body' (Name' ("nah" :| []) ()) (), annLike = ()}, letBody = Match' (Match''' {matchOn = Name' ("check" :| []) (), matchBindigns = MatchL' {matchLPattern = MatchLogic' {matchLogicContents = MatchName' "seven" (), matchLogicNamed = Nothing, annMatchLogic = ()}, matchLBody = Constant' (Number' (Integer'' 11 ()) ()) (), annMatchL = ()} :| [MatchL' {matchLPattern = MatchLogic' {matchLogicContents = MatchName' "eleven" (), matchLogicNamed = Nothing, annMatchLogic = ()}, matchLBody = Constant' (Number' (Integer'' 7 ()) ()) (), annMatchL = ()}, MatchL' {matchLPattern = MatchLogic' {matchLogicContents = MatchName' "f" (), matchLogicNamed = Nothing, annMatchLogic = ()}, matchLBody = OpenExpr' (OpenExpress' {moduleOpenExprModuleN = "Fails" :| [], moduleOpenExprExpr = Do' (Do''' (DoBody' {doBodyName = Nothing, doBodyExpr = Application' (App' {applicationName = Name' ("print" :| []) (), applicationArgs = Name' ("failed" :| []) () :| [], annApp = ()}) (), annDoBody = ()} :| [DoBody' {doBodyName = Nothing, doBodyExpr = Name' ("fail" :| []) (), annDoBody = ()}]) ()) (), annOpenExpress = ()}) (), annMatchL = ()}], annMatch'' = ()}) (), annLet'' = ()}) ()) (),
```
mariari, line 231: Same here

test/Frontend/Parser.hs
```diff
@@ -424,32 +460,168 @@ sumTypeTest :: T.TestTree
...
+                               ]
+                        )
+                        (),
+                    annNonArrowed = ()
+                  },
+              annTyp = ()
+            }
+        )
+        ()
+    ]
```
mariari, line 624: I'll probably just conver this myself later removing so many of the lines

test/examples/Addition.ju.golden
```diff
@@ -0,0 +1 @@
...(), annLike = ()}) ()) (),Signature' (Sig' {signatureName = "default", signatureUsage = Nothing, signatureArrowType = Infix' (Inf' {infixLeft = Name' ("nat" :| []) (), infixOp = "->" :| [], infixRight = Infix' (Inf' {infixLeft = Name' ("storage" :| []) (), infixOp = "->" :| [], infixRight = Name' ("storage" :| []) (), annInf = ()}) (), annInf = ()}) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = "default", functionLikeArgs = [ConcreteA' (MatchLogic' {matchLogicContents = MatchName' "addnat" (), matchLogicNamed = Nothing, annMatchLogic = ()}) (),ConcreteA' (MatchLogic' {matchLogicContents = MatchName' "storage" (), matchLogicNamed = Nothing, annMatchLogic = ()}) ()], functionLikeBody = Body' (Application' (App' {applicationName = Name' ("pair" :| []) (), applicationArgs = Parened' (Application' (App' {applicationName = Name' ("add-storage" :| []) (), applicationArgs = Name' ("storage" :| []) () :| [], annApp = ()}) ()) () :| [Parened' (Application' (App' {applicationName = Name' ("foo" :| []) (), applicationArgs = Name' ("input" :| []) () :| [Name' ("storage" :| []) ()], annApp = ()}) ()) ()], annApp = ()}) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: This is still 1 line!

test/examples/Id-Strings.ju.golden
```diff
@@ -0,0 +1 @@
...efault", signatureUsage = Nothing, signatureArrowType = Infix' (Inf' {infixLeft = Name' ("string" :| []) (), infixOp = "->" :| [], infixRight = Infix' (Inf' {infixLeft = Name' ("storage" :| []) (), infixOp = "->" :| [], infixRight = Name' ("storage" :| []) (), annInf = ()}) (), annInf = ()}) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = "default", functionLikeArgs = [ConcreteA' (MatchLogic' {matchLogicContents = MatchName' "para" (), matchLogicNamed = Nothing, annMatchLogic = ()}) (),ConcreteA' (MatchLogic' {matchLogicContents = MatchName' "storage" (), matchLogicNamed = Nothing, annMatchLogic = ()}) ()], functionLikeBody = Body' (Application' (App' {applicationName = Name' ("pair" :| []) (), applicationArgs = List' (ListLit' [] ()) () :| [Name' ("storage" :| []) ()], annApp = ()}) ()) (), annLike = ()}) ()) (),Signature' (Sig' {signatureName = "init-storage", signatureUsage = Nothing, signatureArrowType = Name' ("storage" :| []) (), signatureConstraints = [], annSig = ()}) (),Function' (Func' (Like' {functionLikedName = "init-storage", functionLikeArgs = [], functionLikeBody = Body' (Name' ("unit" :| []) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Same here!

test/examples/Token.ju.golden
```diff
@@ -0,0 +1 @@
...me' ("NotEnoughFunds" :| []) () :| [], annApp = ()}) (), annCondExpression = ()}]) ()) (), annMatchL = ()} :| [MatchL' {matchLPattern = MatchLogic' {matchLogicContents = MatchCon' ("Mint" :| []) [MatchLogic' {matchLogicContents = MatchName' "_" (), matchLogicNamed = Nothing, annMatchLogic = ()}] (), matchLogicNamed = Nothing, annMatchLogic = ()}, matchLBody = Cond' (C' (CondExpression' {condLogicPred = Application' (App' {applicationName = Name' ("Validation" :| ["mint"]) (), applicationArgs = Name' ("token" :| []) () :| [Name' ("tx" :| []) ()], annApp = ()}) (), condLogicBody = Application' (App' {applicationName = Name' ("Right" :| []) (), applicationArgs = Parened' (Application' (App' {applicationName = Name' ("mint" :| []) (), applicationArgs = Name' ("token" :| []) () :| [Name' ("tx" :| []) ()], annApp = ()}) ()) () :| [], annApp = ()}) (), annCondExpression = ()} :| [CondExpression' {condLogicPred = Name' ("else" :| []) (), condLogicBody = Application' (App' {applicationName = Name' ("Left" :| []) (), applicationArgs = Name' ("NotEnoughFunds" :| []) () :| [], annApp = ()}) (), annCondExpression = ()}]) ()) (), annMatchL = ()}], annMatch'' = ()}) ()) (), annLike = ()}) ()) ()]
```
mariari, line 1: Same here!

test/Main.hs
```diff
@@ -1,49 +1,59 @@
...
+import Core.Parser (coreParser)
+import Core.Typechecker (coreCheckerEval)
+import Frontend.Desugar (allDesugar)
+import Frontend.Golden (contractFiles)
+import Frontend.Parser (allParserTests)
+import FrontendContextualise.Infix.ShuntYard (allInfixTests)
+import FrontendContextualise.Module.Open (openTests)
+import Juvix.Library (IO)
+import Pipeline (tests)
 import qualified Test.Tasty as T
```
mariari, line 20: Can we change all these back?

I really don't want unqualified imports, it is poor that we have redundant names, but we should change the main test name not the qualified itself

test/Frontend/Desugar.hs
```diff
@@ -1,59 +1,101 @@
...
-        (fmap show x T.@=? Right y)
-    ]
-
-shouldDesugar :: T.TestName -> ByteString -> String -> T.TestTree
-shouldDesugar name parseString =
-  desugarTasty name (fmap desugar (parseOnly (many Parser.topLevelSN) parseString))
+shouldDesugar ::
+  T.TestName -> ByteString -> [TopLevel] -> T.TestTree
+shouldDesugar name =
+  shouldParseAs name (parse (many Parser.topLevelSN))
```
mariari, line 47: I'm not sure the desugar test should do this...  I don't see where we desugar

test/Frontend/Desugar.hs
```diff
@@ -1,59 +1,101 @@
...
-    \= ()}], annMatch'' = ()}) (), annMatchL = ()}], annMatch'' = ()}) \
-    \())} :| [],Nothing)) ()]"
+    [ Function'
+        ( Func'
+            ( Like'
+                { functionLikedName = Sym "foo",
+                  functionLikeArgs = [],
+                  functionLikeBody =
+                    Guard'
+                      ( C'
```
mariari, line 61: I don't think we desugar at all here!

test/Frontend/Desugar.hs
```diff
@@ -1,59 +1,101 @@
...
+    Infix' (Inf', annInf, infixLeft, infixOp, infixRight),
+    Numb' (Integer''),
+    TopLevel' (Function'),
+  )
+import Juvix.Library
+  ( Alternative (many),
+    ByteString,
+    NonEmpty ((:|)),
+    Symbol (Sym),
+  )
```
mariari, line 35: I'd much rather keep the qualified imports and deal with the verbosity by using |>'s

mariari (changes requested):
I will approve when

1. Once the *.golden files have been ran through our formatter
2. the desugar tests actually desugar again!
3. the qualified imports for the main are back

I will handle the formatting for the tests to remove the redundant ()'s via |> after this merges

test/Frontend/Parser.hs
mariari: I'll fix this up myself after the pr merges

test/Frontend/Parser.hs
mariari: I'll fix this up later as well

test/Frontend/Desugar.hs
cwgoes: We should be calling `desugar` somewhere, right?

test/Frontend/Golden.hs
```diff
@@ -0,0 +1,130 @@
...
+goldenTest :: T.TestName -> FilePath -> T.TestTree
+goldenTest name file =
+  let goldenFileName = file <> ".golden"
+   in T.goldenTest1
+        name
+        (getGolden goldenFileName)
+        (parsedContract file)
+        compareParsedGolden
+        -- show the golden/actual value, not working atm
+        ( T.ShowText . Text.pack
```
cwgoes, line 115: I don't quite understand, what isn't working here / why?

cwgoes (commented):
Let's ensure that the tests are still testing the right things, I'm not so concerned about the `.golden` file format as long as it is consistent.

test/Frontend/Golden.hs
thealmarty: When a test fails and shows the golden/actual value, it doesn't use this function in the print out. It also doesn't use the next input to update the golden file.

cwgoes: Hmm, that's odd, sounds like either a library bug or we're using the library incorrectly somehow?

test/Frontend/Desugar.hs
thealmarty: Oops corrected.

test/Frontend/Golden.hs
thealmarty: Yeah I don't know if I misunderstood the function or what. There's isn't a lot of documentation about this that I can find.

thealmarty:
- Desugar test actually desugars now. 
- Per meeting more formatting and golden file formats will be fixed by @mariari . 
- Golden test function to be fixed later. If we can't find how to use those "show" and "update golden file" functions, I can change it back to the old test which doesn't take those inputs.

mariari (approved):

cwgoes:
Alright. We should check why the `test_suite` is still failing though.

mariari:
> Alright. We should check why the `test_suite` is still failing though.

agreed

thealmarty:
> > Alright. We should check why the `test_suite` is still failing though.
> 
> agreed

Seems to have timed out. It succeeded in my local run. I re-ran it, some timed out error below:
image

PR #473, opened 2020-08-10
Add `stdlib/Michelson`

### LINKED ISSUE #468: Add Michelson primitives file
`stdlib/Michelson.ju` - should expose all the primitive types from [here](https://hackage.haskell.org/package/morley-1.5.0/docs/Michelson-Untyped-Type.html) and functions from [here](https://hackage.haskell.org/package/morley-1.5.0/docs/Michelson-Untyped-Instr.html) (though not all instructions are functions) using the `%prim` syntax. Then we can import this file in contracts which compile to the Michelson backend.

It would be nice if Juvix can automatically expose the stdlib (whenever `juvix` is invoked), but this can happen later.

Per @mariari best to import qualified for now.

cyberglot:
Closes #468.

### DIFF
--- a/stdlib/Michelson.ju
+++ b/stdlib/Michelson.ju
@@ -1,140 +1,143 @@
-mod Michelson =
+  -- datatypes
+  let option    = %Michelson.option
+  let list      = %Michelson.list
+  let set       = %Michelson.set
+  let operation = %Michelson.operation
+  let contract  = %Michelson.contract
+  let pair      = %Michelson.pair
+  let either    = %Michelson.either
+  let lambda    = %Michelson.lambda
+  let map       = %Michelson.map
+  let big-map   = %Michelson.big-map
 
-    -- datalets
-    let option = %Michelson.option
-    let list = %Michelson.list
-    let set = %Michelson.set
-    let operation = %Michelson.operation
-    let contract = %Michelson.contract
-    let pair = %Michelson.pair
-    let either = %Michelson.either
-    let lambda = %Michelson.lambda
-    let map = %Michelson.map
-    let big-map = %Michelson.big-map
+  -- primitive types
+  let key       = %Michelson.key
+  let unit      = %Michelson.unit
+  let signature = %Michelson.signature
+  let chain-id  = %Michelson.chain-id
+  let int       = %Michelson.int
+  let nat       = %Michelson.nat
+  let string    = %Michelson.string
+  let bytes     = %Michelson.bytes
+  let mutez     = %Michelson.mutez
+  let bool      = %Michelson.bool
+  let hash-key  = %Michelson.hash-key
+  let timestamp = %Michelson.timestamp
+  let address   = %Michelson.address
 
-    -- primitive lets
-    let key = %Michelson.key
-    let unit = %Michelson.unit
-    let signature = %Michelson.signature
-    let chain-id = %Michelson.chain-id
-    let int = %Michelson.int
-    let nat = %Michelson.nat
-    let string = %Michelson.string
-    let bytes = %Michelson.bytes
-    let mutez = %Michelson.mutez
-    let bool = %Michelson.bool
-    let hash-key = %Michelson.hash-key
-    let timestamp = %Michelson.timestamp
-    let address = %Michelson.address
+  -- arithmetic functions
+  sig add : (x: type) -> x -> x -> x
+  let add = %Michelson.add
 
-    -- arithmetic functions
-    sig add : (x: type) -> x -> x -> x
-    let add = %Michelson.add
+  sig mul : (x: type) -> x -> x -> x
+  let mul = %Michelson.mul
 
-    sig mul : (x: type) -> x -> x -> x
-    let mul = %Michelson.mul
+  sig sub : (x: type) -> x -> x -> x
+  let sub = %Michelson.sub
 
-    sig sub : (x: type) -> x -> x -> x
-    let sub = %Michelson.sub
+  sig div : (x: type) -> x -> x -> x
+  let div = %Michelson.div
 
-    sig div : (x: type) -> x -> x -> x
-    let div = %Michelson.div
+  sig abs : (x: type) -> x -> x -> x
+  let abs = %Michelson.abs
 
-    sig abs : (x: type) -> x -> x -> x
-    let abs = %Michelson.abs
+  -- boolean functions
+  sig and' : (x: type) -> x -> x -> x
+  let and' = %Michelson.and
 
-    -- boolean functions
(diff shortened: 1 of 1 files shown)


cyberglot:
Some types are not clear to me but I will go back to them after I'm done with the whole list of functions.

cwgoes:
We need to add all the arithmetic operations, which should be dependently-typed (eventually we'll use typeclasses).

cyberglot:
`add : x : Type -> x -> x -> x` like this or?

cwgoes:
> `add : x : Type -> x -> x -> x` like this or?

Yes, essentially, later on it will be restricted to typeclasses.

stdlib/Michelson.ju
```diff
@@ -0,0 +1,58 @@
...
+    type hash-key	 
+    type timestamp	 
+    type address
+
+    -- instructions
+    -- not clear what the types are from documentation (more research needed)
+    -- `words` are typed as `bytes`
+    -- otherwise `a` 
+    sig ext : a -- ???
+    sig dropn : int -> bytes -- return type???
```
cwgoes, line 35: We won't have these operations (anything which directly manipulates the stack)

cwgoes: (there's no stack in the Juvix semantics, we just compile the lambda calculus to it for Michelson)

stdlib/Michelson.ju
```diff
@@ -0,0 +1,140 @@
...
+    let sub = %Michelson.sub
+
+    sig div : (x: type) -> x -> x -> x
+    let div = %Michelson.div
+
+    sig abs : (x: type) -> x -> x -> x
+    let abs = %Michelson.abs
+
+    -- boolean functions
+    sig and : (x: type) -> x -> x -> x
```
mariari, line 47: And will be reserved

stdlib/Michelson.ju
```diff
@@ -0,0 +1,140 @@
...
+    let chain-id = %Michelson.chain-id
+    let int = %Michelson.int
+    let nat = %Michelson.nat
+    let string = %Michelson.string
+    let bytes = %Michelson.bytes
+    let mutez = %Michelson.mutez
+    let bool = %Michelson.bool
+    let hash-key = %Michelson.hash-key
+    let timestamp = %Michelson.timestamp
+    let address = %Michelson.address
```
mariari, line 28: If we have a bunch of these in a row like this, can we align the `=` ?

stdlib/Michelson.ju
```diff
@@ -0,0 +1,140 @@
+mod Michelson =
```
mariari, line 1: We don't need this mod at the top...

Though we will once we have a top level thing I supsect it'll look like

```haskelll
module Michelson where
...
```

Though for now, I'll just use the filename and capitalize the first

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  sig amount : (x : let) -> x -> ()
+  let amount = %Michelson.amount
+
+  sig balance : (x : let) -> address -> x
+  let balance = %Michelson.balance
+
+  sig sha256 : (x: type) -> x -> hash-key
+  let sha256 = %Michelson.sha256
+
+  sig sha512 :
```
cwgoes, line 73: This and subsequent hash functions are `bytes -> bytes`

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  sig source : address
+  let source = %Michelson.source
+
+  sig sender : address
+  let sender = %Michelson.sender
+
+  sig implicit-account : address
+  let implicit-account = %Michelson.implicit-account
+
+  sig create-contract : () -> signature
```
cwgoes, line 88: This & transfer-tokens create an operation

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  sig empty-set : set a
+  let empty-set = %Michelson.empty-set
+
+  sig empty-map : map a
+  let empty-map = %Michelson.empty-map
+
+  sig empty-big-map : big-map a
+  let empty-big-map = %Michelson.empty-big-map
+
+  sig map : (x: type) -> lambda a b -> x a -> x b
```
cwgoes, line 133: The compiler could do the `lambda` wrapping automatically but maybe explicit is better (like you have here)

mariari: ahh I missed thing, `(a -> b)` works for lambda

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  let balance = %Michelson.balance
+
+  sig sha256 : (x: type) -> x -> hash-key
+  let sha256 = %Michelson.sha256
+
+  sig sha512 :
+  let sha512 = %Michelson.sha512
+
+  sig blake2b :
+  let blake2b = %Michelson.
```
mariari, line 77: This neeeds to be finished

stdlib/Michelson.ju
mariari, line 88: seems like we bind unit, probably best to use that for now

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  sig transfer-tokens : (x: let) -> address -> x -> hash-key
+  let transfer-token = %Michelson.transfer-tokens
+
+  sig set-delegate : address -> ()
+  let set-delegate = %Michelson.set-delegate
+
+  sig none : either () a
+  let none = %Michelson.none
+
+  sig some : b -> either a b
```
mariari, line 100: Same here

mariari: probably best to use b for the second argument of either (doesn't matter, just a nit!)

stdlib/Michelson.ju
```diff
@@ -0,0 +1,143 @@
...
+  let empty-big-map = %Michelson.empty-big-map
+
+  sig map : (x: type) -> lambda a b -> x a -> x b
+  let map = %Michelson.map
+
+  sig update : (x: type) -> key -> a -> x a -> ()
+  let update = %Michelson.update
+
+  sig iter : (x: type) -> lambda a b -> lambda b ()
+  let iter = %Michelson.iter
```
mariari, line 140: same here

mariari:
One last change, we can use normal lambdas for michelson, so you can convert lambda a b to (a -> b)

cyberglot:
> One last change, we can use normal lambdas for michelson, so you can convert lambda a b to (a -> b)

will do.

cwgoes (approved):
ACK, we'll need to write lots of tests

PR #478, opened 2020-08-17
Initial substitution algorithm

mariari:


src/Juvix/Core/IR/Subst.hs
```diff
@@ -0,0 +1,105 @@
...
+data T primTy primVal
+  = T
+      { _depth :: Natural,
+        _sub :: Map.T Key (Types.Elim primTy primVal)
+      }
+  deriving (Show)
+
+makeLenses ''T
+
+-- useless with debrunj indincies ☹
```
cwgoes, line 33:
```suggestion
-- useless with de bruijn indices ☹
```

src/Juvix/Core/IR/Subst.hs
```diff
@@ -0,0 +1,105 @@
...
+-- - take an environment of some kind
+--   + This env should have meta data at some point to better make
+--     decisions. Or better yet tag it to the structure itself in some
+--     extension.
+-- - Make this run on some type of size and usage measurement.
+--   + The current metric just checks if the usage is one. However
+--     this is bad, as we could move the evaluation in the middle of
+--     some computation that degrades performance.
+
+inlineP :: Usage.T -> Types.Elim primTy primVal -> Bool
```
cwgoes, line 99: Add comment, is this a function used to decide if we will inline something?

src/Juvix/Core/IR/Subst.hs
```diff
@@ -0,0 +1,105 @@
...
+instance Hashable Key
+
+data T primTy primVal
+  = T
+      { _depth :: Natural,
+        _sub :: Map.T Key (Types.Elim primTy primVal)
+      }
+  deriving (Show)
+
+makeLenses ''T
```
cwgoes, line 31: alright... are they really worth it? relatively simple transformations

(I don't mind lenses but we only use them sometimes, it's a bit inconsistent)

src/Juvix/Core/IR/Subst.hs
```diff
@@ -0,0 +1,105 @@
...
+f _ (Types.PrimTy ty) =
+  Types.PrimTy ty
+f _ (Types.Star uni) =
+  Types.Star uni
+f subst (Types.Pi usage typ body) =
+  Types.Pi usage (f subst typ) (f (over depth succ subst) body)
+f subst (Types.Lam body) =
+  Types.Lam (f (over depth succ subst) body)
+f subst (Types.Let usage bound body)
+  | inlineP usage bound =
```
cwgoes, line 52: honestly I find it odd that this function makes inlining decisions, won't that result in a fair bit of duplicate work (checking `inlineP` at least) if we substitute several times over the same term?

cwgoes (commented):
Thanks; a few questions.

src/Juvix/Core/IR/Subst.hs
mariari: I'm using it for over, which is useful, nicer than using the record updating syntax for a single field.

src/Juvix/Core/IR/Subst.hs
mariari: Ideally we'll have annotation extension to this, should it should be better. The GHC inliner does something very similar, as it checks things as it goes.

src/Juvix/Core/IR/Subst.hs
mariari: done

cwgoes (approved):
ACK

PR #480, opened 2020-08-17
[String] → Frontend

mariari:
This along with Frontend → Core, will give us the pipeline

### DIFF
--- a/src/Juvix/Frontend.hs
+++ b/src/Juvix/Frontend.hs
@@ -1,0 +1,30 @@
+module Juvix.Frontend where
+
+import qualified Control.Arrow as Arrow
+import qualified Data.ByteString as ByteString
+import qualified Data.Char as Char
+import qualified Juvix.Core.Common.NameSymbol as NameSymbol
+import qualified Juvix.Frontend.Parser as Parser
+import qualified Juvix.Frontend.Types as Types
+import Juvix.Library
+import qualified System.FilePath as FilePath
+import Prelude (String)
+
+toFrontEnd :: [FilePath] -> IO (Either String [(NameSymbol.T, [Types.TopLevel])])
+toFrontEnd files = do
+  read <- ByteString.readFile `traverse` files
+  case traverse Parser.parseOnly read of
+    Left x ->
+      pure (Left x)
+    Right xs ->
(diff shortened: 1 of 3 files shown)


src/Juvix/Pipeline.hs
```diff
@@ -0,0 +1,24 @@
...
+
+toCore :: [FilePath] -> IO (Either Error Target.FinalContext)
+toCore paths = do
+  x <- Frontend.ofPath paths
+  case x of
+    Left er -> pure (Left (ParseErr er))
+    Right x ->
+      case Core.ofFrontend x of
+        Left errr -> pure (Left (PipeLine errr))
+        Right con -> pure (Right con)
```
cwgoes, line 24:
```suggestion
      pure $ case Core.ofFrontend x of
        Left errr -> Left (PipeLine errr)
        Right con -> Right con
```

src/Juvix/Frontend.hs
```diff
@@ -0,0 +1,30 @@
...
+    Right xs ->
+      zip files xs
+        |> fmap
+          ( Arrow.first
+              (NameSymbol.fromSymbol . intern . toUpper . FilePath.takeBaseName)
+          )
+        |> Right
+        |> pure
+
+toUpper :: String -> String
```
cwgoes, line 28: `toUpper = map Char.toUpper` ?

cwgoes (approved):
ACK, two nits

src/Juvix/Frontend.hs
mariari: No, it only does it on the first element, I don't want `FOO` I want `Foo`... this is namely for NTFS reasons...


Over time, We'll just have module Foo where.... and do this technique when it isn't there

src/Juvix/Pipeline.hs
mariari: Will do, but first I'll see how our formatter handles this

mariari: Works fine, I moved it even higher though

src/Juvix/Frontend.hs
cwgoes: oh oops, you're right

PR #483, opened 2020-08-19
Adds IDE info to the build package

mariari:
this will only effect the `./stack-build` directory and not cloud the programming directory with .hie files, unless one is running emacs.

Emacs seems to leverage the local ones(?) quite well, and I notice a significant CPU load decrease when these files are up to date.

Just good to have this flag

Is there any other places this needs to be added, or does the options here effect all (executable, test, benchmarks) options?


cwgoes (approved):
ACK

PR #486, opened 2020-08-24
Proper and implicit opens

mariari:


src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +149,126 @@ lookupModMap s =
 removeModMap ::
   HasState "modMap" ModuleMap m => Symbol -> m ()
 removeModMap s = Juvix.Library.modify @"modMap" (Map.delete s)
+
+--------------------------------------------------------------------------------
+-- fully resolve module opens
```
mariari, line 154: Please read this section carefully.

If there is any way to streamline the logic do let me know

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+-- any place where this is inconsistent will break resolution
+
+-- | @populateModMap@ populates the modMap with all the global opens
+-- in the module
+populateModMap ::
+  WorkingMaps m => m ()
+populateModMap = do
+  old <- get @"old"
+  new <- get @"new"
+  open <- Juvix.Library.ask @"openMap"
```
cwgoes, line 178: Why is `ask` not in scope unqualified? Import it?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
 removeModMap ::
   HasState "modMap" ModuleMap m => Symbol -> m ()
 removeModMap s = Juvix.Library.modify @"modMap" (Map.delete s)
+
+--------------------------------------------------------------------------------
+-- fully resolve module opens
+--------------------------------------------------------------------------------
+
+-- Precondition ∷ old and new context must be in the same context
+-- also terms must be in either the old map or the new map
```
cwgoes, line 168: but not both? or every term must be in one or the other?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+    Just opens ->
+      -- TODO ∷
+      -- explicts and implicits for now work the same
+      put @"modMap" (foldr f modM opens)
+      where
+        f y modMap = foldr addNewSymbol modMap openList
+          where
+            nameSpace =
+              case y of
+                Implicit x -> x
```
cwgoes, line 193: could use a helper function here

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+                Explicit x -> x
+            unpopulate = (() <$)
+            --
+            addNewSymbol x modMap =
+              case unpopulate (Context.lookup (NameSymbol.fromSymbol x) old)
```
cwgoes, line 198: can deduplicate a bit of code here

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+  open <- Juvix.Library.ask @"openMap"
+  modM <- get @"modMap"
+  let curr = pure Context.topLevelName <> Context.currentName old
+  case open Map.!? curr of
+    Nothing ->
+      pure ()
+    Just opens ->
+      -- TODO ∷
+      -- explicts and implicits for now work the same
+      put @"modMap" (foldr f modM opens)
```
cwgoes, line 187: you can just use `modify` instead of `get` and `put`

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+
+-- we don't take a module mapping as all symbols at the point of
+-- resolve can't be anything else, thus we don't have to pre-pend any
+-- qualification at this point... we add qualification later as we add
+-- things, as we have to figure out the full resolution of all opens
+
+resolve :: Context.T a b c -> [PreQualified] -> Either Error OpenMap
+resolve ctx = foldM (resolveSingle ctx) mempty
+
+resolveSingle ::
```
cwgoes, line 236: add a comment describing pre, post-conditions, function behaviour

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,189 @@ lookupModMap s =
...
+        |> ( \map ->
+               foldr
+                 (\mod -> Map.insert mod (fmap Implicit qualifiedNames))
+                 map
+                 implicitInner
+           )
+        |> Right
+
+-- this goes off after the checks have passed regarding if the paths
+-- are even possible to resolve
```
cwgoes, line 256: add a comment describing pre, post-conditions, function behaviour

src/Juvix/FrontendContextualise/ModuleOpen/Transform.hs
```diff
@@ -39,6 +39,7 @@ transformModuleOpenExpr (Old.OpenExpress modName expr) = do
       let NameSpace.List {publicL} = NameSpace.toList innerC
           newSymbs = fst <$> publicL
        in protectOpenPrim newSymbs $ do
+            -- TODO ∷ should we update this to not shad
```
cwgoes, line 42: to not shadow?

test/FrontendContextualise/Module/Open.hs
```diff
@@ -0,0 +1,30 @@
...
+
+import Juvix.Library
+import qualified Juvix.FrontendContextualise.ModuleOpen.Environment as Env
+import qualified Data.Text as Text
+import qualified Juvix.Core.Common.Context as Context
+import qualified Juvix.Core.Common.NameSpace as NameSpace
+import qualified Juvix.Library.HashMap as HashMap
+-- import qualified Test.Tasty as T
+-- import qualified Test.Tasty.HUnit as T
+
```
cwgoes, line 11: needs tasty tests?

cwgoes (commented):
Initial pass, see comments, some functions need documentation or an example before I can review them.

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: it's a conflict with another function in env

src/Juvix/FrontendContextualise/Contextify/Transform.hs
```diff
@@ -5,36 +5,61 @@ module Juvix.FrontendContextualise.Contextify.Transform where
...
+f ,
+  contextify ::
+    Type.Context ->
+    (Context.NameSymbol, [Repr.TopLevel]) ->
+    Either Context.PathError Type.Pass
 contextify cont (nameSymb, xs) =
   case Context.switchNameSpace nameSymb cont of
-    Left errrr -> Left errrr
-    Right cont -> Right (foldr updateTopLevel cont xs)
+    Left errr -> Left errr
```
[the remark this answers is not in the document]
mariari: I think it would read worse

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: it would be odd if the term is in both at this point.

As we go, we depopulate one map while populating the other....

In reality for this function it doesn't matter.. as long as the symbol lives somewhere we are all good

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: I'd need to flip the foldr then

src/Juvix/FrontendContextualise/ModuleOpen/Transform.hs
mariari: yes to now shadow the functions already in scope.

test/FrontendContextualise/Module/Open.hs
mariari: yes, but that will come in a bit, it's commented out atm, as it's more convinent to add it back in the import when I actually do use them

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: I'm not sure there are many preconditions for this function.

I think the `opens` part of the `PreQualified` has to not contain toplevel... but that is something fixable

Outside of that the ctx should be full by this point.

I guess that is a precondition

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: Well this function deals with the single of the above, so any precondition would be from resolve... which doesn't really have any significant ones

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: This issue is now fixed, so there should be no preconditions other than ctx being realized by this point

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: alright

src/Juvix/FrontendContextualise/ModuleOpen/Transform.hs
cwgoes, line 42:
```suggestion
            -- TODO ∷ should we update this to not shadow
```

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,199 @@ lookupModMap s =
...
+    addToModMap map (def, sym) =
+      case def of
+        Context.Record cont _ ->
+          let NameSpace.List {publicL} = NameSpace.toList cont
+           in foldlM
+                ( \map (x, _) ->
+                    case map Map.!? x of
+                      -- this means we already have opened this in another
+                      -- module
+                      Just _ -> Left (AmbigiousSymbol x)
```
cwgoes, line 284:
```suggestion
                      Just _ -> Left (AmbiguousSymbol x)
```

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: still I think it's nicer, nbd though

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +159,199 @@ lookupModMap s =
...
+                  -- we don't shadow it....
+                  -- TODO ∷ later throw an error for
+                  --        explicit opens that do this
+                  modMap
+                Nothing ->
+                  Map.insert x nameSpace modMap
+            --
+            openList =
+              -- haskell gives type error if grabList is not typed
+              let grabList ::
```
cwgoes, line 211: Should this just be a top-level function?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: right, I guess I'm looking for an English description of what the function should do, that I can conceptualise and review the code against, otherwise it's tricky to know if it's correct or not

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: ditto, see above response

src/Juvix/FrontendContextualise/Contextify/Transform.hs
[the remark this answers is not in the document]
cwgoes: hmm alright

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +169,205 @@ lookupModMap s =
...
+    Right resFull
+  | otherwise =
+    Left (OpenNonModule diff)
+  where
+    resFull = resolveWhatWeCan ctx opens
+    resFirst = resolveWhatWeCan ctx (firstName <$> opens)
+    -- O(n log₁₆(n))
+    diff =
+      Set.difference
+        (Set.fromList (notResolved resFull))
```
cwgoes, line 341: some duplicate computation can be avoided here with `let`

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -100,3 +169,205 @@ lookupModMap s =
...
+      case def of
+        Context.Record cont _ ->
+          let NameSpace.List {publicL} = NameSpace.toList cont
+           in foldlM
+                ( \map (x, _) ->
+                    case map Map.!? x of
+                      -- this means we already have opened this in another
+                      -- module
+                      Just _ -> Left (AmbiguousSymbol x)
+                      -- if it's already in scope from below or above us
```
cwgoes, line 301: hmm, will this result in things being in scope with the names the programmer expects?

cwgoes (commented):
there are probably more declarative ways to do this, but they may be less efficient?

the examples & logic look alright to me, still see outstanding comments

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: Which part? 

If the symbol is already in the map, then that means `Open A` has symbol `x`. So when we get to `Open B` which also has symbol `x` we should have an ambigious open.

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: yeah will do

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: I think the limited scope is fine, if we have to use this logic again (we likely won't), then it's an easy move

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: sure I'll make `unpopulatedLookup`. Though it'll likely still be 2 lines

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: for `nameSpace` this is only really a temporary solution as implicit vs explicit should have more of a difference later?

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
mariari: done

mariari:
> there are probably more declarative ways to do this, but they may be less efficient?
> 
> the examples & logic look alright to me, still see outstanding comments

I'm sure there is a more declartive way to do it, however it is not obvious to me.

Maybe with some more standard infastructure I could? Some of my folds are basically, `addToMap` but left on ambi.

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: ah I see

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: alright

src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
cwgoes: alright, maybe add a comment then?

cwgoes (approved):
ACK modulo final remaining comment (add a comment about implicit/explicit)

PR #497, opened 2020-08-31
Add fields to parameterisations for injecting constants

andy-morris:
Basically what is added is a field `intTy :: Integer -> primTy -> Bool` to check whether a literal is valid for a given type (in range, etc), and `intVal :: Integer -> Maybe primVal` to wrap an integer (if possible), and similarly for `Text` and `Double` (which maybe we should switch to `Rational`? either way)

There are some parts which are still missing currently:
  - `intTy` and `intVal` currently reject everything in `Backends.ArithmeticCircuit.Parameterisation[.Integers]`, because I didn't know what the valid range was, or if we want `0` and `1` to type as bool as well. Strings and floats are also rejected but I don't think they exist in ACs.
  - The additions in `Michelson.Parameterisation` should definitely be looked over by someone which knows more about what is valid in michelson values

### DIFF
--- a/src/Juvix/Backends/Michelson/Parameterisation.hs
+++ b/src/Juvix/Backends/Michelson/Parameterisation.hs
@@ -164,8 +164,8 @@ michelson =
     stringVal = Just . Constant . M.ValueString . M.mkMTextUnsafe, -- TODO ?
     intTy = checkIntType,
     intVal = integerToPrimVal,
-    floatTy = \_ _ -> False, -- TODO
-    floatVal = const Nothing -- TODO
+    floatTy = \_ _ -> False, -- Michelson does not support floats
+    floatVal = const Nothing
   }
 
 type CompErr = CompTypes.CompilationError
--- a/src/Juvix/Backends/Michelson/Parameterisation.hs
+++ b/src/Juvix/Backends/Michelson/Parameterisation.hs
@@ -22,6 +22,7 @@ import Juvix.Library hiding (many, try)
 import qualified Michelson.Macro as M
 import qualified Michelson.Parser as M
 import qualified Michelson.Untyped as M
+import qualified Michelson.Text as M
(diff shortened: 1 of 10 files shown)


cyberglot:
>     * `intTy` and `intVal` currently reject everything in `Backends.ArithmeticCircuit.Parameterisation[.Integers]`, because I didn't know what the valid range was, or if we want `0` and `1` to type as bool as well. Strings and floats are also rejected but I don't think they exist in ACs.

I can update them to be valid. Floats are the basic primitive type for ACs, since they have to be field elements. Other types are just for the lib we are using.

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -135,15 +136,36 @@ reservedNames = []
...
-    parseVal
-    reservedNames
-    reservedOpNames
+  Core.Parameterisation {
+    typeOf, apply, parseTy, parseVal, reservedNames, reservedOpNames,
+    stringTy = checkStringType,
+    stringVal = Just . Constant . M.ValueString . M.mkMTextUnsafe, -- TODO ?
+    intTy = checkIntType,
+    intVal = integerToPrimVal,
+    floatTy = \_ _ -> False, -- TODO
```
cwgoes, line 167:
```suggestion
    floatTy = \_ _ -> False, -- Michelson does not support floats
```

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -135,15 +136,36 @@ reservedNames = []
...
+    floatVal = const Nothing -- TODO
```
cwgoes, line 168:
```suggestion
    floatVal = const Nothing
```

cwgoes (approved):
ACK, this is fine for now

PR #499, opened 2020-08-31
Upgrade extensible-data

andy-morris:
A tiny PR to bump the extensible version. You don't have to do that double-import stuff any more 🎉


cwgoes (approved):
ACK, thanks!

PR #500, opened 2020-08-31
Make typechecker (& evaluator) polymorphic over the monad and extension

andy-morris:
Currently extensions are just rejected by the typechecker; a way to integrate them properly is still in progress


thealmarty (approved):
LGTM!

cwgoes (approved):
utACK

PR #501, opened 2020-09-01
Derive Generic, Data, and NFData for core AST

andy-morris:


andy-morris:
(Sorry about the branch name still being wrong, it doesn't seem to be possible to change it :( )

cwgoes (approved):
ACK, seems fine

PR #502, opened 2020-09-02
Use Generics for HasWeak, etc

andy-morris:
This is the first commit of #500, because:

1. it might be easier to work on the substitution stuff with it merged?
2. the unifier also depends on it


cwgoes (approved):
ACK

PR #504, opened 2020-09-02
Move Prim to Term, allow multi-typed primitives, add `builtin` fields

### LINKED ISSUE #326: Refactor parameterisation to type-check in additional to type-infer
Some backends may have `PrimVal` inhabitants which can be type-checked but not type-inferred.

e.g. `V.Value Int` in Michelson/Morley, which can be an `Int`, a `Nat`, or a `Mutez` (maybe also `Timestamp`).

ref https://github.com/cryptiumlabs/juvix/pull/322/files#r382209153

andy-morris:
fixes #326

other notes:

- Since prims are terms now, `*` as multiplication clashes with `*` as type-of-types, so the nat operations have been renamed to `add`, `sub`, `mul`
- For the same reason, `()` clashes with brackets for grouping, so it has been renamed to `tt`
- I think the `builtin__` fields have all the information necessary to remove the `reserved__` and `parse__` fields? If so that should probably happen at some point in the future (edit: no, not quite, because that's where e.g. numeric literals are parsed too)
- Like I mentioned, the builtins in the arithmetic circuit and michelson parameterisations are empty because I don't know them well enough

### DIFF
--- a/test/Pipeline.hs
+++ b/test/Pipeline.hs
@@ -164,25 +164,33 @@ test_partial_erase =
     (EmptyInstr (MT.Seq (MT.Nested (MT.PUSH (MT.VInt 12))) MT.Nop))
 
 twoTerm :: HR.Term PrimTy PrimVal
-twoTerm = HR.Elim (HR.Prim (Constant (M.ValueInt 2)))
+twoTerm = HR.Prim (Constant (M.ValueInt 2))
 
 erasedLamTerm :: HR.Term PrimTy PrimVal
-erasedLamTerm = HR.Lam "x" (HR.Elim (HR.Prim (Constant (M.ValueInt 2))))
+erasedLamTerm = HR.Lam "x" (HR.Prim (Constant (M.ValueInt 2)))
 
 erasedLamTy :: HR.Term PrimTy PrimVal
 erasedLamTy = HR.Pi (Usage.SNat 0) "x" intTy intTy
 
 appLam :: HR.Term PrimTy PrimVal
-appLam = HR.Elim (HR.App (HR.App (HR.Ann (Usage.SNat 1) lamTerm lamTy 0) (HR.Elim (HR.Prim (Constant (M.ValueInt 2))))) (HR.Elim (HR.Prim (Constant (M.ValueInt 3)))))
+appLam = HR.Elim (HR.App (HR.App (HR.Ann (Usage.SNat 1) lamTerm lamTy 0) (HR.Prim (Constant (M.ValueInt 2)))) (HR.Prim (Constant (M.ValueInt 3))))
+
+addTyT :: HR.Term PrimTy PrimVal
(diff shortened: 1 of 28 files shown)


cyberglot (approved):
LGTM. 👍

test/Pipeline.hs
```diff
@@ -164,25 +164,33 @@ test_partial_erase =
...
-lamTerm = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App (HR.Prim AddI) (HR.Elim (HR.Var "x"))) (HR.Elim (HR.Var "y")))))
+lamTerm = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App addElim (HR.Elim (HR.Var "x"))) (HR.Elim (HR.Var "y")))))
 
 appLam2 :: HR.Term PrimTy PrimVal
-appLam2 = HR.Elim (HR.App (HR.App (HR.Ann (Usage.SNat 1) lamTerm2 lamTy2 0) (HR.Elim (HR.Prim (Constant (M.ValueInt 2))))) (HR.Elim (HR.Prim (Constant (M.ValueInt 3)))))
+appLam2 = HR.Elim (HR.App (HR.App (HR.Ann (Usage.SNat 1) lamTerm2 lamTy2 0) (HR.Prim (Constant (M.ValueInt 2)))) (HR.Prim (Constant (M.ValueInt 3))))
 
 lamTerm2 :: HR.Term PrimTy PrimVal
-lamTerm2 = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App (HR.Prim AddI) (HR.Elim (HR.Var "x"))) (HR.Elim (HR.Prim (Constant (M.ValueInt 10)))))))
+lamTerm2 = HR.Lam "x" (HR.Lam "y" (HR.Elim (HR.App (HR.App addElim (HR.Elim (HR.Var "x"))) (HR.Prim (Constant (M.ValueInt 10))))))
```
mariari, line 193: would be nice if we could newline these tests, so they aren't so big... also replacing ()'s with |> to make the empty )'s not eat so much space

That is a nit though

mariari (commented):
Everything looks fine.

Before you merge make sure you run it through the type checker.

I can finish the rest of the michelson mapping from here

cwgoes (commented):
Generally looks fine; we can ditch the parsing functions from the parameterisations entirely.

Mark this R4R whenever you think it is.

andy-morris:
I think it might be best to put the parser changes and builtin lists in separate PRs, since this one has probably ballooned enough already, so it's ready for review whenever people have the chance.

test/Pipeline.hs
andy-morris: I factored out some stuff to try and make these terms more legible. Is this better?

thealmarty (approved):
LGTM.

cwgoes (approved):
ACK

PR #514, opened 2020-09-14
Adding primitive translations for Michelson

mariari:


mariari:
Some overall structure things to do.

Some insturctions like `Left`, `Right` `Nil` need a type or an op to satisfy them.

Meaning that there will be a followup PR that will add the insturctions to the primval with special instructions

mariari:
This is also true for builtin data structures which take a type

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -161,11 +162,66 @@ checkIntType val (PrimTy (M.Type ty _)) = case ty of
   M.TInt -> True -- TODO bounds?
   _ -> False
 
+primify :: Untyped.T -> PrimTy
+primify t = PrimTy (Untyped.Type t "")
+
 builtinTypes :: P.Builtins PrimTy
-builtinTypes = [] -- FIXME
+builtinTypes =
+  [ (NameSymbol.fromSymbol "Michelson.unit-t", primify Untyped.TUnit),
```
cwgoes, line 170: this can be done with a `map (\(x, y) -> (NameSymbol.fromSymbol x, primify y))` for reduced duplication

src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -161,11 +162,66 @@ checkIntType val (PrimTy (M.Type ty _)) = case ty of
...
+    (NameSymbol.fromSymbol "Michelson.bool", primify Untyped.TBool),
+    (NameSymbol.fromSymbol "Michelson.key-hash", primify Untyped.TKeyHash),
+    (NameSymbol.fromSymbol "Michelson.timestamp", primify Untyped.TTimestamp),
+    (NameSymbol.fromSymbol "Michelson.address", primify Untyped.TAddress)
+  ]
 
 builtinValues :: P.Builtins PrimVal
-builtinValues = [] -- FIXME
+builtinValues =
+  [ (NameSymbol.fromSymbol "Michelson.add", AddI),
```
cwgoes, line 187: ditto for the first element

cwgoes (approved):
ACK pending nits & follow-up

src/Juvix/Backends/Michelson/Parameterisation.hs
mariari: Used Arr.*** to do this

src/Juvix/Backends/Michelson/Parameterisation.hs
mariari: done

PR #519, opened 2020-09-22
More permissive symbols and infix declarations

mariari:


src/Juvix/FrontendContextualise/Contextify/Transform.hs
```diff
@@ -42,13 +42,45 @@ updateTopLevel (Repr.Type t@(Repr.Typ _ name _ _)) ctx =
...
+          Repr.NonAssoc n assoc ->
+            (n, Context.Pred Context.NonAssoc (fromIntegral assoc))
+   in case Context.extractValue <$> Context.lookup (pure name) ctx of
+        Just def@(Context.Def {}) ->
+          Type.P
+            { ctx = (Context.add (NameSpace.Pub name) def {Context.precedence = prec}) ctx,
+              opens = [],
+              modsDefined = []
+            }
+        _ ->
```
cwgoes, line 73: throw an error? why are we adding undefined things?

src/Juvix/FrontendDesugar/RemoveModules/Transform.hs
```diff
@@ -37,6 +37,9 @@ transformTop body = foldr combine record body
           New.LetType (New.LetType'' (transformType typ) expression)
         Old.Function (Old.Func func) ->
           New.Let (New.Let'' (transformFunctionLike func) expression)
+        -- TODO ∷ update this later to handle infix declarations better
```
cwgoes, line 40: hmm, will this just error?

cwgoes (commented):
Generally looks fine

src/Juvix/FrontendDesugar/RemoveModules/Transform.hs
cwgoes: i.e., do we expect to hit this code?

src/Juvix/FrontendContextualise/Contextify/Transform.hs
cwgoes: ok, just add a comment

cwgoes (approved):
ACK modulo discussion

PR #524, opened 2020-09-28
frontend to core

andy-morris:


src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+transformTerm :: FE.Expression -> Env primTy primVal (IR.Term primTy primVal)
+transformTerm = fmap hrToIR . transformTermHR
+
+transformTermHR :: FE.Expression -> Env primTy primVal (HR.Term primTy primVal)
+transformTermHR (FE.Constant k) = do
+  p <- ask @"param"
+  case paramConstant p k of
+    Just x  -> pure $ HR.Prim x
+    Nothing -> throwFF $ UnsupportedConstant k
+transformTermHR (FE.Let l) = _
```
cwgoes, line 137: Presumably we can turn this into core `let`?

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+transformTermHR (FE.LetType l) = _
```
cwgoes, line 138: Ditto

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+transformTermHR (FE.Match m) = _
```
cwgoes, line 139: I guess this is the main remaining bit

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+transformTermHR (FE.Name n) = pure $ HR.Elim $ HR.Var _n
+transformTermHR (FE.Lambda l) = _
```
cwgoes, line 141: Also lambdas

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+    HR.Pi <$> transformUsage π
+          <*> x
+          <*> transformTermHR a
+          <*> transformTermHR b
+
+makeList :: [HR.Term primTy primVal] -> HR.Term primTy primVal
+makeList =
+  -- TODO: make lists a builtin syntax form in core (translated to
+  -- e.g. michelson lists)
+  _
```
cwgoes, line 192: This can be saved for a follow-up PR

cwgoes (commented):
Once we have let, lambda, match, and a few simple test-cases, I think we can merge this and continue work on e.g. list syntax, various not-yet-implemented features in follow-up PRs.

cwgoes:
Let's do the following:

- Add anything else done (other than the match stuff)
- Add the function to map a context from the frontend to term/globals (dealing with recursive groups)
- Stub out a function as required for frontend syntax free variables

Then let's merge this so we can work on follow-ups in parallel.

src/Juvix/Core/FromFrontend.hs
andy-morris: Frontend lets can be pattern matching functions, so their body will need to be lifted I think.

But if it's just `let foo a b c = ... in` then it can be transformed to `let foo = λ a b c → ... in` as a special case

src/Juvix/Core/FromFrontend.hs
andy-morris: LetType is a local datatype so it will also need to be lifted.

src/Juvix/Core/FromFrontend.hs
mariari: I presume once we have case, we can forgo lambda lifting and write something akin to `let foo =  λ a b c → case (a,b,c,) of ...`

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+transformTermHR (FE.UniverseName i) =
+  -- TODO for universe polymorphism
+  throwFF $ UniversesUnsupported i
+transformTermHR (FE.Parened e) = transformTermHR e
+
+transformArrow :: FE.ArrowExp -> Env primTy primVal (HR.Term primTy primVal)
+transformArrow f@(FE.Arr' xa π b) =
+  case xa of
+    FE.NamedTypeE (FE.NamedType' x a) -> go π (getName x) a b
+    a                                 -> go π _unusedName a b
```
mariari, line 178: is `_unusedName` a hole that gets fileld? should I use this to replace undefined that go unused?

src/Juvix/Core/FromFrontend.hs
```diff
@@ -0,0 +1,260 @@
...
+  | otherwise = throwFF $ ConstraintsUnsupported cons
+transformValSig def π Nothing = throwFF $ SigRequired def
+
+transformDef :: FE.Final Ctx.Definition
+             -> Env primTy primVal [IR.Global primTy primVal]
+transformDef (Ctx.Def π _ def _)  = _
+transformDef (Ctx.Record rec _)   = _
+transformDef (Ctx.TypeDeclar typ) = _
+transformDef (Ctx.Unknown _)      = pure []
+transformDef Ctx.CurrentNameSpace = pure [] -- FIXME ???
```
mariari, line 260: these get called in `transformInner` in the other passes.

It seems here `transformInner` is not a function, so this function is uneeded? 

I'm not really sure how this is all called.

src/Juvix/Frontend/FreeVars.hs
```diff
@@ -0,0 +1,108 @@
...
+  gfreeVars (x :*: y) = gfreeVars x <> gfreeVars y
+
+instance GFreeVars f => GFreeVars (M1 i t f) where
+  gfreeVars (M1 x) = gfreeVars x
+
+instance FreeVars c => GFreeVars (K1 i c) where
+  gfreeVars (K1 x) = freeVars x
+
+
+class BoundVars a where boundVars :: a -> VarSet
```
mariari, line 106: I'm not fully sure how all these typeclasseses interact just by reading it, is . It seems we need to do this so we don't fix this pass on a specific version of the frontend?

src/Juvix/Library.hs
```diff
@@ -116,6 +117,13 @@ traverseM ::
   f (m a2)
 traverseM f = fmap join . traverse f
 
+foldA ::
+  (Applicative f, Foldable t, Monoid a) =>
+  (b -> f a) ->
+  t b -> f a
+foldA f = foldl' (\acc x -> liftA2 (<>) acc (f x)) (pure mempty)
```
mariari, line 124: Is this function similar to foldmap?

```haskell
foldMap :: (Foldable t, Monoid m) => (a -> m) -> t a -> m
```

src/Juvix/Core/FromFrontend.hs
andy-morris: Yeah, exactly. It's just that would unfortunately be a very invasive change to core

src/Juvix/Core/FromFrontend.hs
andy-morris: It's a GHC hole at the moment because I wasn't sure what to put there. I suppose just `""` would do, if that won't break anything.

src/Juvix/Library.hs
andy-morris: Actually yeah, I got `fold` and `foldMap` mixed up.

I was going to rename this to `foldMapA` but I forgot, sorry.

src/Juvix/Core/FromFrontend.hs
andy-morris: The full thing when it all fits together will split the frontend context into dependency groups, and then for each group:

- get its declared type via `transformSig`, check all of them are actually types
- temporarily add them all to the core context as `GAbstract`s with the declared types
- transform & check the definitions with this temporary context (i.e., knowing the types of everything in the group, but not being able to reduce them)
- finally, extend the real core context with the actual definitions, which will be available to later groups

src/Juvix/Frontend/FreeVars.hs
andy-morris: Basically:

- `FreeVars` instances are needed for everything that can occur in an expression, i.e. until GHC stops saying `Could not deduce (FreeVars (Whatever' ext))`
- `FreeVars` can be empty unless there is some binding going on; the default implemented by `GFreeVars` is just to recurse into subterms and concatenate them together.
- `ExpressionAll'` will need to have a constraint added for each datatype reachable from `Expression'`. Sorry about that :( 
- an instance of `BoundVars` will be needed for everything in binding position, returning the names being bound here and visible to the outside (e.g. the definition name in a `Let`, the type name in `LetType`, but *not* the functions'/types' arguments).

PR #529, opened 2020-10-01
Standard library builitins

mariari:


stdlib/Prelude.ju
```diff
@@ -0,0 +1,7 @@
+-- it seems custom arrow syntax is not operational
```
cwgoes, line 1: hmm why not?

stdlib/Prelude.ju
```diff
@@ -0,0 +1,7 @@
...
+
+let (->) = %Builtin.ArrowOmega
+
+declare infixr (->) 1
+
+declare infixr (:) 2
```
cwgoes, line 7: is `(:)` a builtin?

mariari: yes, not one you can define though.

cwgoes (approved):
ACK, you can merge this if you want

PR #530, opened 2020-10-01
File/Module Headers

mariari:


src/Juvix/Frontend.hs
```diff
@@ -10,22 +9,26 @@ import Juvix.Library
...
-        |> fmap
-          ( Arrow.first
-              (NameSymbol.fromSymbol . intern . toUpper . FilePath.takeBaseName)
-          )
-        |> Right
-        |> pure
+    Right (Types.Header name xs) ->
+      pure (Right (name, xs))
+    Right (Types.NoHeader xs) ->
+      let toName =
```
cwgoes, line 29: This should be top-level?

cwgoes (changes requested):
One nit, and can you explain the intent here? Shouldn't every file have a module name?

mariari:
> One nit, and can you explain the intent here? Shouldn't every file have a module name?

Sure, it basically works like haskell does.

One can give a name for the module at the start or omit it.

src/Juvix/Frontend.hs
cwgoes: Hmm OK I think it's a bit better form but nbd

mariari:
> > > One nit, and can you explain the intent here? Shouldn't every file have a module name?
> > 
> > 
> > Sure, it basically works like haskell does.
> > One can give a name for the module at the start or omit it.
> 
> In which case it will default to the file name, or something like that?

exactly, it works like how it did before we made the change

cwgoes (approved):
ACK

PR #535, opened 2020-10-06
Adding the rest of the primitives to the Michelson backend

mariari:
This was due to certain types or operations taking types or ops


mariari:
last commit fixes the first part of the issue in #537

cwgoes (approved):
utACK

PR #543, opened 2020-10-13
simplified the logic, and fixed the bug where extra -- was not removed

mariari:


cwgoes (changes requested):
Changes look fine but needs a test-case

PR #545, opened 2020-10-14
Update resolver

cwgoes:


mariari:
Is there anything this bump breaks?

cwgoes:
No, but at the moment it's unnecessary. Let's wait until we switch to custom CI.

PR #556, opened 2020-10-15
Moving over FreeVars to be complete

mariari:


cwgoes (commented):
This looks alright but presumably we should be calling it in `FromFrontend`? cc @andy-morris

cwgoes (approved):
ACK from me

PR #558, opened 2020-10-16
hasType cases for the michelson Parameterisation

mariari:


src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -32,13 +33,63 @@ import Text.ParserCombinators.Parsec
...
+checkFirst2AndLast (x :| [y, last]) check
+  | check2Equal (x :| [y]) && check last = True
+  | otherwise = False
+checkFirst2AndLast (_ :| _) _ = False
+
 -- TODO: Add rest of primitive values.
 -- TODO: Add dependent functions for pair, fst, snd, etc.
 typeOf :: PrimVal -> NonEmpty PrimTy
 typeOf (Constant v) = PrimTy (M.Type (constType v) "") :| []
 typeOf AddI = PrimTy (M.Type M.TInt "") :| [PrimTy (M.Type M.TInt ""), PrimTy (M.Type M.TInt "")]
```
cwgoes, line 64: Delete this line?

cwgoes (changes requested):
We should remove `typeOf` completely, except maybe for constants, though at some point we'll need to check those too (e.g. literals can be either ints or nats)

cwgoes (approved):
ACK (for now)

PR #559, opened 2020-10-19
Moving Juvix into Proper Components

mariari:


Makefile
```diff
@@ -18,6 +18,25 @@ build-z3:
 build:
 	stack build --copy-bins --fast -j $(shell nproc)
 
+build-cache:
```
cwgoes, line 21: hmm can we still do this with a master `stack.yaml` file?

Makefile
```diff
@@ -34,9 +53,25 @@ format:
 	find . -path ./.stack-work -prune -o -path ./archived -prune -o -type f -name "*.hs" -exec ormolu --mode inplace {} --ghc-opt -XTypeApplications --ghc-opt -XUnicodeSyntax --ghc-opt -XPatternSynonyms --ghc-opt -XTemplateHaskell \;
 
 org-gen:
-	org-generation app/ doc/Code/App.org test/ doc/Code/Test.org src/ doc/Code/Juvix.org bench/ doc/Code/Bench.org
+	org-generation app/ doc/Code/App.org test/ doc/Code/Test.org src/ doc/Code/Juvix.org bench/ doc/Code/Bench.org library/ doc/Code/Library.org
 
 test:
+	cd library/StandardLibrary; \
```
cwgoes, line 59: ditto

also if we're going to repeat commands like this you can abstract it in a Makefile variable

library/Backends/ArithmeticCircuit/arithmetic-circuit.cabal
```diff
@@ -0,0 +1,91 @@
+cabal-version: 1.12
```
cwgoes, line 1: this file should not be committed, since we are using stack / hpack

cwgoes: ditto for subsequent files

cwgoes (changes requested):
A few comments, haven't yet tested locally.

Makefile
mariari: you can type stack build... however I think the master `stack.yaml` flag has some extra flags(?) it seems to ruin my cache at least for some reason.

library/Backends/ArithmeticCircuit/arithmetic-circuit.cabal
mariari: my bad, it got through, as I'm often adding files fast

stack.yaml
```diff
@@ -98,8 +106,8 @@ extra-deps:
 
 allow-newer: true
 
-extra-lib-dirs:
-  - .stack-work/prefix/lib
+# extra-lib-dirs:
+#   - .stack-work/prefix/lib
```
cwgoes, line 110: this was for Z3 but I guess we can figure it out later

cwgoes (changes requested):
Just a few more nits and still need to abstract the Makefile `test-all` command

Makefile
```diff
@@ -34,11 +34,30 @@ format:
...
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
+	cd library/Translate; \
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
+	cd library/Backends/LLVM; \
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
+	cd library/Backends/Michelson; \
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
+	cd library/Backends/ArithmeticCircuit; \
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
+	stack test --fast --jobs=$(shell nproc) --test-arguments "--hide-successes --ansi-tricks false"
```
cwgoes, line 59: this line is duplicated

mariari: it is not!, it is to build the main, and not arith circuit!

cwgoes: ah I see

Makefile
cwgoes: doing this in a separate PR

cwgoes (approved):
ACK

PR #560, opened 2020-10-23
Add fwrite-ide-info to Frontend

mariari:


cwgoes:
Did you mean to remove attoparsec or do we still require it?

cwgoes (approved):
ACK

Issue #537, opened 2020-10-09
Michelson Backend: Bug in appM.


mariari:
If we remove the optimization for appM for primitives

```haskell
-- this
appM :: Env.Reduction m => Types.NewTerm -> [Types.NewTerm] -> m Env.Expanded
appM form@(Types.Ann _u ty t) args =
  let app = inst form >>= flip applyExpanded args
   in case t of
        -- We could remove this special logic, however it would
        -- result in inefficient Michelson!
        Ann.Prim p ->
          let (f, lPrim) = primToFargs p ty
           in case length args `compare` lPrim of
                EQ -> Env.unFun f args
                LT -> app
                GT ->
                  throw
                    @"compilationError"
                    (Types.InternalFault "Michelson call with too many args")
        _ -> app

-- to this
appM :: Env.Reduction m => Types.NewTerm -> [Types.NewTerm] -> m Env.Expanded
appM form@(Types.Ann _u ty t) args = inst form >>= flip applyExpanded args
```

Then certain examples fail not being able to find the proper elements in the stack.

This is most likely due to inconsistent consing.

One such issue was found in `pushConstant` which did not cons a value to the stack at all, causing an error with non fast tracked logic (fast tracked meaning that extra case).

That issue in particular is now fixed but in the tests `xtwice` is suffering the same issue and is under investigation.

```haskell
> extractTest xtwice
(Left (NotInStack "3"),Env {stack = T {stack' = [(VarE (fromList ["3","4","x"]) (Usage (SNat 1) True) (Just FuncResultE),Type TInt :),(VarE (fromList ["y"]) (Usage (SNat 0) False) (Just FuncResultE),Type TInt :),(VarE (fromList ["x"]) (Usage (SNat 1) False) (Just FuncResultE),Type TInt :),(VarE (fromList ["z"]) (Usage (SNat 0) False) (Just FuncResultE),Type TInt :)], size = 4}, compilationLog = [], ops = [], count = 5, debug = 0})
```

mariari:
it seems the main issue is that it works with usage 4, but not usage 2.

This is because the name didn't propagate properly, and thus only had one of the x's properly named.

mariari:
With the merging of #562 this should be fixed,

PR #562, opened 2020-10-26
Fixing bug where not special casing appM causes usage errors

mariari:
It seems the errors in #537 are related to names not propogating propelry...

However fixing this breaks the identity example @cwgoes 

```haskell
identityApp :: T.TestTree
identityApp =
  shouldCompile
    identityAppTerm
    identityType
    "parameter unit;storage unit;code { { DIG 0;DUP;DUG 1;DIG 0;DUP;DUG 1;CAR;NIL operation;PAIR;DIP 1 { DROP };DIP { DROP } } };"


   but got: Left (DidNotTypecheck (SeqEx [PrimEx DIG 0,PrimEx DUP,PrimEx DUG 1,PrimEx DIG 0,PrimEx DUP,PrimEx DUG 1,PrimEx CAR,PrimEx NIL operation,PrimEx PAIR,PrimEx DIP 2 { DROP },PrimEx DIP 1 { DROP },PrimEx DIP { DROP }]) Error checking expression DROP against input stack type [].)

```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -4,9 +4,9 @@ import qualified Compile as Compile
 import qualified Config as Config
 import Development.GitRev
 import qualified Interactive as Interactive
+import Juvix.Library
 import Options
 import Options.Applicative
-import Juvix.Library
 import System.Directory
 import Text.PrettyPrint.ANSI.Leijen hiding ((<>))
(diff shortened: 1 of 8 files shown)


cwgoes (approved):
ACK

PR #572, opened 2020-11-05
Weakening effects for Context

mariari:
This PR stands to weaken the effects in the `Contextualise` passes, along with making parts of the code more readable and correct.


library/Translate/src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -194,13 +199,15 @@ populateModMap = do
...
       -- thus our map should have implciit explicit on
       -- each symbol. Later when we are done, we remove
       -- these markings as they are no longer useful
-      put @"modMap" (foldr f modM opens)
+      -- TODO ∷
+      -- should we even append modM to this
+      put @"modMap" (modM <> Map.fromList (concatMap f opens))
       where
-        f y modMap = foldr addNewSymbol modMap openList
+        f y = catMaybes (fmap addNewSymbol openList)
```
mariari, line 210: Note the removal of folds for maps, for ease of reasoning.

library/Translate/src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -265,47 +286,26 @@ populateModMap = do
...
       -- should we even append modM to this
-      put @"modMap" (modM <> Map.fromList (concatMap f opens))
+      assocNameWithAlias <- concatMapM f opens
+      put @"modMap" (modM <> Map.fromList assocNameWithAlias)
       where
-        f y = catMaybes (fmap addNewSymbol openList)
+        f y = do
+          openList <- inScopeNames nameSpace
+          maybeAssocNameWithAlias <- traverse addNewSymbol openList
+          pure (catMaybes maybeAssocNameWithAlias)
```
mariari: Due to accessing information is now an effect, we have to promote our `concatMap` to `concatMapM`. However the code is a lot cleaner to read now

library/Translate/src/Juvix/FrontendContextualise/ModuleOpen/Environment.hs
```diff
@@ -83,12 +110,72 @@ newtype Context a = Ctx {antiAlias :: ContextAlias a}
...
+-- Generic Interface Definitions for Environment lookup
+--------------------------------------------------------------------------------
+
+-- | Names encapsulates the idea of looking up all current names in
+-- a context
+class Names a where
+  type NameConstraint a (m :: * -> *) :: Constraint
+  contextNames :: (NameConstraint a m) => NameSymbol.T -> a -> m [Symbol]
+  currentNameSpace' :: NameConstraint a m => a -> m (NameSymbol.T)
+  inCurrentModule' :: NameConstraint a m => NameSymbol.T -> a -> m Bool
```
mariari, line 134: This interface should make it very easy to test `populateModMap` without having any context logic.

Since all calls hit this, we can setup an interface for this that just returns static things, which we can test against.

mariari:
This PR does not cover adding secondary effects for testing nor for a more minimal environment. Those will be had in a follow up PR.

library/Translate/src/Juvix/FrontendContextualise/InfixPrecedence/Environment.hs
```diff
@@ -53,18 +92,71 @@ newtype Context a = Ctx {antiAlias :: ContextAlias a}
...
+  queryInfo' sym SingleDispatch = do
+    undefined
+
+instance Query EnvDispatch where
+  type QueryConstraint _ m = TransitionMap m
+  queryInfo' sym EnvDispatch = do
+    looked <- lookup sym
+    lookedO <- ask sym
+    let -- dealing with the second map winning lookup
+        -- Main rules:
```
cwgoes, line 131: Is this explicitly documented somewhere (somewhere else)?

cwgoes: also wondering if this function should be top-level

cwgoes (commented):
Changes generally look OK. I'm not familiar with this typeclass usage pattern but it's cool.

Do we have tests that cover the functions being changed here?

mariari:
> Changes generally look OK. I'm not familiar with this typeclass usage pattern but it's cool.
> 
> Do we have tests that cover the functions being changed here?

Yes the tests have cases that test the functions being changed and the pipeline itself

library/Translate/src/Juvix/FrontendContextualise/InfixPrecedence/Environment.hs
mariari: This should be gone once we don't have to deal with two maps at once, It'll be promted when I have to reproduce the logic again in another env effect

cwgoes (approved):
ACK

PR #575, opened 2020-11-09
Creating new effects for single function compilation

mariari:


cwgoes (approved):
ACK

mariari:
I think we should merge this now, making a running environment for `ModuleOpen` requires #576 to be implemented.

PR #579, opened 2020-11-19
Subtyping document

mariari:


doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+
+  -- or
+  f1 (coerce {x = 5; y = 3}) {x = 1}
+#+end_src
+
+*** Further Arguments/Unresolved Issues
+As of <2020-11-19 Thu> we still don't have a full answer for the
+following questions
+
+1) How do we handle usages
```
cwgoes, line 137: At minimum, implicit coercion should fail for non-0 non-w record fields - explicit coercion (dropping fields) will be required.

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+intention, with the most serious error still being impossible!
+
+Finally the last claim for f_{1}'s legitimacy rests on  flexibility of a
+library user and the incorrect choices of the author. f_{1} should really
+allow different subtypes, it just needs to specify it so! For this we
+can gain back our expressiveness by adding an =coerce= function which
+forgets fields to make the two records the same, and thus legitimate
+in f_{1}
+#+begin_src ocaml
+  f1 {x = 5; y = 3} (coerce {x = 1})
```
cwgoes, line 127: Wait what does this do? We can't add a `y` field, we don't have any value to put in it

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+
+Another argument is around the flexibility of functors if we disallow
+the above example. However this argument was also flawed in that we
+still have a subtyping relation so f_{3} in our code block can be called
+like
+#+begin_src ocaml
+  f3 {x = 3; y = 4}
+#+end_src
+
+thus the argument on functors in a single argument can be
```
cwgoes, line 112:
```suggestion
Thus the argument regarding functors with a single argument can be
```

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+disregarded. However, for multiple argument the legitimacy of the rest
```
cwgoes, line 113:
```suggestion
disregarded. However, for multiple arguments the legitimacy of the rest
```

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+What 2 tells us is that a function which accepts a record with fields
+=a= through =v= can also accept a record with those fields /and/ extra
+fields. The intuition is that in functions the sub typing relation is
+reversed, but we still need to forget information to get our
+relationship.
+
+
+*** Implications of Rules and Arguments about Behavior
+
+These implications can be quite worrying, consider the longer form
```
cwgoes, line 28:
```suggestion
These implications can be quite worrying, consider the longer-form
```

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
+* Subtyping
+** An argument about the extent of subtyping
+
+Subtyping is an important part of any module or Object system. The
```
cwgoes, line 4:
```suggestion
Subtyping is an important part of any module or object system. The
```

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+  -- note that the t is opaque
+  sig set : int -> mod {x : t,  type t, add : t -> t -> t}
+  mod set x =
+    type t = int
+    add = (+)
+    x
+  end
+
+  -- We include the set
+  sig set : int -> mod {x : t,  type t, add : t -> t -> t}
```
cwgoes, line 55: `sig bag`, perhaps?

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+  mod bag x = include (set x)
```
cwgoes, line 56: is this valid syntax?

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,146 @@
...
+
+  f1 (set 3) (bag 4)
+
+  foo a b c =
```
cwgoes, line 60: if you include this in the example we should explain it

cwgoes (changes requested):
Thanks for the write-up; a few questions and maybe @andy-morris would like to add arguments in favour of allowing `f1`.

cyberglot:
I agree we should probably think about subtyping, but probably go for [algebraic subtyping](https://www.cs.tufts.edu/~nr/cs257/archive/stephen-dolan/thesis.pdf)? it's my leisure reading of this weekend, so I will report back on monday.

doc/Architecture/Subtyping.org
mariari: signature syntax isn't fleshed out yet, but the rest should be?

doc/Architecture/Subtyping.org
mariari: Fair, though I was hoping we could fix the type for the application, but you would be correct

doc/Architecture/Subtyping.org
mariari: yes

doc/Architecture/Subtyping.org
mariari: I did mention it once in regards to oleg's article but it isn't very explicit

mariari: oh, you meant the f1 example.... well I removed it, I did update the section on foo and foo1 however

doc/Architecture/Subtyping.org
mariari: I removed the mod word

thealmarty:
I think we don't have to worry about these issues for Juvix. The Oleg article concerns imperative programming languages and Juvix is not. As you can see in Oleg's [proposed solution](http://okmij.org/ftp/Computation/Subtyping/Preventing-Trouble.html), the fix is basically to make your language functional!

mariari:
> I think we don't have to worry about these issues for Juvix. The Oleg article concerns imperative programming languages and Juvix is not. As you can see in Oleg's [proposed solution](http://okmij.org/ftp/Computation/Subtyping/Preventing-Trouble.html), the fix is basically to make your language functional!

In his article he argues that even a functional object system has this issue! The issue is with subtyping not being functional or imperative

thealmarty:
I didn't think Juvix has an object system either.

mariari:
> I didn't think Juvix has an object system either.

no but the issue still stands!

andy-morris:
I only have a couple of things to add:

- I feel pretty strongly that data structures like these implementations of `Set`/`Bag`, that have invariants not maintained by construction, should be abstract outside of their defining module. (Caveats for things like testing apply, I suppose, but outside of their defining _package_, definitely.) Once a type becomes abstract, it no longer has any subtyping behaviour since its structure is no longer available. In a language where correctness is important I think it should either be guaranteed with proofs embedded in the type, or at least enforced by an abstraction barrier.

- I also think that *not* having subtyping for modules would be a more serious drawback than we give it credit for. Imagine if in OCaml you couldn't write `Set.Make(Int)` but instead had to write

  ```
  Set.Make(struct
    type t = int
    val compare = Int.compare
  end)
  ```

  every single time. Perhaps that could be its representation at a sufficiently low level but I don't think forcing the user to write it like this would be a good idea.

- This is a minor point, but the types of the fᵢ all use bounded polymorphism, which are not something that Juvix *currently* has any specific plans for. At the moment you can only write:

  ```
  f: {x: int} → {x: int} → {x: int}
  f = ...
  ```

  which is less powerful but is at least completely unambiguous what fields the result has.

  If we do decide to go for record subtyping of some sort, we may also want to have bounded polymorphism when we have the infrastructure for it (a constraint solver, mostly), so we should probably still keep it in mind. It might not be necessary if we're only using it for things that are sufficiently module-like though?

mariari:
> I only have a couple of things to add:
> 
> * I feel pretty strongly that data structures like these implementations of `Set`/`Bag`, that have invariants not maintained by construction, should be abstract outside of their defining module.

I think we would agree on this point for the most part, not having it be so could lead into problems. Though I think for repl purposes they should implement some kind of show. 

My exceptions are for built in data types? LIke it's rather nice to have cons and nil when working with lists and what not.  

>  (Caveats for things like testing apply, I suppose, but outside of their defining _package_, definitely.) 

Yeah this is why I was thinking of something akin to private that lets me peek in only for testing.

> Once a type becomes abstract, it no longer has any subtyping behaviour since its structure is no longer available. In a language where correctness is important I think it should either be guaranteed with proofs embedded in the type, or at least enforced by an abstraction barrier.

What kind of abstraction barrier are you thinking of? Also for no subtyping, do you mean even among the explicitly declared subtyping types?

> I also think that not having subtyping for modules would be a more serious drawback than we give it credit for. I...

I completely agree

> This is a minor point, but the types of the fᵢ all use bounded polymorphism, which are not something that Juvix currently has any specific plans for. At the moment you can only write:

hmmm it seems in your example you've given the invariant version
```ocaml
f: {x: int} → {x: int} → {x: int}
```
what would a version with subtyping look like, specifically the three I gave.

Also since these are expression I could always be lazy and start abstracting over the signatures

```haskell
f : let y = {x : int} in y → y → y
```

but it seems like I'm basically just doing foralls by hand?

thealmarty:
@mariari I think you are confused. In a functional language without objects, like Juvix, there is no issue allowing `f2`. See the following Haskell code:

```haskell
class HasX n where
  x :: n -> Int
  setX :: n -> Int -> n

data JustX = JustX Int

instance HasX JustX where
  x (JustX v) = v
  setX _ = JustX

f0 :: (HasX n, HasX m) => n -> m -> n
f0 n m = setX n (x n + x m)

f1 :: (HasX n) => n -> n -> n
f1 n m = setX n (x n + x m)

f2 :: (HasX n, HasX m) => n -> m -> JustX
f2 n m = JustX (x n + x m)
```

mariari:
> @mariari I think you are confused. In a functional language without objects, like Juvix, there is no issue allowing `f2`. See the following Haskell code:
> 
> ```haskell
> class HasX n where
>   x :: n -> Int
>   setX :: n -> Int -> n
> 
> data JustX = JustX Int
> 
> instance HasX JustX where
>   x (JustX v) = v
>   setX _ = JustX
> 
> f0 :: (HasX n, HasX m) => n -> m -> n
> f0 n m = setX n (x n + x m)
> 
> f1 :: (HasX n) => n -> n -> n
> f1 n m = setX n (x n + x m)
> 
> f2 :: (HasX n, HasX m) => n -> m -> JustX
> f2 n m = JustX (x n + x m)
> ```

The issue still persists, the issue with `f2` has naught to do with objects, but with subtyping. 

Your example gives back a `JustX` which is explicit about the return type, which is fine, as it does not imply it's the same type as the arguments. This is in part due to Haskell lacking subtyping, so it can't express the problematic relation that is the `f2` in my example. Thus imagine your type really had a signature that was akin to `f1`

```haskell
(HasX n) => n -> n -> n
```

but the output was constructed with a `JustX`.

This would be a valid signature in fω sub, which is valid fω with sub typing. No objects neededs, to get this problematic behavior.

the idea the document proposes is that we make this invalid, namely all these n's would refer to the **same** subtype. with

```haskell
f0 : λX:>{x : int} -> x1 -> x2 -> x1
f1 : λX:>{x : int} -> X -> X -> X
f2 : λX:>{x : int} -> x1 -> x2 -> {x : int}
```

being the cannonical signatures for the various behaviors on display.

f0 takes 2 different subtypes of `{x : int}` and returns a thing shaped like the first.

f1 takes 2 of the same subtypes of `{x : int}` and returns a thing of the same shape.

f2 takes 2 different variants (like your `f2`!) and returns a `{x : int}` (like your `f2`).


As andy pointed out a proper module system needs some kind of subtyping! Further we are already doing subtyping with usages.

doc/Architecture/Subtyping.org
```diff
@@ -0,0 +1,161 @@
+* Subtyping
+** An argument about the extent of subtyping
+
+Subtyping is an important part of any module or object system. The
```
cyberglot, line 4: I don't think Subtyping is a requirement for module/object system. I believe Row Polymorphism is better suited for this task and can be trivially implemented within the dependent typing discipline [citation needed]*.

* not sure about that but I think it's probably true.

mariari: care describing the technical differences for me? It looks like row polymorphism is like subtyping for records only?

Doe it have the distinction that normal subtyping does not (see the f1, f0, and f2 examples)?

cyberglot: The main "problem" with subtyping is that it sometimes makes important information disappear, even for structural subtyping. I believe we can make this work with parametric polymorphism and row polymorphism, and  we can implement algebraic subtyping as a library or something that happens before core.

mariari: I'd love to see this written up, could you say make a new section about this, and display the arguments and what we could do?

cyberglot: @mariari I will try to write it asap, and we can discuss next in the next architecture meeting.

cyberglot:
@mariari this what you wanted? https://try.purescript.org/?session=f14632a4-a8a6-3663-6ace-c2285f174ea6

cwgoes:
What's the status of this PR?

mariari:
@cyberglot please update the website with any changes you want from this

PR #582, opened 2020-11-25
Start adding typechecking of declarations

thealmarty:


thealmarty:
Close for now. Will reopen after release.

PR #583, opened 2020-11-26
Remove unused libraries

thealmarty:
This removes LLVM, arithmetic circuit, and interaction net, which were added to facilitate future plans. Plans have changed and these won't be needed for a while. They are removed for now to keep Juvix lighter.


thealmarty:
This compiles on my laptop but fails on the CI. Looking into it.

thealmarty:
The latest commit compiles on both my laptop and desktop. I think the CI failures are from caching? @cwgoes any idea?

library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -63,7 +63,7 @@ extensible
         Elim (Elim primTy primVal)
       deriving (Eq, Show, Generic, Data, NFData)
 
-    -- | inferable terms
+    -- inferable terms
```
thealmarty, line 66: Haddock doc not yet supported in template haskell. Removing for now.

cwgoes:
> The latest commit compiles on both my laptop and desktop. I think the CI failures are from caching? @cwgoes any idea?

Looks like `arithmetic-circuit` is still listed as a dependency in `package.yaml` - probably will need to remove that.

thealmarty:
> > The latest commit compiles on both my laptop and desktop. I think the CI failures are from caching? @cwgoes any idea?
> 
> Looks like `arithmetic-circuit` is still listed as a dependency in `package.yaml` - probably will need to remove that.

Oops you're right. Thanks. I guess `stack clean` isn't sufficient! I'm checking with a `/tmp` dir now.

thealmarty:
Still working on it. Not ready for review yet.

thealmarty:
Per meeting pausing this PR to work on other more important issues.

thealmarty:
- cleaned up some warnings.
- removed LLVM/INet/ArithmeticCircuit libraries.
- moved EAC out of core.

cwgoes:
@thealmarty I think you might have some leftover local files since the org-gen is different. Fixed quickly.

cwgoes (approved):
ACK

PR #587, opened 2020-11-30
Add signature pass to contextToCore

### LINKED ISSUE #585: Signature type error on frontend to core conversion
Right now, when trying to convert anything from frontend to core (from the example contracts), I get:

```bash
WrongSigType Nothing
```

It looks like this is coming from [this line](https://github.com/metastatedev/juvix/blob/develop/src/Juvix/Core/FromFrontend.hs#L426).

Any ideas? Maybe we should alter the error message to make it more descriptive?

This issue can be replicated on `develop` with:

```bash
make test-typecheck
```

(which just calls `contextToCore`)

andy-morris:
Fixes #585

- [x] do signature pass in `contextToCore`
- [x] use `topLevelMap` instead of `currentNamespace`
- [x] special-case declarations of built-in syntax to not need a signature  
  `%Builtin.Arrow`, `%Builtin.Pair`, `%Builtin.Omega`, `%Builtin.Colon`, `%Builtin.Type`
  - [x] this specialness needs to be transitive to partial applications
  - [x] update `Prelude.ju`
- [x] dependency sorting for definitions
- [x] dependency sorting for *modules*
- ~~give the fromfrontend errors actual messages~~ move to follow-up pr
- [x] deal with unqualified local names
- [x] transform the remaining specials correctly
- [x] return the dependency order so the typechecker (etc) can also use it


andy-morris:
current problems:

1. The context from the frontend is currently in arbitrary hashmap order. it needs to either remember the declaration order in the source, or do a dependency sort first. Otherwise the signatures of the names encountered might not even have been processed yet.

2. For some reason, `add`, `identity_unit`, and `replace_string` don't seem to see the contents of the prelude, while `swap` and `transfer` seem to just fine???

cwgoes:
@andy-morris Is this ready-for-review?

andy-morris:
@cwgoes it still doesn't work yet, but it's probably close enough that you can start reviewing it.

library/Backends/Michelson/src/Juvix/Backends/Michelson/Parameterisation.hs
```diff
@@ -270,7 +270,7 @@ builtinValues =
     ("Michelson.self", Inst (M.SELF "" "")),
     ("Michelson.unit", Inst (M.UNIT "" "")),
     -- added symbols to not take values
-    ("Michelson.if", Inst (M.IF [] []))
+    ("Michelson.if_builtin", Inst (M.IF [] []))
```
mariari, line 273: Probably best to use a `-` to stay consistent with the rest?

library/StandardLibrary/src/Juvix/Library/NameSymbol.hs
```diff
@@ -59,17 +63,20 @@ hd = NonEmpty.head
...
+mod = fst . split
 
-baseName :: T -> Symbol
-baseName = snd . split
+base :: T -> Base
+base = snd . split
 
-applyBase :: (Symbol -> Symbol) -> T -> T
+applyBase :: (Base -> Base) -> T -> T
 applyBase f n = let (m, b) = split n in qualify1 m (f b)
```
mariari, line 82: Good changes

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
```diff
@@ -0,0 +1,119 @@
...
+      (g, fromV', _) = Graph.graphFromEdges edges
+      fromV v = let (gs, _, _) = fromV' v in gs
+   in Graph.topSort g |> reverse |> concatMap fromV
+
+recGroups' ::
+  (Data term, Data ty, Data sumRep) =>
+  Context.NameSpace term ty sumRep ->
+  Env term ty sumRep ()
+recGroups' ns = do
+  defs <- fmap concat $ for (NameSpace.toList1' ns) \(name, def) ->
```
mariari, line 93: Minor nit: `fmap f $ ==== f <$>`

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
```diff
@@ -0,0 +1,119 @@
...
+        Graph.scc g
+          |> mapAccumL (\xs t -> accum xs (toList t)) HashSet.empty
+  addDeps fvs
+  for_ groups addGroup
+
+fv :: Data a => a -> [NameSymbol.T]
+fv = HashSet.toList . SYB.everything (<>) (SYB.mkQ mempty (FV.op []))
+
+toNameSpace :: HashMap.T Symbol a -> NameSpace.T a
+toNameSpace public = NameSpace.T {public, private = mempty}
```
mariari, line 119: Thankfully when I merge my #593 this will be less hacky!

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
```diff
@@ -0,0 +1,119 @@
...
+      get n = maybe [] toList $ HashMap.lookup n deps
+      edges = map (\(n, gs) -> (gs, n, get n)) $ HashMap.toList groups
+      (g, fromV', _) = Graph.graphFromEdges edges
+      fromV v = let (gs, _, _) = fromV' v in gs
+   in Graph.topSort g |> reverse |> concatMap fromV
+
+recGroups' ::
+  (Data term, Data ty, Data sumRep) =>
+  Context.NameSpace term ty sumRep ->
+  Env term ty sumRep ()
```
mariari, line 91: it seems we are fixing ourselves to an environment here, instead of just relying on the field we care for.

All i can see  is a direct use of "currNameSpace"

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
```diff
@@ -0,0 +1,119 @@
...
+    Group,
+    Groups,
+    recGroups,
+  )
+where
+
+import qualified Data.Graph as Graph
+import qualified Data.HashSet as HashSet
+import qualified Generics.SYB as SYB
+import Juvix.Core.Common.Context.Traverse.Types
```
mariari, line 19: I found this dep a bit hard to figure out what was in it, given how it has more functions than just run, env, etc.

(though this could just be because my jump to source isn't online while reviewing this yet, as I didn't recompile things)

library/Translate/src/Juvix/FrontendContextualise/InfixPrecedence/FreeVars.hs
```diff
@@ -85,8 +85,8 @@ op _ (Type.UniverseName _) = Set.empty
 op s (Type.NamedTypeE nem) = nameifyNamedType s nem
 op s (Type.Application ap) = nameifyApplication s ap
 op s (Type.Name varNam)
-  | Set.member (NameSymb.hd varNam) s = Set.singleton varNam
-  | otherwise = Set.empty
+  | Set.member (NameSymb.hd varNam) s = Set.empty
+  | otherwise = Set.singleton varNam
```
mariari, line 90: My bad

library/Translate/src/Juvix/FrontendContextualise/InfixPrecedence/FreeVars.hs
```diff
@@ -160,7 +160,7 @@ nameifyNamedType s (Type.NamedType' _name exp) =
 
 -- we currently don't care about declares
 nameifyDeclarationExpression :: ExcludedSet -> Type.DeclarationExpression -> Set
-nameifyDeclarationExpression s (Type.DeclareExpession _i e) =
+nameifyDeclarationExpression s (Type.DeclareExpression _i e) =
```
mariari, line 163: Also my bad

src/Juvix/Core/FromFrontend.hs
```diff
@@ -232,56 +276,125 @@ transformPatVar p = pure p
...
-transformTermHR (FE.Name n) = pure $ HR.Elim $ HR.Var n
-transformTermHR (FE.Lambda l) = transformSimpleLambda l
-transformTermHR (FE.Application (FE.App f xs)) = do
-  f' <- toElim f =<< transformTermHR f
-  HR.Elim . foldl HR.App f' <$> traverse transformTermHR xs
-transformTermHR (FE.Primitive (FE.Prim p)) = do
+transformTermHR ::
+  NameSymbol.Mod ->
+  FE.Expression ->
+  Env primTy primVal (HR.Term primTy primVal)
```
mariari, line 282: As previous all these functions shouldn't fix the env, instead use an alias which spans the map, so I can add my own custom environment down later!

src/Juvix/Core/FromFrontend.hs
```diff
@@ -232,56 +276,125 @@ transformPatVar p = pure p
...
+        throwFF UnexpectedOmega
+      Nothing -> do
+        f' <- toElim f =<< transformTermHR q f
+        HR.Elim . foldl HR.App f' <$> traverse (transformTermHR q) xs
+    nargs n xs
+      | (xs, []) <- splitAt n (toList xs),
+        length xs == n =
+        pure xs
+      | otherwise =
+        throwFF $ WrongNumberBuiltinArgs n xs
```
mariari, line 362: Application seems quite special

src/Juvix/Core/FromFrontend.hs
```diff
@@ -232,56 +276,125 @@ transformPatVar p = pure p
...
+        π <- transformUsage q π
+        go (Just (ArrowS (Just π))) [a, b]
+      Just (ArrowS (Just π)) -> do
+        ~[xa, b] <- nargs 2 xs
+        (x, a) <- namedArg q xa
+        HR.Pi π x a <$> transformTermHR q b
+      Just (PairS Nothing) -> do
+        ~[π, a, b] <- nargs 3 xs
+        π <- transformUsage q π
+        go (Just (PairS (Just π))) [a, b]
```
mariari, line 338: This seems a bit odd, what if we have
```
b : B
f : B -> A
(f b : A)
```

This now seems ambi instead of binding `f b` or `b` to type `A`

src/Juvix/Core/FromFrontend.hs
mariari: Care to explain what all these special casings are?

src/Juvix/Core/FromFrontend.hs
mariari: Further I can follow this up, if A is C -> D by
```
(f b : A) c
```

stdlib/Michelson.ju
```diff
@@ -3,32 +3,32 @@ mod Prelude.Michelson where
...
-let pair = %Michelson.pair
-
-sig either : ty -> ty
-let either = %Michelson.either
-
-sig map : ty -> ty
-let map = %Michelson.map
-
-sig big-map : ty -> ty
-let big-map = %Michelson.big-map
```
mariari, line 31: Did the type signatures not work here?

mariari (commented):
Overall seems OK

There is an ambi point in application, and you fix your env in many places

src/Juvix/Core/FromFrontend.hs
andy-morris: These are the interpretations of the magic builtins, e.g. `x : A`, which gets parsed as `App (:) [x, A]`, needs to be translated to `Ann x A` rather than ``(:) `App` x `App` A``.

src/Juvix/Core/FromFrontend.hs
andy-morris: Sorry I don't think I understand the problem. This alternative interpretation is only used after `%Builtin.Pair` (or `Arrow`), in something like `dpair 1 (x: A) (B x)`, and only if the `x` part is a single identifier. In other cases `(x: A)` is just a type annotation (or more precisely an application of `(:)`). Could you clarify?

stdlib/Michelson.ju
andy-morris: The type signatures are fine, the only problem with these is that they're not in the parameterisation yet.

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
andy-morris: Yeah that's true, this was written a little differently at first with a `$` somewhere so the precedences didn't line up. But I can change it to that now

src/Juvix/Core/FromFrontend.hs
mariari: well consider the expression
```haskell

sig f = int -> int -> int
let f = (+)

-- I forget the exact syntax that F* uses for refinements....
let nat = x : int { x >= 0 }

let foo a b = 
  (f a : nat -> nat) a b
```

to me it's not clear that `:` is only used in a signature context so saying that they are Π always seems odd to me?

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
andy-morris: I can move the functions out to the main `Traverse` module if that would be clearer, or rename `Types` to something else?

library/Translate/src/Juvix/Core/Common/Context/Traverse.hs
andy-morris: There are indirect uses of the other features; I've rewritten the type to list the capabilities.

cwgoes (approved):
ACK, thanks @andy-morris!

PR #588, opened 2020-11-30
Some minor makefile fixes

andy-morris:
- Make the `test-*` targets depend on `build`, so they are actually testing the current code
- Use `stack exec juvix` instead of just `juvix`, so the executable need not be installed globally


cwgoes (approved):
ACK

PR #590, opened 2020-12-01
Makefile: Don't copy executable to global path

andy-morris:
This patch just removes the `--copy-bins` flags from the builds, since `stack exec` uses the local copy anyway.


cwgoes (approved):
ACK

PR #593, opened 2020-12-14
Adding Reverse dependency lookup and other goodies to the Context

mariari:


library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -334,68 +359,100 @@ instance MapSym PrivNameSpace where
...
+    -- dumb repeat code but idk how to handle to remove the tuple ☹
+    applyAndSetTop f =
+      t |> applyAndSet f (_topLevelMap, _topLevelMap)
+    applyAndSetCurrent f =
+      t |> applyAndSet f (_currentNameSpace . contents, _currentNameSpace . contents)
+
+applyAndSet f (l1, l2) t = do
+  ret <- f (t ^. l1)
+  pure ((\r -> set l2 r t) <$> ret)
+
```
mariari, line 423: Can someone look at this function and tell me if you can understand the function with the comments?

That alone should be enough but I would like some opinions

thealmarty: `modifySpaceImp` is very long. I think it'll be easier to read if you could split it into smaller functions. Also, could you give a bit more context in the comments?

mariari: I only see one place to split it up, the Nothing case  (without comments the function is only 15 lines long!),, as I just do some map lookups.

But I do agree, I think the main thing that needs to be commented is `recurseImp`, since we just dispatch onto that.

Also I should comment on that we just update the map given to `recurseImp`.

However, the important part is, do the comments properly give you a feeling of what paths things take given what the current name is and the symbol name

(that is what the `currentName : Foo.Bar, symbol: Something.Else` tries to show)

library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -111,55 +128,64 @@ topList T {topLevelMap} = HashMap.toList topLevelMap
...
-              Just __ -> Lib.Left er
+    ret <- addPathWithValue newNameSpace CurrentNameSpace t
+    case ret of
+      -- In this path the CurrentNameSpace is already added
+      Lib.Right t -> do
+        empty <- atomically emptyRecord
+        -- check if we have to qualify Name
+        pure $ case t !? newNameSpace of
+          Just (Current _) ->
+            Lib.Right (addCurrentName t empty qualifyName)
```
cwgoes, line 160: duplicate expression with two lines below, you can simplify

library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -111,55 +128,64 @@ topList T {topLevelMap} = HashMap.toList topLevelMap
...
-                  case addGlobal' t !? newNameSpace of
-                    Just (Outside (Record def _)) ->
-                      Lib.Right (addCurrent (addCurrentName t def newNameSpace))
-                    _ -> error "doesn't happen"
-                | otherwise ->
-                  Lib.Right (addCurrent (addCurrentName t def newNameSpace))
-              Nothing -> Lib.Left er
-              Just __ -> Lib.Left er
+    ret <- addPathWithValue newNameSpace CurrentNameSpace t
+    case ret of
```
cwgoes, line 153: this needs way more comments, reviewing this I am pretty confused as to what all the variables are

try adding a comment for each branch of each case switch, to start

library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -334,68 +359,106 @@ instance MapSym PrivNameSpace where
...
     Just (NameSpace.Priv _) ->
-      updateCurr t . unPriv <$> recurse f (s :| ymbol) (Priv currentNameSpace)
+      applyAndSetCurrent (fmap (fmap unPriv) . recurseImp f symbol . Priv)
+    -- The first part of the name is not in the map, maybe they fully qualified the name
     Nothing ->
-      case NameSymbol.takePrefixOf currentName (removeTopName (s :| ymbol)) of
-        Just subPath ->
-          updateCurr t <$> recurse f subPath currentNameSpace
+      case NameSymbol.takePrefixOf (t ^. _currentName) (removeTopName symbol) of
+        -- currentName : Foo.Bar, symbol: TopLevel.Foo.Bar.baz
```
cwgoes, line 389: this is an example of much better comments :) !

library/Core/src/Juvix/Core/Common/Context/Types.hs
```diff
@@ -58,15 +56,73 @@ data Definition term ty sumRep
   | -- Signifies that this path is the current module, and that
     -- we should search the currentNameSpace from here
     CurrentNameSpace
-  deriving (Show, Generic, Eq, Data)
+  deriving (Show, Generic, Eq)
+
+data Record term ty sumRep
+  = Rec
+      { recordContents :: NameSpace.T (Definition term ty sumRep),
+        -- Maybe as I'm not sure what to put here for now
```
cwgoes, line 64: explicate, why?

cwgoes (commented):
Thanks; needs some more comments in a few places before I can review thoroughly.

library/Core/src/Juvix/Core/Common/Context/Types.hs
mariari: As modules should always have types, however currently we don't know our module typing yet, so I just leave it as maybe typed for now... todo with more core updates

library/Core/src/Juvix/Core/Common/Context.hs
cwgoes: thank you, the new comments are much better

cwgoes (commented):
Thank you, the comments are helpful. At some point I think we should discuss whether or not there are alternative architectures that accomplish the same semantics without all of this complex case-switching and state manipulation logic, but for now this is fine.

mariari:
> Thank you, the comments are helpful. At some point I think we should discuss whether or not there are alternative architectures that accomplish the same semantics without all of this complex case-switching and state manipulation logic, but for now this is fine.

Feel free to give suggestions! A lot of the reference stuff is there so I can keep track of things properly, sort of hard to do things without a lot of information on hand.

I'm always open to changing the context to have/be:
1) less memroy consumption
2) more efficinet
3) easier to implement

library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -35,24 +37,76 @@ nameSymbolFromSymbol = NameSymbol.fromSymbol
...
+  emptyQualificationMap <- STM.new
+  pure
+    Rec
+      { recordContents = NameSpace.empty,
+        recordOpenList = [],
+        recordQualifiedMap = emptyQualificationMap,
+        recordMTy = Nothing
+      }
+
+data AmbigiousDef = AmbigiousDef NameSymbol.T
```
andy-morris, line 73: there's a typo here (ambig**i**ous instead of ambig**u**ous)

library/Core/src/Juvix/Core/Common/Context/RecGroups.hs
```diff
@@ -0,0 +1,32 @@
+-- | Calculate mutually-recursive groups of definitions.
+module Juvix.Core.Common.Context.RecGroups
```
andy-morris, line 2: Git seems to have got confused about this module moving to `J.C.C.Context.Traverse`. This file (and `…Types`) should probably go

library/Core/test/Common/Context.hs
```diff
@@ -1,15 +1,22 @@
...
+import qualified System.IO.Unsafe as Unsafe
 import qualified Test.Tasty as T
 import qualified Test.Tasty.HUnit as T
 
 foo :: Context.T Int Int Int
-foo = Context.empty ("Foo" :| ["Bar", "Baz"])
+foo = Unsafe.unsafePerformIO (Context.empty ("Foo" :| ["Bar", "Baz"]))
+
+unsafeEmpty :: NameSymbol.T -> Context.T term ty sumRep
+unsafeEmpty x = Unsafe.unsafePerformIO (Context.empty x)
```
andy-morris, line 19: according to the docs, you can't use `atomically` inside the dynamic scope of an `unsafePerformIO`. i just tried it in ghci and got a `NestedAtomically`:

```
[1]» :m Control.Concurrent.STM System.IO.Unsafe 
[2]» let x = unsafePerformIO $ atomically $ newTVar 0
x :: Num a => TVar a
[3]» atomically $ writeTVar x 14
*** Exception: Control.Concurrent.STM.atomically was nested
```

so it's strange that this seems to work?

library/Translate/src/Juvix/FrontendContextualise.hs
```diff
@@ -31,42 +31,47 @@ data Error
...
+    Left err -> pure $ Left (PathErr err)
+    Right (context, openList) -> do
+      trans1 <- Module.transformContext context openList
+      case trans1 of
+        Left err -> pure $ Left (ModuleErr err)
+        Right xs -> do
+          infixed <- Infix.transformContext xs
+          case infixed of
+            Left err -> pure $ Left (InfixErr err)
+            Right xs -> pure $ Right xs
```
andy-morris, line 51: is it possible to use `ExceptT` here too?

andy-morris (changes requested):
a couple of comments but other than those it looks good to me!

library/Core/test/Common/Context.hs
mariari: That is odd... Maybe this is differnet if the values are compiled in the file, thus lifting it of some kind of dynamic scoping!?

library/Translate/src/Juvix/FrontendContextualise.hs
mariari: It would be nice, I think I've opted out of that due to the error being differnet types all the way down

mariari:
Most of the issues Andy have brought up have been incorporated into the branch

cwgoes (approved):
tACK (tested locally)

PR #594, opened 2020-12-18
Cleanup of extra files being removed

mariari:
Followup of #583


thealmarty (approved):
Thanks for removing z3 as well.

cwgoes (approved):
ACK - thanks!

PR #595, opened 2020-12-18
Update README.md

thealmarty:
- remove Z3 requirement from readme

I'm also wondering if we can provide our docker images for users. I can list them on the readme, for each public release.


cwgoes (approved):
ACK

Let's worry about Docker images later

PR #596, opened 2020-12-18
updated the primitve types to have singuatres

mariari:


cwgoes (approved):
ACK, looks fine to me

PR #597, opened 2020-12-21
fixed a bug in grabSimilar which discarded values which aren't the same

mariari:


andy-morris:
looks good to me!

andy-morris (approved):

cwgoes (approved):
ACK

PR #598, opened 2020-12-22
Fix some typos in michelson parameterisation

andy-morris:
- `bytes` was called `string`
- `if` is a reserved word so it is renamed to `if_builtin` like in `Michelson.ju`


andy-morris:
actually since this is 2 lines it's probably not worth separating from #587 sorry

PR #609, opened 2020-12-30
Mariari/infix symbol paser fix

mariari:


cwgoes (approved):
utACK, we should write a test later

Issue #617, opened 2021-01-04
Michelson: Better Primitive types


mariari:
Currently we have 

```haskell
data RawPrimVal
  = Constant (M.Value' Op)
  | Inst (Instr.InstrAbstract Op)
  | AddN
  | AddI
  | AddTimeStamp
  | AddMutez
  | NegN
  | NegI
  | SubN
  | SubI
  | SubMutez
  | SubTimeStamp
  | MulI
  | MulN
  | MulMutez
  | EDivI
  | EDivN
  | EDivMutez
  | OrB
  | ORI
  | AndI

...
```
This gets split up into 

``` haskell
primToArgs Types.Nil ty = Instructions.nil <$> typeToPrimType ty
primToArgs Types.EmptyB ty = ....
```

and 

```haskell
instructionOf :: Types.RawPrimVal -> Instr.ExpandedOp
instructionOf x =
  case x of
    Types.AddN -> Instructions.add
    Types.AddI -> Instructions.add
    Types.AddTimeStamp -> Instructions.add
    Types.AddMutez -> Instructions.add
```
depending if it's a function or a constant value.

Thus we should split up the type to have better exhaustion checks

```haskell
data RawPrimVal
  = Constnat 
  | Inst
  | Value
  | Function ...
```

Issue #613, opened 2020-12-30
Support nil with type in the Michelson backend


cwgoes:
Need a version in primitive operations, e.g. for `NIL` which takes a type.

mariari:
Note: there are more operations like nil that I need to bake in.

PR #618, opened 2021-01-04
Adding Nil and other primtives which take types to the Michelson Backend

mariari:
Deals with #613 

Follow up pr will do #617

### DIFF
--- a/library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
+++ b/library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
@@ -748,7 +748,7 @@ constructPrim ::
   (Env.Stack m, Env.Count m, Env.Error m) => Types.RawPrimVal -> Types.Type -> m Env.Expanded
 constructPrim prim ty
   | isNonFunctionPrim prim = do
-    prim <- primToArgs prim ty
+    prim <- Env.Expanded <$> primToArgs prim ty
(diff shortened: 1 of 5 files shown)


library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -193,6 +193,42 @@ nameSymb :: Env.Reduction m => NameSymbol.T -> Types.RawTerm -> m Env.Expanded
 nameSymb symb f@(Types.Ann usage _ _) =
   inst f <* modify @"stack" (VStack.nameTop symb usage)
 
+unboxSingleTypeErr :: Untyped.T -> Untyped.T
```
cwgoes, line 196: Why do these functions end in `Err`? Because we use `error` ? Should it be in a monad?

cwgoes (approved):
ACK although it would be nice not to use `error`

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: yes if you call them wrong they error, but we should never get to that branch, #617 covers why I do this

cwgoes: yup, makes sense, thanks

PR #621, opened 2021-01-05
Test for the new constant values in Michelson and constant fixups

mariari:


library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -444,7 +444,23 @@ onPairGen1 op f =
...
+    ( \instr1 -> do
+        addExpanded instr1
+        case val instr1 of
+          Env.Constant _ -> pure Env.Nop
+          _ -> pure (val instr1)
+    )
+
+--------------------------------------------------------------------------------
+-- On Argument Functions
+--------------------------------------------------------------------------------
```
mariari, line 459: Please review this entire section

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -58,6 +58,8 @@ inst (Types.Ann _usage ty t) =
   case t of
     Ann.Var symbol -> var symbol
     Ann.AppM fun a -> appM fun a
+    Ann.UnitM -> undefined
```
cwgoes, line 61: this should probably not be undefined? or will this never get called? at least `error` if so?

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+
+------------------------------------------------------------
+-- Generator functions
+------------------------------------------------------------
+
+onTwoArgsGen :: OnTerm2Gen m Protect Env.Expanded
+onTwoArgsGen applyOperation typ arguments = do
   -- last argument evaled first
-  v <- traverse (protect . (inst >=> promoteTopStack)) (reverse instrs)
+  v <- traverse (protect . (inst >=> promoteTopStack)) (reverse arguments)
```
cwgoes, line 471: we could abstract this, there's code in common between `onTwoArgsGen` and `onOneArgGen`

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+onOneArgsNoConst ::
+  Env.Reduction m => Instr.ExpandedOp -> Types.Type -> [Types.RawTerm] -> m Env.Expanded
+onOneArgsNoConst op =
+  onOneArgGen (noOneConstantCase op)
+
+onTwoArgs :: OnTerm2 m (V.Value' Types.Op) Env.Expanded
+onTwoArgs op f =
+  onTwoArgsGen
+    ( \instr2 instr1 -> do
+        -- May be the wrong order?
```
cwgoes, line 516: why does it matter? we're just calling `allConstants`, no?

cwgoes: I don't really see why we need to make a temporary list here at all tbh

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+  -- add the op
+  addInstr op
+  -- return a nop to signal an already added output
+  pure Env.Nop
+
+noOneConstantCase ::
+  Env.Ops m => Instr.ExpandedOp -> Protect -> m Env.Expanded
+noOneConstantCase op instr1 = do
+  addExpanded instr1
+  -- copyAndDrop 1
```
cwgoes, line 559: what does this mean?

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+    )
+
+------------------------------------------------------------
+-- Helper predicate/abstractions for the on functions
+------------------------------------------------------------
+
+noTwoConstantCase ::
+  Env.Ops m => Instr.ExpandedOp -> Protect -> Protect -> m Env.Expanded
+noTwoConstantCase op instr2 instr1 = do
+  -- May be the wrong order?
```
cwgoes, line 546: this will result in the stack `res1 : res2 : ...` after execution presumably, is that what we want?

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+  let instrs = [instr2, instr1]
+  -- add the protected instructions
+  traverse_ addExpanded instrs
+  -- add the op
+  addInstr op
```
cwgoes, line 551: does this appropriately alter the virtual stack?

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -423,71 +447,124 @@ onPairGen1 op f =
...
+  -- return a nop to signal an already added output
+  pure Env.Nop
+
+noOneConstantCase ::
```
cwgoes, line 555: why do we need `noTwoConstantCase` and `noOneConstantCase`, seems like they could be one function?

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -745,12 +822,18 @@ valToBool _ = error "called valToBool on a non Michelson Bool"
...
+  -- TODO ∷ determine if this should be a constant, they should all be in
+  -- vstack...
+  -- also once we do this, we can remove the StackOps effect of this function!
   | isNonFunctionPrim prim = do
-    prim <- Env.Expanded <$> primToArgs prim ty
-    consVal prim ty
-    pure prim
+    primToArgs prim ty
+    -- cons to show we have it on the stack
+    consVal Env.Nop ty
```
cwgoes, line 834: honestly I just find this pattern confusing (having to use `Nop` like this when we're not actually emitting it as an instr), I guess it's a topic for a later refactor though

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
```diff
@@ -173,9 +177,9 @@ var symb = do
       pure (Env.Curr lamPartial)
     Just (VStack.Position (VStack.Usage usage _saved) index)
       | usage == one ->
-        Env.Expanded <$> moveToFront index
+        Env.Nop <$ moveToFront index
```
cwgoes, line 180: I'm confused as to why this change is in this PR, was it just wrong? What does this have to do with the other changes?

cwgoes (changes requested):
The test-case looks correct; confused by some of the code changes.

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: For abstraction purposes we did this in the past, as we send it into the all check function... basically to scale it up to N arguments with the same pattern... since haskell's type system makes it hard to abstract

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Comment was from old code, that I propagated, was something we did that was wrong, but still may serve to be helpful, I can delete if you wish

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Yes, that is what we want, at least that is what we've had in the past.

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: You are correct, abstracting it out, leads to another optimization!

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: This makes 0 difference, in the past you've mentioned adding `Nop` as you found functions returning values and doing stack operations confusion. This would be another consistent place to add it

mariari: Like with many changes, I found this while reasoning through the backend again.

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Yeah, I could put an expanded from the `primToArgs`, however since this pushes to the stack, we just have a nop for anything which likes to process arguments (it's already been processed)

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: no, if you look at the signature
```haskell
noOneConstantCase ::
  Env.Ops m => Instr.ExpandedOp -> Protect -> m Env.Expanded
```
they only touch the ops, the stack effect happens in the gen function

```haskell
onTwoArgsGen :: OnTerm2Gen m Protect Env.Expanded
onTwoArgsGen applyOperation typ arguments = do
  -- last argument evaled first
  v <- traverse (protect . (inst >=> promoteTopStack)) (reverse arguments)
  case v of
    instr2 : instr1 : _ -> do
      -- Get the result of the operation on the arguments
      res <- applyOperation instr2 instr1
      -- drop the two arguments on the stack
      modify @"stack" (VStack.drop 2)
      -- add the result to the stack
      consVal res typ
      -- return the operation
      pure res
    _ -> throw @"compilationError" Types.NotEnoughArguments

```

namely
```
      modify @"stack" (VStack.drop 2)
```
is what changes the vstack before we say the top is res

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: yeah I could probably with this, didn't think of it before

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: Sure I"ll change it to error, I was going to fix it in this pr anyways, since it's just a pair and unit call

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: let's either delete it or clarify

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: right OK

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: it's fine for now

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
cwgoes: ah ah I see, yes this is better, thanks!

library/Backends/Michelson/src/Juvix/Backends/Michelson/DSL/InstructionsEff.hs
mariari: both converted, just do a quick look over at the pair and unit translations

cwgoes (approved):
LGTM.

PR #631, opened 2021-01-12
added a proper handleInfix function that properly preserves infix syms

### LINKED ISSUE #608: NameSymbol Noms dots on infix expression.

mariari:
Tests to come.

With this, we satisfy the other way of the identiy

```haskell
id :: NameSymbol.T -> NameSymbol.T
id = NameSymbol.fromSymbol . NameSymbol.toSymbol
``` 

Fixes #608

### DIFF
--- a/doc/Code/Library.org
+++ b/doc/Code/Library.org
@@ -886,20 +886,17 @@ Tests that weak works as expected
   + [[Frontend/Types]]
   + [[Library]]
   + [[NameSymbol]]
-***** Lexer
-- _Relies on_
-  + [[Library]]
 ***** Parser <<Frontend/Parser>>
 - The front end parser for the Juvix Programming language
 - Parsers with S at the end, eat the spaces at the end of the parse
 - Parsers with SN at the end, eats the spaces and new lines at the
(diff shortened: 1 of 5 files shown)


mariari:
Currently

```haskell
 T..&&. eq dotStart
```
is broken in the test, seeing why it's broken now

mariari:
Can't do an idr test as 
```haskell
λ> idR ("Foo" :| ["Bar","(_Foo_-_)","Bar","(Foo)","(.).-..>.."])
"Foo" :| ["Bar","(_Foo_-_)","Bar","(Foo)","(",")","-..>.."]
```
since we don't get terms in ()'s I don't think an idr function works.

This is true for other such edge cases where the . is imporperly placed.

The PR is good as is

cwgoes:
Can we do a test of a previously broken case, though?

mariari:
> Can we do a test of a previously broken case, though? At least the expected output (which was wrong before).

will do

mariari:
Test case added

cwgoes (approved):
ACK

PR #633, opened 2021-01-13
Add variables as arguments for primitives

### LINKED ISSUE #630: Turn primVal into a neutral at least for now so we can apply primitives to variables
This will change `vapp` to fix the following error:

Lam' (Elim' (App' (App' (Free' (Global ("TopLevel" :| ["Identity","makePair"])) ()) (Elim' (Free' (Global ("TopLevel" :| ["Identity","nil"])) ()) ()) ()) (Elim' (App' (Free' (Global ("TopLevel" :| ["Identity","car"])) ()) (Elim' (Bound' 0 ()) ()) ()) ()) ()) ()) ()
TypecheckerError CannotApply {fun = VPrim' (Return {retType = Application Pair (PrimTy (Type TInt :) :| [PrimTy (Type TInt :)]) :| [PrimTy (Type TInt :)], retTerm = Inst CAR}) (), arg = VNeutral' (NBound' 0 ()) (), paramErr = NoApplyError}

andy-morris:
Fixes #630.

Concretely, this PR adds a new datatype for primitive arguments, which can contain variables. This seems to be a cleaner solution than making `Prim` a neutral?

Bear in mind that this patch means that `Cont {args, numLeft=0}` is now a well-formed `Return`, as long as at least one of `args` is a variable. (Previously `numLeft` had to be non-zero since a fully-saturated application was always reducible.)

- [ ] Add `HasSubst` for `Application` stuff, and use it in terms
  - [ ] (Remember to retry reduction after substituting!)
- [x] `ErasedAnn` doesn't use de Bruijn. Address that....... somehow

### DIFF
--- a/library/Core/src/Juvix/Core/Application.hs
+++ b/library/Core/src/Juvix/Core/Application.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DeriveTraversable, UndecidableInstances #-}
 -- |
 -- Types to support partial application and polymorphic primitives.
 module Juvix.Core.Application where
@@ -6,19 +6,18 @@ module Juvix.Core.Application where
 import Juvix.Library
 import qualified Juvix.Library.Usage as Usage
 import qualified Juvix.Core.IR.Types as IR
-import Data.Bifunctor
 import Data.Bifoldable
 import Data.Bitraversable
 
 -- |
 -- A primitive along with its type, and possibly some arguments.
-data Return ty term
+data Return' ext ty term
   = -- | Partially applied primitive holding the arguments already given
     Cont
       { -- | head of application
         fun :: Take ty term,
         -- | arguments
-        args :: [Arg ty term],
+        args :: [Arg' ext ty term],
         -- | number of arguments still expected
         numLeft :: Natural
       }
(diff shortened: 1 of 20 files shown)


library/Core/src/Juvix/Core/Application.hs
```diff
@@ -1,19 +1,25 @@
...
-data Return ty term
+data Return' ext ty term
   = -- | Partially applied primitive holding the arguments already given
     Cont
       { -- | head of application
         fun :: Take ty term,
         -- | arguments
-        args :: [Take ty term],
+        args :: [Arg' ext ty term],
         -- | number of arguments still expected
```
mariari, line 23: What do these cascades of new types here give use exactly?

library/Core/src/Juvix/Core/Application.hs
```diff
@@ -22,8 +28,70 @@ data Return ty term
...
+
+-- | The representation of variables used in IR.Term' ext
+type family ParamVar ext :: Type
+
+data DeBruijn
+  = BoundVar IR.BoundVar
+  | FreeVar IR.GlobalName
   deriving (Show, Eq, Generic)
 
+type instance ParamVar IR.NoExt = DeBruijn
```
mariari, line 66: How do type instances like these work?

library/Core/src/Juvix/Core/Erasure/Algorithm.hs
```diff
@@ -9,24 +9,28 @@ import Juvix.Library hiding (empty)
 import qualified Juvix.Library.NameSymbol as NameSymbol
 import qualified Juvix.Library.Usage as Usage
 
-type ErasureM primTy primVal m =
+type ErasureM primTy1 primTy2 primVal1 primVal2 m =
```
mariari, line 12: I'm guesing we take 4 aprams as we have to track before and after types?

cwgoes: I guess? Why are before and after types different (now)?

cwgoes (approved):
ACK although this would benefit from some comments

library/Core/src/Juvix/Core/Erasure/Algorithm.hs
andy-morris: The base primitives themselves (Michelson instructions, or whatever) don't change, but the variable type in partial applications does, so `primVal` goes from e.g. `Return' IR.NoExt primTy primVal` (where arguments can contain dB indices) to `Return' ErasedAnn.T primTy primVal` (where they can't).

PR #636, opened 2021-01-16
Fixes RecGroup module resolution issues.

### LINKED ISSUE #614: WrongSigType issues with modules
For example, when including `stdlib/Prelude.ju` with `test/examples/demo/identity.ju`, I get:

```bash
WrongSigType ("TopLevel" :| ["Prelude","->"]) Nothing
```

whereas putting the definitions directly in `test/examples/demo/identity.ju` works fine.

mariari:
closes #614

### DIFF
--- a/library/Core/src/Juvix/Core/Common/Context.hs
+++ b/library/Core/src/Juvix/Core/Common/Context.hs
@@ -217,10 +217,7 @@ removeTop sym t@T {topLevelMap} =
 ------------------------------------------------------------
 
 changeCurrentModuleWith ::
-  Cont (Definition term ty sumRep) ->
-  NameSpace.T (Definition term ty sumRep) ->
-  NonEmpty Symbol ->
-  Cont (Definition term ty sumRep)
+  T ter ty sr -> NameSpace.T (Definition ter ty sr) -> NameSymbol.T -> T ter ty sr
 changeCurrentModuleWith t startingContents newCurrName =
   let queued = queueCurrentModuleBackIn t
    in t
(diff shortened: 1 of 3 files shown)


cwgoes:
Does this fix https://github.com/metastatedev/juvix/issues/614? If so, it should be marked as doing so, and we should add a testcase or two.

mariari:
> Does this fix #614? If so, it should be marked as doing so, and we should add a testcase or two.

Yes this does, wrote closes so it will close #614 automatically


for note for future testing
```haskell
λ> ordered
"switching modules to:"
"properly hit name: " :| ["TopLevel","Helper"]
"Did we fail to switch?we did switch"
"Helper" :| []
"TopLevel" :| ["Helper"]
"we should be in module:"
"starting from module:"
"Identity" :| []
["TopLevel" :| ["Helper","ty_"],"TopLevel" :| ["Helper","ty"],"TopLevel" :| ["Helper","pair"],"TopLevel" :| ["Helper","operation"],"TopLevel" :| ["Helper","omega"],"TopLevel" :| ["Helper","int"],"TopLevel" :| ["Helper","arr"],"TopLevel" :| ["Helper","arrow"],"TopLevel" :| ["Helper",":"],"TopLevel" :| ["Helper","->"],"TopLevel" :| ["Helper","list"],"TopLevel" :| ["Helper","nil"],"TopLevel" :| ["Helper","pairTy"],"TopLevel" :| ["Helper","car"],"TopLevel" :| ["Helper","makePair"],"TopLevel" :| ["Identity","main"]]
```

This was done using the `.ju` files in in #586 so tests will be done shortly after

PR #641, opened 2021-01-22
Readme amendments

Asamartino:


README.md
```diff
@@ -70,6 +70,8 @@ After cloning Juvix into a local directory, go into the local Juvix directory, a
 ```bash
 make
 ```
+Expect the installation to take some time (~ 40 minutes).
```
thealmarty, line 73: the time will depend on the machine's specs. Maybe mention you machine's spec? 
```suggestion
Expect the installation to take some time (~ 40 minutes on a ?? machine).  
```

Asamartino: ok so let's not give any time indication, but just mention that it might take some time.

Issue #589, opened 2020-11-30
More frontend test-cases for parsing, typechecking, and Michelson compilation


cwgoes:
To be added to the `test/examples/demo` folder, complementing the existing test examples and a few more in https://github.com/metastatedev/juvix/pull/586.

1. More complex conditionals (multi-clause `if`)
1. Integer/natural comparison (maybe)
1. `FAILWITH` used in an example
1. Very simple proofs (we may need a `Refl`)
1. An example with a Michelson list
1. An example with a Michelson map

cwgoes:
We have enough for now.

PR #647, opened 2021-01-26
Some more example contracts

mariari:
Covers #589


cwgoes (approved):
utACK, this is fine for now as long as everything compiles

PR #649, opened 2021-01-26
Test Group Ordering Fix and tests

mariari:
This was an issue mentioned in  #614 in which #636 was meant to fix, however no test cases were given

Now that we have a test case we found a bug and have hopefully fixed it for good


cwgoes (approved):
utACK

PR #655, opened 2021-01-28
Module Opens caused weird bugs in other modules

### LINKED ISSUE #654: Modules pick the wrong value
```haskell
open Prelude.Michelson

sig cons-pair : list operation -> int -> pair (list operation) int
let cons-pair = %Michelson.pair

sig main : pair int int -> pair (list operation) int
let main = \params -> cons-pair nil 0
```

Here `Michelson.cons-pair` is favored over the local `cons-pair`

mariari:
fixes #654 . Wrong orders were appened this should be fixed?

Exact tests to come!


mariari:
Test cases added

cwgoes (approved):
utACK

PR #656, opened 2021-01-29
HasSubstElim -> HasSubstTerm

andy-morris:
(The test failure is from `test-typecheck`, everything that used to pass still does)


cwgoes (approved):
utACK

PR #657, opened 2021-01-29
Michelson Backend extra param to all stdlibs functions

mariari:


andy-morris:
The definitions also need to bind the tyoe arguments, e.g.

```
sig none : (a : ty) -> either unit-t a
let none a = %Michelson.none
```

(I know it looks a bit strange, but it's because the primitives are polymorphic without actually taking the type arguments anywhere.) Sorry this wasn't clear before.

mariari:
> The definitions also need to bind the tyoe arguments, e.g.
> 
> ```
> sig none : (a : ty) -> either unit-t a
> let none a = %Michelson.none
> ```
> 
> (I know it looks a bit strange, but it's because the primitives are polymorphic without actually taking the type arguments anywhere, but Juvix itself still needs to pass the type info around.) Sorry this wasn't clear before.

Ahh thanks, just bind it and throw it away with _?

cwgoes (approved):
tACK; merging for now

PR #663, opened 2021-02-01
Get number of threads based on OS

Fraccaman:
Tried to compile juvix on OSX but `nproc` it not a valid command. 
Modified a little bit the `Makefile` in order to get the number of logical processors checking first the OS.


cwgoes (approved):
ACK, thanks!

PR #666, opened 2021-02-02
fix typo in makefile

andy-morris:
It seems like the space after `ifeq` is mandatory, otherwise I get an error:

```
Makefile:7: extraneous text after 'else' directive
```


Fraccaman:
Yes, a bad typo, you are right.

fdedden:
It might not be that simple: https://stackoverflow.com/a/39343918.

I highly doubt anyone would compile this under Windows, but it seems like sensible syntax to me. Weird that it works correctly outside of Windows.

cwgoes (approved):
ACK

PR #667, opened 2021-02-03
Tests for evaluator, etc

andy-morris:
- [x] `weak` family
- [x] `eval`
- [x] `toLambda`
- [ ] `subst`
- [ ] `substTerm`?
- [ ] `substV`
- [ ] `substValue`?
- [ ] `vapp`
- [ ] `patSubst`
- [ ] `patSubstTerm`?
- [ ] `patSubstValue`?

(this is just all the main eval-related functions; the ones with "?" might not actually need separate tests)


andy-morris:
status update on this: it's stalled at the moment because I've been working on other stuff, but the tests that are already there don't lead us towards the eval bug, so we're waiting to merge until it's more complete

PR #668, opened 2021-02-03
Integration of type checking of data type and function declarations

thealmarty:


library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -122,38 +125,101 @@ type GlobalAllWith (c :: * -> Constraint) ty ext primTy primVal =
     c primVal,
     TermAll c ext primTy primVal,
     ElimAll c ext primTy primVal,
+    ValueAll c ext primTy primVal,
+    NeutralAll c ext primTy primVal,
     PatternAll c ext primTy primVal
   )
 
-data DatatypeWith ty primTy primVal
+data DatatypeWith ty ext primTy primVal
```
cwgoes, line 133: let's consider splitting this into `DatatypeWithPreTypecheck` and `DatatypeWithPostTypecheck` so we don't have the duplicate `dataArgs` / `nfDataArgs` fields?

cwgoes: (or figure out how to use the `ty` argument)

library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -122,38 +125,101 @@ type GlobalAllWith (c :: * -> Constraint) ty ext primTy primVal =
...
+deriving instance
+  GlobalAllWith NFData ty ext primTy primVal =>
+  NFData (DatatypeWith ty ext primTy primVal)
 
 type RawDatatype' ext = DatatypeWith (Term' ext)
 
 type Datatype' extV = DatatypeWith (Value' extV)
 
-data DataArgWith ty primTy primVal
+data RawArgType' ext primTy primVal
```
cwgoes, line 168: we can abstract this (diff Term vs Value)

library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -276,5 +338,56 @@ type GlobalsWith ty ext primTy primVal =
...
+        --   or that @clauseBody@ contains meta-variables;
+        --   these could be filled with recursive calls later!
+        --   @Just False@ means definitely no recursive call.
+        --   @Just True@ means definitely a recursive call.
+        clauseCatchall :: Bool,
+        clauseUnreachable :: Maybe Bool
+      }
+  deriving (Generic)
+
+type Signature ty ext primTy primVal = Map.Map Name (SigDef ty ext primTy primVal)
```
cwgoes, line 377: check if we have existing datatypes for this

library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -276,5 +338,56 @@ type GlobalsWith ty ext primTy primVal =
...
+
+-- Return type of all type-checking functions.
+-- state monad for global signature
+type TypeCheck ty ext primTy primVal a = StateT (Signature ty ext primTy primVal) IO a
+
+data SigDef ty ext primTy primVal
```
thealmarty, line 383: Merge with fromfrontend later.

thealmarty:
This PR aims to finish stage 1 of [this plan](https://gist.github.com/thealmarty/2e58f3949d1ff2bac926a2313c7d6a75).

library/Core/src/Juvix/Core/IR/Types/Base.hs
thealmarty: Andy and I discussed further, we will have a data type with `Value'` and a separate with with `Term'`.

library/Core/src/Juvix/Core/IR/Types/Base.hs
thealmarty: Checked with Andy, there isn't.

library/Core/src/Juvix/Core/IR/Types/Base.hs
thealmarty: Andy and I agree that the one from fromFrondEnd is not quite the same and so we'll keep a separate one.

thealmarty:
Per discussion, outstanding changes:
- [ ] (cont') taking out `ty` parameter in `datacon` and `datatype`
- [ ] make a `Value'` and a `Term'` `datatype`
- [ ] `Signature` is a map of `GlobalName` (not `Name`) and `SigDef`
- [ ] take out `IsParam` field of `DataArg` etc because it's no longer needed

thealmarty:
Fixups of `FromFrontend.hs` to follow (@andy-morris ):
- add back signatures of some transform functions
- look into `transformHR`: linter suggests that there are over 200,000 recursions in that function

thealmarty:
@andy-morris @cwgoes please follow up:
- Fixups of `FromFrontend.hs`: look into `transformTermHR`: "Pattern match checker exceeded (2000000) iterations in an equation for ‘transformTermHR’"
- lots of pattern matching in-exhaustive warnings.

app/Compile.hs
```diff
@@ -95,10 +95,10 @@ convGlobal g =
     GFunction (Function n u t cs) -> GFunction (Function n u (baseToReturn t) (map funClauseReturn cs))
     GAbstract (Abstract n u t) -> GAbstract (Abstract n u (baseToReturn t))
 
-funClauseReturn (FunClause patts term) = FunClause (map pattEval patts) (baseToReturn term)
+funClauseReturn (FunClause tel patts term rhs catchall unreachable) = FunClause undefined (map pattEval patts) (baseToReturn term) undefined undefined undefined -- TODO
 
 funClauseEval :: IR.FunClause' IR.NoExt Param.PrimTy Param.RawPrimVal -> IR.FunClause' IR.NoExt Param.PrimTy (TypedPrim Param.PrimTy Param.RawPrimVal)
-funClauseEval (FunClause patts term) = FunClause (map pattEval patts) (baseToReturn term)
+funClauseEval (FunClause tel patts term rhs catchall unreachable) = FunClause undefined (map pattEval patts) (baseToReturn term) undefined undefined undefined --TODO
```
andy-morris, line 101: Is it possible to fill in the `undefined`s yet, or is more work elsewhere needed first?

library/Core/src/Juvix/Core/IR/Typechecker.hs
```diff
@@ -1,434 +1,42 @@
...
+-- import Data.List.NonEmpty ((<|))
+-- import qualified Juvix.Core.Application as App
+-- import qualified Juvix.Core.IR.CheckTerm as IR
+-- import qualified Juvix.Core.IR.Evaluator as Eval
 import Juvix.Core.IR.Typechecker.Env as Env
 import Juvix.Core.IR.Typechecker.Types as Typed
 import qualified Juvix.Core.IR.Types as IR
 import qualified Juvix.Core.IR.Types.Base as IR
-import qualified Juvix.Core.Parameterisation as Param
+-- import qualified Juvix.Core.Parameterisation as Param
```
andy-morris, line 22: Can these commented imports be removed now?

library/Core/src/Juvix/Core/IR/Types/Base.hs
```diff
@@ -244,8 +362,10 @@ deriving instance
...
 
 data GlobalWith ty ext primTy primVal
-  = GDatatype (DatatypeWith ty primTy primVal)
-  | GDataCon (DataConWith ty primTy primVal)
+  = GDatatype (Datatype' ext primTy primVal)
+  | RawGDatatype (RawDatatype' ext primTy primVal)
+  | GDataCon (DataCon' ext primTy primVal)
+  | RawGDataCon (RawDataCon' ext primTy primVal)
   | GFunction (FunctionWith ty ext primTy primVal)
   | GAbstract (AbstractWith ty primTy primVal)
```
andy-morris, line 370: Would this work better as separate `Global` and `RawGlobal` (also containing `RawFunction` and `RawAbstract`, analogous to the data stuff) types and just removing the `ty` argument completely? It seems that at any given point you are only consuming/producing either all raw things, or all checked things. If this would work it would also fix the exhaustiveness warnings I think.

app/Compile.hs
thealmarty: Unfortunately (much!) more work elsewhere is needed first.

library/Core/src/Juvix/Core/IR/Types/Base.hs
thealmarty: I was wondering if you would want that. As I mentioned, I think removing `ty` completely would simplify things. I'll go ahead and remove it then.

library/Core/src/Juvix/Core/IR/Typechecker.hs
thealmarty: Yup. Done.

library/Core/src/Juvix/Core/IR/Types/Base.hs
thealmarty: actually could I roll back to [this](https://github.com/heliaxdev/juvix/pull/668/commits/8ed00d70720f41df06b6e51d3bd3b9684a860f1c) commit and merge to develop so that all other changes are in? Especially with #697 coming. Removing `ty` the rest of the way can come in a later PR.

andy-morris: Yeah that should be fine I think! Sorry I ended up having to touch the same code as you.

library/Core/src/Juvix/Core/IR/Types.hs
```diff
@@ -66,7 +61,7 @@ type Function = Function' NoExt NoExt
 
 type RawFunction = RawFunction' NoExt
 
-type FunClause = FunClause' NoExt
+type FunClause primTy primVal = FunClause' primTy primVal NoExt
```
andy-morris, line 64: same as above; I think `NoExt` should be the first argument?

src/Juvix/Core/FromFrontend.hs
```diff
@@ -416,7 +416,6 @@ transformNormalSig _ _ (Ctx.Record _) = pure Nothing -- TODO
 transformNormalSig q x (Ctx.TypeDeclar typ) = Just <$> transformTypeSig q x typ
 transformNormalSig _ _ (Ctx.Unknown sig) =
   throwFF $ UnknownUnsupported $ FE.signatureName <$> sig
-transformNormalSig _ _ Ctx.SumCon {} = pure Nothing
```
andy-morris, line 419: this case needs to be put back I think

src/Juvix/Core/FromFrontend.hs
```diff
@@ -519,7 +517,6 @@ transformNormalDef q x (Ctx.TypeDeclar dec) = transformType q x dec
 transformNormalDef _ _ (Ctx.Unknown _) = pure []
 transformNormalDef _ _ Ctx.CurrentNameSpace = pure []
 transformNormalDef _ _ (Ctx.Information {}) = pure []
-transformNormalDef _ _ Ctx.SumCon {} = pure []
```
andy-morris, line 522: this case needs to be put back I think

andy-morris (changes requested):
just a couple of tiny things but the rest looks good! thanks!

andy-morris (approved):

PR #673, opened 2021-02-07
docs: split Debian and Ubuntu install commands.

RichAyotte:
https://packages.debian.org/search?keywords=haskell-stack&searchon=names&suite=all&section=all


cwgoes (approved):
ACK, thanks!

Issue #675, opened 2021-02-08
There is no build-opt.


hgsipiere:
Hi, I tried to do the 'build-opt' as described for 'Building with optimisations' but there is no such choice. Should it be 'build-prod' instead?

fdedden:
You are correct, I will update the Makefile accordingly.

fdedden:
The README has been fixed, but it created another problem: #679 (at least for me).

PR #680, opened 2021-02-11
Renamed `make build-opt` -> `make build-prod` in README (#675)

fdedden:
`build-opt` was still used in the README as the way to build a production optimised
Juvix compiler. This has been matched to the Makefile's `build-prod`.


cwgoes (approved):
ACK

PR #682, opened 2021-02-11
Add `handler` and `via` to the frontend language and desugar it later

cyberglot:


cwgoes (changes requested):
Looks relatively straightforward modulo extensible-data questions, also needs testcases.

cyberglot:
@mariari I want to squash it into two or three commits, lmk if it's ok by you.

mariari:
> @mariari I want to squash it into two or three commits, lmk if it's ok by you.

Fine by me, though note that it gets squashed into master, so it'll only show as 1 commit there.

cwgoes (approved):
ACK, seems straightforward to me, should see a quick review from @mariari before merge

cyberglot:
Rejected in favour of #722.

Issue #671, opened 2021-02-05
Timeline of the Feature plans and implementation of the S-expressions 


mariari:
This is a follow up and plans of issue #634

### Phase 1
- Implementing the SEXP data structure
- Writing Traversal over this structure
### Phase 2
- Begin Frontend Syntax Data structure -> SEXP conversion
- Start of parallel pipeline that does the same order 
- Write some passes back to make sure the passes agree
### Phase 3
- Finish all such passes
- Deprecate old transition pipeline
    + Remove old code
- Re-order the pipeline
### Phase 4
- Get core on the new format
### Phase 5 Reserach 
From This phase onwards, it's more of a research phase in which we can take a few paths

#### Plan 1 Eval Apply
Since core already has to eval terms, it'd be quite nice to get a macro expansion sysetm going. This has a lot of practical benefits.
#####  Phase1:
- This phase is all about getting core Eval Apply
##### Phase 2: 
- Here we wil convert our passes into macros for our eval apply system to discharge

#### Plan 2 The Juvix Macro System
Plan 2 here requires plan 1 to a degree.

We will also need to update the pipeline to handle inner definitions before we should attempt this.

##### Phase 1:
- Figure out all spots of the code base which would break with inner definitions or what code relies on expansion to not be had.
- Write an issue and mention anyone who it may concern
##### Phase 2:
- Begin fixes and start the implementation

#### Plan 3 The Reader hacker to extend the Juvix Frontend Syntax
This is the most far out idea and requires a lot of research and thought. But there is likely a way we can hack our reader.

Look into CL and racket for inspiration

cwgoes:
Thank you for the write-up. In general the ordering of phases seems reasonable.

Did you have a chance to write a few more concrete examples of what macros would look like? That would be helpful for me at least in understanding how users will interact with this system.

mariari:
> Thank you for the write-up. In general the ordering of phases seems reasonable.
> 
> Did you have a chance to write a few more concrete examples of what macros would look like? That would be helpful for me at least in understanding how users will interact with this system.

I have made one just now!

```juvix
-- Note the :: here is more of a sexp than list match here
macro sig-forall (fun-name :: xs) =
  let all-syms [] acc =
    acc
  and all-syms ((x :: ys) :: xs) acc = do
     new-acc <- all-syms (x :: ys) acc
     all-syms xs new-acc
  and all-syms (x :: xs) acc = do
    check <- in-context x
[...]

Issue #672, opened 2021-02-05
Smplifcation on the passes down


mariari:
Currently the Translation passes require 2 or 3 maps to go down. On whole program compilation this is just the transition from the old context type to the new one. However for the interpreter we currently need 3. One for the current context, and 2 for the transition functionality. Issue #634 Lays out a proposed solution.

Once we have SEXPs in some form down our pipeline we could instead do a much simpler task.

Since our translation steps now act more as macros, we could instead just run it over the new forms that come in, checking the context for various bits of information we need to fully do this.

Once we are done we can then commit it to the context as the proper fully fledged form.

This will likely happen in Phase 2 of #671

cwgoes:
Superseded by idealised pipeline discussions.

PR #685, opened 2021-02-12
Sexpression Structure of Juvix

mariari:
This Finishes Phase 1 of #671.

Further the tests given show the signs of doing part of phase2. (Namely basic transformations testing if the structures I've written are correct!).

### DIFF
--- a/library/StandardLibrary/src/Juvix/Library/LineNum.hs
+++ b/library/StandardLibrary/src/Juvix/Library/LineNum.hs
@@ -1,0 +1,11 @@
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE TemplateHaskell #-}
+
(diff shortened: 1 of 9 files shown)


mariari:
This PR covers all of phase 1 of #671, along with all of the passes for phase 2 of #672.

library/StandardLibrary/src/Juvix/Library/LineNum.hs
```diff
@@ -0,0 +1,11 @@
+{-# LANGUAGE FunctionalDependencies #-}
```
cwgoes, line 1: Should we just put this in `package.yaml` ?

cwgoes (approved):
ACK, seems relatively straightforward

andy-morris (approved):

PR #691, opened 2021-02-16
Frontend ⟶ Sexpression conversion and testing pass

mariari:
Finishes the Conversion of phase 2 in #671

### DIFF
--- a/doc/Architecture/Sexp.org
+++ b/doc/Architecture/Sexp.org
@@ -1,5 +1,6 @@
 * Generation
-** Let Top Level
+** Top Level
+*** Let
 1. basic
    - input
      #+begin_src haskell
@@ -14,7 +15,7 @@
            ((:infix == x 2) (:infix + 3 (:infix + 4 x)))
            (else            (:infix + x 2))))
      #+end_src
-** Top Level Sig
+*** Sig 
 1. basic
    - input
      #+begin_src haskell
@@ -40,7 +41,7 @@
      #+end_src
      + However for now I think it's fine... Though this is a place for
        improvement for the future
-** Top Level Types
+*** Types
 1. record
    - input
      #+begin_src haskell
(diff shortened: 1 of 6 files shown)


doc/Architecture/Sexp.org
```diff
@@ -0,0 +1,399 @@
...
+   - output
+     #+begin_src lisp
+       (:defun add (x)
+         (:cond ((:infix == x 3) 1)
+                ((:infix == x 4) 5)
+                (else            0)))
+     #+end_src
+* Notes
+** Design Principles
+Overall the goal of the S-expression incantation is twofold.
```
cwgoes, line 307: incantation? Odd word choice, do you just mean the syntax? Or the system as a whole?

doc/Architecture/Sexp.org
```diff
@@ -0,0 +1,399 @@
...
+
+1) An unambiguous syntax that is faithful to the original ML syntax
```
cwgoes, line 309: What does it mean to be faithful? Is there a bijection (ideally)?

cwgoes (changes requested):
Concept ACK, code looks fine, a few questions on the docs

Presumably we should have tests for `transTopLevel` and `transExpr`? Or do you intend to test those in another way or separate PR?

doc/Architecture/Sexp.org
mariari: Yes we preserve properties. Though when I wrote this i was using more colloquial meaning of the word

doc/Architecture/Sexp.org
mariari: As in the expression of this form. Of the reasoning of the form of the syntax, yes.

Part of the reason for the word choice is to pay homage to the imagery given by the dragon book [1] and the legendary SICP course [2]. 

[1] https://en.wikipedia.org/wiki/Compilers:_Principles,_Techniques,_and_Tools
[2] https://www.youtube.com/watch?v=-J_xL4IGhJA&list=PLE18841CABEA24090

mariari:
> Concept ACK, code looks fine, a few questions on the docs
> 
> Presumably we should have tests for `transTopLevel` and `transExpr`? Or do you intend to test those in another way or separate PR?

Yes, as this PR doesn't include the parser, I've put that into a seperate PR... writing tests without the parser is a pain

doc/Architecture/Sexp.org
cwgoes: OK, but we should define it or link to those things, it's not particularly standard usage

doc/Architecture/Sexp.org
cwgoes: Can you be more specific about what properties are preserved?

cwgoes (approved):
ACK

PR #695, opened 2021-02-19
started work to implement the pattern match case in fronFrontend

mariari:


cwgoes:
Per @mariari closing.

PR #698, opened 2021-02-23
Sexp Conversion Tests

mariari:
This PR tests the conversion of #691 to make sure everything stays good


cwgoes (approved):
ACK, but do we have / should we have tests for complex nested pattern matching?

PR #707, opened 2021-03-03
Fix underflow when inlining functions of arity 1

### LINKED ISSUE #702: Arithmetic underflow in toLambda
While playing around with Juvix today I stumbled across a `juvix: arithmetic underflow` error message. For example the following program:
```
mod ArithUnderflow where

open Prelude
open Michelson

sig make-pair : list operation -> int -> pair (list operation) int
let make-pair = %Michelson.pair

sig nil : list operation
let nil = %Michelson.nil

-- Inlining this function is problematic:
sig int-id : int -> int
let int-id x = x

sig main : pair int int -> pair (list operation) int
let main = \ps -> make-pair nil (int-id 10)
```
It seems that the error is happening when inlining global functions of arity 1, such as `int-id` in the above example. There is a simple fix here: https://github.com/heliaxdev/juvix/compare/develop...andreaslyn:andreaslyn/arity1-inline-fix

andy-morris:
Thanks @andreaslyn!

Fixes #702.

### DIFF
--- a/library/Core/src/Juvix/Core/IR/Evaluator.hs
+++ b/library/Core/src/Juvix/Core/IR/Evaluator.hs
@@ -192,7 +192,7 @@ toLambda (IR.GFunction (IR.Function {funUsage, funType, funClauses}))
   | IR.FunClause _clauseTel pats rhs _rhsTy _catchall _unreachable :| [] <- funClauses = do
     patVars <- traverse singleVar pats
     let len = fromIntegral $ length patVars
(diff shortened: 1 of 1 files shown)


andy-morris:
I'm not sure why the CI isn't working on this PR, but the tests pass locally (unsurprisingly).

@Fraccaman this branch isn't even listed in the CI, do you know why that might be?

thealmarty (approved):

cwgoes (approved):
ACK, thank you!

PR #710, opened 2021-03-04
Added pull_request event

Fraccaman:
This should fix the case where we want to run the CI if a PR is made from a fork.


cwgoes:
Now it looks like it's running twice, though?

Fraccaman:
It's because drone receives two events from the github webhook, 1 is the creation of the PR and one is the push, so starts two build. I'll check how to run one of the other.

cwgoes:
Yes - we want to run on every push to _this_ repository and on any PR to _this_ repository but not twice.

cwgoes:
@Fraccaman Is this updated to our latest discussions (just on PR, for now)?

Fraccaman:
@cwgoes I still have to do some stuff. I will ping as soon as it is ready :)

cwgoes (approved):
utACK

PR #715, opened 2021-03-08
Explicit pipeline - Feedback

fdedden:
I think that understanding the code as a newcomer is quite difficult. One reason is a lack of comments, but more importantly I think that the whole structure of the code could be more self-explanatory. There are a whole lot of steps the code goes through: parsing, contextToCore, annotation, typechecking etc. Each of these functions rely on the previous ones, e.g. typecheck relies on parse (which makes sense from the perspective of typecheck). The big downside to this is that the pipeline is very implicit: there is no way to instantaneously see where a certain action is done, nor is it easy to add a phase somewhere in-between. Propagating errors or warnings is more difficult as well.

Having an explicitly programmed pipeline, where every phase of the compiler is called, would help greatly with making the code easier to read for newcomers, and ourselves as well.


library/StandardLibrary/src/Juvix/Library/Feedback.hs
```diff
@@ -0,0 +1,61 @@
+module Juvix.Library.Feedback where
+
+import qualified Control.Monad.Trans as Trans
+import Juvix.Library
+
+-- | Keep track of messages during the compilation or REPL.
+-- The implementation is based on
+-- [[https://github.com/UU-ComputerScience/uu-cco/blob/master/uu-cco/src/CCO/Feedback.hs]].
+data Feedback msg a
+  = Success msg a -- Indicate success.
```
Acentelles, line 10: I can't see any use of the `msg` parameter in Success. In which cases are we planning to provide a message to Success?

app/Compile.hs
```diff
@@ -21,65 +22,83 @@ import qualified Juvix.Core.Pipeline as CorePipeline
...
-  core <- Pipeline.toCore ["stdlib/Prelude.ju", "stdlib/Michelson.ju", "stdlib/MichelsonAlias.ju", fin]
+type Code = Text
+
+type OutputCode = Text
+
+type Message = [P.String]
+
+type Pipeline = Feedback.FeedbackT Message IO
+
+-- | Function that parses code gives as input.
```
Acentelles, line 42: typo `given`

library/StandardLibrary/src/Juvix/Library/Feedback.hs
fdedden: In case of success, there are still reasons to provide feedback. Compilers typically present warnings or advice, even though the compilation is successful.

Acentelles: Yes, although using it in its current form looks difficult. Have you considered having a Writer instance?

app/Compile.hs
```diff
@@ -21,65 +22,83 @@ import qualified Juvix.Core.Pipeline as CorePipeline
...
-      pure ctx
-    Left err -> do
-      T.putStrLn (show err)
-      exitFailure
+    Right ctx -> return ctx
+    Left err -> Feedback.FeedbackT $ return $ Feedback.Fail [show err]
+  where
+    toCore_wrap :: Code -> IO (Either Pipeline.Error FE.FinalContext)
+    toCore_wrap code = do
+      fp <- Temp.writeSystemTempFile "juvix-toCore.ju" (Text.unpack code)
```
Acentelles, line 52: Will you be able to write to "juvix-toCore.ju" if you run this code twice (i.e. the file already exists)?

app/Compile.hs
```diff
@@ -3,6 +3,7 @@
 module Compile where
```
Acentelles, line 3: Wonder whether this module should live in src. It would be great to have it backend agnostic. What do you think of having a Backend typeclass?

app/Compile.hs
fdedden: Note this solution is just a workaround to be able to pass data into the 'old' pipeline, by means of a file. Ideally these functions will take actual code instead of a filepath.

The writeSystemTempFile function takes into account the file already existing by adding a process id and possible a number.

app/Compile.hs
fdedden: Yes, this definitely has to go and should move to, I guess, `src/`. Having a Backend typeclass is what I was thinking about as well, but I need to think it through with regards to how this will work together with the  Juvix-module level backends.

library/StandardLibrary/src/Juvix/Library/Feedback.hs
```diff
@@ -0,0 +1,61 @@
...
+  | Fail msg -- Indicate a failure.
```
Acentelles: I don't think it's a good idea having same `msg` for both Success and Fail.

I would expect having a CompileError datatype that encapsulates all errors. E.g.
```
data PipelineError =  ParseError ... | TypecheckError ... | CompileError ... | ...
```

library/StandardLibrary/src/Juvix/Library/Feedback.hs
```diff
@@ -0,0 +1,101 @@
...
+import qualified Control.Monad.Trans as Trans
+import qualified Data.String as String
+import Juvix.Library
+
+-- | Keep track of messages during the compilation or REPL.
+-- The implementation is based on
+-- [[https://github.com/UU-ComputerScience/uu-cco/blob/master/uu-cco/src/CCO/Feedback.hs]].
+data Feedback (app :: * -> *) msg a
+  = Success (app msg) a -- Indicate success.
+  | Fail (app msg) -- Indicate a failure.
```
Acentelles, line 20: I don't think it's a good idea having the same `msg` for both Success and Fail.

It'd be great to have a datatype that encapsulates all pipeline errors. E.g.
```
data PipelineError = ParseError ... | TypecheckError ... | CompileError ... | ...
```

What do you think?

fdedden: Mind that the current message implemented as a string is something to be improved with new implementations of the functions in the pipeline.

app/Compile.hs
fdedden: For the Backend class, see: https://github.com/heliaxdev/juvix/issues/157#issuecomment-793964127

fdedden:
Merge is pending a fix for the problems with the CI.

library/StandardLibrary/src/Juvix/Library/Feedback.hs
```diff
@@ -0,0 +1,101 @@
...
+    String.IsString msg
+  ) =>
+  Fail.MonadFail (Feedback app msg)
+  where
+  fail = Fail . pure . String.fromString
+
+-- | Monad transformer of Feedback.
+data FeedbackT app msg m a = FeedbackT {runFeedbackT :: m (Feedback app msg a)}
+
+instance Monad m => Functor (FeedbackT app msg m) where
```
Acentelles, line 53: Can you derive some of these instances automatically (e.g. Functor, Applicative)?

cwgoes (approved):
ACK, tested locally, let's note unaddressed comments in issues

PR #720, opened 2021-03-09
Sexp contextify pass

mariari:


library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
```diff
@@ -0,0 +1,223 @@
...
+updateTopLevel (name Sexp.:> body) ctx
+  | named ":defsig-match" = defun body ctx
+  | named "declare" = declare body ctx
+  | named "type" = type' body ctx
+  | named "open" = open body ctx
+  where
+    named = Sexp.isAtomNamed name
+updateTopLevel _ ctx = pure $ Type.PS ctx [] []
+
+defun :: Sexp.T -> Context.T Sexp.T Sexp.T Sexp.T -> IO Type.PassSexp
```
cwgoes, line 58: needs explanatory comment

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
```diff
@@ -0,0 +1,223 @@
...
+      decideRecordOrDef forms name (Context.currentName ctx) precendent actualSig
+    pure $
+      Type.PS
+        { ctxS = Context.add (NameSpace.Pub name) def ctx,
+          opensS = [],
+          modsDefinedS
+        }
+defun _ _ctx = error "malformed defun"
+
+declare :: Sexp.T -> Context.T Sexp.T Sexp.T Sexp.T -> IO Type.PassSexp
```
cwgoes, line 82: needs explanatory comments

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
```diff
@@ -0,0 +1,223 @@
...
+                    (NameSpace.Pub atomName)
+                    (Context.Information [Context.Prec prec])
+                    ctx,
+                opensS = [],
+                modsDefinedS = []
+              }
+declare _ _ = error "malformed declare"
+
+-- TODO ∷ don't ignore the assoc information given with type'
+type' :: Sexp.T -> Context.T Sexp.T Sexp.T Sexp.T -> IO Type.PassSexp
```
cwgoes, line 120: needs explanatory comment

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
```diff
@@ -0,0 +1,223 @@
...
+-- - once we have type checking, rely on that
+-- - for dep types where inference is undecidable, force signature
+-- - for functions like (f x) where that evals to a module, where it
+--   is dependent but decidable, force signature?
+
+-- The NameSymbol.T return returns back all record names that are found
+-- we don't return a map, as we can't do opens in a record
+
+-- | decideRecordOrDef tries to figure out
+-- if a given defintiion is a record or a definition
```
cwgoes, line 161:
```suggestion
-- if a given definition is a record or a definition
```

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
```diff
@@ -0,0 +1,223 @@
...
+decideRecordOrDef ::
+  Sexp.T ->
+  Symbol ->
+  NameSymbol.T ->
+  Context.Precedence ->
+  Maybe Sexp.T ->
+  IO (Type.DefinitionSexp, [NameSymbol.T])
```
cwgoes, line 168: is there a reason this is in `IO` instead of `MonadIO` ?

cwgoes (changes requested):
See comments, and do we have tests for `declare` / records?

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
mariari: Probably so I dont' have to lift in emptyRecord and other such functions.

cwgoes: hmm let's make sure this is standardised?

cwgoes (approved):
ACK modulo the question about MonadIO

library/Translate/src/Juvix/FrontendContextualise/Contextify/Sexp.hs
mariari: I moved it to MonadIO... all the functions inside are stil IO except for that one, but since we really only want to export a few, it would make sense to have exported things be in MonadIO?

PR #722, opened 2021-03-10
Witch Frontend updated to use S-expressions

cyberglot:
Summary: new syntax for effect handlers in Juvix.

Main contributions: 
 - adds `effect` and `handler` syntaxes (also `_ via _`);
 - removes do notation as `Prelude.»=`.
 
### Added Syntaxes
```ocaml
effect Print where 
  let print : string -> unit
  let pure  : x -> string
```
A new effect Print is defined, and its interface signature. All effects must have a return clause (here named `pure`), and hopefully operation signatures as well (an effect without operations is just a pure computation).

```ocaml
handler printer where 
  let print   = %IO.print
  let pure x  = toString x
```
A handler provides an implementation of at least one effect and a return clause (here named `pure`), and it can implement as many operations as the user wants.

```ocaml
let main = prog via printer
```
Handlers can be applied to programs using the `via` operator, which is just a shortcut for function application.

## Do Notation
The do notation no longer gets translated into `Prelude.»=`, it is just another effect construct now, to be translated by the ANF pass – since every command inside a do notation feeds the next continuation.


mariari:
For the syntax it might be best to have

```juvix
effect Print = 
  | op print : string -> ()
  | return   : x -> string
```
just to mimic what type looks like, same with the handler. At least for now

library/Frontend/src/Juvix/Frontend/Sexp.hs
```diff
@@ -185,6 +187,33 @@ transGuardBody _alsee trans (Types.Body b) = trans b
...
+--------------------------------------------------------------------------------
+-- Effect Expansion
+--------------------------------------------------------------------------------
+
+transHand :: Types.Handler -> Sexp.T
+transHand (Types.Hand name ops ret) =
+  Sexp.list
+    [ Sexp.atom ":defhandler",
+      Sexp.atom (NameSymbol.fromSymbol name),
+      Sexp.listStar [Sexp.atom ":ops", Sexp.list (fmap transOperation ops)],
```
mariari, line 199: What does `(:ops ops)` give us rather than doing it with just the second argument of defhandler being a list directly?

library/Frontend/src/Juvix/Frontend/Sexp.hs
```diff
@@ -185,6 +187,33 @@ transGuardBody _alsee trans (Types.Body b) = trans b
...
+      transOperation ret
+    ]
+
+transEffect :: Types.Effect -> Sexp.T
+transEffect Types.Eff {effName, effOps, effRet} =
+  Sexp.list
+    [ Sexp.atom ":defeff",
+      Sexp.atom (NameSymbol.fromSymbol effName),
+      Sexp.listStar [Sexp.atom ":ops", Sexp.list (fmap transSig effOps)],
```
mariari, line 208: Same here.

would make sense if the given form doens't know but here it seems always fixed, or if the form was in an style like

```juvix
(test :a 3 :b 4)
```
but in this case it's not?

library/Frontend/test/Sexp.hs
```diff
@@ -394,6 +396,40 @@ caseTest =
...
+    [ T.testCase "basic" (basic T.@=? basicExpected)
+    ]
+  where
+    basic =
+      Parser.parse
+        "handler print where op print x = print x , return x = toString x"
+        |> singleEleErr
+    basicExpected =
+      Sexp.parse
+        "(:defhandler print (:ops (:defop print (x) (print x))) (:defop return (x) (toString x)))"
```
mariari, line 416: I see what you are trying to go for, yeah you may want an alist style like
```
(:defhandler print
  :ops ((print (x) (print x)))
  :defop ((return (x) (toString x))))
```

if we can have more than one, if we can only have one then

```lisp

(:defhandler print
  :ops (print (x) (print x))
  :defop (return (x) (toString x)))
```

mariari (changes requested):
Overall Looks okay, a few comments though

I think the way the `:defhandler` and the `:defeff` form is needs to be tweaked in terms of being either an a-list or or dropping the tags

library/Translate/src/Juvix/Desugar/Passes.hs
```diff
@@ -259,46 +257,6 @@ combineSig [] = []
...
-            Just arr ->
-              Sexp.list
-                [ Sexp.atom "Prelude.>>=",
-                  arr ^. body,
-                  Structure.Lambda (Sexp.list [arr ^. name]) acc
-                    |> Structure.fromLambda
-                ]
-            Nothing ->
-              Sexp.list [Sexp.atom "Prelude.>>", bodyOf, acc]
-
```
mariari, line 301: We should keep this in until we have the new one in.

library/Frontend/src/Juvix/Frontend/Sexp.hs
mariari: Yeah I'm not sure this comes out as you expect it to, as now upon reasoning about it you want something like 

```lisp
(:defhandler print
  (:defops print (x) (print x))
  (:defops fud (x) (print x))
  (:defun return (x) (toString x)))
```

right?

library/Translate/src/Juvix/Desugar/Passes.hs
cyberglot: Why? I need `:do` atoms to transform into continuations.

library/Frontend/src/Juvix/Frontend/Sexp.hs
cyberglot: `(:ops ops)` is more clearly structured than an anonymous list, also it's easier to query directly.

library/Frontend/src/Juvix/Frontend/Sexp.hs
cyberglot: no, return is also an operation, but a special one.

cyberglot:
> what is op? is everything but return not op????

`op` stands for operation, and `return` is one as well. Theoretically, all operations including `return` are called a `clause`, I can change it to `:defclause` but I don't think we need to be this pedantic.

> Also it seems you want your forms to be something like
> 
> ```lisp
> (:defhandler print
>   (:defops bar (x) (print x))
>   (:defops print (x) (print x))
>   (:defun return (x) (toString x)))
> ```
```lisp
(:defhandler print
  (:ops ((:defop print (x) (print x))
             (:defop fud (x) (print x)))
  (:defop return (x) (toString x)))
```
This is the form I coded and what I want. I think it's best to have a way to query all operations (including `return`) in one go, and a way to get just the effectful ones (meaning, no `return`). 

> the current strategy you give seems like it can't handle it?

I don't understand what you mean.

> Also can you update the changelog

Will do.

library/Translate/src/Juvix/Desugar/Passes.hs
mariari: Yes, though we should keep the current one until the next pr, where the continuation strategy is in.

mariari:
> > what is op? is everything but return not op????
> 
> `op` stands for operation, and `return` is one as well. Theoretically, all operations including `return` are called a `clause`, I can change it to `:defclause` but I don't think we need to be this pedantic.
> 

Hmm, the thing I found odd was that they all had `op` except `return`. I think it'd be rather easy to intuitively write `op` for the `return` case as well. When we separate it we can look for `retrun` explicitly for now, so it should be just as easy to work over, without the user qurik.

Note :: that if we keep `,` as our separator might be good to change `op` to `let` to just reflect the rest of the language. However if we take `|` instead I think it would be more consistent for now, to drop `op` entirely.

> > Also it seems you want your forms to be something like
> > ```lisp
> > (:defhandler print
> >   (:defops bar (x) (print x))
> >   (:defops print (x) (print x))
> >   (:defun return (x) (toString x)))
> > ```
> 
> ```lisp
> (:defhandler print
>   (:ops ((:defop print (x) (print x))
>              (:defop fud (x) (print x)))
>   (:defop return (x) (toString x)))
> ```
> 
> This is the form I coded and what I want. I think it's best to have a way to query all operations (including `return`) in one go, and a way to get just the effectful ones (meaning, no `return`).

I don't disagree with your logic here, though I came to the form I suggested after looking at the test. Mainly due to the first layer of the s-expression is visible to the end user for writing furter trasnformations. Hence I think that that there should be a layer that exapnds it, like the following. 

```lisp
(:defhandler print
  (:defops bar (x) (print x))
  (:defops print (x) (print x))
  (:defops return (x) (toString x)))

(:let-handler print
  (:ops ((:defop print (x) (print x))
         (:defop fud (x) (print x)))
  (:defop return (x) (toString x)))
```

One thing that may change in the future is more forms can be included in this, IIRC we discussed types in there, and I believe the drawn out form supports extensions a bit better.


As for the form itself, the thing that gives me pause, is that it seems you want keyword arguments like the following

```lisp
(:defhandler print
  :ops ((:defop print (x) (print x))
        (:defop fud (x) (print x)))
  :defop (return (x) (toString x)))
```

Which I then simplified to, since we only had `:defops`

```lisp
(:defhandler print
  :ops ((print (x) (print x))
        (fud (x) (print x)))
  :defop (return (x) (toString x)))
```

However upon thinking about it some more, we are saying `(:defop return (x) (to-string x))`

so the keyword thought doesn't really work out the greatest.

So maybe something more positional would be better?

```lisp
(:defhandler print
    ((:defop print (x) (print x))
     (:defop fud (x) (print x)))
  (return (x) (toString x)))
```

```lisp
(:defhandler print
    ((print (x) (print x))
     (fud (x) (print x)))
  (return (x) (toString x)))
```

Overall though, I think if we have a direct translation into `defhandler` from the frontend version then do the grouping in a tiny expansion, then these matters matter less, as they won't be exposed at all.

> 
> > the current strategy you give seems like it can't handle it?
> 
> I don't understand what you mean.

I was incorrect, I tested a form with multiple `op`'s in there and it works fine!

library/Frontend/src/Juvix/Frontend/Parser.hs
```diff
@@ -449,6 +451,64 @@ nameParser =
...
+--------------------------------------------------
+
+handlerParser :: Parser Types.Handler
+handlerParser = do
+  reserved "handler"
+  name <- prefixSymbolSN
+  reserved "where"
+  ops <- P.sepEndBy opParser (skipLiner J.comma)
+  ret <- retParser
+  pure $ Types.Hand name ops ret
```
mariari, line 465: This seems rather inflexible as you have to include the `return` parser as the last parser, otherwise we get an error.

library/Frontend/src/Juvix/Frontend/Parser.hs
```diff
@@ -449,6 +451,64 @@ nameParser =
...
+          pure (Types.Like name args guard)
+      )
+
+retParser :: Parser Types.Operation
+retParser =
+  Types.Op <$> do
+    reserved "return"
+    args <- P.many argSN
+    guard <- guard expression
+    pure (Types.Like "return" args guard)
```
mariari, line 483: Can this be turned into an `opParser`?

mariari (commented):
I think the current strategy of special casing `return` makes this a bit more inflexible, that can easily be teased out without a bunch of special logic

library/Frontend/src/Juvix/Frontend/Parser.hs
cyberglot: I do agree, the order shouldn't be inflexible.

library/Frontend/src/Juvix/Frontend/Parser.hs
cyberglot: Oooh, I thought I had refactored this to be an `opParser`, thanks.

cyberglot:
I'm fine with replacing `,` with `|`, no worries here.

 I'd, however, prefer to keep the `op` reserved word rather than `let` cuz I think it's important to keep a cognitive signifier that you're defining an effectful operation that has a different significance in the language. Then, I'd prefer to stick with `op foo ... | return ...` or remove it altogether and go with `foo ... | return ...`, I want the users to make operations thinking of explicit effects, and `return` as the monadic return that connects pure values with effectful computations.

I don't understand that part about s-expressions being available to the user, but I can macro-expand it to have better query results, no worries.

library/Frontend/src/Juvix/Frontend/Parser.hs
systemf-frank: Do we need to consider `return` part of the syntax, or could it just be a regular function? I guess for parsing it shouldn't matter, the lack of a `return` I guess is a something that could/should be checked later on in the compilation process? Depends on how we look at these things.

library/Frontend/src/Juvix/Frontend/Types/Base.hs
```diff
@@ -148,6 +150,29 @@ data NameType = NameType'
...
+  deriving (Show, Read, Eq)
+
+data Operation = Op (FunctionLike Expression)
+  deriving (Show, Read, Eq)
+
+--  A 'Handler', as implemented here, is a set of functions that implement
+-- (at least) one `Effect` interface.
+-- However, Handlers are NOT scoped, meaning that they can't be defined
+-- defined within another function. We CAN fix that, but it requires
+-- us to make some choices, it's wise to have Witch up and running first.
```
systemf-frank, line 171: Handlers not being scoped might actually be a good thing? Having each and every function use it's own handler definition for the same effect could be very confusing. On the other hand it would provide more flexibility.

Also note the double `defined defined` in 169-170.

ChangeLog.org
```diff
@@ -1,5 +1,44 @@
...
+  | op print : string -> ()
+  | return   : x -> string
+#+END_SRC
+A new effect Print is defined, and its interface signature. All
+effects must have a `return` clause, and hopefully operation
+signatures as well (an effect without operations is just a pure
+computation).
+
+#+BEGIN_SRC ocaml
+handler print where
```
systemf-frank, line 23: Shouldn't `print` start with a capital here?

ChangeLog.org
```diff
@@ -1,5 +1,44 @@
...
+  | op print  = %IO.print
+  | return x  = toString x
+#+END_SRC
+A handler provides an implementation of at least one effect and
+a return clause, and it can implement as many operations as the
+user wants.
+
+#+BEGIN_SRC ocaml
+let main = prog via print
```
systemf-frank, line 32: The `print` here refers to handler named `print` and not to the `print` operation, right?

ChangeLog.org
cyberglot: handlers are lowercase, while effects are uppercase (convention, not enforced). you can also name handlers whatever you want regarless of the effect you handling

ChangeLog.org
cyberglot: You're correct, I'm gonna change the name of the handler to `printer` so it won't be confusing anymore. Thanks.

ChangeLog.org
systemf-frank: So the connection between an effect and it's handler is the fact that the handler has a operation with the correct name and type available?

cyberglot: exactly, same operation name and type. you can even handle multiple different effects within a single handler if you want to, or not handle some operations of an effect... as long as some other handler handles that operation you omitted -- otherwise it's a handling error.

ChangeLog.org
```diff
@@ -1,5 +1,44 @@
...
+
+ - _Main contributions_:
+   + adds `effect` and `handler` syntaxes (also `_ via _`);
+   + removes do notation as `Prelude.»=`.
+
+*** Added Syntaxes
+#+BEGIN_SRC ocaml
+effect Print where
+  | op print : string -> ()
+  | return   : x -> string
```
cwgoes, line 15: Why `return` as opposed to `pure` ?

I think the latter avoids bad associations (`return` in JS, e.g.) and is a bit more self-evident.

cwgoes (commented):
The syntax seems mostly fine, but I'm wondering about `return` vs `pure` .

ChangeLog.org
cyberglot: Changed it to `pure`already 👍🏽

systemf-frank (approved):

mariari (approved):

PR #723, opened 2021-03-10
General Map and Traversal functions over context and Sexpressions

mariari:


library/StandardLibrary/src/Juvix/Library/Sexp.hs
```diff
@@ -111,9 +153,9 @@ nameFromT (Atom (A name _)) = Just name
 nameFromT _ = Nothing
 
 assoc :: T -> T -> Maybe T
-assoc t (car :> cdr)
-  | t == car = Just (cadr car)
-  | otherwise = assoc t cdr
+assoc t (car' :> cdr')
+  | t == car car' = Just (cadr car')
```
cyberglot, line 157: wait, `cadr`? is that correct? 🤔  am I missing something?

cyberglot:
I have just one comment, otherwise LGTM.

cyberglot (commented):
Just have one question, otherwise LGTM.

library/StandardLibrary/src/Juvix/Library/Sexp.hs
mariari: yeah, as we are searching a list like

((:a 3) (:b 4)) 

so if we do

assoc :a then we should get 3 back so

(cadar *) ===>  3.

we have an extra a as we are doing this cadr on the car thus forming cadar

cyberglot (approved):

cwgoes (commented):
ACK

PR #724, opened 2021-03-11
Plonk IR

Acentelles:
- Implements an arithmetic circuit backend for Juvix 
- Creates an IR that expresses Plonk operations (see IR.md)
- Tests polynomial equations written in Juvix

Other changes:
- Updates Pipeline
- Tests circuit against Groth (although I would remove Groth from this PR)
- Adds JSON serialization
- Adds a prettified output of the circuit
- Outputs a dot graph


library/Backends/Plonk/src/Juvix/Backends/Plonk/Expr.hs
```diff
@@ -0,0 +1,388 @@
...
+  -- gte
+  CGte :: CompOp f
+  -- lt
+  CLt :: CompOp f
+  -- lte
+  CLte :: CompOp f
+  -- eq
+  CEq :: CompOp f
+
+opPrecedence :: BinOp f a -> Int
```
cwgoes, line 101: Huh, are we planning to parse thsi?

cwgoes (commented):
Thanks! My primary questions concern how we plan to translate erased core terms to circuit expressions - would we have a `termToExpr` function which compiles erased core to the `Expr` term you have here? We'll need to translate case statements to conditionals (nested), inline lambdas, convert datatypes to some field element representation etc. 

If `Expr` is intended to be the start of the IR, I think it would be worth writing down a quick English spec describing the operations - this needs to be agreed upon between the Juvix codebase and the PLONK/plookup codebase, we'll also need a standard serialisation format (JSON or something is fine for now)

library/Backends/Plonk/src/Juvix/Backends/Plonk/Expr.hs
Acentelles: It was meant for pretty printing. See
```
parensPrec :: Int -> Int -> Doc -> Doc
parensPrec opPrec p = if p > opPrec then parens else identity
```

app/Main.hs
```diff
@@ -103,27 +115,42 @@ run :: Context -> Options -> IO ()
...
-            >>= Pipeline.typecheck @Pipeline.BMichelson
-            >>= liftIO . print
-        Compile fin fout (Michelson backend) -> do
-          liftIO (readFile fin)
-            >>= Pipeline.parse
-            >>= Pipeline.typecheck @Pipeline.BMichelson
-            >>= Pipeline.compile @Pipeline.BMichelson
-            >>= Pipeline.writeout fout
-            >>= liftIO . print
+        Parse fin backend -> case backend of
```
cwgoes, line 124: Can some code be de-duplicated here?

app/Main.hs
```diff
@@ -103,27 +115,42 @@ run :: Context -> Options -> IO ()
...
+          (Michelson b) -> readAndPrint fin (Pipeline.parse b)
+          (Plonk b) -> readAndPrint fin (Pipeline.parse b)
+        Typecheck fin backend -> case backend of
```
cwgoes, line 127: Ditto?

library/Backends/Plonk/IR.md
```diff
@@ -0,0 +1,103 @@
+# PLONK IR
+
+This is mainly a translation of https://hackmd.io/@zacwilliamson/r1z1yZTs8 into Haskell
+
+## Binary operations
+
+| Opcode | Opcode Bytes    | Types | Description     | Notes |
+| ------ | --------------- | ----- | --------------- | ----- |
+| add    | `[a], [b], [o]` |       | o = a + b       |       |
```
cwgoes, line 9: Should we fill in the types column?

library/Backends/Plonk/IR.md
```diff
@@ -0,0 +1,103 @@
...
+| sub    | `[a], [b], [o]` |       | o = a - b       |       |
+| mul    | `[a], [b], [o]` |       | o = a * b       |       |
+| div    | `[a], [b], [o]` |       | o = a / b       |       |
+| mod    | `[a], [b], [o]` |       | o = a % b       |       |
+| and    | `[a], [b], [o]` |       | `o = a &amp; b` |       |
```
cwgoes, line 14: This treats the element as a boolean value, right?

What happens if it is out of range?

Do we have or want opcodes for bitwise `and`, `or`, `xor` ?

library/Backends/Plonk/IR.md
```diff
@@ -0,0 +1,103 @@
...
+  CEq :: CompOp f
+```
+
+## Unary operations
+
+| Opcode   | Opcode Bytes  | Description                    | Notes                 |
+| -------- | ------------- | ------------------------------ | --------------------- |
+| setpub   | `[p], [o]`    | o = p                          | from public calldata  |
+| setpriv  | `[p], [o]`    | o = p                          | from private calldata |
+| dup      | `[a], [o]`    | `o = a`                        |                       |
```
cwgoes, line 58: i.e., not from public or private calldata?

library/Backends/Plonk/IR.md
```diff
@@ -0,0 +1,103 @@
...
+| shr      | `[a], x, [o]` | `o = a &gt;&gt; x`             |                       |
+| rotl     | `[a], x, [o]` | `o = a &lt;&lt;&lt; x`         |                       |
+| rotr     | `[a], x, [o]` | `o = a &gt;&gt;&gt; x`         |                       |
+| asserteq | `[a], [o]`    | throw if `a != 0`              |                       |
+| assertlt | `[a], x`      | throw if `a &gt;= 2^x`         |                       |
+| mload    | `[x], [o]`    | read `o` from RAM location `x` |                       |
+| mstore   | `[x], [a]`    | write `a` to RAM location `x`  |                       |
+
+```haskell
+data UnOp f a where
```
cwgoes, line 71: Seems we're missing `setpub` and `setpriv` and the memory operations, is that intentional?

library/Backends/Plonk/IR.md
```diff
@@ -0,0 +1,103 @@
...
+| dup      | `[a], [o]`    | `o = a`                        |                       |
+| iszero   | `[a], [o]`    | `o = (a == 0)`                 |                       |
+| not      | `[a], [o]`    | `o = !a`                       |                       |
+| shl      | `[a], x, [o]` | `o = a &lt;&lt; x`             | x is const            |
+| shr      | `[a], x, [o]` | `o = a &gt;&gt; x`             |                       |
+| rotl     | `[a], x, [o]` | `o = a &lt;&lt;&lt; x`         |                       |
+| rotr     | `[a], x, [o]` | `o = a &gt;&gt;&gt; x`         |                       |
+| asserteq | `[a], [o]`    | throw if `a != 0`              |                       |
+| assertlt | `[a], x`      | throw if `a &gt;= 2^x`         |                       |
+| mload    | `[x], [o]`    | read `o` from RAM location `x` |                       |
```
cwgoes, line 67: What is this RAM referred to here?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Builder.hs
```diff
@@ -0,0 +1,174 @@
...
+-- | Apply function to variable names.
+mapVarsIR :: (i -> j) -> IR i f ty -> IR j f ty
+mapVarsIR f ir = case ir of
+  IVar i -> IVar $ f i
+  IConst v -> IConst v
+  IBinOp op e1 e2 -> IBinOp op (mapVarsIR f e1) (mapVarsIR f e2)
+  IUnOp op e1 -> IUnOp op (mapVarsIR f e1)
+  IIf b tr fl -> IIf (mapVarsIR f b) (mapVarsIR f tr) (mapVarsIR f fl)
+
+-- EEq lhs rhs -> EEq (mapVarsIR f lhs) (mapVarsIR f rhs)
```
cwgoes, line 174: What's this comment about?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Circuit.hs
```diff
@@ -0,0 +1,224 @@
...
+-- describe affine transformations.
+data AffineCircuit i f
+  = Add (AffineCircuit i f) (AffineCircuit i f)
+  | ScalarMul f (AffineCircuit i f)
+  | ConstGate f
+  | Var i
+  deriving stock (Read, Eq, Show, Generic)
+  deriving anyclass (FromJSON, ToJSON)
+
+fetchVars :: AffineCircuit Wire f -> [Wire]
```
cwgoes, line 20: Missing comment, though this function is pretty obvious

library/Backends/Plonk/src/Juvix/Backends/Plonk/Circuit.hs
```diff
@@ -0,0 +1,224 @@
...
+  -- | circuit to evaluate
+  AffineCircuit i f ->
+  f
+evalAffineCircuit lookupVar vars = \case
+  ConstGate f -> f
+  Var i -> fromMaybe 0 $ lookupVar i vars
+  Add l r -> evalAffineCircuit lookupVar vars l + evalAffineCircuit lookupVar vars r
+  ScalarMul scalar expr -> evalAffineCircuit lookupVar vars expr * scalar
+
+-- | Generate enough roots for a circuit
```
cwgoes, line 63: What is "enough" roots?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Compiler.hs
```diff
@@ -0,0 +1,118 @@
...
+import qualified Data.Map as Map
+import Juvix.Backends.Plonk.Builder as P
+import Juvix.Backends.Plonk.Circuit as P
+import Juvix.Backends.Plonk.IR as P
+import Juvix.Backends.Plonk.Types as P
+import qualified Juvix.Core.ErasedAnn.Types as Ann
+import Juvix.Library
+import qualified Juvix.Library.NameSymbol as NameSymbol
+
+-- translate case statements to conditionals (nested)
```
cwgoes, line 15: Is this intended to be a comment on `compileBinOp`? Seems misplaced.

library/Backends/Plonk/src/Juvix/Backends/Plonk/Compiler.hs
```diff
@@ -0,0 +1,118 @@
...
+compileTermWithWire term = do
+  compileOut <- compileTerm term mempty mempty
+  case compileOut of
+    Left wire -> pure wire
+    Right circ -> do
+      wire <- freshOutput
+      emit $ MulGate (ConstGate 1) circ wire
+      pure wire
+
+-- do
```
cwgoes, line 114: What's this comment?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Compiler.hs
```diff
@@ -0,0 +1,118 @@
...
+
+compileTerm :: (Integral f, Show f) => FFAnnTerm f -> Map NameSymbol.T Wire -> [FFAnnTerm f] -> IRM f (Either Wire (AffineCircuit Wire f))
+compileTerm _term@(Ann.Ann _ _ t) m a =
+  case t of
+    Ann.Prim p -> compilePrim p m a
+    Ann.Var symbol -> case Map.lookup symbol m of
+      Just v -> pure . Left $ v
+      Nothing -> panic $ "Unable to find variable " <> show symbol
+    Ann.AppM fun@(Ann.Ann _ _ v) args -> compileTerm fun m args
+    -- case v of
```
cwgoes, line 88: Remove these two lines?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Compiler.hs
```diff
@@ -0,0 +1,118 @@
...
+  n -> panic $ show n
+
+compileTerm :: (Integral f, Show f) => FFAnnTerm f -> Map NameSymbol.T Wire -> [FFAnnTerm f] -> IRM f (Either Wire (AffineCircuit Wire f))
+compileTerm _term@(Ann.Ann _ _ t) m a =
+  case t of
+    Ann.Prim p -> compilePrim p m a
+    Ann.Var symbol -> case Map.lookup symbol m of
+      Just v -> pure . Left $ v
+      Nothing -> panic $ "Unable to find variable " <> show symbol
+    Ann.AppM fun@(Ann.Ann _ _ v) args -> compileTerm fun m args
```
cwgoes, line 87: Should we be doing something with `v` ?

library/Backends/Plonk/test/Test/Example/Juvix/Polynomial.ju
```diff
@@ -0,0 +1,24 @@
...
+
+sig (^) : field -> field -> field
+let (^) = %Circuit.exp
+declare infixl (^) 8
+
+sig (*) : field -> field -> field
+let (*) = %Circuit.mul
+declare infixl (*) 7
+
+
```
cwgoes, line 18:
```suggestion

```

test/examples/Polynomial.ju
```diff
@@ -0,0 +1,24 @@
...
+
+sig (^) : field -> field -> field
+let (^) = %Circuit.exp
+declare infixl (^) 8
+
+sig (*) : field -> field -> field
+let (*) = %Circuit.mul
+declare infixl (*) 7
+
+
```
cwgoes, line 18:
```suggestion

```

scripts/run-formatter.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
cwgoes, line 1: why are these scripts renamed in this PR?

scripts/run-org-generation.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
cwgoes, line 1: why are these scripts renamed in this PR?

library/Translate/src/Juvix/Contextify/ToContext/Sexp.hs
```diff
@@ -129,6 +129,27 @@ declare (Sexp.List [inf, n, i]) ctx
               }
 declare _ _ = error "malformed declare"
 
+mkCtx :: Symbol -> [(Symbol, Sexp.T)] -> Context.T Sexp.T Sexp.T Sexp.T -> Context.Record Sexp.T Sexp.T Sexp.T -> Context.T Sexp.T Sexp.T Sexp.T
```
cwgoes, line 132: Why are these changes in this PR, are they necessary for some reason?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Parameterization.hs
```diff
@@ -0,0 +1,238 @@
...
+import qualified Juvix.Core.IR.Types.Base as IR
+import qualified Juvix.Core.Parameterisation as Param
+import qualified Juvix.Core.Types as Core
+import Juvix.Library hiding (many, show, try)
+import qualified Juvix.Library.HashMap as Map
+import qualified Juvix.Library.NameSymbol as NameSymbol
+import qualified Juvix.Library.Usage as Usage
+import Prelude (Show (..))
+
+check3Equal :: Eq a => NonEmpty a -> Bool
```
cwgoes, line 26: I think some of these can be moved into code common to all parameterisations or a stdlib, this can happen later though

cwgoes (changes requested):
Thank you! See comments, a few more general notes:

- Some code in this PR seems to be related to s-expressions. Is it related to this PR or not?
- Can we add a testcase that covers `&&`, `||`, and `xor` ?
- It looks like the JSON serialisation is for the _circuit_ at the moment. Is that just convenient, or are there choices in converting the IR to a circuit which we want to make use of higher-level information from Juvix for? Otherwise, should we be serialising the IR instead?

library/Frontend/src/Juvix/Frontend/Parser.hs
```diff
@@ -146,10 +144,6 @@ expressionGen p =
 expression' :: Parser Types.Expression
 expression' = expressionGen app''
 
--- used to remove app from parsing
```
systemf-frank, line 149: How are the changes in this file related to the new backend? I think these could better be dealt with in their own issue/PR.

library/Translate/src/Juvix/Contextify/ToContext/Sexp.hs
```diff
@@ -230,15 +230,27 @@ decideRecordOrDef xs _ _ pres ty =
...
-     in case foo of
-          Nothing -> []
-          Just xs ->
-            -- filter out the record constructors, which really aren't constructors
-            filter (\x -> x /= ":record-d" && x /= ":") xs
+    case traverse (eleToSymbol . Sexp.car) s of
+      Nothing -> []
+      Just xs -> xs
+  -- filter out the record constructors, which really aren't constructors
+  -- filter (\x -> x /= ":record-d" && x /= ":") xs
```
systemf-frank, line 253: ?

systemf-frank (changes requested):
In general the changes are looking good. As I am not into all the semantics and details, my review is mostly a check for surprising things / sanity check / chance to nitpick ;).

My main point of complaint is that it is a very big list of changes, not all of them actually directly related to the new backend (for example moving the test files under Translate, among other things). While sometimes it is difficult to limit yourself to make changes, dealing with these changes in separate issues keeps the PR more to the point. Only the reviewers will read through you code, while other devs will most likely only see the title of the PR. Based on the title they don't expect all these other modifications, and thus might be surprised that they did occur at some point. With separate issues for these changes, they will encounter the 'Move tests under Translate' issue for example, and know what happened. With the squash merging that we do, all these changes will be compiled into one commit, further hiding all the changes.

Some nitpicking:
- Not all functions / datatypes / modules had comments, but luckily the most important ones do.
- Not all modules had exports.

I know I am a bit an ass on these two points, but it will save us a lot of work in the future.

In general it looks OK, and it is good to see that there's partly support for multiple backends now.

library/Backends/Plonk/IR.md
cwgoes: Let's just list the operand types and output type (field or boolean)

library/Backends/Plonk/IR.md
cwgoes: Let's just note this somewhere, e.g. in the types

cwgoes: Bitwise operands seem likely to be expensive, let's avoid them for now

library/Backends/Plonk/IR.md
cwgoes: Per discussion seems like this can be removed

library/Frontend/src/Juvix/Frontend/Parser.hs
Acentelles: It's just cleaning up warnings that are clutter. Having a separate PR for just removing warnings is an overkill in my opinion.

Acentelles: I was told early that refactoring is something that happens along the way as a means to achieve something, not a goal (PR in this case) in itself. In this case, these changes are not even distracting

app/Interactive.hs
```diff
@@ -1,143 +0,0 @@
-module Interactive (interactive) where
```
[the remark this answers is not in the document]
Acentelles: A PR to remove this file?

systemf-frank (commented):
Looking good!

systemf-frank (approved):

app/Interactive.hs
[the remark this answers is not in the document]
systemf-frank: Yes, normally I would do that, but it is fine for now.

library/Backends/Plonk/test/Test/Example/Juvix/And.ju
```diff
@@ -0,0 +1,11 @@
+mod Polynomial where
+
+open Prelude
+open Circuit
+
+sig (&) : bool -> bool -> bool
```
cwgoes, line 6:
```suggestion
sig (&&) : bool -> bool -> bool
```

library/Backends/Plonk/test/Test/Example/Juvix/And.ju
```diff
@@ -0,0 +1,11 @@
...
+let (&) = %Circuit.and
```
cwgoes, line 7:
```suggestion
let (&&) = %Circuit.and
```

cwgoes (approved):
ACK modulo existing comments

PR #725, opened 2021-03-12
Binder aware traversal function for context passes

mariari:
Ts PR gives us a generic traversal function that is aware of binders and properly handles each form that might bind.

This means that any pass that cares about closures just needs to supply a proper predicate and transform function.

We also test and document this function in this PR.

### DIFF
--- a/library/Translate/test/Context/Environment.hs
+++ b/library/Translate/test/Context/Environment.hs
@@ -16,7 +16,7 @@ import qualified Juvix.Library.NameSymbol as NameSymbol
 import qualified Juvix.Library.Sexp as Sexp
 import qualified Test.Tasty as T
 import qualified Test.Tasty.HUnit as T
-import Prelude (error, fail)
+import Prelude (error)
 
 --------------------------------------------------------------------------------
 -- Top Level Test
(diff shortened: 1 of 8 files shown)


cyberglot (approved):
LGTM.

cyberglot (approved):

library/Translate/test/Context/Environment.hs
```diff
@@ -0,0 +1,250 @@
...
+
+capture :: ByteString -> (NameSymbol.T -> Bool) -> IO [Env.Closure']
+capture str trigger = do
+  Right (ctx, _) <-
+    contextualizeFoo str
+  let (_, Cap _ capture) =
+        runCtx (Env.passContextSingle ctx trigger recordClosure) emptyClosure
+  pure capture
+
+-- Right (ctx,_) <- contextualizeFoo "let f = 3"
```
cwgoes, line 226: What are these comments for?

cwgoes (approved):
ACK modulo nit

library/Translate/test/Context/Environment.hs
mariari: Just for quickly copy and pasting something into the repl

PR #727, opened 2021-03-16
Parameterisation clean-up

fdedden:
The parameterisation code needs a cleanup:

- [x] Remove the parsing functions, they aren't actually used anywhere.
- [x] Ensure `apply` works (particularly in the Michelson backend).
- [ ] Consider changing `apply` per `applyProper` in the Michelson backend or unify them somehow.


mariari (approved):
Everything here looks fine, thanks for adding all the extra comments.

Should merge master and see if anything is in conflict

systemf-frank:
`apply`/`applyProper` has not been changed yet. Basically `apply` in the Michelson backend deals with three cases:
* Too little arguments are passed.
* The right amount of arguments are passed.
* Too many arguments are passed.
Only in the case where the right amount of arguments are passed to a function, `applyProper` is called to actually perform the application.

By having `apply` do this decision, we need to reimplement this in every backend. Ideally we want to replace `apply` by the implementation of `applyProper`, and have the currying be done beforehand, so it can be shared among backends. This will require `apply`'s type to be changed to force evaluated terms is input (similar to how `applyProper` uses `Take` instead of `PrimVal'`).

These modifications haven't been made yet, as at the moment it is not easy to see how the exact implementation should work. While developing a second backend, it will most likely become clearer how we can unify code amongst backends in a better way.

cwgoes:
Minor note: please mark PR requests as ready-for-review when you'd like them to be reviewed, this ensure that Github's notification system works properly. I've done it for you in this case.

.drone.yml
```diff
@@ -1,6 +1,7 @@
+---
```
cwgoes, line 1: why are these changes in this PR?

cwgoes (approved):
ACK modulo rebase and removing unrelated bits

.drone.yml
systemf-frank: These aren't modifications I made myself, but I merged these from the develop branch to stay up to date with changes in the CI.

systemf-frank:
> Minor note: please mark PR requests as ready-for-review when you'd like them to be reviewed, this ensure that Github's notification system works properly. I've done it for you in this case.

Sorry, I wanted to do a merge first. Do you prefer to do it before merging?

cwgoes:
> > Minor note: please mark PR requests as ready-for-review when you'd like them to be reviewed, this ensure that Github's notification system works properly. I've done it for you in this case.
> 
> Sorry, I wanted to do a merge first. Do you prefer to do it before merging?

Oh no, that's fine then!

library/Core/src/Juvix/Core/Application.hs
```diff
@@ -103,22 +127,12 @@ instance Bitraversable (Arg' ext) where
...
--- fully-applied itself.
+-- | An argument to a partially applied primitive, which must be fully-applied
+-- itself.
 data Take ty term = Take
-  { usage :: Usage.T,
-    type' :: ty,
-    term :: term
+  { usage :: Usage.T, -- < Usage annotation for quantified types.
+    type' :: ty, -- < The type of the term.
+    term :: term -- < The term itself.
```
andy-morris, line 135:
```suggestion
  { usage :: Usage.T, -- ^ Usage annotation for quantified types.
    type' :: ty, -- ^ The type of the term.
    term :: term -- ^ The term itself.
```

app/Interactive.hs
```diff
@@ -40,8 +40,9 @@ mainLoop func = do
           H.outputStrLn =<< liftIO (func inp)
           mainLoop func
 
+-- TODO: remove
 parseString :: String -> Maybe (HR.Term Nat.Ty Nat.Val)
-parseString = HR.generateParser Nat.t
+parseString = undefined
```
andy-morris, line 45: Should this be removed, or is it still mentioned somewhere else?

andy-morris (commented):
a couple of nitpicks, but otherwise looks good to me!

app/Interactive.hs
systemf-frank: A quick grep shows that it isn't used anymore and can be removed, however I think it is outside the scope of this PR.

andy-morris: oh sorry, I didn't notice the filename, never mind

library/Core/src/Juvix/Core/Application.hs
systemf-frank: I think they are actually both incorrect, as the comments are located after the comma. Maybe one of these files where the formatter fails? I will have a look.

systemf-frank: Ormolu is not fixing it, so I will do it by hand.

PR #729, opened 2021-03-16
replace positivity in raw datatype with number of params

andy-morris:
since the positivity of parameters will be calculated by the typechecker, it doesn't make sense in raw datatypes. but it is still necessary to know how many arguments are parameters in the first place.

### DIFF
--- a/app/Compile.hs
+++ b/app/Compile.hs
@@ -114,7 +114,7 @@ unsafeEvalGlobal ::
   IR.Global primTy primVal
 unsafeEvalGlobal globals g =
   case g of
-    RawGDatatype (RawDatatype n pos a l cons) -> undefined
+    RawGDatatype (RawDatatype n a ps l cons) -> undefined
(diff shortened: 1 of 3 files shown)


thealmarty:
I believe the plan was that the user can specify positivity of the parameters. It was discussed a while back though. Could that still be supported?

andy-morris:
actually you're right, i'd forgotten about that. it should probably be replaced with `[Maybe Pos]` in raw in that case rather than just a length.

thealmarty:
> actually you're right, i'd forgotten about that. it should probably be replaced with `[Maybe Pos]` in raw in that case rather than just a length.

It was already a list in raw I think? We can maybe put them in `Maybe`. My thought was to default it to a value that makes sense.

andy-morris:
Actually I was misremembering what this flag is for.

I was thinking it was for marking parameters which are used strictly-positively, but in fact it is for parameters which can only be passed SP arguments. So it's correct as is, sorry.

PR #730, opened 2021-03-16
Add type checking of data declarations

thealmarty:
This PR aims to implement part of phase 2 of [this plan](https://gist.github.com/thealmarty/2e58f3949d1ff2bac926a2313c7d6a75).

This PR adds type checking of data type declarations. Positivity checks will be in a separate PR.

### DIFF
--- a/doc/Code/Library.org
+++ b/doc/Code/Library.org
@@ -836,6 +836,12 @@ Typechecked declarations are added to the signature.
   + [[Typechecker/Env]]
   + [[Typechecker/Types]]
   + [[IR/Types]]
+  + [[Globals]]
+  + [[Library]]
+******* CheckDatatype
+- _Relies on_
+  + [[CheckTerm]]
+  + [[IR/Evaluator]]
   + [[IR/Types/Base]]
   + [[Globals]]
   + [[Library]]
@@ -856,7 +862,6 @@ Typechecked declarations are added to the signature.
   + [[IR/Evaluator]]
   + [[IR/Types]]
   + [[IR/Types/Base]]
-  + [[Globals]]
   + [[Core/Parameterisation]]
   + [[Library]]
   + [[Usage]]
@@ -872,7 +877,6 @@ Quantitative type implementation inspired by
 ******* Base <<IR/Types/Base>>
 - _Relies on_
   + [[Library]]
-  + [[HashMap]]
   + [[NameSymbol]]
   + [[Usage]]
 ******* Globals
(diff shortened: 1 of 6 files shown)


thealmarty:
- [ ] explicit export (to conform to new style guide).
- [ ] for error messages, add to `TypecheckError'`.
- [ ] change telescopes to lists of records and include usages and other info.
- [ ] make typechecking functions accumulate a list of [Global' extV extT primTy primVal]`.
- [ ] make all typechecking functions return `[Global' extV extT primTy primVal]` i.e., the output type is `TypeCheck ext primTy primVal IO (Global' extV extT primTy primVal)`.

thealmarty:
Also:
- [ ] make sure `AllowAmbiguousTypes` flag is not on.

cwgoes:
Is this ready-for-review? If so, please mark it as such.

thealmarty:
> Is this ready-for-review? If so, please mark it as such.

I'll mark so as soon as the CI finishes.

library/Core/src/Juvix/Core/IR/Types/Globals.hs
```diff
@@ -410,26 +459,37 @@ deriving instance
...
+  StateT (Signature ext primTy primVal)
 
 -- Return type of all type-checking functions.
 -- state monad for global signature
 -- TODO move this somewhere
-type TypeCheck ty ext primTy primVal a =
-  StateT (Signature ty ext primTy primVal) IO a
-
-data SigDef extV extT primTy primVal
+newtype TypeCheck ext primTy primVal m a
```
andy-morris, line 471: would it be possible to move this somewhere? maybe even merge it with `Typecheck.Env.EnvTypecheck'`?

andy-morris (commented):
just the one note, otherwise looks good!

library/Core/src/Juvix/Core/IR/Types/Globals.hs
thealmarty: Not sure if it's appropriate to merge it with `EnvTypecheck`. The return type shouldn't return a type that's more than it needs I think. I could move it to the `Env` file?

andy-morris: Actually that's a good point. Moving it to `Env` as a distinct type is fine. But maybe it should be renamed to `TypecheckDecl` or something since it's going to be specific to that?

scripts/pre-run.sh
```diff
@@ -16,8 +16,10 @@ CHECK="[ci]"
```
cwgoes, line 16: why is this file changing

scripts/update-drone-signature.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 PROFILE=$1
-DRONE_URL=${2:-juvix.ci.heliax.dev}
```
cwgoes, line 4: why is this file changing

.drone.yml
```diff
@@ -15,7 +15,7 @@ steps:
     image: alpine/git:v2.30.1
     pull: if-not-exists
```
cwgoes, line 16: why is this file changing

cwgoes (approved):
ACK to the Haskell changes

I don't understand why the CI changes are in this PR, those should be separate

scripts/pre-run.sh
thealmarty: reverted

scripts/update-drone-signature.sh
thealmarty: reverted

library/Core/src/Juvix/Core/IR/Types/Globals.hs
thealmarty: Moved to `Env`. As for the name, is it OK if I leave it as is because it's the main type checking function.

PR #732, opened 2021-03-17
Run CI when a PR is merged on develop

Fraccaman:
Drone doesn't support the merge webhook sent by Github when a PR is merged. This PR triggers the CI when a PR is merged on develop. 
Not the best solution as the the 2 configurations in `.drone.yml` are duplicated (would be better to use a preprocessing language to avoid this issue) but I think this will be an issue when we will have more branches with different conditions.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -128,7 +128,138 @@ trigger:
   event:
     - pull_request
 ---
+kind: pipeline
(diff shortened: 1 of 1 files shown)


cwgoes (approved):
ACK

PR #736, opened 2021-03-18
Fix format and Haddock comments.

thealmarty:


cwgoes:
This fails CI, maybe the formatter version is different?

thealmarty:
> This fails CI, maybe the formatter version is different?

Yes I'm looking into it. I didn't change my formatter version or anything so that's a bit odd.

thealmarty:
Will reopen with a different branch.

PR #737, opened 2021-03-19
Update style guide

fdedden:
The style guide currently leaves to much room for interpretation, we need to clearify the following:

- [x] Explicit exports.
- [x] Folder names as nouns.
- [x] Folder names directly under `library/` shouldn't start with a capital.
- [x] Module names should not contain prefixes, that is what folders are for.
- [x] Use American English for folders, modules, functions and comments.
- [x] Have argument comments use `-- |` on the line before, instead of `-- ^` on the same line.
- [x] Prefix unused pattern match variables with `_` to avoid large amounts of `-Wunused-matches` warnings, while still keeping the code self-documenting.

See https://github.com/heliaxdev/juvix/issues/734 for more details on folder naming.

### DIFF
--- a/doc/STYLEGUIDE.md
+++ b/doc/STYLEGUIDE.md
@@ -41,6 +41,20 @@ library/frontend           <-- frontend is not part of the module structure.
 library/frontend/src/Juvix <-- Juvix is a module.
 ```
 
+
+### Pattern matches
+
+When pattern matching a data constructor, using variables even for unused
+fields does improve the readability of the code. To avoid numerous
+`-Wunused-matches` warnings, we prefix unused variables with `_`:
+
+```haskell
+func (RawDatatype _ _ _ lvl _) = ...
+
+func_better (RawDatatype _n _pos _a lvl _cons) = ...
+```
+
+
 Formatting
 ----------
 
--- a/doc/STYLEGUIDE.md
+++ b/doc/STYLEGUIDE.md
@@ -19,6 +19,28 @@ General naming convention
 For anything code related: modules, functions and variables, filenames etc. we
 will use the American English spelling.
 
+### Modules and directories
+
+Module and directory names should be written as nouns:
+
+```
+Translation
+Contextualization
+Desugaring
+```
+
+We should avoid prefixes in module names as much as possible, this is what sub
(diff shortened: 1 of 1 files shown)


fdedden:
Any comments on the changes?

cyberglot (commented):
We should probably rename `library` to `lib` so it is in accordance with `src` and `app`? What do you think, @cwgoes ?

fdedden:
Yes, that is actually a good one. But I don't think there is a necessity for it in the style guide.

cyberglot (approved):

cyberglot:
It's true, you're not updating any other code here. Anyway, approved now. 👍🏽

doc/STYLEGUIDE.md
```diff
@@ -13,6 +13,48 @@ Set up [HLint](https://github.com/ndmitchell/hlint)
...
+
+```
+Translation
+Contextualization
+Desugaring
+```
+
+We should avoid prefixes in module names as much as possible, this is what sub
+directories are for, i.e. `Frontend.Contextualize` instead of
+`FrontendContextualize`.
```
cwgoes, line 34:
```suggestion
We should avoid prefixes in module names as much as possible, this is what sub-directories are for, i.e. `Frontend.Contextualize` instead of `FrontendContextualize`.
```

doc/STYLEGUIDE.md
```diff
@@ -60,6 +102,37 @@ plus4 n = n + 4  -- whitespace on either side of `+`
...
+```haskell
+op ::
+  Int -> -- ^ left hand side of operator.
+  Int -> -- ^ right hand side of operator.
+  Int
+```
+
+Note that the comments are not pointing to any argument, as there are placed
+after the `->`. To fix this, the `->` should be placed at the beginning of the
+lines, and the code formatter will correctly work.
```
cwgoes, line 134:
```suggestion
lines, and the code formatter will work correctly.
```

cwgoes (approved):
ACK, thanks!

cwgoes:
> We should probably rename `library` to `lib` so it is in accordance with `src` and `app`? What do you think, @cwgoes ?

Yes, that would also be good.

PR #739, opened 2021-03-19
Converts the from frontend pass from ML to Sexp

mariari:
This PR is intended to replace the current `FromFrontend` with an identical pass in the translate package that works over SEXPS instead.


test/RecGroups.hs
```diff
@@ -33,6 +33,8 @@ pipelineOpen :: T.TestTree
 pipelineOpen =
   let correctOrder =
         [ "Foo-Helper" :| ["bar"],
+          "Foo-Helper" :| ["Bah", "fi"],
+          -- DUPLICATE!?
           "Foo-Helper" :| ["Bah", "fi"],
```
mariari, line 38: Odd for the dependency list to have repeat elements, is this intended or does this mean there is a bug somewhere else?

library/Translate/src/Juvix/Contextify/Environment.hs
```diff
@@ -219,6 +291,16 @@ lookupPrecedence name ctx = do
...
 ------------------------------------------------------------
 
+primitive :: HasClosure m => Sexp.T -> (Sexp.T -> m Sexp.T) -> m Sexp.T
+primitive (Sexp.List [name]) cont
+  | Just Sexp.A {atomName} <- Sexp.atomFromT name = do
+    -- this is bad, however it's FINE, as we only have the primitive
+    -- which no operation should disrupt
+    local @"closure" (Closure.insertGeneric (NameSymbol.hd atomName)) $ do
+      prim <- cont name
+      pure $ Sexp.list [prim]
```
andy-morris, line 301: sorry could you clarify what could (or couldn't) go wrong here?

library/Translate/test/Contextualise/Contextify.hs
```diff
@@ -36,39 +36,39 @@ sexpression =
...
+--   where
+--     test str = do
+--       Right (ctx, _) <-
+--         Contextualize.contextify (("Foo", desugared) :| [])
+--       ctx Context.!? str
+--         |> fmap Context.extractValue
+--         |> (T.@=? Just (Context.SumCon (Context.Sum Nothing "bool")))
+--     Right desugared =
+--       Desugar.op . AST.extractTopLevel
+--         <$> Parser.parse "type bool = True | False"
```
andy-morris, line 71: should these (and the imports above) be removed?

test/RecGroups.hs
andy-morris: i don't think this should happen, no

library/Translate/src/Juvix/ToCore/FromFrontend.hs
```diff
@@ -0,0 +1,799 @@
...
+  p <- ask @"param"
+  case paramConstant' p k of
+    Just x -> pure x
+    Nothing -> throwFF $ UnsupportedConstant (Sexp.Atom k)
+
+-- | N.B. doesn't deal with pattern variables since HR doesn't have them.
+-- 'transformTermIR' does that.
+transformTermHR ::
+  ReduceEff primTy primVal m => NameSymbol.Mod -> Sexp.T -> m (HR.Term primTy primVal)
+transformTermHR _ (Sexp.Atom a@Sexp.N {}) =
```
andy-morris, line 78: there's a case for `Nil` missing here

library/Translate/src/Juvix/ToCore/FromFrontend.hs
```diff
@@ -0,0 +1,799 @@
...
+  a ->
+  m [IR.RawGlobal primTy primVal]
+transformType q name _ = do
+  (ty, conNames) <- getDataSig q name
+  let getConSig' x = do (ty, def) <- getConSig q x; pure (x, ty, def)
+  conSigs <- traverse getConSig' conNames
+  cons <- traverse (uncurry3 $ transformCon q) conSigs
+  (args, ℓ) <- splitDataType name ty
+  let dat' =
+        IR.RawDatatype
```
andy-morris, line 347: The field `rawDataPos` is missing here (which was already true, sorry, but might as well fix it now).

Until the frontend has an idea of parameters vs indices, i suppose it has to just be `[]` as a placeholder though

andy-morris (changes requested):
Unless git messed up on my end, it looks like the `src/.../FromFrontend` is still there? That should probably be removed.

mariari:
> Unless git messed up on my end, it looks like the `src/.../FromFrontend` is still there? That should probably be removed.

I was getting the PR ot a reviewable point, everything from the current PR onis just going to be removing code

library/Translate/test/Contextualise/Contextify.hs
mariari: Yes, just commented them out before I was getting ready to do mass deletion and moving things around

library/Translate/src/Juvix/Contextify/Environment.hs
mariari: So we add the hd of the name so in 
```lisp
(:primitive Michelson.Foo)
```
we add Michelson to the closure, which is wrong.

However, since we just have the primitive at this point, we can't have any other symbol named Michelson in this scope, so it's not an actual issue.

mariari:
Ormulo formatting has to be done by hand, we should have a followup pr that uses the pretty printer on it

library/Translate/src/Juvix/Contextify/Environment.hs
```diff
@@ -147,9 +165,62 @@ bindingForms x =
...
+searchAndClosureNoCtx ::
+  (HasClosure f) =>
+  -- | the atom to dispatch on
+  Sexp.Atom ->
+  -- | The sexp form in which the atom is called on
+  Sexp.T ->
+  -- | the continuation of continuing the changes
+  (Sexp.T -> f Sexp.T) ->
+  f Sexp.T
+searchAndClosureNoCtx a as cont
```
cwgoes, line 208: Can we deduplicate this a little bit (the `f as cont` bits)?

cwgoes (commented):
Quick look-over, logic looks alright but needs more tests :)

library/Translate/src/Juvix/Contextify/Environment.hs
mariari: yeah I just rushed copy and pasted it cause I set it up poorly. I should have returned a function that I then applied in a more general pattern.

mariari: Okay fixed

andy-morris (approved):

PR #743, opened 2021-03-23
Handle escape characters in string parser

Acentelles:
Fixes https://github.com/heliaxdev/juvix/issues/718

Not sure if we accept both single and double quotes, but from the look of the code we do, so this PR handles both cases
```
let foo = "hello"
let bar = 'world'
```

It also fixes parsing universes. E.g.
```
let fi a = f u#a
```


mariari (approved):
Looks fine

cwgoes (commented):
ACK, thanks!

PR #745, opened 2021-03-25
Embrace multiple backends

Acentelles:
Sets pipeline infrastructure to add multiple backends through
```
class HasBackend b where
  type Ty b = ty | ty -> b
  type Val b = val | val -> b

  typecheck :: FE.FinalContext -> Pipeline (ErasedAnn.AnnTerm (Ty b) (CoreApp.Return' ErasedAnn.T (NonEmpty (Ty b)) (Val b)))
  compile :: ErasedAnn.AnnTerm (Ty b) (CoreApp.Return' ErasedAnn.T (NonEmpty (Ty b)) (Val b)) -> Pipeline Text
```

In every backend, we define a datatype that encapsulates all the different types of the backend and another datatype for the different values.

It decouples Core from Michelson by abstracting the functions involved in Compile.hs and FromFrontend.hs


app/Main.hs
```diff
@@ -109,17 +111,42 @@ run ctx opt = do
...
+  FilePath ->
+  FilePath ->
+  b ->
+  Compile.Pipeline ()
+compile fin fout _b = do
+  (liftIO $ readFile fin)
+    >>= Compile.parse
+    >>= Compile.typecheck @b
+    >>= Compile.compile @b
+    >>= Compile.writeout fout
```
mariari, line 136: Makes sense so far, rather cute how you can use `@` in this way

app/Options.hs
```diff
@@ -16,11 +22,12 @@ data Options
       }
 
 data Backend
-  = Unit
-  | Naturals
-  | Michelson
+  = Michelson BMichelson
+  -- | Plonk (BPlonk Fr)
```
mariari, line 26: Probably for this branch, best to just remove it, then get it back in your plonk branch

app/Options.hs
```diff
@@ -103,10 +110,9 @@ backendOptions =
   option
     ( maybeReader
         ( \case
-            "unit" -> pure Unit
-            "naturals" -> pure Naturals
-            "michelson" -> pure Michelson
+            -- "plonk" -> pure $ Plonk (BPlonk)
```
mariari, line 113: ditto

app/Options.hs
```diff
@@ -1,6 +1,12 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+
 module Options where
 
+import Data.Curve.Weierstrass.BLS12381 
+import Data.Data
+import Data.Field.Galois
```
mariari, line 7: Do we need curves and galois fields for this PR?

package.yaml
```diff
@@ -147,14 +147,16 @@ executables:
       - -rtsopts
       - -with-rtsopts=-N
     dependencies:
+      - aeson
+      - ansi-wl-pprint
+      - elliptic-curve
+      - galois-field
+      - gitrev
+      - haskeline
```
mariari, line 155: Probably would be best if we comment what each section of deps are for, I forgot to do that the last time I was here

src/Juvix/Core/FromFrontend.hs
```diff
@@ -41,8 +41,7 @@ paramConstant k = do
     Nothing -> throwFF $ UnsupportedConstant k
```
mariari, line 41: This entire file is going to change, so you might have to merge and deal with conflicts wtih the sexp changes that got merged

src/Juvix/Pipeline/Backend/Internal.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LiberalTypeSynonyms #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
```
mariari, line 7: Are some of these not in our default pacakge?

mariari (commented):
Yeah this seems better, always good to see things moved out of app.

There are a few points i noted about dependencies, and potential merge conflicts, I'll look at it again after that, but overall seems nice

app/Main.hs
```diff
@@ -109,17 +111,42 @@ run ctx opt = do
...
+  forall b.
+  ( Compile.HasBackend b,
+    Show (Compile.Ty b),
+    Show (Compile.Val b)
+  ) =>
+  FilePath ->
+  FilePath ->
+  b ->
+  Compile.Pipeline ()
+compile fin fout _b = do
```
cwgoes, line 131: Do we want to call `typecheck` in `compile` ? Seems there is some repeated code

src/Juvix/Pipeline/Backend/Internal.hs
cwgoes: Some of them definitely should be if they aren't, though not UndecidableInstances

src/Juvix/Pipeline/Backend/Michelson.hs
```diff
@@ -0,0 +1,75 @@
+{-# LANGUAGE FlexibleContexts #-}
```
cwgoes, line 1: Ditto on the question

src/Juvix/Pipeline/Compile.hs
```diff
@@ -0,0 +1,168 @@
+{-# LANGUAGE FlexibleContexts #-}
```
cwgoes, line 1: Ditto

cwgoes (approved):
ACK modulo minor nits - thank you @Acentelles! Much cleaner :tada:

src/Juvix/Core/Pipeline.hs
```diff
@@ -15,27 +14,19 @@ import qualified Juvix.Core.Translate as Translate
...
-import qualified Michelson.Untyped as Michelson
 
-type RawMichelson f = f Michelson.PrimTy Michelson.RawPrimVal
+type RawTerm ty val = IR.Term ty val 
 
-type RawMichelsonTerm = RawMichelson IR.Term
+type RawElim ty val = IR.Elim ty val 
 
-type RawMichelsonElim = RawMichelson IR.Elim
+type Term ty val = HR.Term ty val
```
systemf-frank, line 22: Just a general remark: do we want to use these kind of aliases? Outside of the module it is not clear that for example a `RawTerm` is equal to a `IR.Term`. I think there are some other places where similar aliases are used, and unless we expect the aliases to change, I don't see a reason to use them.

src/Juvix/Pipeline/Backend/Internal.hs
```diff
@@ -0,0 +1,30 @@
...
+import qualified Juvix.Pipeline.Internal as Pipeline
+import Juvix.Pipeline.Compile
+import qualified Juvix.Library.Sexp as Sexp
+import qualified Juvix.Core.Common.Context as Context
+
+class HasBackend b where
+  type Ty b = ty | ty -> b
+  type Val b = val | val -> b
+
+  typecheck :: Context.T Sexp.T Sexp.T Sexp.T -> Pipeline (ErasedAnn.AnnTerm (Ty b) (CoreApp.Return' ErasedAnn.T (NonEmpty (Ty b)) (Val b)))
```
systemf-frank, line 27: Do we still need `typecheck` to be part of a backend?I think the last time we discussed this, we came to the conclusion that it might be better to leave it out. Maybe it is better to do this in a separate issue though, and keep this one focused on rewriting the old backends into HasBackend.

src/Juvix/Pipeline/Backend/Michelson.hs
```diff
@@ -0,0 +1,74 @@
...
+import qualified Juvix.Backends.Michelson.Parameterisation as Param
+import qualified Juvix.Core.Application as CoreApp
+import qualified Juvix.Core.ErasedAnn as ErasedAnn
+import Juvix.ToCore.FromFrontend as FF
+import qualified Juvix.Core.IR as IR
+import qualified Juvix.Core.IR.TransformExt.OnlyExts as OnlyExts
+import qualified Juvix.Core.IR.Typechecker.Types as TypeChecker
+import Juvix.Core.IR.Types.Base
+import Juvix.Core.IR.Types.Globals
+import Juvix.Core.Parameterisation
```
systemf-frank, line 23: In the styleguide it says that will be only including `Juvix.Library` directly, and the rest in a quantified manner. I have seen lots of code that doesn't work this way, so I am not sure how strict we stick to this.

src/Juvix/Pipeline/Internal.hs
```diff
@@ -0,0 +1,61 @@
...
+import qualified Juvix.Library.Sexp as Sexp
+import qualified Juvix.ToCore.FromFrontend as FF
+import Prelude (String)
+
+data Error
+  = PipeLine Core.Error
+  | ParseErr ParserError
+  deriving (Show)
+
+toCore :: [FilePath] -> IO (Either Error (Context.T Sexp.T Sexp.T Sexp.T))
```
systemf-frank, line 22: I think we want these functions to use `Feedback`/`Pipeline` as well, but that should be done in a separate issue as well.

systemf-frank (commented):
General remark: please add explicit imports to the files.

app/Main.hs
cwgoes: I agree that we should chain them, I just want to deduplicate the code

src/Juvix/Core/Pipeline.hs
Acentelles: I agree

src/Juvix/Pipeline/Backend/Michelson.hs
Acentelles: Thanks! I'll follow the styleguide

src/Juvix/Pipeline/Backend/Internal.hs
Acentelles: Yeah, the type signature of `typecheck` involved a backend. I guess it'll be tricky to decoupled from a backend

PR #746, opened 2021-03-25
Generate golden files from parsed files.

thealmarty:
As requested by @mariari .

Once we have the `prettyPrint` instance for `TopLevel` we can generate the human readable version as well. But I don't think we have the instance now?


thealmarty:
@mariari do the updated golden files look right to you?

thealmarty:
- golden files now auto update each time the `.ju` test files are changed.
- changed the files to track `Header` instead of `[topLevel]`.
- `Header` has a show instance, using `pretty-simple` (as suggested by @Acentelles ) to generate the HRGolden files for easier debugging.
- since the files are all auto generated, git is no longer tracking them.

thealmarty:
As requested by @mariari pretty print `sexp` to `.HRGolden` instead.

thealmarty:
Track `.golden` files otherwise tests won't pass. If the `.ju` files are passed correctly, the tests will always pass because the golden files are the same as the parsed files in that case. Otherwise, errors will be put in the parsed file and no golden or `.HRGolden` files will be generated.

mariari (approved):
Seems good!

PR #747, opened 2021-03-26
Language BNF update

mariari:
This update starts the syntax changes for the language.

This pr will contain both the BNF changes, which we should talk about, and the parser changes to the best of my ability.

### DIFF
--- a/doc/Frontend/syntax.org
+++ b/doc/Frontend/syntax.org
@@ -37,9 +37,17 @@
 
   ;; Match-logic could just be a name, and so is thus fine
   ;; # is for implicit
+  ;; : is for named symbols... AKA Keywords
   ;; Should we have implicit pattern matching like here?
-  args ::= <match-logic>
-         | #<match-logic>
+  args ::= <args-bare>
+         | #<args-no-#>
+         | :<args-no-:>
+
+  args-no-: ::= #<args-bare> | <args-bare>
+
+  args-no-# ::= :<args-bare> | <args-bare>
+
+  args-bare ::= <match-logic>
 
   ;; sig only here while indent sensitivity is not here
   ;; if we only have usage in arrows, we should allow a way to have
@@ -133,8 +141,7 @@
 
 
   ;; we use mod here to disambiguate it for parsing speed
-  moduleE ::= mod <name> <args>+ = <top-level>* end in <expression>
-            | mod <name> <args>+ <cond-top>* end in <expression>
+  moduleE ::= <module> in <expression>
 
   ;; what should we allow in a module signature?
   ;; begin and end are only here while we aren't indent sensitive
@@ -184,6 +191,7 @@
                  | <list>
                  | <parens>
                  | <symbol>
+                 | <keyword>
                  ; This is useful for having nested do's or matchs
                  | <block>
                  | <do>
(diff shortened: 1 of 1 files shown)


andy-morris:
as a little bikeshed i would prefer if there was something joining the name to the value so it doesn't look like two arguments, but it's not a big deal.

mariari:
> a couple of questions:
> 
> * how are these written in types? e.g. for `let id-with-label #A :x = x`, what's the accompanying `sig`?

```haskell
sig id-with-label : #a : typ -> :x : a -> a
let id-with-label #a :x = x
```

though for real id with label we shouldn't need to 

> * it looks like `f !x` is sugar for either `f #x x` or `f :x x`. so what happens if you have `let f #x :x = …`? is that just disallowed?

yes this is disallowed, as both take the name x

> * how do you distinguish between a `:name` being a named argument vs a literal symbol? e.g. if `f :x 4` corresponds to idris `f {x=4}` (one argument, called `x`, with value `4`), then how would you write the equivalence of `f 'x 4` (two arguments, no names, passed `'x` and `4` respectively)?


You'd have to name the x if it's conflicting so

`f :x :x 4`

CL works on this same principle 

> * the BNF looks like it accepts something like `let f :(Cons x xs) = …`. is there a name missing, or what is the name of this argument?

I was just mistaken, this should just be a re-assignment of names, so `:(<symbol> = <args-bare>)`

and `:<args-bare>` should just be `:<symbol>`


> * it also accepts `let f :#x #:y = …`. what are `:#` and `#:`?

this should have gotten changed in the last push, but if it ahsn't then my bad!

mariari:
> as a little bikeshed i would prefer if there was something joining the name to the value so it doesn't look like two arguments, but it's not a big deal.

That might be hard, as I already have used named arguments in other ways, like

```
mod (foo :backend Michelson.backend) ... = ...
```

where the forms themselves can be alists, further how would that look like in the literal? Another reason is that when one wants to write a macro over stuff, they can inspect the argument list, and grab any named arguments easily enough as they are also first class values

can you show me what you're thinking of?

cwgoes:
> as a little bikeshed i would prefer if there was something joining the name to the value so it doesn't look like two arguments, but it's not a big deal.

Yeah, I also had this sense when reading - it's a bit different in e.g.

```
mod (foo :backend Michelson.backend) ... = ...
```

because there are parentheses I guess, even here I would like some kind of joiner. What about just `=`?

mariari:
> > as a little bikeshed i would prefer if there was something joining the name to the value so it doesn't look like two arguments, but it's not a big deal.
> 
> Yeah, I also had this sense when reading - it's a bit different in e.g.
> 
> ```
> mod (foo :backend Michelson.backend) ... = ...
> ```
> 
> because there are parentheses I guess, even here I would like some kind of joiner. What about just `=`, e.g.
> 
> `foo :bar=3 :baz=2 4`
> 
> To me this is clearer. It is a little extra "stuff" though, which I guess is a bit less elegant, I can see why you like this way (no `=`).

The main objection I see to using = here is that is `:bar=` the valid symbol? I think it isn't so and plus we'd want users to be able to write `:bar=3` or `:bar = 3`. While this on the surface seems fine, one will realize that in order to insert named arguments at a meta level into things it'll be disconnected into a form like `(named-argument :bar 3)`. I think for the SEXP evaluation this can go either way usability wise (a lot more verbose), as now we don't have to group the keywords by two for functions that expect them to be, however I think we'd lose something in the ML, as the user couldn't form such form at a meta level due to `=` being a trick of the parser here, rather than just being a fist class value like the various keyword symbols.

Hard to fully put my finger on,

PR #761, opened 2021-04-09
got the codebase upgraded, removed arithmetic circuit backend

mariari:
Removed the circuit backend as we are using a version not relying on it for the current one @Acentelles .

We are pegged to 17.3 and not the latest 17.9 due to morley dependencies.

The excess line changes are due to the formatter being updated


cyberglot (approved):
no major changes, most files have been re-indented and it removes files related to arithmet circuits.

cwgoes (approved):
ACK

PR #762, opened 2021-04-12
Testing Library for quickly moving around code

mariari:
This PR seeks to move all code out of Juvix/src into it's own library and then use it to create a testing target that allows us to quickly play with steps in our compiler


Acentelles:
LGTM!

Acentelles (approved):

PR #763, opened 2021-04-13
CI improvements

Fraccaman:
Improving the CI in 2 ways:

- The *cache rebuild step* is run even if one of the next steps fail
- *Formatting* and *Org-generation* steps are run by the CI if commits are coming from a PR

I was thinking about, instead of just committing changes from the formatting and org-generations steps, would it be better to squash it into the last commit? Something like:

```bash
git reset --hard HEAD~2
git merge --squash HEAD@{1}
git commit
git push
```


cwgoes (approved):
ACK

I think we should avoid the squash for now, since squashing should happen when branches are merged to develop.

Issue #765, opened 2021-04-13
Lack of Exhaustion and unknown forms (Liquid Haskell)


mariari:
When traversing down complex strands of code where you are passing forms around like a mad man, one may forget what form exactly are they taking?

1. Am I taking an atom?
2. Am I taking a list?
3. Do I have an alist or is it grouped for me already?

These kinds of questions slow down development and may cause some confusion/bugs while it's being sorted out. This is less than ideal.

However we are in luck, with newer versions of GHC there is Liquid Haskell. So With this we can very well type the size of forms and even give them names! Further if we can break down the total amount of  forms, we may get back exhaustion.

However this will need to be investigated.

Thus we will set off and try to make abstractions that make working over concrete forms easier... even if it bears in compile time costs or run time costs, that is fine and can always be optimized away later.

mariari:
Currently the blocker for this is that it turns on every module and thus behaves very poorly

mariari:
Some Issues I've found

1. In the main module of Tests, you need to have measures and the protolude in scope see  https://github.com/ucsd-progsys/liquidhaskell/issues/770
2. Lexical scoping is **not** respected

Module Juvix.Library.Sexp.Types
```haskell
{-@ type Test = {s : T | lengthM s <= 4 } @-} 
```
Juvix.Frontend.Types.Sexp
```haskell
-- Importing everything qualified
import Juvix.Library
import qualified Juvix.Library.Sexp as Sexp
import qualified Juvix.Library.Sexp as Sexp
import qualified Juvix.Library.Sexp.Types as Sexp
{-@ type Test = {s : Sexp.T | Sexp.lengthM s > 5} @-}
type Test = Sexp.T

{-@ test :: Test @-}
test :: Sexp.T
test =
  Sexp.list [Sexp.number 1, Sexp.number 2, Sexp.number 3, Sexp.number 4, Sexp.number 5]
```
If lexical scope were respected it would grab the latest test here

However we get a refinement error
```haskell
src/Juvix/Frontend/Types/Sexp.hs:(14,1)-(15,87): error: …
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : Juvix.Library.Sexp.Types.T | Juvix.Library.Sexp.Types.lengthM v == Juvix.Library.lengthL ?j}
    .
    is not a subtype of the required type
      VV : {VV : Juvix.Library.Sexp.Types.T | Juvix.Library.Sexp.Types.lengthM VV <= 4}
```

[...]

PR #766, opened 2021-04-13
Getting Refinements up and running

mariari:
Things that need to get done before we merge this PR

- [ ] Get a demo of this used on the Frontend ML --> Lisp Translation
- [ ] Figure out the trouble points
- [ ] Learn how to use Liquid more


mariari:
Tries to implement #765

cyberglot:
I think we might have uncovered a bug in liquid haskell, I will report back soon.

cyberglot:
I just had to remove termination and totality checks that it compiles normally now.

mariari:
Some other issues to look out for
```haskell
import Juvix.Library (IO)
```
this type of import is dangerous as it tries to hardwire proofs so we have to import all of our library

mariari:
Seems `Juvix.ToCore.Types.Errors` takes forever to compile with liquid, we'll likely want a flag to have it ignore such large adts... probably a max depth option

PR #767, opened 2021-04-26
Naming the various Sexp Strcutures we have

mariari:
This should help facilitate understanding of the form of the data we process

### DIFF
--- a/library/Translate/src/Juvix/Desugar/Passes.hs
+++ b/library/Translate/src/Juvix/Desugar/Passes.hs
@@ -20,9 +20,11 @@ module Juvix.Desugar.Passes
   )
 where
 
+import Control.Lens hiding ((|>))
 import qualified Data.Set as Set
 import Juvix.Library
 import qualified Juvix.Library.Sexp as Sexp
+import qualified Juvix.Sexp.Structure as Structure
 import Prelude (error)
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


library/Translate/src/Juvix/Desugar/Passes.hs
```diff
@@ -20,9 +20,12 @@ module Juvix.Desugar.Passes
   )
 where
 
+import Control.Lens hiding ((|>))
 import qualified Data.Set as Set
 import Juvix.Library
 import qualified Juvix.Library.Sexp as Sexp
+import qualified Juvix.Sexp.Structure as Struct
+import qualified Juvix.Sexp.Structure as Structure
```
Acentelles, line 28: Why do we want different names for the same imported module?

mariari: Just because `longname ^. Structure.lensName` I found was too long, however I do want to keep the explicitness for `Structure.Constructor`, `Structure.to<Name>` etc etc.

thealmarty: probably not worth it for the extra line. It creates confusion. `Structure.lensName` is probably fine.

mariari: Well the issue is, this cascades into many lines taking 3~ lines more... I might just make a lens module under it to import unqualified, as it's essentially a record lookup which shouldn't be `qualifeid` to begin with. How do others feel about modules re-exporting othes (people can import the lens module for unqualified, or have it qualified in the normal?)

library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -37,6 +37,7 @@ nameSymbolFromSymbol = NameSymbol.fromSymbol
 -- Body
 --------------------------------------------------------------------------------
 
+-- |
```
Acentelles, line 40: hehe is this on purpose?

library/Translate/src/Juvix/Desugar/Passes.hs
mariari: I've added a Lens module now that is unqualified, hope that is fine with everyone

library/Core/src/Juvix/Core/Common/Context.hs
mariari: Not particularly, I was going to comment it for the website but got sidetracked and didn't have a custom branch setup for it.

library/Translate/src/Juvix/Sexp/Structure.hs
```diff
@@ -0,0 +1,745 @@
...
+isCond (Sexp.Cons form _) = Sexp.isAtomNamed form nameCond
+isCond _ = False
+
+toCond :: Sexp.T -> Maybe Cond
+toCond form
+  | isCond form =
+    case form of
+      _Cond Sexp.:> predAns1
+        | Just predAns1 <- toPredAns `fromStarList` predAns1 ->
+          Cond predAns1 |> Just
```
Acentelles, line 437: Fascinated that your brain parses `Cond predAns1 |> Just` better that `Just $ Cond predAns1`

library/Translate/src/Juvix/Sexp/Structure.hs
```diff
@@ -0,0 +1,745 @@
+{-# LANGUAGE DeriveTraversable #-}
```
Acentelles, line 1: This is great

Acentelles (approved):
Wonderful changes. Some questions:

How does one generate the Structure.hs code?

I assume parts of the Structure.hs file are generated (some functions) and some other parts are not (helpers and data structures). What happens if you change something on the lisp generator?

library/Core/src/Juvix/Core/Common/Context.hs
Acentelles: Found it funny because the function name is `empty`

library/Translate/src/Juvix/Desugar/Passes.hs
```diff
@@ -40,44 +43,58 @@ import Prelude (error)
...
+    condToIf atom cdr
+      | Just cond <- Structure.toCond (Sexp.Atom atom Sexp.:> cdr),
+        Just last <- lastMay (cond ^. entailments) =
+        let acc =
+              Structure.IfNoElse (last ^. predicate) (last ^. answer)
+                |> Structure.fromIfNoElse
+         in foldr generation acc (initSafe (cond ^. entailments))
+              |> Sexp.addMetaToCar atom
+      | otherwise = error "malformed cond"
     --
```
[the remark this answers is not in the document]
mariari: these shouldn't happen just due to the frontend parse, which excludes these kinds of errors. So we leave it here due to exhaustion and potential mishap. but no. One thing we will do when we have flexible syntax is turn these into eithers and have good error reporting at  these spots with a generic structure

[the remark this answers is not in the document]
thealmarty: And can the error messages be more informative?

PR #778, opened 2021-05-05
New LLVM Backend

systemf-frank:
The new LLVM backend will be implemented in a gradual fashion, i.e. start simple and expand upon that.


library/Backends/llvm/stack.yaml
```diff
@@ -0,0 +1,72 @@
...
+# These entries can reference officially published versions as well as
+# forks / in-progress versions pinned to a git hash. For example:
+#
+# extra-deps:
+# - acme-missiles-0.3
+# - git: https://github.com/commercialhaskell/stack.git
+#   commit: e7b331f14bcffb8367cd58fbfc8b40ec7642100a
+#
+extra-deps:
+- ../../Core
```
mariari, line 43: You'll want this in the Packages: list. I've found this makes it so the REPL does not update when you update the Core package. took a while to figure out that Package was the right place

systemf-frank: This has been fixed by moving to the yaml-generator.

systemf-frank:
The current backend is only very basic, it supports the `add` and `sub` primitives along with int8 literals. More features will be added after the release, along with more documentation to go with the backend.

I have removed the old LLVM backend to avoid confusion in the release. If we need to refer to the code, we can do so via git. The merge is not complete 

The build is failing on the tests, as it is missing golden values for the tests that I added. Surprisingly it says these are part of the `translate` package. @Acentelles was going to take a look at those.

I will set the issue as ready for review now, as I am done for the release, unless changes need to be made.

systemf-frank:
Clarification: some tests under `test/examples/positive/llvm` do not work:
- `Const.ju` works.
- `ConstAdd.ju` fails due #806.
- `ConstAddPrim.ju` works.
- `Identity.ju` fails due to #809.
- `MonoIdentity.ju` fails due to #809.

scripts/yaml-generator/src/config.lisp
```diff
@@ -178,19 +178,8 @@
...
-                      (string->dep-sha
-                       "llvm-hs-pretty-0.6.2.0@sha256:4c600122965e8dff586bdca0044ec2b1896f2875c2da5ad89bbab9799c9697cd,1670"))))
-
-(defparameter *llvm-hs-extra-deps*
-  (make-groups :comment "LLVM-HS Library Extra dependencies"
-               :deps (list
-                      (string->dep-sha
-                       "derive-storable-0.2.0.0@sha256:a5bb3fb8feb76e95c713f3a5e401f86b7f622dd598e747cad4324b33933f27e1,2422")
-                      (string->dep-sha
-                       "derive-storable-plugin-0.2.3.0@sha256:11adeef08d4595cfdfefa2432f6251ba5786ecc2bf0488d36b74e3b3e5ca9ba9,2817"))))
```
mariari, line 192: Oddly this doesn't break any sub-packages from building!

mariari (approved):
Overall looks fine, glad to see how it evolves going forward

PR #783, opened 2021-05-07
GPL 3 to all sub packages

mariari:


andy-morris:
The files in Michelson and Core are called LI**S**CENSE atm, but otherwise this looks good

andy-morris (approved):

PR #784, opened 2021-05-12
Clean up warnings.

thealmarty:


library/Core/src/Juvix/Core/Erasure/Algorithm.hs
```diff
@@ -79,10 +78,10 @@ eraseFunction ::
   IR.Function primTy1 primVal1 ->
   m (Erasure.Function primTy2 primVal2)
 eraseFunction (IR.Function name usage ty clauses) = do
-  let (tys, ret) = piTypeToList (IR.quote0 ty)
-  clauses <- flip mapM clauses $ \(IR.FunClause _tel patts term _rhsTy _catchAll _unreachable) -> do
-    let ty_ret = listToPiType (drop (length patts) tys, ret)
-    (patts, ty) <- erasePatterns (patts, (tys, ret))
+  let (tys, ret) = piTypeToList (IR.quote ty)
+  clauses <- flip mapM clauses $ \(IR.FunClause _tel patts _term _rhsTy _catchAll _unreachable) -> do
+    -- let ty_ret = listToPiType (drop (length patts) tys, ret)
```
thealmarty, line 83: this is not used, and is the only thing that calls `piTypeToList`. @andy-morris can we delete these lines?

thealmarty, line 83: This is not used, and is the only thing that calls `listToPiType`. @andy-morris can we delete this line?

library/Core/src/Juvix/Core/Erasure/Algorithm.hs
```diff
@@ -153,11 +152,11 @@ piTypeToList ty =
       let (rest, res) = piTypeToList ret in ((usage, arg) : rest, res)
     _ -> ([], ty)
 
-listToPiType ::
-  ([(Usage.Usage, IR.Term primTy primVal)], IR.Term primTy primVal) ->
-  IR.Term primTy primVal
-listToPiType ([], ret) = ret
-listToPiType ((u, x) : xs, ret) = IR.Pi u x (listToPiType (xs, ret))
+-- listToPiType ::
```
thealmarty, line 155: Not called anywhere, except mentioned above. Can we delete these lines if there are no future plans?

cwgoes (approved):
ACK modulo the question for @andy-morris

library/Core/src/Juvix/Core/Erasure/Algorithm.hs
thealmarty: If there are no future plans, please delete all commented out lines before merging this. Thanks!

andy-morris:
removed the commented parts, as well as some redundant constraints, etc. this is ok to merge unless someone wants to have another quick look at it

PR #786, opened 2021-05-12
Yaml File Generator

mariari:
Currently we have 6~ libraries that have overlapping dependencies. Currenlty it is a pain to change all versions at the same time.

This PR stands to generate those various yaml files while giving a good way to share dependency code and extend our depdencies


scripts/yaml-generator.lisp
```diff
@@ -0,0 +1,304 @@
...
+  (make-groups :comment "Morley Specific dependencies"
+               :deps (list
+                      *tezos-bake-monitor*)))
+
+
+;; -----------------------------------
+;; stack-yaml for the YAML generation
+;; -----------------------------------
+
+(defparameter *standard-library*
```
Acentelles, line 257: Missing some libraries? (e.g. translate, core, ...)

Makefile
```diff
@@ -73,4 +73,7 @@ clean:
 clean-full:
 	stack clean --full
 
+stack-yaml:
+	ros -Q scripts/yaml-generator.lisp
```
cyberglot, line 77: This ensures it works out of the box, right?

doc/CONTRIBUTING.md
```diff
@@ -100,3 +100,16 @@ Please put the following in `.git/hooks/pre-commit` and run `chmod +x .git/hooks
...
+## New Libraries
+
+Juvix works in a rather modular design, with most loosly coupled
+features being a library in the lanugage. One can start a new library
+with `stack new <Projectname>`. This will generate a lot of cruft that
+you'll want to remove to make it look like the other projects.
+
+Please note that we have a special `stack.yaml` generator because
+there was a lot of repeat code. See the file itself for how to mock
+(big comment at the top).
```
cyberglot, line 113: Could you add a link here? 😬

scripts/yaml-generator.lisp
```diff
@@ -0,0 +1,779 @@
...
+
+;; --------------------------------------
+;; Interaction Net Groups Depencenies
+;; --------------------------------------
+
+(defparameter *interaction-net-extra-deps*
+  (make-groups :comment "For Interaction Nets json-schema"
+               :deps (list
+                      (make-dependency-github
+                       :name "cryptiumlabs/jsonschema-gen"
```
cyberglot, line 528: Shouldn't it be `heliaxdev` here?

scripts/yaml-generator.lisp
```diff
@@ -0,0 +1,779 @@
...
+
+;; --------------------------------------
+;; Stadnard Library Style Dependencies
+;; --------------------------------------
+
+(defparameter *capability*
+  (string->dep-sha "capability-0.4.0.0@sha256:d86d85a1691ef0165c77c47ea72eac75c99d21fb82947efe8b2f758991cf1837,3345"))
+
+(defparameter *extensible*
+  (make-dependency-github :name "metastatedev/extensible-data"
```
cyberglot, line 399: `heliaxdev` as well?

cyberglot (commented):
I'm a bit worried about hashes being hardcoded as strings, how do we update them if necessary?

cyberglot (approved):

Makefile
mariari: yes, as we all have `ros` on our machines due to the org generator. This just tells it to run it when you say `make stack-yaml`

doc/CONTRIBUTING.md
mariari: I've added a link to the generator file, if that is what you meant.

scripts/yaml-generator.lisp
mariari: Probably, but I was keeping the semantics the same from the old file, thus not changing any dependency

Acentelles:
Should also update https://github.com/heliaxdev/juvix/blob/develop/package.yaml#L2 accordingly

Acentelles:
Do you want to follow the convention of feat, fix, etc?

library/EasyPipeline/stack.yaml
```diff
@@ -6,92 +6,89 @@ packages:
 - ../Frontend
 - ../Core
 - ../Translate
+- ../Backends/Michelson
```
Acentelles, line 9: Remove

mariari: It's needed to make it compile

mariari: This was part of the chnages needed to make `EasyPipeline` compile

PR #787, opened 2021-05-13
Add ANF transformation for Effect Handlers

cyberglot:
CPS-transforms effect handlers and ANFfify the rest of the language.

Blocks
----
 - #888 

Blocked by
----
 - None


cwgoes:
Could you add a few testcase examples?

PR #791, opened 2021-05-20
Sexp: Documentation and Fixes

mariari:
Documented all the functions in the Sexp.hs module. Along with fixing
flatten and foldr1

### DIFF
--- a/ChangeLog.org
+++ b/ChangeLog.org
@@ -1,12 +1,12 @@
 * Changelog for Juvix
 ** 0.1.0.2
 - _Changes to =Library/Sexp.hs=_
-  + Added documentation to all functions
-  + Changed =Sexp.flatten= to be a full flatten
-  + Fixed a small bug in =Sexp.foldr1= where it behaves poorly with
+  + Adds documentation to all functions
+  + Changes =Sexp.flatten= to be a full flatten
+  + Fixes a small bug in =Sexp.foldr1= where it behaves poorly with
     dotted lists
-    * Also created a test to accommodate for this new change
-  + Added dotted list syntax to the show.
+    * creates a test to accommodate for the new change
+  + Addes dotted list syntax to the show.
     #+begin_src haskell
       λ> (Cons (atom "hi") (atom "hi"))
       ("hi" . "hi")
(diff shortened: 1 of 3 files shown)


ChangeLog.org
```diff
@@ -1,4 +1,16 @@
 * Changelog for Juvix
+** 0.1.0.2
+- _Changes to =Library/Sexp.hs=_
+  + Added documentation to all functions
```
Acentelles, line 4: Can we use present tense?

ChangeLog.org
```diff
@@ -1,4 +1,16 @@
...
+  + Changed =Sexp.flatten= to be a full flatten
+  + Fixed a small bug in =Sexp.foldr1= where it behaves poorly with
+    dotted lists
+    * Also created a test to accommodate for this new change
```
Acentelles, line 8: Words like `also` seem redundant to me and rely on context

Acentelles: I'd also avoid using `this` to refer to a previous message

ChangeLog.org
mariari: Sure will change it to present tense

ChangeLog.org
mariari: How should I mention the bullet this one is the child of?

Acentelles: I would avoid nesting. In this case:
```
- Fix: bug in =Sexp.foldr1= on dotted lists
- Test:  =Sexp.foldr1= on dotted lists
- Feat: Add dotted list syntax to the Sexp Show instance
```

Acentelles: These are just suggestions. E.g. Not sure if `Test: XXX` is a good idea. Want to know your thoughts as well

mariari: Semes to me that it just removes any structure that may exist, as now the context between the two is not obvious from a quick look. However if it is nested, then you can assume the nested items refer to the one above it, thus a test on it would intuitively refer to the entry above

mariari: Could have the nested section just start wtih `Tests` much like what I've just pushed

Acentelles (approved):

PR #793, opened 2021-05-24
Switches over the Contextify Passes to use the Structure generator

mariari:
We should not be matching on the S-expression sturcture directly as we did in this file, this pr fixes that mistake.

Finishing the rest of Desugar will also be in the PR, in the second commit.

### DIFF
--- a/library/Translate/src/Juvix/Sexp/Structure.hs
+++ b/library/Translate/src/Juvix/Sexp/Structure.hs
@@ -317,7 +317,6 @@ toArgBody form =
 fromArgBody :: ArgBody -> Sexp.T
 fromArgBody (ArgBody sexp1 sexp2) =
   Sexp.list [sexp1, sexp2]
-
 ----------------------------------------
 -- Type
 ----------------------------------------
@@ -333,7 +332,7 @@ toType :: Sexp.T -> Maybe Type
 toType form
   | isType form =
     case form of
-      _Type Sexp.:> sexp1 Sexp.:> sexp2 Sexp.:> sexp3 ->
+      _nameType Sexp.:> sexp1 Sexp.:> sexp2 Sexp.:> sexp3 ->
         Type sexp1 sexp2 sexp3 |> Just
       _ ->
         Nothing
@@ -343,7 +342,6 @@ toType form
 fromType :: Type -> Sexp.T
 fromType (Type sexp1 sexp2 sexp3) =
   Sexp.listStar [Sexp.atom nameType, sexp1, sexp2, sexp3]
-
 ----------------------------------------
 -- LetType
 ----------------------------------------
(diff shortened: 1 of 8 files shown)


Acentelles (approved):

cyberglot:
I'd need to update my stuff as well. Thanks, it looks way better now.

library/Translate/src/Juvix/Sexp/Structure.hs
```diff
@@ -702,3 +856,185 @@ toRecordNoPunned form
...
+
+isDefModule :: Sexp.T -> Bool
+isDefModule (Sexp.Cons form _) = Sexp.isAtomNamed form nameDefModule
+isDefModule _ = False
+
+toDefModule :: Sexp.T -> Maybe DefModule
+toDefModule form
+  | isDefModule form =
+    case form of
+      _DefModule Sexp.:> sexp1 Sexp.:> sexp2 Sexp.:> sexp3 ->
```
Acentelles, line 1005: This _DefModule is a weird name. It refers to the name I think. Maybe prepending name to it? E.g. _nameDefModule?

library/Translate/test/Test/Desugar/Sexp.hs
```diff
@@ -234,6 +256,6 @@ modLetWorkAsExpected =
         "(:defun foo ()\
         \   (let foo ()\
         \         (let bar () 3 \
-        \            (:let-type foo (() (XTZ))\
```
Acentelles, line 237: Was it working before as expected?

library/Translate/src/Juvix/Sexp/Structure.hs
mariari: Do you want this to be true for all names or just this one?

library/Translate/test/Test/Desugar/Sexp.hs
mariari: I'd argue no, as the form is harder to deal with in general, as the car is arguments yet the cdr is a bunch of sum types, it's just odd as is.

library/Translate/src/Juvix/Sexp/Structure.hs
Acentelles: All names. I guess it'd be a small change as it's autogenerated

library/Translate/test/Test/Desugar/Sexp.hs
mariari: Yes, I understand, the form of these aren't fully fit for purpose yet, the :module and :let-mod forms i think need a bit of tweaking. 

I'll likely have to redesign a whole lot when the syntax changes come in (the tests get the brunt of the annoying work, while the code should stay the same!)

library/Translate/src/Juvix/Sexp/Structure.hs
mariari: Those changes are now in

Issue #531, opened 2020-10-01
Pass full context through all passes instead of having special term


cwgoes:
# Description

Per discussion in architecture meeting.

Pass the full context to the typechecker, through erasure, etc. instead of having a separate term.

# Blocks

- #973 

# Blocked by

# See also

cwgoes:
To be clear, this should be using the context datastructure.

@mariari do you want to clarify any further requirements from the REPL side?

cwgoes:
Part of idealised pipeline discussions.

PR #795, opened 2021-05-26
Compile with context

Acentelles:


cwgoes:
(this aims to thread datatypes through to backends so that variable lookups of fields work)

cwgoes:
ref https://github.com/anoma/juvix/issues/531 perhaps

PR #803, opened 2021-06-05
Create a `to-fix` folder within `test/examples`

cyberglot:
`Transfer.ju` is breaking my PR #722, and after talking with @Acentelles, we've decided that just not running it is enough since it's still `to-fix`. I could alternatively move it somewhere else, but this PR documents why it is `.error` (it will be linked in the file's `git blame`) and I'm not sure whether I understand the polarity added to the test folders.

This change makes all tests pass in #722.


Acentelles:
I think it's best if we move the folder `demo-to-fix` outside `positive` so that the golden tests discovery function doesn't find them and tests don't apply to them.
```
- examples
  |- to-fix
  |- positive
  |- negative
```

Acentelles (approved):

PR #813, opened 2021-06-09
Cleaning up Context and pipeline code

mariari:
Currently we remove some repeat code in the context, `mapWithContext'`

More cleanup soon to follow


library/Core/src/Juvix/Core/Common/Context.hs
```diff
@@ -301,137 +301,120 @@ mapWithContextPure ::
...
-        Unknown u -> pure ctx
-        SumCon s@(Sum def name') -> do
-          newDef <- f def name' (NameSpace.extractValue name) s ctx
-          pure $ add name (SumCon (Sum newDef name')) ctx
-        TypeDeclar t -> pure ctx
-        CurrentNameSpace -> pure ctx
-        Information _ -> pure ctx
+    fOnSum (SumCon s) name ctx =
+      f s (NameSpace.extractValue name) ctx
+    fOnSum def _ _ = pure def
```
Acentelles, line 351: Not handling Record?

mariari: The record gets handled by the switcher, namely in mapWithCurrentName

```lisp
      -- tuple return is so we don't have to repeat work in recurseOnName
      (updatedCtx, defMaybeRecord) <- do
        newDef <- f form name ctx
        pure (add name newDef ctx, newDef)
      if
          | isRecord defMaybeRecord -> recurseOnName updatedCtx
          | otherwise -> pure updatedCtx
```
Now we don't have to worry about manually recursing it, we have a bit cleaner interface for that!

library/Pipeline/src/Juvix/Pipeline/Internal.hs
```diff
@@ -52,7 +52,7 @@ mkDef ::
   Context.SumT term1 ty1 ->
   Context.T term2 ty2 Sexp.T ->
   Maybe (Context.Def Sexp.T Sexp.T)
-mkDef typeCons dataConstructor s@Context.Sum {sumTDef, sumTName} c = do
+mkDef typeCons dataConstructor _s@Context.Sum {sumTDef = _, sumTName} c = do
```
Acentelles, line 55:
```suggestion
mkDef typeCons dataConstructor _s@Context.Sum {sumTName} c = do
```

Acentelles (approved):

PR #810, opened 2021-06-09
add pattern variable processing to transformTermIR

andy-morris:
- [ ] tests
- [ ] changelog entry

PR #821, opened 2021-06-11
Allow main to have any signature

Acentelles:
Addresses https://github.com/heliaxdev/juvix/issues/820


library/EasyPipeline/stack.yaml.lock
```diff
@@ -3,262 +3,273 @@
 # For more information, please see the documentation at:
 #   https://docs.haskellstack.org/en/stable/lock_files
 
-snapshots:
```
rokopt, line 6: Are the `stack.yaml.lock` changes related to the rest of the PR?

rokopt: A guess: maybe it's that the `import Debug.Pretty.Simple ( pTraceShowM )` requires package `pretty-compact` and that adding that caused `stack` to rearrange the `stack.yaml.lock` file for some reason?

If so, would you mind splitting the change to add that package into a separate commit (or PR) with an explanation of that, so that it's clear that the new functionality itself doesn't depend on package changes (only a temporary trace message does)?

library/Pipeline/src/Juvix/Pipeline.hs
```diff
@@ -47,9 +65,58 @@ class HasBackend b where
...
+      IR.HasSubstValue IR.NoExt (Ty b) (TypedPrim (Ty b) (Val b)) (Ty b),
+      IR.HasPatSubstTerm (OnlyExts.T IR.NoExt) (Ty b) (TypedPrim (Ty b) (Val b)) (Ty b),
+      Show (ApplyErrorExtra (Ty b)),
+      Show (ApplyErrorExtra (TypedPrim (Ty b) (Val b))),
+      Show (Arg (Ty b)),
+      Show (Arg (TypedPrim (Ty b) (Val b))),
+      IR.HasPatSubstTerm (OnlyExts.T IR.NoExt) (Ty b) (Val b) (Ty b),
+      IR.HasPatSubstTerm (OnlyExts.T IR.NoExt) (Ty b) (Val b) (Val b),
+      IR.HasPatSubstTerm (OnlyExts.T TypeChecker.T) (Ty b) (TypedPrim (Ty b) (Val b)) (Ty b)
+    ) =>
```
mariari, line 88: Kind of an unwiedly signature, this shows we may need better instnaces in some regards? Cool how GHC can include it here

library/Translate/src/Juvix/ToCore/FromFrontend/Transform/IR.hs
```diff
@@ -0,0 +1,74 @@
+{-# OPTIONS_GHC -fdefer-typed-holes #-}
```
mariari, line 1: Seems we are including #810?

library/Translate/src/Juvix/ToCore/FromFrontend/Transform/Def.hs
```diff
@@ -103,9 +104,6 @@ transformDef x def = do
         patts <- traverse transformArg args
         clauseBody <- transformTermIR q body
         pure $ IR.RawFunClause [] patts clauseBody False
-      where
-        transformTermIR q fe = do
-          hrToIR <$> transformTermHR q fe
```
mariari, line 108: Ditto here

mariari (approved):
Overall this is fine, However it seems that this PR includes #810 first, so should we just merge that then this?

The algorithm difference seems fine, though it was a bit hard to spot with all the other misc changes that were includes with this PR

PR #828, opened 2021-06-15
Context out of core

mariari:
Moves the Context code out of core and into it's own package


library/Translate/src/Juvix/FreeVars.hs
```diff
@@ -54,7 +54,7 @@ module Juvix.FreeVars where
 
 import qualified Data.HashSet as Set
 import qualified Juvix.Contextify.Environment as Env
-import qualified Juvix.Core.Common.Closure as Closure
+import qualified Juvix.Closure as Closure
```
cyberglot, line 56: shouldn't it be `Juvix.Context.Closure`?

library/Translate/src/Juvix/ToCore/FromFrontend/Transform/Def.hs
```diff
@@ -2,7 +2,7 @@ module Juvix.ToCore.FromFrontend.Transform.Def (transformDef) where
 
 import qualified Data.HashMap.Strict as HM
 import qualified Data.List.NonEmpty as NonEmpty
-import qualified Juvix.Core.Common.Context as Ctx
+import qualified Juvix.Context as Ctx
```
cyberglot, line 5: Same here.

cyberglot (commented):
I think it should be prefixed with `Juvix.Context`, so we know where the package is more easily.

mariari:
> I think it should be prefixed with `Juvix.Context`, so we know where the package is more easily.

I thought about the same, but I think in this case it'd imply the Closure system is part of the Context module.

I think the it should stay separate as it's tangentially related to the context, something we may split off into a small package some day if it makes sense.

cyberglot (approved):
G2G then.

scripts/yaml-generator/src/config.lisp
```diff
@@ -413,7 +422,8 @@ common ones to include"
                    *michelson*
                    *easy-pipeline*
                    *plonk*
-                   *llvm*)
+                   *llvm*
+                   *Context*)
```
Acentelles, line 426: Minor detail: why is it the only capitalized variable?

Acentelles (approved):

scripts/yaml-generator/src/config.lisp
mariari: Not any particular reason since it's called `Context` unlike some of the packages (which are mostly lower case in the package name) I capitalized the name there. CL doesn't care, so it's just a style thing

Issue #816, opened 2021-06-10
Michelson hasType' doesn't know LSL instruction


rokopt:
# Description

`Juvix.Backends.Michelson.Parameterisation.hasType'` doesn't yet know how to type-check some instructions, at least `LSL` -- see https://github.com/heliaxdev/juvix/issues/814#issue-917706958. 

# Blocks

- #814 

# Depends On

PR #830, opened 2021-06-15
Typecheck more Michelson instructions

rokopt:
Typecheck the Michelson LSL, SHA256, SHA512, NOW, and SET_DELEGATE instructions.

Also fail typechecking of EDIV if the type doesn't match the expected pattern (rather than throwing an "unsupported instruction" in that case).

Add tests for the new typechecking.

This addresses #816 as far as `LSL` specifically is concerned, along with a few other instructions, but there remain more Michelson instructions that aren't typechecked yet.


cwgoes (approved):
utACK

PR #836, opened 2021-06-17
Extra Documentation for HR IR conversion

mariari:


library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -37,6 +37,9 @@ hrToIRWith pats term =
     |> exec pats mempty
     |> fst
 
+-- | @hrToIR'@ transforms an HR term into an IR term. The is achieved
+-- by pushing binder's variables to the stack, then getting the index
```
rokopt, line 46: binder's -> binders'

library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -37,6 +37,9 @@ hrToIRWith pats term =
...
+-- when names occur for the proper De-bruijn index
```
rokopt, line 47: De-bruijn -> de Bruijn

rokopt, line 47: I think this is correct, but could be made more clear by pointing out that `hrToIR'` does the pushing and `hrElimToIR'` does the getting.  Based on the comment, I first looked into `hrToIR'` for where it ever got the index, and I couldn't find where -- it took me a little while to think of reading down, and found that that happens in `hrElimToIR'`'s call to `lookupName`, in its `HR.Var` case.

So I'd emphasize that the two mutually recursive functions split the functionality, analogously to other operations:  `Term`s have their types checked and `Elim`s have their types synthesized, for example; here, `hrToIR'` pushes binders onto the stack (using `withName`) and `hrElimToIR'` turns them into de Bruijn indices (using `lookupName`).

At least, I think that's what's happening :)

library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -66,6 +69,7 @@ hrToIR' = \case
     pure (IR.Let π l b)
   HR.Elim e -> IR.Elim |<< hrElimToIR' e
 
+-- | @hrElimToIR'@ works the same as @hrToIR'@ but for Elims
```
rokopt, line 77: See above -- I'd point out that there is actually an asymmetry here, with `hrToIR'` pushing, and `hrElimToIR'` generating indices.

library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -95,6 +99,10 @@ irToHRWith ::
   HR.Term primTy primVal
 irToHRWith pats t = fst $ exec pats (varsTerm t) $ irToHR' t
 
+-- | @irToHR'@ transforms an IR term into an HR one. works like
+-- @hrToIR'@ but in reverse. Namely we have binders introduce a new
+-- name to the stack, then the free var index, @n@, grabs the @nth@
+-- item in the stack
```
rokopt, line 114: I think I'd prefer the same style as I suggested in comments above -- explain the dual purposes of `irToHR'` and `irElimToHR'`, where the former generates names (using `withFresh`) while the latter looks them up (`getPatToSym`; `lookupIndex`).

library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -126,6 +134,7 @@ irToHR' = \case
       pure (HR.Let π n l b)
   IR.Elim e -> HR.Elim |<< irElimToHR' e
 
+-- | @irElimToHR'@ works the same as @irToHR'@ but for Elims
```
rokopt, line 146: (Analogous comments to above.)

rokopt (approved):

library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -143,7 +151,9 @@ irToHR' = \case
       pure (HR.Let π n l b)
   IR.Elim e -> HR.Elim |<< irElimToHR' e
 
--- | @irElimToHR'@ works the same as @irToHR'@ but for Elims
+-- | @irElimToHR'@ is the @Elim@ form of the algoirthm
+-- @irToHR'@. Namely this is responsible for relating the de bruijn
```
rokopt, line 155: Capitalize "bruijn"

rokopt (approved):

PR #842, opened 2021-06-18
Clean up usage module

andy-morris:
- [x] `Usage.T` now has one name, not three
- [x] fix bug in multiplication (0 × ω should be 0 not ω)
- [ ] come up with a test that actually fails for 0×ω=ω
- [x] make `pred` partial (what is `pred (SNat 0)`? ⊥ currently!!!) and fix uses in michelson backend
  - edit: "make it partial" as in change the return type to `Maybe T` instead of it being the _bad_ kind of partial like now


andy-morris:
replaced by #966

Issue #844, opened 2021-06-18
Pattern Vars in HR IR Conversion are dropped


mariari:
### Description

Issue #809 describes a bug where the global x is found in.

```
sig foo : int Prelude.-> int 
let foo x = x
```

However if we carefully watch this we find

```haskell
printTimeLapse "sig foo : int Prelude.-> int let foo x = x" defMichelson

...
Rec
    { recordContents = T
        { public = fromList
            [
                ( "foo"
                , Def
                    ( D
                        { defUsage = Nothing
                        , defMTy = Just
                            ( "Prelude.->" "int" "int" )
                        , defTerm =
                            ( ":lambda-case" ( ( "x" ) "x" ) )
                        , defPrecedence = Pred Left 9 } ) ) ]
        , private = fromList [] }
    , recordMTy = Nothing
    , recordOpenList = []
    , recordQualifiedMap = map }
Just
    ( CoreDef
        ( RawGFunction
            ( RawFunction
                { rawFunName = "Juvix-User" :| [ "foo" ]
                , rawFunUsage = GOmega
                , rawFunType = Pi' Omega
                    ( Elim' ( Free' ( Global ( "int" :| [] ) ) () ) () )
                    ( Elim' ( Free' ( Global ( "int" :| [] ) ) () ) () ) ()
                , rawFunClauses = RawFunClause
                    { rawClauseTel = []
                    , rawClausePats =
                        [ PVar' 0 () ]
                    , rawClauseBody = Elim'
                        ( Free' ( Global ( "x" :| [] ) ) () ) ()
                    , rawClauseCatchall = False } :| [] } ) ) )
```

We can see that it thinks `x` is a global! However if we look at [Translate](https://github.com/heliaxdev/juvix/blob/develop/library/Core/src/Juvix/Core/Translate.hs), we can see argument patterns are ignored.

Thus we can fix this bug by having `hrToIR` consider the pattern given to it.

### Blocked by
- #810 
  - This starts to substantiate some pattern variables in a generic way 
### See Also
- #809 
- #810

mariari:
Solved by #846

Issue #809, opened 2021-06-09
`Global name "x" :|[] not in scope` error for simple identity function.


systemf-frank:
When defining a simple, non polymorphic identity function, type checking fails:
```ocaml
mod Identity where

open Prelude
open Prelude.Michelson

sig intid : int -> int
let intid x = x

sig main : int -> int
let main = intid
```
Apparently, `x` is recognised as a global, which is incorrect:
```
Elim' (Free' (Global ("Identity" :| ["intid"])) ()) ()
"TypecheckerError Global name "x" :| [] not in scope"
```
Note that there is not global called `x` at all, so it is not an issue of variable shadowing.

Adding an unused polymorphic argument:
```ocaml
sig intid : (x : ty) -> int -> int
let intid _ x = x

sig main : int -> int
let main = intid int
```
gives us the same result:
```
Elim' (App' (Free' (Global ("Identity" :| ["intid"])) ()) (Elim' (Free' (Global ("Prelude" :| ["Michelson","int"])) ()) ()) ()) ()
"TypecheckerError Global name "x" :| [] not in scope"
```

Making the function fully polymorphic, gives us (obviously) the error described in #799.

systemf-frank:
Interestingly, rewriting it with an alias for `intid`:
```ocaml
sig intid : int -> int
let intid x = x

sig intid2 : int -> int
let intid2 = intid

sig main : int -> int
let main = intid2
```
Gives us a new error:
```
"InvalidInputType "Ty (TLambda (Ty TInt :) (Ty TInt :)) :""
```
But I don't think this is related to this.

For clarification, this is without the initial fix in #805. I am not sure if a proper fix for that issue could fix this particular error as well. I am pretty sure the 'not in scope' in the first post is unrelated to #805.

systemf-frank:
The initial program:
```ocaml
sig intid : int -> int
let intid x = x

sig main : int -> int
let main = intid
```
Works fine now. The one with the added polymorphism fail for a reason unrelated to this issue. The code using the `intid2` alias works as well.

This issue can be closed now.

PR #846, opened 2021-06-21
PR Fixing Pattern Variables being dropped, see #844

mariari:
Fixes the bug laid out in #809 where function arguments are being treated as global variables.

#844 is the root cause of this, and this PR serves to fix and create test cases

### DIFF
--- a/library/Core/src/Juvix/Core/Translate.hs
+++ b/library/Core/src/Juvix/Core/Translate.hs
@@ -6,6 +6,7 @@ module Juvix.Core.Translate
     irToHR,
     irToHRWith,
     hrPatternToIR,
+    hrPatternsToIR,
     hrPatternToIRWith,
     irPatternToHR,
     irPatternToHRWith,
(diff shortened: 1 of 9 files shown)


library/Core/src/Juvix/Core/Translate.hs
```diff
@@ -244,12 +259,20 @@ exec pats avoid (M env) =
...
     Env
       { nameSupply = filter (`notElem` avoid) names,
-        nextPatVar = 0,
+        nextPatVar =
+          case PM.lookupMax pats of
+            Nothing -> 0
+            Just (k, _) -> succ k,
         nameStack = [],
         patToSym = pats,
         symToPat = PM.toList pats |> map swap |> HM.fromList
```
Acentelles, line 268: This map inversion can be a handy util function

library/Core/test/Conv.hs
```diff
@@ -48,7 +56,11 @@ hrToirConversion =
               IR.Bound 0
                 `IR.App` IR.Lam (IR.Elim $ IR.Bound 0)
                 `IR.App` IR.Lam (IR.Elim $ IR.Free (IR.Global "x"))
-        )
+        ),
+      shouldConvertHRWith
+        [HR.PVar "a", HR.PVar "hi"]
+        (HR.Elim (HR.Var "a"))
+        (IR.Elim (IR.Free (IR.Pattern 0)))
```
Acentelles, line 63: It'd be nice to test your change, where you have also managed to update the counter!

library/EasyPipeline/test/FromFrontend.hs
```diff
@@ -0,0 +1,48 @@
...
+    [ T.testCase
+        "pattern vars should be tracked"
+        ( do
+            pattern' <- rawPattern
+            pattern' T.@=? rawPatternShouldBe
+        )
+    ]
+  where
+    rawPattern = do
+      Right x <- Easy.coreify "sig foo : int let foo x = x" Easy.defMichelson
```
Acentelles, line 36: It'd be nice to test `foo x y` for example as it tests the counter as well?

What about `... let foo Nothing = 0; foo (Just x) = x`?

Acentelles (approved):

Acentelles:
Maybe update the changelog as well?

Acentelles:
Can you make some golden tests as well that reflect this new change? Is there any file in the to-fix folder that can be now moved out of there?

library/EasyPipeline/test/FromFrontend.hs
mariari: I had issues `coreifying` the type declaration, but I did add the other

mariari:
> Can you make some golden tests as well that reflect this new change? Is there any file in the to-fix folder that can be now moved out of there?

No files in `to-fix` exhibit a condition that would trigger this, but let me generate a new test

library/EasyPipeline/test/FromFrontend.hs
```diff
@@ -28,21 +28,49 @@ patternVarTests =
...
+      pure (grabSingleBody (Easy.lookupCoreFunction x Easy.defMichelson "foo"))
     rawPatternShouldBe =
       IR.Elim (IR.Free (IR.Pattern 0))
+    rawPatternAdd = do
+      Right x <-
+        Easy.coreify
+          "open Prelude\
+          \ open Michelson\
+          \ open Alias\
+          \ sig foo : int -> int\
```
Acentelles, line 52: This should be `foo : int -> int -> int`

mariari: FIxed

PR #851, opened 2021-06-23
Put S-expressions in their own library

mariari:


library/Sexp/test/Main.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import qualified Sexp
+import qualified Sexp.Parser
+import qualified Sexp.SimplifiedPasses
+import qualified Test.Tasty as T
+
+allCheckedTests :: T.TestTree
+allCheckedTests =
+  T.testGroup
+    "All tests that are checked"
```
rokopt, line 12: Trivial comment here, and I know this is taken from existing code, but `"All tests that are checked"` sounds odd and non-specific to me.  Would something like simply `"S-expression library tests"` read better?

doc/Code/Library.org
```diff
@@ -1467,6 +1467,113 @@ Can be added to via core translation
...
+  + [[Parser/Internal]]
+  + [[Sexp/Types]]
+***** Types <<Sexp/Types>>
+- _Relies on_
+  + [[Library]]
+  + [[LineNum]]
+  + [[NameSymbol]]
+***** Structure
+****** Frontend <<Structure/Frontend>>
+Hardeing S-expressions into a more readable form. Here we use a
```
rokopt, line 1497: Hardeing -> Hardening

doc/Code/Library.org
```diff
@@ -1467,6 +1467,113 @@ Can be added to via core translation
...
+mixture of record structures and aliases. Each cover a form that we
+wish to talk about rather than just match away at
+- _The form for transformation follows this structure_
+#+begin_src haskell
+  -- the data type
+  data Form = ... deriving (Show)
+  is<Form>   :: Sexp.T -> Bool
+  to<Form>   :: Sexp.T -> Maybe <Form>
+  from<Form> :: <Form> -> Sexp.T
```
rokopt, line 1506: I'm guessing there are invariants that we might want to note here, such as:

```haskell
`ForAll s : SExp.T, is<Form> s = True <-> isJust (to<Form> s)`

`to<Form> . from<Form> = Just`
```

doc/Code/Library.org
```diff
@@ -1467,6 +1467,113 @@ Can be added to via core translation
...
+ 1. Figure out if we can even express a spec system in
+    Haskell... =to<Form>= and =From<From>= have the exact same signature
+ 2. replace the repeat code with the =to<Form>= with an abstraction
+ 3. put the meta data with the form so we don't have to do it by
+    hand in the code that uses this
+    1. Use =Juvix.Library.LineNum=
+    2. append the =Form= With this
+    3. have to<Form> fill this
+    4. Have extra smart consturctors that are =<form>=, so that we
+       can automatically fill in this meta data
```
rokopt, line 1554: This comment looks like it's repeated between `Frontend` and `Transition` -- should it be in a single place?

library/Sexp/src/Juvix/Sexp/Structure/generator.lisp
```diff
@@ -263,71 +263,73 @@ and a rhs that may contain a guard, so no = is assumed for the rhs"
                        (body from-body))
                   "")))))))
 
-(generate-haskell "ArgBody" '("sexp" "sexp") nil)
+(defun frontend-types ()
+  (generate-haskell "Type" (repeat 3 "sexp") "type" :list-star t)
```
rokopt, line 267: If I'm reading this file correctly, a bunch of spaces were added at the beginnings of lines, and some of the definitions were re-ordered.  Is any of that necessary, or is this some kind of reformatting accident?  Or is the file auto-generated and some structure changes made it reformat/reorder everything?

library/Sexp/src/Juvix/Sexp/Structure/Transition.hs
```diff
@@ -0,0 +1,345 @@
+{-# LANGUAGE DeriveTraversable #-}
```
rokopt, line 1: Nice, I like how splitting it up into two files makes it read.

rokopt (approved):

library/Sexp/src/Juvix/Sexp/Structure/generator.lisp
mariari: I made functions, and split it between a few files!

Only 2 forms differ both the handlers which were incorrect

doc/Code/Library.org
mariari: perhaps but for now it should stay like this I think?

PR #852, opened 2021-06-24
Minor core cleanups

andy-morris:


andy-morris:
see #894 for the only important part of this PR

PR #859, opened 2021-06-30
Separate Core.Base from Core.IR

Acentelles:
IR is a specific state of the pipeline, whereas there is a base datatype `Term` from which most stages in the pipeline (e.g. HR and Erased) extended their types using extensible-data.

It led to some confusing such as `HR.Term` being equivalent to `IR.Term' HR.T`, and one could construct signatures mixing, say, Erased and IR, where IR meant the Base Type file from which other terms are extended.

In summary, this proposes separating Core.Base from Core.IR, where IR follows the pattern seen in HR, meaning a specific extension of Core.Base. 

It also renames IR.NoExt to IR.T
```
IR.Term <=> Core.Term' IR.T
```

Open question: Do you think we can move the IR.Evaluator and IR.TransformExt  modules as well?


andy-morris (approved):
looks good!

about your question: `TransformExt` can definitely be moved into Core, but I think maybe the evaluator should actually be specialised to only work on typed syntax (later, once we can typecheck decls and `unsafeEval` becomes no longer necessary). @mariari @cwgoes does this sound right to you?

Acentelles:
Great! I'll create a PR for TransformExt. Thank you

PR #861, opened 2021-07-01
Perform substitution when typechecking dependent sigmas

### LINKED ISSUE #853: `typeTerm'`'s `Sig'` case doesn't handle dependency
The `Sig'` case of `typeTerm'` currently reads:

```haskell
    IR.Sig' π a b _ -> do
      requireZero σ
      void $ requireStar ty
      a' <- typeTerm' a ann
      b' <- typeTerm' b ann
      pure $ Typed.Sig π a' b' ann
```

That doesn't handle the possibility that the second type might depend on the first.  It should be enhanced to do so.

rokopt:
Resolves #853 

The `Sig'` case of `typeTerm'` had been missing the substitution of the argument into the possibly-dependent type.  That would cause a dependent sigma to fail when typechecked as `VStar 1`, because the typechecker would think that there was an unbound local in the type.

This PR adds tests for typechecking of dependent sigmas, as well as dependent pairs, after extending the Typechecker tests to be able to expect failure.  Then it fixes the `Sig'` case of `typeTerm'`, which is exercised by the new test of typechecking `allSig` against `starAnn`.

### DIFF
--- a/library/Core/src/Juvix/Core/IR/CheckTerm.hs
+++ b/library/Core/src/Juvix/Core/IR/CheckTerm.hs
@@ -175,7 +175,8 @@ typeTerm' term ann@(Annotation σ ty) =
       requireZero σ
       void $ requireStar ty
       a' <- typeTerm' a ann
-      b' <- typeTerm' b ann
+      av <- evalTC a'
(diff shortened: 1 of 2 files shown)


andy-morris (approved):

andy-morris:
looks good!

PR #862, opened 2021-07-01
Move TransformExt to Core.Base

Acentelles:
TransformExt is a module that converts Terms of different extensions. As it is not (or shouldn't) be bound to IR, it is moved to Core.Base.

Some functions inside that module are now abstracted.


systemf-frank (approved):

systemf-frank:
Please, for changes like these, provide some information in the PR:
- The reason for these changes.
- A short description of what has changed.

Acentelles:
Thanks @systemf-frank . Updated description

PR #863, opened 2021-07-01
Combine Erased pieces into a single module

Acentelles:
Erased is currently divided into Erased, ErasedAnn and Erasure. They constitute a set of conversions and flavours of the Erased term, that is, a term that is already typechecked and terms of some usage (0 usage terms have been removed).

- [X] Move Core.Pipeline to Erased.Ann.Convert as functions there are ErasedAnn related
- [X] Rename coreToAnn to irToErasedAnn, which now lives in Erased.Ann.Convert module


library/Backends/Plonk/src/Juvix/Backends/Plonk/Pipeline.hs
```diff
@@ -15,15 +15,15 @@ import qualified Juvix.Backends.Plonk.Dot as Dot
 import qualified Juvix.Backends.Plonk.Parameterization as Parameterization
 import qualified Juvix.Backends.Plonk.Types as Types
 import qualified Juvix.Core.Base.TransformExt.OnlyExts as OnlyExts
-import qualified Juvix.Core.ErasedAnn.Types as CoreErased
+import qualified Juvix.Core.Erased.Ann as ErasedAnn
+import qualified Juvix.Core.Erased.Ann.Types as ErasedAnn
```
andy-morris, line 19: it looks like `Erased.Ann` reexports `Types`, so this import might be redundant?
```suggestion
```

andy-morris (changes requested):
just one tiny thing, but the rest looks good to me!

library/Backends/Plonk/src/Juvix/Backends/Plonk/Pipeline.hs
Acentelles: Updated @andy-morris

andy-morris (approved):

Issue #854, opened 2021-06-28
Golden:  `stack test --test-arguments "--accept"` Does not always generate new files


mariari:
One Annoying Thing is that running

```
 stack test --test-arguments "--accept"
```

does not always force new files, we should have a `make generate-golden-tests` that does this for us.

PR #875, opened 2021-07-08
Force Golden Tests to be updated even when files exist

mariari:
Adresses the issue in #854

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,9 @@ stack-yaml:
 accept-golden:
 	rm -rf test/examples-golden/positive
 	rm -rf test/examples-golden/negative
-	-stack test --test-arguments "--accept"
+	# If we want further commands, put - at the start, to ignore
+	# the error from this command
+	stack test --test-arguments "--accept"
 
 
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -78,6 +78,10 @@ stack-yaml:
 
 # Overwrite existing golden files
 accept-golden:
+	rm -rf test/examples-golden/positive
+	rm -rf test/examples-golden/negative
+	# If we want further commands, put - at the start, to ignore
+	# the error from this command
 	stack test --test-arguments "--accept"
```
Acentelles, line 85: Do you have any reproducible error?

mariari: Not any easy ones. This mainly comes out of when we refactor tests that the  `stack test --test-arguments "--accept"` line alone does not cause already written files to always be overwritten.

Also I think this has the side benefit of catching when we delete some tests like `.circuit`, as they won't be re-generated

Acentelles: Yes, I can see its benefits 👍

Acentelles (approved):

Issue #882, opened 2021-07-13
Move Code into Experimental


mariari:
Namely we need to move

- Encoding (inside pipline)
- Interaction Net Libraries

PR #883, opened 2021-07-13
Move Interactionnet and Encoding code to the experimental directory

mariari:
See #882


scripts/yaml-generator/src/config.lisp
```diff
@@ -320,26 +320,6 @@ common ones to include"
                      *standard-library-extra-deps*
                      *eac-solver*)))
 
-(defparameter *interaction-net*
```
Acentelles, line 323: I assume it's too much work to also keep the experimental libraries under the yaml generator

Acentelles (approved):

scripts/yaml-generator/src/config.lisp
mariari: I thought about it.

In this case we easily could. As `*interaction-net*` and `*interaction-net-ir*` only depend on packages in libraries. However until I allow packages to be anywhere and resolved in a smarter way, I'd rather not rely on this fact (I think it is likely wrong as `*interaction-net*` should rely on it's own IR!)

PR #884, opened 2021-07-15
HR Sexp Layer

mariari:
This PR creates the HR Sexp layer according to the chart documentation, [here](https://heliaxdev.github.io/juvix-docs/docs/design/s-expression/hr/)


library/Sexp/src/Juvix/Sexp/Structure/CoreNamed.hs
```diff
@@ -0,0 +1,441 @@
...
+--     4. Have extra smart consturctors that are =<form>=, so that we
+--        can automatically fill in this meta data
+module Juvix.Sexp.Structure.CoreNamed where
+
+import Juvix.Library hiding (Meta, fromInteger, toInteger)
+import qualified Juvix.Library.NameSymbol as NameSymbol
+import qualified Juvix.Sexp as Sexp
+import Juvix.Sexp.Structure.Helpers
+
+newtype Star = Star {starUniverse :: Integer} deriving (Show)
```
Acentelles, line 38: Why don't we use the types from Core.Base directly, instead of creating new ones here?

Acentelles:
I don't understand the motivation of this PR. How will the s-expression equivalent of core value help us?

mariari:
> I don't understand the motivation of this PR. How will the s-expression equivalent of core value help us?

This gives us the basis of the target that the code that was in `fromFrontend` should target. This makes it so the step from `Sexp.T` to `Core HR.T` very easy, further it gives us direction on the forms we should desugar to. This means that we have a very clear goal of what the S-expressions should start to morph into.

For example take the `Let` form, currently we have pattern matching in let, however what we have to go down to is [to this form here](https://heliaxdev.github.io/juvix-docs/docs/design/s-expression/hr/#headline-10). In the idealized pipeline this is a few steps (namely `localLetsIntoLambdas`, `lambdaLiftLocalMatches`, and `removeGuardsAndSimplifyLetToSingleCase`). However if we look at what core currently can support these don't work too well.  Thus any attempt to do it as a straight shot to core (like we currently do), will end with a lot of complicated logic or simply not support it.

Further, I'd like to note this is the level @rokopt is targeting with his formalization of Sexpressions. This gives him a very direct target so he doesn't have to do the more laboious task.

library/Sexp/src/Juvix/Sexp/Structure/CoreNamed.hs
mariari: Sadly targeting `Core.Base` is hard in the sense that we have many unsupported forms, and this leads into targetting them easier!

rokopt (approved):
I think this looks really good -- thanks for taking care of it.

The code duplication in the `to` forms is actually the exact kind of thing which should be possible to factor out as a transformer with the new S-expression library.  I've written code to match S-expressions against other S-expressions in other contexts, so I'm reasonably sure that will be feasible. :)

doc/Code/Library.org
```diff
@@ -1279,6 +1279,39 @@ sexp type and all the various helper functionality one can need
...
+  from<Form> :: <Form> -> Sexp.T
+#+end_src
++ With the following properties of the forms
+  #+begin_src haskell
+    ∀ s : Sexp.T. is<Form> s = True ⟷ is-just (to<Form> s)
+    to<Form> 。 from<Form> = Just
+  #+end_src
+_TODO_
+ 1. Figure out if we can even express a spec system in
+    Haskell... =to<Form>= and =From<From>= have the exact same signature
```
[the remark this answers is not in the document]
mariari: We could generate tests that do this actually!

PR #886, opened 2021-07-16
Remove FromFrontend/ToCore

Acentelles:
Continuation of the work of making the code reflect the diagram below. Some of the work was discussed during EthCC with @mariari 
Screenshot 2021-07-26 at 11 23 09

- Remove ToCore inside the Translate module and create Pipeline/ToHR instead
- Separate temporary functions, datatypes, ... that live only inside a certain algorithm, such as in frontendToHR, to persisting code that belongs to a state in the pipeline. For example, Special def is only used in the algorithm from Sexp to HR and discarded afterwards, whereas the so called CoreDef is a RawGlobal that defines the HR state
- Remove `'` (ticks) from Core.Base types: This requires Andy's approval, since he's the creator of the library and this PR overwrites the default config value of `extensible`
- Add qualified imports
- Refactor Pipeline implementation such as typecheck
- Remove some re-exporting for consistency of imports. E.g. 
    - TC.Annotation (TC referring to CheckTerm)
    - IR.Annotation
    - Typed.Annotation (Typed referring to Typechecker)


mariari (approved):
A lot of code changes, overall it seems good. It's a good step forward for the compiler

cwgoes (commented):
Concept ACK; should see review from @andy-morris

library/EasyPipeline/src/Easy.hs
```diff
@@ -45,9 +45,10 @@ import qualified Juvix.Library.HashMap as Map
 import qualified Juvix.Library.NameSymbol as NameSymb
 import qualified Juvix.Pipeline as Pipeline
 import qualified Juvix.Pipeline.Compile as Compile
-import qualified Juvix.Pipeline.Core as Core
+import qualified Juvix.Pipeline.ToHR as ToHR
+import qualified Juvix.Pipeline.ToIR as ToIR
 import qualified Juvix.Sexp as Sexp
-import qualified Juvix.ToCore.Types as ToCore.Types
+-- import qualified Juvix.ToCore.Types as ToCore.Types
```
andy-morris, line 51: can this just be removed?
```suggestion
```

library/Pipeline/src/Juvix/Pipeline/ToHR/Sig.hs
```diff
@@ -1,35 +1,29 @@
-module Juvix.ToCore.FromFrontend.Transform.Sig (transformSig) where
+{-# LANGUAGE LiberalTypeSynonyms #-}
+{-# LANGUAGE RecordWildCards #-}
```
andy-morris, line 2: can these be moved to the `package.yaml`? `UndecidableInstances` i was asked before to only enable explicitly in files that need it but these other two are probably fine

library/StandardLibrary/src/Juvix/Library/LineNum.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE TemplateHaskell #-}
```
andy-morris, line 3: can these be moved to `package.yaml` too?

andy-morris (approved):
looks good! a couple of tiny nitpicks but i selected approve anyway since what you want to do with those things is up to you

PR #889, opened 2021-07-26
HTTP Server

Acentelles:
There are some advantages to provide an HTTP server that exposes the compilation steps in its API:
- Online web editor
- Visualize diffs between compilation steps
- Identify why a certain step didn't work (e.g. toIR was successful but toErased wasn't. We could edit the IR code)
- Language Server Protocol
- Witch interactive mode tooling

Due specifically to Servant (https://docs.servant.dev/en/stable/) and swagger:
- Generate client code (e.g. typescript)
- Documentation


library/HTTP/src/Juvix/HTTP/Routes.hs
```diff
@@ -0,0 +1,72 @@
...
+           toCircuit
+    where
+        toML = notImplemented
+        toSexp = notImplemented
+        toHR = notImplemented
+        toIR = notImplemented
+        toErased = notImplemented
+        toCircuit = notImplemented
+
+
```
mariari, line 70: Does this do dynamic type checking to properly be able to determine which alternative to call?

scripts/yaml-generator/src/config.lisp
```diff
@@ -182,7 +182,7 @@
                       (make-dependency-bare :name "llvm-hs-pretty-0.9.0.0"))))
 
 ;; --------------------------------------
-;; Interaction Net Groups Depencenies
+;; Interaction Net Groups Dependencies
```
mariari, line 185: Thanks for these catches

mariari (commented):
Interested in Seeing how This turns out.

It seems the HTTP is a template, but should look good once we put our own data in

Acentelles:
Future short term improvements:
- [ ] Choose backends
- [ ] Use pretty show instead of the json one (it's too verbose) for the compiled steps
- [ ] Format error messages
- [ ] Show circuit graph

cwgoes:
@Acentelles should we review and potentially merge this, or do you want to make more changes first?

Acentelles:
Yes, please. I cleaned it up and removed any unfinished work like swagger related code for future work

cwgoes:
> Yes, please. I cleaned it up and removed any unfinished work like swagger related code for future work

Thank you!

mariari:
iI can't get the `compiler-debugger` to work namely due to not having react-scripts, what dependency leads to that?

mariari:
Upon running the server (not the debugger) it seems to just have the page load forever, am I doing anything wrong?

mariari (approved):

PR #892, opened 2021-07-28
Records into Σ types

mariari:
This PR gives us the transformation to have records be considered Σ types.

The Pass happens in the `Context` pass level, and not `Desugar` as we wish to record infromation before running it (see #895). The pass itself does not rely on anything related to the context besides that

### Blocking
- #893
- #895 
   - This needs the structure general changes, along with this PR to have the passes done make any sense 

### Blocked on
- #884


library/Translate/src/Juvix/Contextify.hs
```diff
@@ -37,14 +37,14 @@ op names = do
   case context of
     Left err -> pure $ Left err
     Right ctx -> do
-      (newCtx, _) <- Environment.runMIO (Passes.resolveModule ctx)
+      (newCtx, _) <- -- only Passes.resolveModule requires IO in these passes
+        Environment.runMIO $
+          Passes.resolveModule ctx
+            >>= Passes.inifixSoloPass
+            >>= Passes.recordPi
```
mariari, line 44: Should we mention somehow which are `runM` or `runMIO`? 

It becomes more complciated, but it's good to note which need IO or not

library/Translate/src/Juvix/Contextify.hs
```diff
@@ -37,14 +37,14 @@ op names = do
...
       case newCtx of
         Left err -> pure $ Left $ PassErr err
-        Right t ->
-          let (infix', _) = Environment.runM (Passes.inifixSoloPass t)
-           in case infix' of
-                Left err -> pure $ Left $ PassErr err
-                Right x -> pure (Right x)
+        Right t -> pure (Right t)
```
Acentelles, line 47: Wonder if we could have in out standard library a function `left` and `right` 
```
left :: (a -> c) -> Either a b -> Either c b
left = flip bimap identity
```

so that these cases
```
      case newCtx of
        Left err -> pure $ Left $ PassErr err
        Right t -> pure (Right t)
```
become
```
      pure $ left PassErr newCtx
```

library/Translate/src/Juvix/Contextify/Passes.hs
```diff
@@ -9,25 +14,28 @@ import qualified Juvix.Contextify.InfixPrecedence.ShuntYard as Shunt
...
 
 type ExpressionIO m = (Env.ErrS m, Env.HasClosure m, MonadIO m)
 
 type Expression m = (Env.ErrS m, Env.HasClosure m)
 
+--------------------------------------------------------------------------------
+-- Open Transformation
+--------------------------------------------------------------------------------
+
+resolveModule ::
```
Acentelles, line 34: Can we add a brief description?

library/Translate/src/Juvix/Contextify/Passes.hs
```diff
@@ -105,3 +122,45 @@ convertShunt :: Shunt.Application Sexp.T Sexp.T -> Sexp.T
...
+-- | @recordPi@ - Transforms record declarations into @Pi@ Type
+-- expressions
+-- - BNF Input Form
+--   1. (:record-d arg-1 type-1 … arg-n type-n)
+-- - BNF Output Form
+--   1. (:sigma (arg-1 :usage-hole type-1)
+--         (:sigma (arg-2 :usage-hole type-2)
+--            … (:sigma (arg-n :usage-hole type-n)
+--                  :unit)))
+recordPi ::
```
Acentelles, line 145: Shouldn't this be recordSigma (or recordPiSigma)?
Screenshot 2021-08-02 at 00 18 34

mariari: Yes I should call it recordSigma instead of `recordPi` good find.

The `Pi` part shouldn't be done here, as there is a generic encoding for the arguments that holds true regardless if it's a record or not

library/Translate/test/Test/Golden.hs
```diff
@@ -258,22 +259,33 @@ contextifySexp names = do
...
+  m Environment.SexpContext
+resolveRecordPi =
+  "can't resolve record to Sigma"
+    |> runPass (resolveInfixContext, Contextify.recordPi)
+
+runPass ::
+  (MonadIO m, MonadFail m) => (t1 -> m t2, t2 -> Environment.MinimalMIO b) -> String -> t1 -> m b
+runPass (previousPass, currentPass) errString names = do
+  ctx <- previousPass names
+  -- Only the module transform requires IO here!
```
Acentelles, line 285: Can you remind me why `recordQualifiedMap` is an STM map?

library/Translate/test/Test/Golden.hs
```diff
@@ -207,7 +207,8 @@ discoverContext =
   discoverPrefix
     [ (contextifySexp, "contextify-sexp"),
       (resolveModuleContext, "resolve-module"),
-      (resolveInfixContext, "resolve-infix")
+      (resolveInfixContext, "resolve-infix"),
+      (resolveRecordPi, "recordToPi")
```
Acentelles, line 211:
```suggestion
      (resolveRecordPi, "record-to-pi")
```

library/Translate/test/Test/Golden.hs
mariari: need pointers for open maps and reverse dependencies. It allows us when we have an interactive mode to quickly be able to determine if new definitions cause ambiguities among other benefits

Acentelles:
It'd be great to add test examples on records and check their golden tests

Acentelles:
What are our plans on named records?

mariari:
> It'd be great to add test examples on records and check their golden tests

This might be hard, do we have golden tests, that succeed up to pass `N`, I could add those but they'll fail the next passes

library/Translate/src/Juvix/Contextify.hs
Acentelles: Maybe 
```
mapLeft :: (a -> c) -> Either a b -> Either c b
mapLeft = flip bimap identity
```
and
```
left :: (a -> c) -> Either a b -> c
left = flip either identity
```
(same on the right)

library/Translate/src/Juvix/Contextify/Passes.hs
mariari: Will do in #895! Should have done that a while ago!

mariari:
Closing for the last comment in #893, and merging the useful changes into #895.

PR #894, opened 2021-07-29
remove reduntant translation ir→hr→ir

andy-morris:
in `Juvix.Core.Erased.Ann.Conversion`, `typecheckErase'` was converting an IR term to HR, and passing it to `typecheckEval`, which just converted it right back


andy-morris:
ugh sorry this is also part of #886

PR #897, opened 2021-07-30
Add S-exp docs and create ToSexp module

Acentelles:
- Update the pipeline documentation with the Translate module (i.e. S-exp transformations).

- Remove FrontendDesugar and move ML AST to S-exp pass outside Frontend module. Remove sexp dependencies from Frontend


mariari (approved):
Overall Looks good!

Fits the overall plan of the PDF. While we're at it, can we split the First Sexp Diagram into
```
S-expression
  |
  | (desugar)
  |
  ∨
Desugared S-expressions
  |
  | (Contextify.op)
  |
  ∨
Context S-expression
  |
  | (toHR)
  |
  ∨
Core HR
```

Maybe we can add one more layer to this?

Either way this was a needed change

Acentelles:
Thanks! Yes, good suggestion

PR #750, opened 2021-04-01
Debuggging tests

mariari:
provided in this pr is every step to reproduce the bug for applications of 3 or greater errroing

simply load `app/Test.hs` into your repl and call coreify you'll get

```haskell
λ> coreify
"going into res"
"start term"
"typecheckErase': ty started"
"start Ir.typeTerm"
"in typeTerm!' :-----"
Pi' Omega (Elim' (App' (App' (Free' (Global ("Add" :| ["pair"])) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) ()) (Elim' (App' (App' (Free' (Global ("Add" :| ["pair"])) ()) (Elim' (App' (Free' (Global ("Add" :| ["list"])) ()) (Elim' (Free' (Global ("Add" :| ["operation"])) ()) ()) ()) ()) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) ()) ()
"in typeTerm!' :-----"
Elim' (App' (App' (Free' (Global ("Add" :| ["pair"])) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) ()
"call into elim"
"start elim"
App' (App' (Free' (Global ("Add" :| ["pair"])) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()
"in application("
"start elim"
App' (Free' (Global ("Add" :| ["pair"])) ()) (Elim' (Free' (Global ("Add" :| ["int"])) ()) ()) ()
"in application("
"start elim"
Free' (Global ("Add" :| ["pair"])) ()
"in a global"
"Out of the global freely!"
"in typeTerm!' :-----"
Elim' (Free' (Global ("Add" :| ["int"])) ()) ()
"call into elim"
"start elim"
Free' (Global ("Add" :| ["int"])) ()
"in a global"
"Out of the global freely!"
"finished elim"
"end of application)"
"in typeTerm!' :-----"
Elim' (Free' (Global ("Add" :| ["int"])) ()) ()
"call into elim"
"start elim"
Free' (Global ("Add" :| ["int"])) ()
"in a global"
"Out of the global freely!"
"finished elim"
"end of application)"
"finished elim"
[...]

PR #900, opened 2021-08-02
Fix nesting operations issue

Acentelles:
Simple operations like
```
sig main : field -> field -> field -> field
let main x y z = x + y
```
compiled, whereas more complex (nested) ones didn't
```
sig main : field -> field -> field -> field
let main x y z = x^2 + y^2 + z^2
```

This fix allows to compile an arbitrary number or operations.
See https://github.com/anoma/juvix/pull/900/files#diff-89b66698a8860c8b48518210cac3b4a57140cb82355d5240a8235fb189958ae6 as an example file that compiles.

See https://github.com/anoma/juvix/pull/750 for reference.


library/Backends/Plonk/src/Juvix/Backends/Plonk/Parameterization.hs
```diff
@@ -154,16 +155,15 @@ arityRaw = \case
...
-  Just $
-    App.TermArg $
-      App.Take
-        { usage = Usage.Omega,
-          type' = retType,
-          term = retTerm
-        }
+toArg :: App.Return' ext1 ty term -> Maybe (App.Arg' ext1 ty term)
+toArg = Just . App.TermArg
+
```
mariari, line 160: I'm guessing we can do this more minimal version as we keeping the same extension rather than changing it?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Parameterization.hs
```diff
@@ -174,7 +174,10 @@ toTakes App.Return {retType, retTerm} = (fun, [], arityRaw retTerm)
 fromReturn :: Return' ext f -> PrimVal' ext f
 fromReturn = identity
 
-instance App.IsParamVar ext => Core.CanApply (PrimVal' ext f) where
+takeToReturn :: App.Take ty term -> App.Return' ext ty term
+takeToReturn (App.Take {type', term}) = App.Return {retType = type', retTerm = term}
```
mariari, line 178: Seems we have this function here as well!

library/Backends/Plonk/src/Juvix/Backends/Plonk/Parameterization.hs
```diff
@@ -197,14 +200,87 @@ instance App.IsParamVar ext => Core.CanApply (PrimVal' ext f) where
...
+    goPi (ErasedAnn.Pi _ s t) = NonEmpty.cons <$> goPrim s <*> go t
+    goPi _ = Nothing
+    goPrim (ErasedAnn.PrimTy p) = Just p
+    goPrim _ = Nothing
+
+applyPrimOnArgs :: Types.AnnTerm f -> [Types.AnnTerm f] -> Types.AnnTerm f
+applyPrimOnArgs prim arguments =
+  let newTerm = ErasedAnn.AppM prim arguments
+      retType = ErasedAnn.piToReturnType (ErasedAnn.type' prim)
+   in ErasedAnn.Ann one retType newTerm
```
mariari, line 283: Are these related to the fix at all?

library/Core/src/Juvix/Core/Erased/Ann/Conversion.hs
```diff
@@ -182,13 +192,21 @@ toRaw t@(ErasedAnn.Ann {term}) = t {ErasedAnn.term = toRaw1 term}
...
           -- a well typed application can't have more arguments than arrows
           undefined
+        returnToTerm (App.Return ty term) =
+          ErasedAnn.Ann
+            { usage = Usage.Omega, -- FIXME should usages even exist after erasure?
+              type' = ErasedAnn.fromPrimType ty,
+              term = ErasedAnn.Prim term
+            }
+        returnToTerm (App.Cont take args nat) = notImplemented
+
```
mariari, line 209: This function shows up again. 

We should unify this with the other one, as it's hard to maintain copies.

Also in this case, could we pass in an extra usage like we did with the other one, or do we not have that in scope anywhere?

library/Backends/Plonk/src/Juvix/Backends/Plonk/Parameterization.hs
```diff
@@ -197,14 +200,87 @@ instance App.IsParamVar ext => Core.CanApply (PrimVal' ext f) where
...
+
+-- | Translate a 'Return' into a 'RawTerm'.
+takeToAnnTerm :: Take f -> AnnTerm f
+takeToAnnTerm App.Take {type', term} =
+  ErasedAnn.Ann
+    { usage = Usage.Omega,
+      type' = Prim.fromPrimType type',
+      term = ErasedAnn.Prim term
+    }
+
```
mariari, line 241: Seems this function also shows up here as well

library/Core/src/Juvix/Core/IR/Evaluator/SubstV.hs
```diff
@@ -405,7 +416,9 @@ argToValue ::
   App.Arg (Param.PrimType primTy) primVal ->
   IR.Value primTy (Param.TypedPrim primTy primVal)
 argToValue = \case
-  App.TermArg (App.Take {type', term}) ->
-    IR.VPrim $ App.Return {retType = type', retTerm = term}
+  App.TermArg (App.Return {retType, retTerm}) ->
+    IR.VPrim $ App.Return {retType, retTerm}
+  App.TermArg (App.Cont {fun, args, numLeft}) ->
+    notImplemented
```
mariari, line 422: We should probably note where all these unimplemented parts are and try to see if we can centralize this notion, as it seems to spread out all over the place, which makes maintaining harder.

mariari (changes requested):
It seems overall this PR does fix the nesting issues, however it's hard to see exactly how with all the other unreleated changes along with them.

Also it seems the conversion code is spread out between many different modules with varying levels of comments about baked in behavior.

Would be cool if we had a single module that had this change along with a variant that took a usage or non at all

library/Core/src/Juvix/Core/IR/Evaluator/SubstV.hs
Acentelles: The `notImplemented` bits have to do with handling `Cont`. Could we leave the implementation of those to a separate PR?

mariari: I'll make a PR myself after merging this, so I don't lose where this particular issue scatters to

mariari (approved):
I will merge this come monday after merging develop

PR #902, opened 2021-08-03
Feedback Error on HR Error

mariari:


systemf-frank (approved):

library/Pipeline/src/Juvix/Pipeline/ToHR/Env.hs
```diff
@@ -117,6 +117,16 @@ evalEnv ::
 evalEnv ctx param env =
```
Acentelles, line 117: Don't we want this applied to evalEnv and execEnv too?

mariari: `execEnv` Already acts like this, giving back the either. As for evalEnv, since `eval<Monad>` and `exec<Monad>` have a standard meaning, I decided not to change it's typical meaning, but rather make a newer more explicit version

Acentelles: I can't see why we want to ignore the errors as the default case

PR #805, opened 2021-06-08
Recurse after inlining a name

andy-morris:
Fixes #806

Previously, the `inlineAllGlobals` function stopped after performing an inlining. Now it continues on names found inside the definition being inlined.

Issue #905, opened 2021-08-04
InsufficientUsage error for `main = nil-int`


systemf-frank:
The following program fails with an `InsufficientUsage` error:

```ocaml
mod Nil-Fail where

open Prelude
open Michelson

sig nil-int : list int
let nil-int = nil int

sig main : list int
let main = nil-int
```

```
$ stack run -- typecheck -b michelson nil-fail.ju 
"Error: TypecheckerError
    ( InsufficientUsage
        { usageNeeded = Omega
        , usageActual = SNat 0
        }
    ) on Term: Elim
    ( Free
        ( Global
            ( "Nil-Fail" :| [ "nil-int" ] )
        ) ()
    ) ()"
```

(Note that this only occurs after fixing #904)

systemf-frank:
During a discussion with @mariari, we came up with a more self-contained version of the code above that uses some explicit usage information:
```ocaml
mod Nil-Fail where

open Prelude

let (->>) = arr 0

sig list : ty ->> ty
let list = %Michelson.list

sig nil : (a : ty) ->> list a
let nil _ = %Michelson.nil

sig int 0 : ty
let int = %Michelson.int

sig nil-int : list int
let nil-int = nil int

sig main : list int
let main = nil-int
```
Gives us (with some strategic debugging traces):
```
"Prim Ty Call"
VPi (SNat 0) (VStar 0 ()) (VStar 0 ()) ()
"requires the star"
"Prim Ty Call"
VStar 0 ()
"requires the star"
"Prim Ty Call"
VPi (SNat 0) (VStar 0 ()) (VStar 0 ()) ()
"requires the star"
"Error: TypecheckerError
    ( EvalError
        { evalErr = CannotApply
            { fun = VPrimTy List ()
            , arg = VNeutral
                ( NBound 0 () ) ()
            , paramErr = NoApplyError
            }
        }
    ) on Term: Elim
    ( Free
        ( Global
            ( "Nil-Fail" :| [ "nil-int" ] )
        ) ()
```

systemf-frank:
> During a discussion with @mariari, we came up with a more self-contained version of the code above that uses some explicit usage information:
> 
> ```ocaml
> mod Nil-Fail where
> 
> open Prelude
> 
> let (->>) = arr 0
> 
> sig list : ty ->> ty
> let list = %Michelson.list
> 
> sig nil : (a : ty) ->> list a
> let nil _ = %Michelson.nil
> 
> sig int 0 : ty
[...]

PR #906, opened 2021-08-05
Implemented correct typechecking of polymorphic primitives.

systemf-frank:
The current implementation expected every primitive to be of sort `*`. Polymorphic primitives were not supported because of that. This commit fixes this by checking if all arguments to the primitive are `*`.

### Note that this PR should be merged to andy-morris/inlineAllGlobals (see #805), not to develop.


systemf-frank:
This fix works fine for now, but we should discuss these TODO's:
- Check if being right-associative is enough (that all primitives are defined this way).
- Check what to do with the information that we throw away using void.
- What to do with the usage information?

cwgoes:
Needs tests.

andy-morris (approved):

systemf-frank:
For now the nil test (part of https://github.com/anoma/juvix/pull/928) will be used as a test once that has been merged here.

systemf-frank:
The tests are failing because of the issue mentioned in https://github.com/anoma/juvix/issues/905. Taking that example:
```haskell
mod Nil-Fail where

open Prelude

let (->>) = arr 0

sig list : ty ->> ty
let list = %Michelson.list

sig nil : (a : ty) ->> list a
let nil _ = %Michelson.nil

sig int 0 : ty
let int = %Michelson.int

sig nil-int : list int
let nil-int = nil int

sig main : list int
let main = nil-int
```
throws another error now:
```haskell
Error: TypecheckerError
    ( EvalError
        { evalErr = ErrorValue
            ( CannotApply
                { fun = VPrimTy
                    ( Return
                        { retType = PrimType
                            { getPrimType = STAR :| [] }
                        , retTerm = List
                        }
                    ) ()
                , arg = VPrimTy
                    ( Return
                        { retType = PrimType
                            { getPrimType = STAR :| [] }
                        , retTerm = PrimTy ( Ty TInt : )
                        }
                    ) ()
                , paramErr = ApplyErrorT
                    ( Extra
                        ( TooManyPrimArguments
                            ( Take
                                { usage = SAny
                                , type' = PrimType
                                    { getPrimType = STAR :| [] }
                                , term = List
                                }
                            )
                            ( Take
                                { usage = SAny
                                , type' = PrimType
                                    { getPrimType = STAR :| [] }
                                , term = PrimTy ( Ty TInt : )
                                } :| []
                            )
                        )
                    )
                }
            )
        }
    ) on Term: Elim
    ( Free
        ( Global
            ( "Nil-Fail" :| [ "nil-int" ] )
        ) ()
    ) ()
```

systemf-frank:
I noticed something strange:
https://github.com/anoma/juvix/blob/c4d0da1a5972ae34f27e32ed585a7f61472132b9/library/Backends/Michelson/src/Juvix/Backends/Michelson/Parameterisation.hs#L156-L165
In line 160 we extract `ty` and `tys` from `tys0`, but why do we skip the head of the non-empty list? `list` has only a single argument of sort `*`, so obviously this match fails we end up in the `otherwise` case resulting in the error.

Any idea @andy-morris?

library/Core/src/Juvix/Core/IR/CheckTerm.hs
```diff
@@ -375,6 +375,21 @@ requireStar ::
 requireStar (IR.VStar j) = pure j
 requireStar ty = Error.throwTC (Error.ShouldBeStar ty)
 
+-- | Given the arity of a type, require all of them to be Star. We expect the
+-- arguments to be written in a right-associative way, e.g: @* -> (* -> *)@.
+-- TODO: check that pi == 0.
```
Acentelles, line 380: Stale TODO?
```suggestion
```

Acentelles (approved):

Acentelles (changes requested):
Can you add tests?

library/Core/src/Juvix/Core/IR/CheckTerm.hs
```diff
@@ -375,6 +375,21 @@ requireStar ::
...
+requirePrimStars ::
+  Error.HasThrowTC' IR.T ext primTy primVal m =>
+  Natural ->
+  Typed.ValueT IR.T primTy primVal ->
+  m Core.Universe
+requirePrimStars 0 ty = requireStar ty
+requirePrimStars n ty = do
+  (π, l, r) <- requirePi ty
+  requireZero π -- Existentially quantified variables should be 0.
```
Acentelles, line 389:
```suggestion
```

I don't think dependent function types should necessarily be of usage 0. See https://bentnib.org/quantitative-type-theory.pdf

Screenshot 2021-09-23 at 12 31 09

andy-morris: oh you're right! i got my πs mixed up so i didn't notice that was a problem

systemf-frank: Hmm I discussed this with Andy at some point. Can't remember exactly, I thought that he pointed out that it should be zero, because these types are dropped during compilation. What do you think @andy-morris?

library/Core/src/Juvix/Core/IR/CheckTerm.hs
systemf-frank: That is the `requireZero pi` one. Guess I forgot to remove the TODO.

library/Core/src/Juvix/Core/IR/CheckTerm.hs
systemf-frank: I think you were correct. I removed `requireZero pi`, now we don't have to do any special usage stuff anymore.

systemf-frank:
> The tests are failing because of the issue mentioned in #905. Taking that example where work around the issue by providing explicit usage:
> 
> ```haskell
> mod Nil-Fail where
> 
> open Prelude
> 
> let (->>) = arr 0
> 
> sig list : ty ->> ty
> let list = %Michelson.list
> 
> sig nil : (a : ty) ->> list a
> let nil _ = %Michelson.nil
> 
> sig int 0 : ty
> let int = %Michelson.int
> 
> sig nil-int : list int
> let nil-int = nil int
> 
> sig main : list int
> let main = nil-int
> ```
> 
> throws another error now:
> 
> ```haskell
> Error: TypecheckerError
>     ( EvalError
>         { evalErr = ErrorValue
>             ( CannotApply
>                 { fun = VPrimTy
>                     ( Return
>                         { retType = PrimType
>                             { getPrimType = STAR :| [] }
>                         , retTerm = List
>                         }
>                     ) ()
>                 , arg = VPrimTy
>                     ( Return
>                         { retType = PrimType
>                             { getPrimType = STAR :| [] }
>                         , retTerm = PrimTy ( Ty TInt : )
>                         }
>                     ) ()
>                 , paramErr = ApplyErrorT
>                     ( Extra
>                         ( TooManyPrimArguments
>                             ( Take
>                                 { usage = SAny
>                                 , type' = PrimType
>                                     { getPrimType = STAR :| [] }
>                                 , term = List
>                                 }
>                             )
>                             ( Take
>                                 { usage = SAny
>                                 , type' = PrimType
>                                     { getPrimType = STAR :| [] }
>                                 , term = PrimTy ( Ty TInt : )
>                                 } :| []
>                             )
>                         )
>                     )
>                 }
>             )
>         }
>     ) on Term: Elim
>     ( Free
>         ( Global
>             ( "Nil-Fail" :| [ "nil-int" ] )
>         ) ()
>     ) ()
> ```

We can see here that `List` has type `*`. List has arity 1, so it should have type `* -> *`.

PR #909, opened 2021-08-06
Easy pipeline erasure

systemf-frank:
The erase has been implemented, but is not usable at this moment due to type constraints being too strict and a missing instance.

```
> printTimeLapseFile "../../nil-fail.ju" defMichelson                                                         

<interactive>:156:1-51: error:
    • No instance for (HasWriter
                         "log" [PipelineLog Juvix.Backends.Michelson.PrimTy RawPrimVal] IO)
        arising from a use of ‘printTimeLapseFile’
    • In the expression:
        printTimeLapseFile "../../nil-fail.ju" defMichelson
      In an equation for ‘it’:
          it = printTimeLapseFile "../../nil-fail.ju" defMichelson
```

I am not sure if this can be fixed. I suspect the easypipeline diverges too much from the main pipeline for this to be solved in an easy way. `defMichelson` should be defined in terms of `Pipeline.Ty _` etc instead of being hardcoded, but I am not entirely sure.


mariari (approved):
Overall looks fine, we could probably revert inline giving back 4 arguments, but that isn't a big deal

PR #913, opened 2021-08-09
Getting Easy Pipeline to compile to ErasedAnn

mariari:
This implements some fixes on top of #909 to let erasure work in the easy pipeline.


systemf-frank (approved):
I have updated the code a bit, fixed a problem with not being able to remove the header from code. This was caused because I implemented the new `*File` functions in terms of the bytestring version. Down the line this would give us code that error in the `ignoreHeader` function.

For a quick fix I just did the stupid thing: copy and update the `*File` implementations from the bytestring implementations.

PR #919, opened 2021-08-10
Push changes error handling

Fraccaman:
When the CI gets triggered from a fork, pushing back changes depends on the source repository permissions. 
As this is not reliable, people pushing changes from a fork should run the formatter locally. 
This PR adds a better log message to inform the use about the error.


cwgoes (approved):
utACK

PR #922, opened 2021-08-10
Clean up of the pipeline from #895

mariari:
Takes the useful infrastruction work from #895 without taking the changes from the record representation


library/Sexp/src/Juvix/Sexp/Structure/Frontend.hs
```diff
@@ -225,6 +230,15 @@ data DefHandler = DefHandler
 -- Not Generated, due to limited generation
 -------------------------------------------
 
+--------------------
+-- NotPunned no Grouping
+--------------------
+fromNotPunnedGroup :: [NotPunned] -> Sexp.T
+fromNotPunnedGroup = Sexp.unGroupBy2 . toStarList fromNotPunned
+
```
rokopt, line 238: Would it make sense to have a `NotPunned with Grouping` comment here, since there's a `no Grouping` one above?

library/Translate/test/Test/Golden.hs
```diff
@@ -258,22 +258,25 @@ contextifySexp names = do
 
 resolveModuleContext ::
   (MonadIO m, MonadFail m) => NonEmpty (NameSymbol.T, [Sexp.T]) -> m Environment.SexpContext
-resolveModuleContext names = do
-  ctx <- contextifySexp names
-  (newCtx, _) <- liftIO $ Environment.runMIO (Contextify.resolveModule ctx)
-  case newCtx of
-    Right ctx -> pure ctx
-    Left _err -> Feedback.fail "not valid pass"
+resolveModuleContext =
```
rokopt, line 261: Good refactoring 👍

library/Translate/src/Juvix/Contextify/Passes.hs
```diff
@@ -13,20 +13,42 @@ import qualified Juvix.Sexp.Structure.Frontend as Structure
...
+--   1. (:open-in Foo body)
+--   2. unqualified-foo
+-- - BNF output form:
+--   1. body
+--   2. qualified-foo
+
+-- Note that the qualification of foo to Bar.foo if the symbol is from
+-- an open module
+
 resolveModule ::
```
rokopt, line 46: Not a question for this PR, but more long-term thought:  do we search by most-recently-opened modules first?  Or least-recently?  Or something else?  Would it make sense (if we don't already) to check whether there's only exactly one possible resolution, so that people can't accidentally intend to mean one thing but get another?

rokopt (approved):
I made tiny comments which I don't need any follow-up on before merging.  LGTM.

library/Translate/src/Juvix/Contextify/Passes.hs
mariari: If a user opens 2 overlapping modules, the code will error currently. I have logic in the context that deals with overlapping names. If a user introduces an ambiguity via interactive definition creation, the context will catch this and error if in some module it introduces an ambiguity.

With that said, the system is fine with `TopLevel` ambiguities, and `Opens` replacing local resolution, as those are puropsfully intended

Issue #923, opened 2021-08-10
Negative golden tests don't work because they expect _every_ phase to fail


rokopt:
In the Michelson and LLVM backends, we have `typecheck` and `compile` tests, and in the Plonk backend we also have `HR`, `IR`, and `Erased` tests.  Those phases are each run on any positive or negative test.

The above scheme doesn't make sense for negative tests (we don't currently have any).  If we were to add a test of code that, for example, typechecks but doesn't compile, or should only fail during erasure, then the test suite would fail because it expects a negative test to fail every phase.  We need it to expect only that a negative test will fail _some_ phase.

rokopt:
Also, there's a common shape among the tests for each backend that it would be good to factor out.  As things stand, we have to change the negative-test checking in three places.

mariari:
#936 Shoud alleviate this issue

rokopt:
I think I have finally realized that with the way that the tests are currently organized, there's nothing that the infrastructure can do to make them work correctly.  Even separating negative and positive tests into different test groups won't be enough, because an existential quantifier on negative tests could succeed when there are other negative tests in the group that should have failed that didn't.  So we would need to have only one test per "negative group" that was expected to fail.

And that in turn doesn't make sense with our current organization of multiple phases (typecheck and compile at least, and on the Plonk backend, also HR, IR, and Erased) -- if an earlier phase fails, then the later phases will also fail, and exactly the same way.

One thing we could do is reorganize the test groups so that there's one group per test case, rather than one group per phase.  Then each test case would be expected to have either all successes (positive) or at least one failure (negative).  That's possible, albeit a bit of a pain, especially with three backends that replicate the same code structure.

But there's something much simpler, it occurs to me, which would also work and which is actually _more_ precise anyway:  just make the negative test cases indicate _which_ phase they expect to fail in, for example by having subdirectories of `negative/<backend>` for each phase -- `negative/llvm/compile`, `negative/circuit/hr`, etc.

If we do that, then each negative test will run in only one phase, so we won't have to worry about the existential quantification -- it'll be the same as the universal.  Furthermore, that's a more precise test -- it's good to specify which phase we're expecting to fail; with the current organization where negative tests run all phases, then if the phase that we expected to fail _didn't_ fail but some other bug caused a later phase to fail, then we'd conceal the bug by seeing an expected "at least one" failure.

rokopt:
[...]

Issue #916, opened 2021-08-09
Standard library of universally-available types would enable backend-independent modules


rokopt:
Currently, I can write a `.ju` file that only opens `Prelude` and I can get types such as `->`.  However, to write any real code, I need some primitive types such as `int`.  Those I can only get from specific backends, so I have to open at least one backend module, such as `LLVM`, and compile that.  But once I open `LLVM`, I can't compile to any other backend.

It might or might not be a short-term goal to be able to write multi-backend validity predicates -- at first I'd expect that LLVM is the one main priority, in particular in Stygarctus.  But I would think that people would want to write at least some code that could be shared by, for example, Plonk as well as LLVM, and in the long run we want to be the best general-purpose programming language that we can be, and to be that we will certainly want people to be able to write code -- indeed, I would expect, the huge majority of their code -- without having to know or specify which backend it's going to be compiled to.

rokopt:
Actually, maybe scratch the "long-term" aspect of this bug.  As @Acentelles pointed out, this might enable an immediate productivity improvement:  currently, when we want to golden-test common core code, we have to either write tests for all backends or write a test for only one and hope that it also works on others.  If we had a standard library, we could also have "all-backend' golden tests, which would allow us to write one test that would be compiled on all backends.

Acentelles:
I think this is an excellent idea

PR #924, opened 2021-08-10
Various mostly-skeletal datatype golden tests and fixes

rokopt:
This PR adds a couple datatype tests and cleans up others and tries to make them as parallel as possible across backends.

It also adds some skeletons to scope out places to add new datatype tests, and to try to make it easy to keep them as consistent as possible across backends.  (It would be good at some point to have backend-independent golden tests, perhaps depending on #916 if/when it becomes available.)

It also fixes a bug in the LLVM golden test code, which had not been expecting negative tests to fail.  (However, see #923 for a more fundamental problem with negative golden tests, which still remains.)


mariari (approved):
Looks good, the extension to the golden test itself is nice

Issue #898, opened 2021-08-02
'Failed to eval term' for `%Michelson.nil`


systemf-frank:
During the process of finding a smaller program that still triggers https://github.com/anoma/juvix/pull/805#issuecomment-889850153, I found a new problem when using the `nil` function:

```ocaml
mod Nil-Fail where

open Prelude
open Michelson

sig nil-int : list int
let nil-int = nil int

sig main : int -> list int
let main _ = nil-int
```

```
$ stack run -- typecheck -b michelson nil-fail.ju
juvix: FatalError {fatalErrorMessage = "Failed to eval term"}
```

systemf-frank:
This error is thrown by the `apply1` function in the Michelson backend.

PR #925, opened 2021-08-11
Merge from #900 as fix for insufficient usage error

systemf-frank:
#900 is merged here, so that I can use the fix to develop further on #898.


systemf-frank:
Closing the PR without merge as the current state doesn't fix the issue anymore.

PR #927, opened 2021-08-12
records in core

andy-morris:


andy-morris:
i force-pushed this to just have the document because the tiny amount of actual code written was not worth the merge conflicts

andy-morris:
the document is now [on hackmd](https://hackmd.io/Xlvu82eFQ_m-cebUWAtGsw?view) so this PR doesn't have anything

Issue #910, opened 2021-08-09
Inductive datatypes and pattern-matching


rokopt:
# Description

This is an indexing issue for tracking (in the form of linked issues) what's incomplete or broken in the typechecking and compilation of inductive datatypes, and pattern-matching to define functions on those datatypes.  In particular we can define for each linked issue whether it's a requirement for the [Stygarchus release](https://github.com/anoma/juvix/issues/890) or whether it will be deferred until after that.

# Blocks

# Blocked by

# See also

- #890

cwgoes:
I suggest that we focus first on the part of the pipeline prior to backends, and in parallel get the LLVM backend to where ti can compile basic datatypes & pattern-matching.

rokopt:
Regarding pattern-matching, see #492 , and in particular https://github.com/anoma/juvix/issues/492#issuecomment-719644992.

rokopt:
See #985 for a proposed start (product/coproduct types), along with a path towards a full-featured implementation (with recursive and dependent types, for example, and guards on pattern-matches).

PR #940, opened 2021-08-13
More tests for variations on datatype declarations

rokopt:
Add tests for:

- Recursive pattern-matches
- Function types in datatype constructors
- Parameterized datatypes


test/examples/to-fix/circuit/Datatypes.ju
```diff
@@ -25,6 +25,19 @@ sig patternMatch : Bar -> field
...
+-- ( probably a bug in the elaborator from frontend )"
+type AtomExp a = Atom a
+
+sig testAExp : AtomExp field
+let testAExp = Atom 1
+
+-- Fails to typecheck:
+-- "Error on toHR: Name "SExp" :| [] not in scope
+-- ( probably a bug in the elaborator from frontend )"
+type SExp a = Atom a | SPair (SExp a) (SExp a)
```
mariari, line 39: This brings up an intersting test case we should have.

Namely in a working example can we break it with

```haskell
type atom-exp a = Atom a

type more-fun-exp = Atom Int.t
```

I think this is an error waiting to happen

test/examples/to-fix/llvm/Datatypes.ju
```diff
@@ -18,6 +18,37 @@ sig patternMatch : Bar -> int
 let patternMatch (P x y) = x + y
 let patternMatch (Q x y z) = x + y + z
 
+type FunctionTypeInConstructor :
```
mariari, line 21: Nit : For consistency can we use lower cased type names?

test/examples/to-fix/michelson/Datatypes.ju
```diff
@@ -44,4 +75,32 @@ let main =
...
 
+  -- Typechecks, but fails to compile:
+  --  juvix: src/Juvix/Backends/Michelson/Compilation.hs:53:13-75:
+  --  Non-exhaustive patterns in Ann _ (Pi argUsage _ _) (LamM _ [name] body)
+  let ftict = fticTestTerm in
+
+  -- Typechecks, but fails to compile:
+  --  juvix: src/Juvix/Backends/Michelson/Compilation.hs:53:13-75:
+  --  Non-exhaustive patterns in Ann _ (Pi argUsage _ _) (LamM _ [name] body)
+  let sexpTest = ssize s1 in
```
mariari, line 86: Yeah, currently we only compile Paired functions for Michelson, as those are the valid contract types

mariari (commented):
Seems like a few nice test cases with variations that make it break.

Would be good to make some smaller tests from the failures here, along with some slightly evil tests (not too evil yet).

Also as a small nit, I'd like it if we can keep type names lower case.

test/examples/to-fix/circuit/Datatypes.ju
rokopt: Good idea, I added that one.

test/examples/to-fix/llvm/Datatypes.ju
rokopt: Sure, I've made the type names lowercase now.

rokopt:
> Seems like a few nice test cases with variations that make it break.
> 
> Would be good to make some smaller tests from the failures here, along with some slightly evil tests (not too evil yet).

I'll definitely accumulate more over the course of #910 (and beyond!).

mariari (approved):
Seems good

Issue #938, opened 2021-08-12
Disallowing Implicit Module FIle Names


mariari:
Currently, if the file does not have an explicit module name, we pun it off the file name. This idea is heavily inspired from OCaml.

However in an image based language I'm not sure if this makes much sense. Thus we should disable this feature until further research is had

### Related Issues

#941

cwgoes:
Agreed; let's require explicit naming for now.

PR #941, opened 2021-08-13
Disabling Implicitly named modules

mariari:
This PR removes the implicit module naming from files.

Hopefully we can enable it once again in the future, but until then, this will have to do.

### See
#938


cwgoes (approved):
utACK

PR #942, opened 2021-08-13
Generalized Context Map

mariari:
In order to properly support forms that Change

Type Declaration --> Definitions

or create many definitions from a single one

Declaration --> [Declarations]

we need a more general mapping form. This PR seeks to give the infrastructure to properly do that.

### Blocking
#926


library/Translate/src/Juvix/Contextify/Environment.hs
```diff
@@ -151,6 +126,59 @@ onExpression form trigger func =
...
+singlePass f = Pass f f f
+
+class
+  Context.ToContextFormGeneral contextTo =>
+  ToContextTraversal (pass :: (* -> *) -> *) contextTo
+    | pass -> contextTo
+  where
+  promotePass ::
+    (HasReader "closure" Closure.T m, HasThrow "error" ErrorS m) =>
+    pass m ->
```
mariari, line 149: Should the `m` be part of the interface like MTL/Capabilities does it? Or should we just fix the effect in here?

rokopt (approved):
This looks like a good refactoring to me!

library/Translate/src/Juvix/Contextify/Environment.hs
rokopt: I think I like it where it is, as it's closest to where it's used, but it's not a strong feeling.

Issue #939, opened 2021-08-13
Negative tests that induce typecheck errors fail with whitespace disagreement


rokopt:
# Description

With code from https://github.com/anoma/juvix/tree/terence/initial-negative-datatype-tests to address https://github.com/anoma/juvix/issues/923 (as described in https://github.com/anoma/juvix/issues/923#issuecomment-898089407), I still suffer from failures in the negative tests that result from a disagreement between the expected and actual golden test results, where the disagreement consists solely of spaces surrounding the names of globals:

```lisp
llvm            >         TooManyArguments.typecheck: FAIL (0.54s)
llvm            >           Output doesn't match golden file.The new result is
llvm            >           [ "Error: TypecheckerError
llvm            >               ( ShouldBeFunctionType
llvm            >                   { typeActual = VNeutral
llvm            >                       ( NFree
llvm            >                           ( Global
llvm            >                               ( "TooManyArguments" :| [ "VerySimpleType" ] )
llvm            >                           ) ()
llvm            >                       ) ()
llvm            >                   }
llvm            >               ) on Term: Let Omega
llvm            >               ( App
llvm            >                   ( App
llvm            >                       ( Free
llvm            >                           ( Global
llvm            >                               ( "TooManyArguments" :| [ "One" ] )
llvm            >                           ) ()
llvm            >                       )
llvm            >                       ( Prim
llvm            >                           ( LitInt 0 ) ()
llvm            >                       ) ()
llvm            >                   )
llvm            >                   ( Prim
llvm            >                       ( LitInt 1 ) ()
llvm            >                   ) ()
llvm            >               )
llvm            >               ( Prim
llvm            >                   ( LitInt 0 ) ()
llvm            >               ) ()" ]
llvm            >            but the expected result is
llvm            >           [ "Error: TypecheckerError
llvm            >               ( ShouldBeFunctionType
llvm            >                   { typeActual = VNeutral
llvm            >                       ( NFree
llvm            >                           ( Global
llvm            >                               ( " TooManyArguments " :| [ " VerySimpleType " ] )
llvm            >                           ) ()
llvm            >                       ) ()
llvm            >                   }
llvm            >               ) on Term: Let Omega
llvm            >               ( App
llvm            >                   ( App
llvm            >                       ( Free
llvm            >                           ( Global
llvm            >                               ( " TooManyArguments " :| [ " One " ] )
llvm            >                           ) ()
llvm            >                       )
llvm            >                       ( Prim
[...]

PR #943, opened 2021-08-14
Make negative golden tests work, and add some initial simple ones

rokopt:
Some negative golden tests that I tried to add for Stygarctus datatype requirements turned out to be the first negative tests in the codebase, and revealed two reasons that negative tests didn't work:

- #923 
- #939 

This PR attempts to fix both, the first by making negative tests specify which _phase_ they expect to fail in (see https://github.com/anoma/juvix/issues/923#issuecomment-898089407 ) and the second by checking only for failure and not for specific output (see https://github.com/anoma/juvix/issues/939#issuecomment-898782374).  The former I think I'm reasonably satisfied with.  The latter is definitely a compromise, but in the short run it's the best one I can think of to have negative tests usable for Stygarctus without spending much more time debugging the negative-test infrastructure itself.  See #939 for discussion of longer-term improvements.


mariari (approved):
Yeah It makes sense how this works, it really is a simpler solution than I thought! Being explicit about where it fails may be a good idea, though I wonder if it's a good fit in the long run.

PR #954, opened 2021-08-20
LLVM lambda term support

systemf-frank:
This PR implements support for the lambda calculus, but is currently still limited in terms of support for primitives.


systemf-frank:
The implementation is ready for review. Currently I only added two new tests. Tomorrow I will see if I can add some more.

systemf-frank:
I am working on a wrapper script to execute the LLVM IR directly. This is currently working fine for programs that do not take any arguments in the main function, but it is causing an issue for a lambda case (where I provide the args on the commandline):
```
LLVM ERROR: Invalid type for second argument of main() supplied
```
I thought that LLVM didn't enforce a predefined type for the main function, maybe I need to tell that somehow.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
 import qualified LLVM.AST.Type as LLVM
 import qualified LLVM.IRBuilder as LLVM
 import qualified LLVM.Pretty as LLVM
 import qualified Prelude as P
 
+-- | A mapping between Juvix variable names and their operands in LLVM.
+type Env = Map.Map NameSymbol.T LLVM.Operand
+
+-- | Compile the input program to an LLVM module.
+-- TODO: maybe do something smarter with the module name?
```
cwgoes, line 27: at some point we might want to map juvix modules to llvm modules or at least include metadata

this is low-priority however

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
+-- body of the main function by compiling @t@. The main function itself does
+-- not have any parameters in this case.
+mkMain ::
+  LLVM.MonadModuleBuilder m =>
+  -- | Term to compile.
+  ErasedAnn.AnnTerm PrimTy RawPrimVal ->
+  m LLVM.Operand
+mkMain t@(ErasedAnn.Ann usage ty t') = do
+  let types = map typeToLLVM (functionTy ty)
+      returnType = P.last types
```
cwgoes, line 58: this can fail, right?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
   ErasedAnn.AnnTerm PrimTy RawPrimVal ->
-  LLVM.IRBuilderT LLVM.ModuleBuilder LLVM.Operand
-compileTerm (ErasedAnn.Ann usage ty t) = case t of
-  ErasedAnn.Prim t' -> mkPrim t
-  ErasedAnn.AppM f xs -> mkApp (ErasedAnn.term f) ty xs
+  m LLVM.Operand
+compileTerm env (ErasedAnn.Ann usage ty t) = case t of
+  ErasedAnn.Var symbol -> case Map.lookup symbol env of
+    Nothing ->
+      P.error $ "Variable not found: " <> show symbol <> " in " <> show env -- TODO improve error message.
```
cwgoes, line 88: we should use something other than Prelude.error I think?

cwgoes: or is this a compiler fault?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
+  params <- mapM mkParam (zip args (functionTy ty))
+  LLVM.function funname params returnTy $ \refs -> do
+    let env' = env `Map.union` Map.fromList (zip args refs)
+    body' <- compileTerm env' body
+    LLVM.ret body'
+  return funname
+  where
+    -- Given a Juvix name and type, construct an LLVM function parameter
+    -- definition.
+    mkParam ::
```
cwgoes, line 120: this should be a top-level function perhaps?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
+-- TODO: implement other primitives.
+mkPrim ::
+  LLVM.MonadIRBuilder m =>
+  -- | Term that contains the primitive.
+  ErasedAnn.Term PrimTy RawPrimVal ->
+  -- | Type of the primitive.
+  ErasedAnn.Type PrimTy ->
+  m LLVM.Operand
+mkPrim (ErasedAnn.Prim prim) ty = case prim of
+  LitInt i -> case ty of
```
cwgoes, line 166: seems like we need more cases or some kind of error case?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
 typeToLLVM (ErasedAnn.PrimTy (PrimTy ty)) = ty
+typeToLLVM (ErasedAnn.Pi _usage f xs) =
+  LLVM.FunctionType
+    { LLVM.resultType = typeToLLVM resultType,
+      LLVM.argumentTypes = map typeToLLVM argumentTypes,
+      LLVM.isVarArg = False
+    }
+  where
+    tyList = f : functionTy xs
+    resultType = P.last tyList
```
cwgoes, line 210: can this fail?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,65 +1,216 @@
...
+    argumentTypes = init tyList
```
cwgoes, line 211: ditto?

cwgoes (changes requested):
Thanks! See comments.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: Yeah, currently there is no way to do so either. Everything is currently a single module in the input program, and there is no way to access a module or input filename.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: I am not a fan of using error, but it is a quick solution for now.

This case should actually never occur, as a program referring to an unknown variable shouldn't typecheck. For now this problem occurs when using a global variable that we do not have access to in the backend.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: I am not particularly happy with it either, there is also some overlap with mkParameterName.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: Currently the set of supported primitives is extremely limited and should definitely be extended either right before or right after the release.

I can add an error here, but there no other primitives that could be used. Having a fall-through case keeps GHC from spitting compiler warnings as well, but we actually want them for functions like this, to remind us of the intotality.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: No, `tyList` has atleast two elements, and therefore we can always execute `last` and `init` on it.

rokopt:
Note: In #954, we add some new LLVM golden tests, and in #985, we make the LLVM golden tests check more phases.  Thus, when we merge the two, we must add the new phases for the new tests, for example by cherry-picking https://github.com/anoma/juvix/commit/9eb0601fe2fa66a48df18d3319cfca7e98220667.

rokopt:
I've tried this in a view that merges #954 and #985:

```ml
mod AlgebraicTypes where

open Prelude
open LLVM

sig (+) : int -> int -> int
let (+) = %LLVM.add
declare infixl (+) 5

sig int-plus-two : int -> int
let int-plus-two i = i + 2

sig main : int
let main = int-plus-two 3
```

```haskell
juvix: Variable not found: "AlgebraicTypes" :| ["+"] in fromList [("0" :| [],LocalReference (IntegerType {typeBits = 8}) (Name "arg0_0"))]
  error, called at src/Juvix/Backends/LLVM/Compilation.hs:89:7 in llvm-0.1.0.0-4fXym3EbBO67IfisJChZAk:Juvix.Backends.LLVM.Compilation
```

Is that expected?  If not, does it happen for you with only #954 (did I break that with #985, even though that code doesn't use products/coproducts)?

library/Backends/llvm/src/Juvix/Backends/LLVM/Parameterization.hs
```diff
@@ -8,24 +9,81 @@ module Juvix.Backends.LLVM.Parameterization
...
+              }
+      -- If there are too many arguments to apply, raise an error.
+      GT -> Left $ Param.ExtraArguments fun args
+    where
+      toTakes App.Cont {App.fun, App.args, App.numLeft} = (fun, args, numLeft)
+      toTakes App.Return {App.retType, App.retTerm} = (fun, [], arityRaw retTerm)
+        where
+          fun =
+            App.Take
+              { App.usage = Omega,
```
rokopt: Is this because the `usage` in `App.Take` isn't actually used?  Or is it used but `Omega` is always correct?  Or is usage-tracking just not implemented here yet?

rokopt (commented):
> Is that expected? If not, does it happen for you with only #954 (did I break that with #985, even though that code doesn't use products/coproducts)?

I just saw your review comment response:

> This case should actually never occur, as a program referring to an unknown variable shouldn't typecheck. For now this problem occurs when using a global variable that we do not have access to in the backend.

Ah, so maybe what I saw is just a new form of an old bug?  (The old bug being, maybe the `Return'` one -- I'm not sure whether there's an issue open for it?)

rokopt:
Super nice sequence of clear separate commits!  Thanks!  👍

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
cwgoes: Hmm sure, I'm not against this but I don't think it's very well standardised across the codebase.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
cwgoes: Can we change it to a top-level function?

cwgoes (approved):

cwgoes:
ACK, but see follow-up comments.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: I have addressed this and replaced it by the clearer `mkParameter` global function.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
systemf-frank: No it isn't really standardised across the codebase. It doesn't help that GHC gives a rather generic `Non-exhaustive patterns in case` instead of telling the exact case that didn't have a match, because that would be the nicest solution as it doesn't require us to do anything.

library/Backends/llvm/src/Juvix/Backends/LLVM/Parameterization.hs
systemf-frank: This code was incomplete and therefore removed from the branch for now. I will need to add something back later.

systemf-frank:
> > Is that expected? If not, does it happen for you with only #954 (did I break that with #985, even though that code doesn't use products/coproducts)?
> 
> I just saw your review comment response:
> 
> > This case should actually never occur, as a program referring to an unknown variable shouldn't typecheck. For now this problem occurs when using a global variable that we do not have access to in the backend.
> 
> Ah, so maybe what I saw is just a new form of an old bug? (The old bug being, maybe the `Return'` one -- I'm not sure whether there's an issue open for it?)

I am not entirely sure what you mean with the `Return'` one (https://github.com/anoma/juvix/pull/979 ?).

The reason this fails is simply because there is no way for the backends to access any global definitions. This will be fixed by inlining once the other issues are fixed.

rokopt:
> I am not entirely sure what you mean with the `Return'` one (#979 ?).

Yes, sorry, that's the one.

systemf-frank:
@mariari found a bug: https://github.com/anoma/juvix/issues/993

systemf-frank:
Something for later: https://github.com/anoma/juvix/issues/996

systemf-frank:
I will close and merge this PR now. The open issues will be taken care of in their own PR's.

PR #956, opened 2021-08-20
Remove extension for Terms in Values in Subst

mariari:
This clears up the evaluator substv code by removing an unused constraint. Namely the `extT` for term extensions. This extension was placed due to the `Error` type being too broad. For clarity and debugging reasons this was removed.

The main drawback besdies readability is that this under-constrained the `extT` type which made it imposible to print any terms handed to the code.


library/EasyPipeline/src/Easy.hs
```diff
@@ -564,9 +564,11 @@ printTimeLapse byteString option = do
   --
   (_, cored) <- coreify byteString option
   traverse_ (printCoreFunction cored option) currentDefinedItems
+  print "finished Cored"
```
cwgoes, line 567: remove this before merge? or do you want to keep the debugging info in the easy pipeline?

mariari: This is useful to have for the timelapse as it came under question what a pass was.

cwgoes:
Code looks fine but can you reference an issue or briefly explain in the top comment the reason for this change?

mariari:
> Code looks fine but can you reference an issue or briefly explain in the top comment the reason for this change?

Updated

cwgoes (approved):
utACK

PR #959, opened 2021-08-23
Update README.md

jonaprieto:


mariari (approved):
These seem good, surprsied we didn't change this in the past.

PR #918, opened 2021-08-09
Record Core Representation Generator

mariari:
## Blocking
#926

PR #974, opened 2021-08-31
Record Transformation

mariari:
This PR writes the context passes and toHR changes needed for record transformations

### Blocked on
#918


rokopt:
Is this still blocked on #918 , or did you figure out a way to do this without that one having been merged?  Or does this one include the changes from that one?

mariari:
> Is this still blocked on #918 , or did you figure out a way to do this without that one having been merged? Or does this one include the changes from that one?

This one was always branched off, however since it never got reviewed we can just review this one instead

library/Sexp/src/Juvix/Sexp.hs
```diff
@@ -98,6 +103,40 @@ foldSearchPred t p1@(predChange, f) p2@(predBind, g) =
...
+  (NameSymbol.T -> Bool, Atom -> T -> (T -> f T) -> f T) ->
+  f T
+foldSearchPredManualRecurse t p1@(predChange, f) p2@(predBind, g) =
+  case t of
+    Cons a@(Atom atom@(A name _)) xs
+      -- this case is a bit special as we wish to remove the form but
+      -- it's a binder! So we must run it then run the transform on it!
+      | predBind name && predChange name -> do
+        bindedTerm <- bindCase
+        changeCase (cdr bindedTerm)
```
rokopt, line 122: Would `bound` be a more standard thing to call it than `binded`?

library/Sexp/src/Juvix/Sexp.hs
```diff
@@ -98,6 +103,40 @@ foldSearchPred t p1@(predChange, f) p2@(predBind, g) =
       | predChange ":atom" -> f a t
       | otherwise -> pure $ Atom a
 
+-- | @foldSearchPredManualRecurse@ is just like @foldsSearchPred@
+-- except the function which changes the sexp form manually has to
+-- recurse, it no longer automatically does this.
```
rokopt, line 108: There does seem to be significant overlap in the code -- is it possible to factor out the commonality between them?

library/Translate/src/Juvix/Contextify/Environment.hs
```diff
@@ -150,18 +176,38 @@ class
...
 
+instance ToContextTraversal PassManual Context.ContextForms where
+  promotePass (PassManual pass) trigger =
+    let f form ctx =
+          Sexp.foldSearchPredManualRecurse
+            form
+            (trigger, pass ctx)
+            -- Hack? We unbind the primtiive form for being a
+            -- binder. Why was this a binder in the first place? If we
+            -- change it we get errors throughout a resolution step.
```
rokopt, line 187: Is this a question that you've been able to answer by now?  I'm not exactly sure what this comment means or what the implications are, but if it's unclear why the code didn't work in some other form in which you expected it to work, then I'd worry that there would still be a bug that could have other consequences that we're just working around here.

rokopt (changes requested):
I'm not sure whether any code changes will come out of them or not, but there's at least one question I'd want to understand the answer to.

library/Translate/src/Juvix/Contextify/Environment.hs
mariari: Yeah I remembered last night. Due to having no quotes, we have to do resolution on all symbols, thus when we see `(:primitive foo)` we need to say it's a binder as otherwise we will try to resolve foo to a module that we've opened

library/Sexp/src/Juvix/Sexp.hs
mariari: Probably can make one in the other, by making a new f such that it always recurses

rokopt: That'd probably be a good thing, then, although I wouldn't say it needs to be done for this PR, maybe just for some low-priority "refactoring"-tag issue afterwards.

library/Translate/src/Juvix/Contextify/Environment.hs
rokopt: Cool, if you update the comment to reflect that, that'll be helpful.

Good stuff!  Thanks.

rokopt (approved):

PR #979, opened 2021-09-02
Support type application in core

andy-morris:
- Adds support for type-level application by reusing the infrastructure in `J.C.Application` in primitive types too.
- Adds a new typeclass `J.C.Parameterisation.CanPrimApply`, for applying a primitive function to known-well-typed arguments, as well as asking the arity of a function. The former is what Michelson's `applyProper` function did before.
    - Backends provide instances for `CanPrimApply Ty Val`, and `CanApply () Ty` (or, in a hypothetical backend with a kind system, `CanApply Kind Ty` or something........)

### DIFF
--- a/library/Backends/Michelson/src/Juvix/Backends/Michelson/Compilation/Types.hs
+++ b/library/Backends/Michelson/src/Juvix/Backends/Michelson/Compilation/Types.hs
@@ -1,14 +1,17 @@
 -- |
 -- - Types used internally by the Michelson backend.
 module Juvix.Backends.Michelson.Compilation.Types
-  ( PrimTy (..),
+  ( RawPrimTy (..),
+    PrimTy',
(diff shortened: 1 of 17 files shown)


mariari (approved):
Looks fine overally, some good changes.

Hard to review as there is a lot of changes, I probably missed a lot in this review

library/Backends/Michelson/src/Juvix/Backends/Michelson/Compilation/Types.hs
```diff
@@ -44,7 +48,8 @@ data PrimTy
   | List
   | Set
   | ContractT
-  | Application PrimTy (NonEmpty PrimTy)
+    -- FIXME is this still needed?
```
Acentelles, line 51: What's the outcome of this?

Issue #975, opened 2021-08-31
`main` signature for each backend is restricted to the signature provided by that backend's call to `typecheck'`


rokopt:
# Description

We've done some work to try to allow `main` to have any signature under #848 .  However, there is still a limitation deriving from the implementation of the `HasBackend` interface and of `typecheck'`.

The `HasBackend` interface which each backend must implement includes a `typecheck` function with the following signature:

```haskell
  typecheck :: Context.T Sexp.T Sexp.T Sexp.T -> Pipeline (ErasedAnn.AnnTermT (Ty b) (Val b))
```

All three pipelines implement `typecheck` in the same way (so I guess this is the intent), by calling the generic `Pipeline` module's `typecheck'` with backend-specific parameters.

Michelson:
```haskell
  typecheck ctx = Pipeline.typecheck' ctx Param.michelson Param.Set
```

LLVM:
```haskell
  -- Copied over from the Michelson backend, and adapter where necessary.
  typecheck ctx = Pipeline.typecheck' ctx llvm Set
```

Plonk:
```haskell
  typecheck ctx = Pipeline.typecheck' ctx (Parameterization.param @f) Types.PField
```

`typecheck'` requires the backend to pass in a type to typecheck the term against, and at this point the backend knows nothing about the term yet -- the input parameter is just a context.  So the backends simply choose some type.  This is particularly strange, I'd say, in the case of LLVM, which I think could support a wide variety of signatures, and for which `Set` is probably not sensible even as a default, let alone as the only option.

# Blocks

- #910 

# Blocked by

# See also

- #848

rokopt:
It turns out that the signature of `typecheck'` has had me fooled.

What's been not at all clear to me is that the type provided as a parameter to `typecheck'` is _not_ the type to check the term against!  Rather, chasing that parameter's use through the code, `typecheck'` passes it only to `toErased`, and `toErased` uses it only as a parameter to this call:

```haskell
typedGlobals = map (typePrims ty) globalDefs
```

`typePrims` makes a bunch of mutually-recursive calls, but most of those, too, use the `ty` parameter only as a pass-through to the other functions in the family.  In the end, the only places where the parameter is actually _used_ are these:

```haskell
pattEval ty patt =
[ ... ]
IR.PPrim p -> IR.PPrim (CoreApp.Return (Param.PrimType $ ty :| []) p)
[... ...]
baseToReturn ty t =
[ ... ]
[...]

Issue #983, opened 2021-09-07
`toLambdaR` does not handle functions with multiple clauses (while inlining globals)


rokopt:
# Description

Closely related to https://github.com/anoma/juvix/issues/981#issue-990194549 , the `toLambdaR` function's only `RawGFunction` case has a guard which matches only functions with exactly one `RawFunClause`.  Consequently, it does not support (total) functions from coproduct types (such as datatypes with multiple constructors).

# Blocks

- #910 

# Blocked by

# See also

- #981 
- #982

PR #985, opened 2021-09-10
Introduce algebraic product and coproduct types

rokopt:
This PR represents a first step towards full user-defined datatypes and pattern-matching (#910) by introducing simple non-dependent (though parametric) product and coproduct types, with builtin functions for introduction and elimination.  The elimination functions correspond to the simplest forms of pattern-matching on algebraic data types.

Included in separate commits are a change to remove the useless parameter described in #975 (that's https://github.com/anoma/juvix/commit/8a1645c0a285ff4147d4576603859355b5825ebf) and a number of stylistic fixes (some made by the pre-commit script).  The main commit is https://github.com/anoma/juvix/commit/3cfc0b823bfd5d8c1dd9484c525afa16aacfbe08, with pretty-printing added in https://github.com/anoma/juvix/commit/960a7b80983908dad93f15f8db0f2273ab7788c5.

With this PR, we can compile this to LLVM:
```ml
cat-product-elim-left int (cat-product-intro 1 2) -- 1
```
(And similarly -right.)

This currently crashes with a non-exhaustive match in LLVM's `mkApp`:
```ml
cat-coproduct-elim int int (cat-coproduct-intro-left 3) int-plus-two int-plus-three -- 5
```
However, that's because coproducts eliminate with a lambda (a case statement).  The term that it's passing to LLVM is an application of a lambda that doesn't contain any coproduct types anymore.  So I think that with @systemf-frank 's changes for #954 , we might be able to evaluate that too.

There will be many more enhancements required, but I think this is a start.  Some next steps:

- Review the usages together -- @jonaprieto  and I aren't sure whether we have them right.  We will probably have to add some usage-checking where we currently don't have any.
- Figure out how to do more elimination at compile-time -- currently the above won't work if you stash the introduction into a let binding, for example.  We should be able to reduce more elim/intro pairs to allow for more programs to work on back-ends that don't support structs/unions.  (For any backends that can support structs/unions, once they are able to use them, we won't need to do those reductions for the programs to work, though they'd still be optimizations.)
- More features.  Maybe in some order like this, though I'm not sure:
  - Lists of products/coproducts instead of just pairs (maybe using an explicit list type, as I think @April suggested a while ago) 
  - Recursion (through fixed points)
  - Dependent types
  - Richer pattern-matching (e.g. guards)
  - User-defined datatypes, which give you:
    - names
    - optionally more initial algebras than just the fixed points of substitution (i.e. an extension of the type theory beyond just what you get from datatypes that compile completely to anonymous algebraic types)


cwgoes:
> Figure out how to do more elimination at compile-time -- currently the above won't work if you stash the introduction into a let binding, for example. We should be able to reduce more elim/intro pairs to allow for more programs to work on back-ends that don't support structs/unions. (For any backends that can support structs/unions, once they are able to use them, we won't need to do those reductions for the programs to work, though they'd still be optimizations.)

Won't work because the backend doesn't support the actual evaluation, or won't work for some other reason?

library/Core/src/Juvix/Core/Erased/Ann/Pretty.hs
```diff
@@ -21,11 +54,19 @@ instance PrimPretty1 primVal => PP.PrettySyntax (Term primTy primVal) where
     Prim p -> fmap HR.toPPAnn <$> PP.pretty' p
     LamM {arguments, body} -> HR.ppLams (arguments, body)
     p@(PairM _ _) -> HR.ppPairs $ getPairs p
+    p@(CatProductIntroM _ _) -> HR.ppPairs $ getPairs p
+    p@(CatProductElimLeftM _ _) -> panic "pretty-printing productelimleft not yet implemented"
```
cwgoes, line 58: why? seems easy enough

library/Core/src/Juvix/Core/IR/CheckTerm.hs
```diff
@@ -208,6 +208,56 @@ typeTerm' term ann@(Typed.Annotation σ ty) =
...
+      let sAnn = Typed.Annotation (σ <.> π) a
+      s' <- typeTerm' s sAnn
+      pure $ Typed.CatCoproductIntroLeft s' ann
+    Core.CatCoproductIntroRight s _ -> do
+      (π, _a, b) <- requireCatCoproduct ty
+      let sAnn = Typed.Annotation (σ <.> π) b
+      s' <- typeTerm' s sAnn
+      pure $ Typed.CatCoproductIntroRight s' ann
+    Core.CatCoproductElim a b cp s t _ -> do
+      a' <- typeTerm' a (Typed.Annotation σ (IR.VStar 0))
```
cwgoes, line 253: very minor, but we can have a const `(Typed.Annotation σ (IR.VStar 0))`

test/examples/positive/llvm/Datatypes.ju
```diff
@@ -3,5 +3,18 @@ mod Datatypes where
...
+type pq : ty = | P int int | Q int int int
+
+sig patternMatch : pq -> int
+let patternMatch (P x y) = x + y
+let patternMatch (Q x y z) = x + y + z
+
 sig main : int
-let main = 0
+let main =
+  -- let pm = patternMatch (P 1 2) in
```
cwgoes, line 18: why is this commented? do these work?

cwgoes (commented):
Thank you for the clear description & commit sequence. Left a few minor comments on the implementation (I didn't review in detail the formatter changes, which presumably are fine).

> However, that's because coproducts eliminate with a lambda (a case statement). The term that it's passing to LLVM is an application of a lambda that doesn't contain any coproduct types anymore. 

Where does this conversion actually happen? I looked for it but perhaps I missed a diff.

rokopt:
> Won't work because the backend doesn't support the actual evaluation, or won't work for some other reason?

Because the back-end doesn't know how to evaluate it.  The term it gets looks sensible to me; it just doesn't know what a product is.  :)  (Similarly to how it currently crashes on the coproduct evaluation because it doesn't know what to do with a lambda, but with #954 I am hoping that will start working!)

rokopt:
> > However, that's because coproducts eliminate with a lambda (a case statement). The term that it's passing to LLVM is an application of a lambda that doesn't contain any coproduct types anymore.
> 
> Where does this conversion actually happen? I looked for it but perhaps I missed a diff.

That's the two cases in `eraseTerm` where we have a `CoproductElim` of either a `CoproductIntroLeft` or `CoproductIntroRight`:

https://github.com/anoma/juvix/pull/985/commits/3cfc0b823bfd5d8c1dd9484c525afa16aacfbe08#diff-bb186dfb35d33d6e3070828689d37896654f7c5792d6bad84139df67b5e493d5R190-R196

library/Core/src/Juvix/Core/Erased/Ann/Pretty.hs
rokopt: I think that was just running out of time!  :) @jonaprieto has some more pretty-printing development in his local repo -- he might be able to let us know whether he has something ready for this case (or might by whenever release hour is).

april:
I think you may have the wrong @april tagged. Did you mean @cyberglot?

rokopt:
> I think you may have the wrong @april tagged. Did you mean @cyberglot? 😅

You're right!  I apologize for the spurious notification.  I will eventually learn the difference between Github and Slack.

library/Core/src/Juvix/Core/IR/CheckTerm.hs
rokopt: Good idea.  Done.

test/examples/positive/llvm/Datatypes.ju
rokopt: No; that, for example, doesn't work because of #983 .   I shouldn't have commented-out stuff in `positive` golden tests.  Or `negative` ones, for that matter.  Maybe in `to-fix` tests they're tolerable.  I've removed the comments.

rokopt:
Note: In #954, we add some new LLVM golden tests, and in #985, we make the LLVM golden tests check more phases.  Thus, when we merge the two, we must add the new phases for the new tests, for example by cherry-picking https://github.com/anoma/juvix/commit/9eb0601fe2fa66a48df18d3319cfca7e98220667.

cwgoes:
> > > However, that's because coproducts eliminate with a lambda (a case statement). The term that it's passing to LLVM is an application of a lambda that doesn't contain any coproduct types anymore.
> > 
> > 
> > Where does this conversion actually happen? I looked for it but perhaps I missed a diff.
> 
> That's the two cases in `eraseTerm` where we have a `CoproductElim` of either a `CoproductIntroLeft` or `CoproductIntroRight`:
> 
> [3cfc0b8#diff-bb186dfb35d33d6e3070828689d37896654f7c5792d6bad84139df67b5e493d5R190-R196](https://github.com/anoma/juvix/commit/3cfc0b823bfd5d8c1dd9484c525afa16aacfbe08#diff-bb186dfb35d33d6e3070828689d37896654f7c5792d6bad84139df67b5e493d5R190-R196)

Ah I see, thank you!

cwgoes (approved):
ACK, thanks for the responses & updates!

library/Core/src/Juvix/Core/Erased/Ann/Types.hs
```diff
@@ -67,6 +75,8 @@ data AnnTerm primTy primVal = Ann
 -- TODO ∷ make usageFromType Fold!
 usageFromType :: Type primTy -> [Usage.T]
 usageFromType (Pi usage _x xs) = usage : usageFromType xs
+usageFromType CatProduct {} = []
+usageFromType CatCoproduct {} = []
 usageFromType Sig {} = []
```
jonaprieto, line 80: Sig contains usage information, why is [] here?

library/Core/src/Juvix/Core/Erased/Ann/Types.hs
```diff
@@ -67,6 +75,8 @@ data AnnTerm primTy primVal = Ann
 -- TODO ∷ make usageFromType Fold!
 usageFromType :: Type primTy -> [Usage.T]
```
jonaprieto, line 76: I think we can simplify all the trivial cases with `usageFromType _ = []`

jonaprieto (changes requested):

jonaprieto (approved):

PR #992, opened 2021-09-16
Higher Order Function Golden Tests: Michelson

mariari:


rokopt (approved):
Given we can't test this with a non-constant `g` because of #975, this looks good to me.

cwgoes (approved):
ACK

Issue #1004, opened 2021-09-23
EvalError TooManyPrimArguments for Michelson `primApply`.


systemf-frank:
The tests for Michelson are failing after merging the changes in
- https://github.com/anoma/juvix/pull/979
into
- https://github.com/anoma/juvix/pull/906.

Running the following code:
```haskell
mod Identity where

open Prelude
open Michelson

sig make-pair : list operation -> int -> pair (list operation) int
let make-pair = %Michelson.pair

sig nil : list operation
let nil = %Michelson.nil

sig car : pair int int -> int
let car = %Michelson.car

sig main : pair int int -> pair (list operation) int
let main = \params -> make-pair nil (car params)
```

Gives us the following error:
```haskell
Error: TypecheckerError
    ( EvalError
        { evalErr = ErrorValue
            ( CannotApply
                { fun = VPrimTy
                    ( Return
                        { retType = PrimType
                            { getPrimType = STAR :| [] }
                        , retTerm = List
                        }
                    ) ()
                , arg = VPrimTy
                    ( Return
                        { retType = PrimType
                            { getPrimType = STAR :| [] }
                        , retTerm = PrimTy ( Ty TOperation : )
                        }
                    ) ()
                , paramErr = ApplyErrorT
                    ( Extra
                        ( TooManyPrimArguments
                            ( Take
                                { usage = SAny
                                , type' = PrimType
                                    { getPrimType = STAR :| [] }
                                , term = List
                                }
                            )
                            ( Take
                                { usage = SAny
                                , type' = PrimType
                                    { getPrimType = STAR :| [] }
                                , term = PrimTy ( Ty TOperation : )
                                } :| []
                            )
                        )
                    )
                }
            )
        }
    ) on Term: Lam
    ( Elim
        ( App
            ( App
                ( Free
                    ( Global
                        ( "Identity" :| [ "make-pair" ] )
                    ) ()
                )
                ( Elim
                    ( Free
                        ( Global
                            ( "Identity" :| [ "nil" ] )
                        ) ()
                    ) ()
                ) ()
            )
            ( Elim
                ( App
                    ( Free
                        ( Global
                            ( "Identity" :| [ "car" ] )
                        ) ()
                    )
                    ( Elim
                        ( Bound 0 () ) ()
                    ) ()
                ) ()
            ) ()
        ) ()
    ) ()
```

We can see here that `List` has type `*`. List has arity 1, so it should have type `* -> *`.

systemf-frank:
[...]

PR #1005, opened 2021-09-24
Fix too many arguments error

Acentelles:
Fixes https://github.com/anoma/juvix/issues/1004


andy-morris:
ah good catch, thanks!

andy-morris:
though, maybe this and [the other place](https://github.com/anoma/juvix/blob/alberto/fix-too-many-arguments/library/Pipeline/src/Juvix/Pipeline/Compile.hs#L193-L195) could be factored out or something, since the source of the problem was two occurrences of the same logic?

andy-morris (approved):

systemf-frank:
Huh both of the commits are done by drone? Something wrong with you git setup?

cyberglot:
> Huh both of the commits are done by drone? Something wrong with you git setup?
When they said robots would take over programming I laughed, but they're indeed onto us.

library/Core/src/Juvix/Core/IR/CheckTerm.hs
```diff
@@ -179,7 +180,11 @@ typeTerm' term ann@(Typed.Annotation σ ty) =
     Core.PrimTy t _ -> do
       requireZero σ
       void $ requirePrimStars (Param.primArity t) ty
-      let t' = App.Return {retTerm = t, retType = Param.PrimType $ Param.STAR :| []}
+      let t' =
+            App.Return
+              { retTerm = t,
+                retType = Param.PrimType $ NonEmpty.fromList $ replicate (fromIntegral (Param.primArity t + 1)) Param.STAR
```
systemf-frank, line 186: A recursive function of over `ty` should also work, I guess. Would that be nicer?

PR #1030, opened 2021-10-22
Simple universe checking

andy-morris:
- [x] add `UAny` syntax to core
- [x] remove universe field from `Ann`
- [x] check against `UAny` instead of `U 0` in `Ann` branch of typechecker and when checking a signature in the pipeline


library/Core/src/Juvix/Core/Erased/Ann/Conversion.hs
```diff
@@ -128,7 +128,7 @@ typecheckErase' ::
   m (Erasure.TermT primTy primVal, Typed.ValueT IR.T primTy primVal)
 typecheckErase' term usage ty = do
   -- FIXME: allow any universe!
```
Acentelles, line 130:
```suggestion
```

Remove fixme?

Acentelles (approved):

library/Core/src/Juvix/Core/Base/Types/Base.hs
```diff
@@ -14,7 +14,23 @@ import Juvix.Library.Usage
 
 ------------------------------------------------------------------------------
 
-type Universe = Natural
+-- | a concrete universe level is just a natural number.
+type ConcUniverse = Natural
```
Acentelles, line 18: Would it be too verbose to have `ConcUniverse` as a `newtype`?

andy-morris: oops, yeah, thanks!

andy-morris: it's only a synonym because the old `Universe` was. I'll make it a `newtype` now and whether it's a pain or not.

PR #1035, opened 2021-11-01
Get correct type kind on typeToTerm

Acentelles:


andy-morris:
Good catch, but also, the result is only well-typed if `ty` does actually have kind `STAR`, so, might it be worth adding an assertion or something to make sure it is actually true?

Acentelles:
I'm trying to find a way to check whether `ty` has type `STAR` but it seems to me that will imply introducing the typechecker into the evaluator. For example, if I attempt to use `requireStar`, the signature already needs `HasThrowTC'` where TC stands for `TypeChecker`.
Do you have any suggestions? @andy-morris

andy-morris:
You can just check its `primArity` is zero. (Sorry, I didn't see this reply until now)

PR #1036, opened 2021-11-01
Remove a comment that breaks the Haddock build

rokopt:
This PR deletes a commented-out line of code because it breaks the Haddock build (at least when I try it).


library/StandardLibrary/src/Juvix/Library/Trace/Format.hs
```diff
@@ -35,7 +36,6 @@ traceLog ::
   T -> Integer -> (Integer -> Integer) -> [Stack] -> [String]
 traceLog t level indentationIncrement stacks =
   (stacks >>= traceStack t level indentationIncrement)
-    -- |> reverse
```
mariari, line 38: Does this break it.... intersting

mariari (approved):

PR #1038, opened 2021-11-02
Stdlibs

Acentelles:
Fetch stdlibs from repository and cache them locally under `~/.juvix/stdlib/`

Standard libraries live in https://heliax-juvix-artifacts-v1.s3.eu-west-1.amazonaws.com (e.g. see https://heliax-juvix-artifacts-v1.s3.eu-west-1.amazonaws.com/Prelude.ju).

Addresses https://github.com/anoma/juvix/issues/1033

Removes the need to specify the path to the standard libraries in the tests. E.g. instead of
```
Pipeline.parseWithLibs ["../../../stdlib/Prelude.ju", "../../../stdlib/Circuit.ju", "../../../stdlib/Circuit/Field.ju"]
```
we do `Pipeline.parse`

It would be nice to start thinking about a package manager.


library/StandardLibrary/src/Juvix/Library/StdLib.hs
```diff
@@ -0,0 +1,56 @@
...
+localStdLibs = do
+  d <- getJuvixHome "stdlib/"
+  doesDirectoryExist d
+
+loadStdLibs :: IO ()
+loadStdLibs = do
+  -- TODO: How long do we want to cache this?
+  -- TODO: Checksum
+  -- success <- localStdLibs
+  downloadStdLibs
```
mariari, line 56: Would be best to have `loadStdLibs` to not download and install the libs, but rather loads them from disk only.

With a separate command to download them.

Within the make file we can run `make install` to install the standard library in the desired directory

app/Main.hs
```diff
@@ -96,5 +100,6 @@ main = do
   home <- getHomeDirectory
   let ctx = Context pwd home
   progVersion <- progNameVersionTag
+  loadStdLibs
```
mariari, line 103: Is this just used to donwload it upon any action, or do we ingest it into memory via this somehow?

mariari (commented):
Would be nice to have a separate command to fetch built into the make file.

It seems at the moment that it will try upon every command to install the standard-library.

Thus I propose

- `make install` which builds and installs the standard library
- `make fetch-stdlib` to fetch the standard library alone

Upon starting up juvix it will look at the directory if its empty it will instruct you to run one of these commands.

mariari:
> It would be nice to start thinking about a package manager.

A package manager would be nice, however I think we should wait to have a working juvix Image/Binary, as the process for making such a tool should follow a process as follows.

1. Juvix gives primitives for reading code
2. Juvix gives tools for ingesting files
3. With 1. and 2. we then create a buil tool that abstracts over the process 1. and 2.
4. with 3. done, we can abstract upon this for packages. This tool can be used in:
    1. the REPL itself to manage and load libraries quickly while in development or experimenting
    2. the CommandLine to give ways to install things system wide when not in the repl
    3. Juvix Package files. These files will outline how projects ought to be loaded and can be executed by 1. or 2.
    4. A gui that allows easy management (which is just a Juvix GUI program that presents the tools representation of packages, etc.).

library/BerlinPipeline/test/Main.hs
```diff
@@ -11,4 +12,6 @@ allCheckedTests =
     []
 
 main :: IO ()
-main = T.defaultMain allCheckedTests
+main = do
+  loadStdLibs
```
mariari, line 16: ditto here with the load/download distiniction.

app/Main.hs
Acentelles: We check they are installed first
```
loadStdLibs :: IO ()
loadStdLibs = do
  success <- localStdLibs
  unless success downloadStdLibs
```

mariari: Checking the diffs, this `loadStdLibs` gets run upon any command (it's been propagated to test main functions as well). I don't think it's a good decision to have the compiler grab files online if some directory does not exist. I'd be more comfortable wtih it if it were a package manager as there yes you have to interact with the internet.

However since we are building from source and not distributing a binary, the make install command can cover this case. 

Upon the files not being there we can repurpose the injections to tell the users to run a command, and even have a restart prompt with the following options

```
1: grab files from local source

2. grab online automatically

3. error and don't continue

4. continue at your own risk 
```
Pressing `1` prompts the user with giving a path, if no path is given it shows a default of the current path

Acentelles: That can be a separate PR

PR #1039, opened 2021-11-02
Add golden and HUnit test infrastructure for LLVM types and instructions, plus support for mul

rokopt:
This PR proposes to add golden-test and HUnit infrastructure for testing LLVM types and instructions, to try to make it straightforward and repeatable to add new ones.  It also adds support for `i16` and `mul` (and fills in a missing bit of support for `sub`) as a test of the tests (as well as just to have those things in and of themselves).

This should be followed up with a PR to add whatever more types and instructions we think we should add support for immediately, but I think that this is a good point for a first PR so that people can decide whether the test infrastructure itself is right, before I start adding a bunch more instances of code that _isn't_ right!


library/Backends/llvm/test/Test/Parameterization.hs
```diff
@@ -0,0 +1,71 @@
...
+    typecheckLLVMSubTest,
+    typecheckLLVMMulTest
+  ]
+
+typecheckLLVMBinOp :: TestName -> RawPrimVal -> TestTree
+typecheckLLVMBinOp instrName instr =
+  testCase ("Typecheck the LLVM " ++ instrName ++ " instruction") $ do
+    hasLLVMType instr int8BinOp @?= True
+    hasLLVMType instr int16BinOp @?= True
+    hasLLVMType instr (primFunc1 LLVM.i8 LLVM.i8) @?= False
```
Acentelles, line 55: Are these necessary? i.e. the arity is fixed and `binOp` already guarantees that the two arguments and return type are of the same type

Acentelles (approved):

library/Backends/llvm/test/Test/Parameterization.hs
rokopt: I might be misunderstanding your comment , but `hasLLVMType` is the function that I'm testing here, so I'm specifically not trusting that it behaves sensibly even in the presence of `binOp`'s guarantees.

PR #1042, opened 2021-11-03
Update core docs

andy-morris:


Acentelles (approved):
Looks much better. Let's continue in a separate PR

PR #1043, opened 2021-11-04
Berline Pipeline: initial Step implementation

cyberglot:


library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
```diff
@@ -0,0 +1,48 @@
...
+    ( ComputationalInput compin,
+      WorkingEnv env
+    ) =>
+    compin ->
+    ComputationalOutput env
+
+data Named = NameStep
+  { name :: NameSymbol.T,
+    step :: forall compin env compout. T compin env
+  }
```
mariari, line 31: we should avoid a forall hidden inside the step, as it makes stopping at a random step much harder

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
```diff
@@ -0,0 +1,48 @@
...
+  )
+where
+
+import qualified Juvix.BerlinPipeline.Meta as Meta
+import qualified Juvix.Sexp as Sexp
+import qualified Juvix.StandardLibrary.NameSymbol as NameSymbol
+
+data Env
+  = InContext NameSymbol.T
+  | SExp Sexp
```
mariari, line 18: I think this belings in `Juvix.BerlinPipeline` not here

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
```diff
@@ -0,0 +1,48 @@
+module Juvix.BerlinPipeline.Step.Types
```
mariari, line 1: We probably don't need a types file for `Step`

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
```diff
@@ -0,0 +1,48 @@
...
+class WorkingEnv (a :: * -> *) where
+  currentExp :: [Env]
+  context :: Context.T Sexp.T
+
+class ComputationalInput a where
+  languageData :: WorkingEnv a
+
+class StepMetaInfo a where
+  name :: Maybe NameSymbol.T
+  meta :: Meta.T
```
mariari, line 42: I think these belong in `Juvix.Pipeline` and without the generics

mariari (changes requested):

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
cyberglot: how so?

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
cyberglot: Without typeclasses, they will use the same types, that would make the structure too rigid.

mariari: Yes, they will have to be the same type. If we look at the type of `WorkingEnv` it has two fixed contents a list of `EnvOrSexp` which contains either the S-expression itself, or a reference to the object in the `Context`. And the context itself, fixed to only hold `Sexpressions`

Since `ComputationalInput` only contains this and the `SurroundingEnvironment`, which is also fixed to have consistent data to flow to each step of the pass, it will also be fixed.

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
mariari: Namely since we get a fixed environment and a fixed input, we can avoid the issue with our library where a pass needs to change one kind of `Step.T compin1 env1` to `Step.T compin2 env2` which would create friction in the pipeline itself. Namely since we have `register_step` which will induce an effect like follows

```haskell
registerStep :: HasState "registeredPipeline" CircularList.T Step.Named => CircularList.T Step.Named -> ()
```

Following the hidden forall this still works, however since the input and output are directly fed by the next and previous step, trying to call these functions would make the haskell type checker force us to coerce it to the right type, as `eval` is responsible from taking this `CirularList.T` that we are building up and executing it. Meaning that we will have to coerce the types as we run.

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step/Types.hs
mariari: (I guess it's fine if we have an `OO` style overloading of data and it's presentation, but I'm unsure of what it buys us here. I think that is my main issue?)

mariari (approved):
Approved, though a lot of changes will need to be had in a followup PR

PR #1047, opened 2021-11-05
Added BerlinPipeline to yaml generator

systemf-frank:
Although there was a definition for BerlinPipeline, it wasn't added as a dependency to juvix. This PR fixes that.

This PR also removed an import in the tests.


jonaprieto (approved):
Looks good to me.

PR #1049, opened 2021-11-05
Fix docs path in Makefile's org-generation target

rokopt:
The `doc` directory was recently renamed to `docs`; the `Makefile`'s `org-generation` target doesn't reflect that yet, though.  This in particular blocks running `precommit.sh`.


Acentelles (approved):
Good catch! Thank you

PR #1050, opened 2021-11-05
LLVM Custom Monad Revival

mariari:
This PR brings back the custom monad infrastructure and DSL over the primitives of the `LLVM-HS-Pure` library.


library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Constants.hs
```diff
@@ -0,0 +1,25 @@
...
+
+import Juvix.Library hiding (Type)
+import qualified LLVM.AST.Type as Type
+
+portLength :: Num p => p
+portLength = 32
+
+-- | An abbreviation for 'IntegerType' 4
+i4 :: Type.Type
+i4 = Type.IntegerType 4
```
Acentelles, line 12: Do we want `i4`?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Constants.hs
```diff
@@ -0,0 +1,25 @@
...
+
+-- | An abbreviation for 'IntegerType' 8
+i8 :: Type.Type
+i8 = Type.IntegerType 8
```
Acentelles, line 16: We can use `Type.i8` right?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
```diff
@@ -0,0 +1,308 @@
...
+  = -- | Error when a block does not exist
+    NoSuchBlock Text
+  | -- | Error when a Variant does not exist
+    NoSuchVariant Text
+  | -- | Error that should never happen
+    DoesNotHappen Text
+  | -- | Error that happens when a variable out of scope is called
+    VariableNotInScope Text
+  | -- | Error that happens when a block lacks a terminator when it should have one
+    BlockLackingTerminator Int
```
Acentelles, line 62: Where do some of these errors come from?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
```diff
@@ -0,0 +1,308 @@
...
+  }
+  deriving (Show)
+
+--------------------------------------------------------------------------------
+-- LLVM Type Operations
+--------------------------------------------------------------------------------
+
+-- TODO :: Replace with safe lens call instead!
+intoStructTypeErr :: Integral a => Type -> a -> Type
+intoStructTypeErr typ' i = elementTypes typ' !! fromIntegral i
```
Acentelles, line 234: Is it possible to have this unsafe lookup removed from this PR?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,34 @@
...
+
+type Names = Map.T Symbol Int
+
+uniqueName :: Symbol -> Names -> (Symbol, Names)
+uniqueName nm ns =
+  case Map.lookup nm ns of
+    Nothing -> (nm, Map.insert nm 1 ns)
+    Just ix -> (intern (show nm <> show ix), Map.insert nm (succ ix) ns)
+
+instance Hashable Name
```
Acentelles, line 34: How does this instance behave?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,34 @@
+-- | Shared between Types and Sum
+module Juvix.Backends.LLVM.Codegen.Types.Shared where
+
+import Juvix.Library hiding (Type)
+import qualified Juvix.Library.HashMap as Map
+import LLVM.AST
+import qualified LLVM.AST as AST ()
```
Acentelles, line 7: Why do we do this?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,34 @@
...
+import qualified LLVM.AST.Constant as C ()
+import LLVM.AST.Global as Global ()
```
Acentelles, line 9: I understand this as only importing instances. Which are we importing?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,34 @@
+-- | Shared between Types and Sum
+module Juvix.Backends.LLVM.Codegen.Types.Shared where
+
+import Juvix.Library hiding (Type)
+import qualified Juvix.Library.HashMap as Map
+import LLVM.AST
+import qualified LLVM.AST as AST ()
+import qualified LLVM.AST.Constant as C ()
```
Acentelles, line 8: Remove?

Acentelles (approved):
Ack modulo minor changes

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
mariari: The first one happens when the variable is not in the map when we wish to call it, so we throw variable not in scope erorr.

In the second case, LLVM demands every block has a terminator of some kind (either in a ret or a jump), and thus this occurs if we don't generate one

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
mariari: The instance or the unique name code above it?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
mariari: We probably should remove them

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
mariari: Agreed

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Constants.hs
mariari: We could rexport them from `Type` perhaps. This is more of a constant LLVM type, I think there is a distinction here between Haskell types and LLVM types hence this split.

However I'm fine in breaking it up to have constants in Types.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
systemf-frank: Should be good to go, right?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,34 @@
+-- | Shared between Types and Sum
+module Juvix.Backends.LLVM.Codegen.Types.Shared where
```
systemf-frank, line 2: Explicit exports.

library/Backends/llvm/stack.yaml
```diff
@@ -9,6 +9,7 @@ packages:
 - ../../Frontend
 - ../../Sexp
 - ../../Translate
+- ../../Test/DataStructures
```
systemf-frank, line 12: This is unrelated to the rest of the PR, right?

scripts/yaml-generator/src/config.lisp
```diff
@@ -370,7 +370,7 @@ common ones to include"
    :name "Backends/llvm"
    :resolver 17.3
    :path-to-other "../../"
-   :packages (list *standard-library* *core* *context* *pipeline* *frontend* *sexp* *translate*)
+   :packages (list *standard-library* *core* *context* *pipeline* *frontend* *sexp* *translate* *datastructures*)
```
systemf-frank, line 373: Same here?

systemf-frank (changes requested):
Looks good, see comments for specifics.

In general I would love to see haddock-style comments to the functions, at least the ones that are exported.

Additionally I think that there are a lot of unused functions, am I right (taking into account that some will only be used once #1052)?

library/Backends/llvm/stack.yaml
mariari: Yes, but needed to actually run `stack test` on the LLVM side

scripts/yaml-generator/src/config.lisp
mariari: Needed to run `stack test` on the llvm package

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -0,0 +1,814 @@
...
+        (mkName label)
+
+--------------------------------------------------------------------------------
+-- Printing facility
+--------------------------------------------------------------------------------
+
+definePrintf :: External m => m ()
+definePrintf = do
+  let name = "printf"
+  op <- externalVar Type.i32 name [(Types.pointerOf Type.i8, "n")]
```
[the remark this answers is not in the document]
mariari: For functions like `prtinf` we can use it for debugging code. I think in the long run we move away from C style `printf`, however it can be quite useful when we actaully have contracts that do any amount of computation.

As for `malloc`, we can view it as an allocation api that we can build ourselves, so should be viewed as an opaque interface to a `malloc` function

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -0,0 +1,814 @@
...
+--------------------------------------------------------------------------------
+-- Module Level
+--------------------------------------------------------------------------------
+
+emptyModule :: ShortByteString -> Module
+emptyModule label = AST.defaultModule {moduleName = label}
+
+-- TODO :: Figure out the semantics of a function like this.
+inModule :: HasState "moduleAST" Module m => Module -> m ()
+inModule = undefined
```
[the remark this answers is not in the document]
mariari: I'm not sure, The reason for it sometimes, is say you want to spell out an API. Seems Haskell gives no `MLI` like functionality so how do we do that? To me it's not clear how to spell out an API for others to fill in if we don't accept undefined at the top level for those instances.

This function is supposed to signal that we should think about the LLVM module level definitoins, so I'm inclined to make something, but I'll delete this specific case for now.

mariari:
> Looks good, see comments for specifics.
> 
> In general I would love to see haddock-style comments to the functions, at least the ones that are exported.
> 
> Additionally I think that there are a lot of unused functions, am I right (taking into account that some will only be used once #1052)?

There are unused functions with the current backend, but they are quite important functions that will occur as we actually make the backend properly feature. 

These for example include
- `getElementPtr`
- `loadElementPtr`
- `addBlock`
- `switch`
- `icmp`
- `load`
- `getBlock`
- `addBlock`
- ETC.

They should not be removed, as it's needless to remove them now then add them back as we actually use them. Namely because the application writer should not have to think often about extending the base layer as he writes the applcaiton. If he were concerned with this, then a lot of time will be spent reinventing the same calls that are already given here. The IRBuilder monad often provides these same functions [the gep function for example is here](https://hackage.haskell.org/package/llvm-hs-pure-9.0.0/docs/LLVM-IRBuilder-Instruction.html#v:gep) for the same job, but can provide a somewhat complete layer

mariari:
Failing CI, will be fixed when the PR is approved, to avoid excess rebasing every time the make file changes.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -0,0 +1,43 @@
...
+data SumInfo = S
+  { sum' :: Symbol,
+    offset :: Int,
+    tagSize' :: Word32
+  }
+  deriving (Show, Eq)
+
+-- | a mapping between the variant and the sum type along with
+-- the tag associated with it
+type VariantToType = Map.T Symbol SumInfo
```
systemf-frank, line 29: Why is this necessary for the custom monad?

library/Backends/llvm/src/Juvix/Backends/LLVM/Parameterization.hs
```diff
@@ -52,7 +53,12 @@ llvm =
     builtinTypes :: Param.Builtins PrimTy
     builtinTypes =
       [ ("LLVM.int8", PrimTy LLVM.i8),
-        ("LLVM.int16", PrimTy LLVM.i16)
+        ("LLVM.int16", PrimTy LLVM.i16),
+        ("LLVM.string", PrimTy (Codegen.Type.pointerOf LLVM.i8)),
+        -- PLEASE FIX --
+        ("LLVM.set", PrimTy (Codegen.Type.pointerOf LLVM.i8)),
+        -- PLEASE FIX --
+        ("LLVM.list", PrimTy (Codegen.Type.pointerOf LLVM.i8))
```
systemf-frank, line 61: Why is this added in this PR?
Why are the pointers fixed to be of type i8? How would this work on other targets / architectures? Is this the thing to fix?

mariari: > Why is this added in this PR?
The codebase refuses to build without adding these in the latest pr, thus it must be addressed

> Why are the pointers fixed to be of type i8? How would this work on other targets / architectures? Is this the thing to fix?
I'ts fine to represent strings as pointers to `i8`, as when we consider ascii chars they are i8, ideally we should probably use `i16` to fit alll of unicode, however that is not here nor there. As for architectures, it should not matter, as LLVM has integers as primitives and can compile this to all sorts of architectures.

The `PLEASE FIX` is for sets and lists at we should not encode them this way in any functional way, unlike strings which can work with ascii

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
mariari: It's left in the shared file, we probably can move it to `1051` though.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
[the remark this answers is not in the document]
mariari: the `Type.i8` should always be fine here, as for why the types are hte way they are, they are even that way in the llvm language reference manual for an example

https://llvm.org/docs/LangRef.html#parameter-attributes

systemf-frank (approved):

PR #1051, opened 2021-11-05
Sum Type InfraStructure

mariari:
This PR is a followup of #1050 which adds Sum type information to the custom monad.

Documentation will be added later showing some design patterns on how to use sum type creation.


library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Sum.hs
```diff
@@ -0,0 +1,143 @@
...
+tagSizeIntExn :: Type -> Word32
+tagSizeIntExn (Type.IntegerType i) = i
+tagSizeIntExn _ = Prelude.error "unsupported operation"
+
+createVariantName :: Symbol -> Symbol -> Symbol
+createVariantName sumName varName = sumName <> "-" <> varName
+
+-----------------------------------------------------------------------------------------
+-- Important functions
+-----------------------------------------------------------------------------------------
```
rokopt, line 71: Aren't they all?  ;)  (If there's a common theme to the functions below this block comment, it might be more informative to call that out.)

rokopt (commented):
Jeremy and I are going to walk through this on Monday and maybe make some further updates then; for now I'm just leaving some small stylistic comments while reading through the changes in preparation for that.

rokopt (approved):
I've rebased and responded to (my own) code review comments (Jeremy said I could feel free to do so).

PR #1052, opened 2021-11-05
Replacing Current LLVM Backend with the Builder

mariari:
Currently the LLVM backend uses the Monad builder from `llvm-hs-pure` this PR seeks to use #1050 for this.

This grants us a more independent platform that gives us the freedom of deciding 
1. Calling conventions
2. Garbage collection
3. Flags for: add, mul, sub, div
4. Attribute parameters
5. etc.

Along with extra features that having our own custom monad has like:
1. inner function compilation
2. Unique name generation
3. Better preparation for the Berlin Pipeline.

This PR also fixes the bug where multiple lambdas share the same name.


test/examples-golden/positive/llvm/Instructions/Instructions.llvm
```diff
@@ -4,46 +4,53 @@
  
 
 
-define external ccc  i8 @lam_0(i8  %arg_0)    {
-  ret i8 99 
+define external fastcc  i8 @"\22lambda\226"(i8  %"5")    {
+"\22lambda\226":
```
systemf-frank, line 8: Hmm what is up here? Problem with variable name printing?

test/examples-golden/positive/llvm/Instructions/Instructions.llvm
```diff
@@ -4,46 +4,53 @@
...
+    ret i8 99 
 }
 
 
-define external ccc  i8 @lam_0(i8  %arg_0)    {
-  %1 = mul   i8 50, 51 
-  %2 =  call ccc  i8  @lam_0(i8  %1)  
-  ret i8 %2 
+define external fastcc  i8 @"\22lambda\225"(i8  %"4")    {
+"\22lambda\225":
```
systemf-frank, line 14: Same

test/examples-golden/positive/llvm/Instructions/Instructions.llvm
```diff
@@ -4,46 +4,53 @@
...
+  %0 = mul nuw nsw i8 50, 51 
+  %1 =  call fastcc  i8  @"\22lambda\226"(i8  %0)
```
systemf-frank, line 16: Same (and more down below)

test/examples-golden/positive/llvm/MainApply/MainApply.llvm
```diff
@@ -4,12 +4,14 @@
  
 
 
-define external ccc  i8 @lam_0(i8  %arg_0, i8  %arg_1)    {
-  ret i8 %arg_0 
+define external fastcc  i8 @lambda(i8  %"0", i8  %"1")    {
+"\22lambda\221":
```
systemf-frank, line 8: Same here (and below)

test/examples-golden/positive/llvm/MainLambda/MainLambda.llvm
```diff
@@ -4,12 +4,14 @@
  
 
 
-define external ccc  i8 @lam_0(i8  %arg_0, i8  %arg_1)    {
-  ret i8 %arg_1 
+define external fastcc  i8 @lambda(i8  %"0", i8  %"1")    {
+"\22lambda\221":
```
systemf-frank, line 8: Again

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -1,38 +1,47 @@
...
+  case Block.runEnvState (mkMain t) mempty of
+    (Right _, state) ->
+      state
+        |> Types.moduleAST
+        |> LLVM.ppllvm
+        |> toStrict
+        |> return
+    (Left err, _) -> do
+      show err
+        |> Feedback.fail
```
systemf-frank, line 40: Is there any way for the codegen to fail, without it being a bug in the codegen?

In my mind feedback is only used for errors where the user is involved. Regular `error` is for compiler implementation errors. For example a failed lookup on a variable should not never happen in the backends, and is therefore an `error` instead of a `Feedback`. What do you think?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -48,115 +57,145 @@ compileProgram t = do
...
+        y <- compileTerm (xs P.!! 1)
+        Block.sub (typeToLLVM ty) x y
+      Mul -> do
+        x <- compileTerm (xs P.!! 0)
+        y <- compileTerm (xs P.!! 1)
+        Block.mul (typeToLLVM ty) x y
+  | otherwise =
+    throw @"err"
+      ( Types.WrongNumberOfArguments
+          ("Was expecsting " <> show (arityRaw f) <> "but got " <> show (lengthN xs))
```
systemf-frank, line 192: expecsting -> expecting

systemf-frank (dismissed):
Looking good, see comments for specifics.

test/examples-golden/positive/llvm/Instructions/Instructions.llvm
mariari: Seems the unique name generator I have generates
`"\"lambda\"1"`

we can create an issue after this PR to clean it up, as it just ugly rather than stopping operation

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
mariari: How else do you express errors to the user? It seems like `feedback` is the only mechanism we can give back here. We don't return an either, and the way it's used is used for the pipleine itself does give fatal errors IIRC.

rokopt (approved):
Jeremy and I went through these (and he gave me a tour of the rest of the LLVM backend too!) on a call, and it looks good to me.  Thanks!

PR #1053, opened 2021-11-05
Fix bug where directory is always created, Also prompt user for dl

mariari:
This PR fixes a slight bug in the remote stdlibs download, along with prompting the user if they want to download the files online


library/StandardLibrary/src/Juvix/Library/Fetch.hs
```diff
@@ -55,4 +54,9 @@ localStdLibs = do
 loadStdLibs :: IO ()
 loadStdLibs = do
   success <- localStdLibs
-  unless success downloadStdLibs
+  unless success $ do
+    print "Standard Library not found; do you want \
+          \ Juvix to download and install the Standard library to ~/.juvix? [yes/no]"
+    continue <- getLine
+    when (continue == "yes" || continue == "y")
+      downloadStdLibs
```
Acentelles, line 62: Every test `main` function of every library runs this function. I don't think it's a good idea to prompt the users for every library test suite

mariari: Well, the user should know if they are grabbing files online from a compiler, and afterall once it's there it won't prompt again as it only need to be told yes once

Acentelles: I believe it adds extra complexity and overhead for the user who runs the tests, including the CI

Acentelles: With the CI we can make sure that the libraries exist first, so I think it'll be fine

Acentelles (approved):

library/StandardLibrary/src/Juvix/Library/Fetch.hs
mariari: > I believe it adds extra complexity and overhead for the user who runs the tests, including the CI

Well if it's an user of the language, then they already have the library installed, and thus no prompt will be given.

If they are running the tests before using the language [1], then they will be prompted to make sure they are comfortable grabbing files off the internet[2].

In the first case, the user will not see a prompt nor it will effect them due to already running `make install` and grabbing the local files. In the other case the user will be rpompted once that they can answer yes, with the caveat that this scenario is unlikely and the overhead for this scenario is very low. 

[1] does anyone even run tests of languages they wish to use??? Typically you just grab it off the package manager or repo and use it

[2] that mind you since they are running tests, you have the entire repo where we can move files without connecting to the internet

PR #1054, opened 2021-11-05
Mariari/llvm new golden tests

mariari:
Creates a new golden test for the new monad


test/examples/positive/llvm/MultipleLambda.ju
```diff
@@ -0,0 +1,11 @@
...
+mod MultipleLambdas where
+
+open Prelude
+open LLVM
+
+sig main : int
+let main =
+  let foo = 3 : int in
+  let bar = 5 : int in
+  bar
```
Acentelles, line 11: Nice! Are there more examples that show what else we can compile now that we couldn't compile before this work?

systemf-frank (approved):

PR #1060, opened 2021-11-08
Use requirements.txt in documentation readme

andy-morris:
At the moment the installation instructions say to install all of the dependencies separately (also, `rtd-tikz` is missing). Instead we should say to just use the dependency list.


andy-morris:
oops, based this on the wrong branch the first time; fixed

Acentelles (approved):

PR #1064, opened 2021-11-09
Allow multiple backends in the HTTP module

Acentelles:
And use LLVM as the default one


library/Playground/HTTP/src/Juvix/Playground/HTTP/Routes.hs
```diff
@@ -27,56 +33,21 @@ import qualified Juvix.Library.NameSymbol as NameSymbol
...
-    :> Post '[JSON] (Result (Core.RawGlobals HR.T (Plonk.PrimTy Fr) (Plonk.PrimVal Fr)))
-
-type ToIR =
-  Description "Convert HR to IR"
-    :> ReqBody '[JSON] (Core.RawGlobals HR.T (Plonk.PrimTy Fr) (Plonk.PrimVal Fr))
-    :> Post '[JSON] (Result (Core.PatternMap Core.GlobalName, Core.RawGlobals IR.T (Plonk.PrimTy Fr) (Plonk.PrimVal Fr)))
+data Backend
+  = Plonk (Plonk.BPlonk Fr)
+  | LLVM LLVM.BLLVM
+  deriving (Eq, Show, Generic, A.ToJSON)
```
mariari, line 43: Guessing the Michelson backend can't work here at all?

mariari (approved):
Looks fine, played with the webiste

library/Playground/HTTP/src/Juvix/Playground/HTTP/Routes.hs
Acentelles: Let me add it

PR #1066, opened 2021-11-10
Update README example

Acentelles:
And remove `string`, `list`, `set` from LLVM stdlib since they are not supported


mariari (approved):
Removes the proper types allowing the build to work :+1:

PR #1070, opened 2021-11-12
More core documentation updates

andy-morris:
Adds `let`, coproducts, and records, and expands on what was already there


andy-morris:
The treatment of primitives I am less certain about, so if that makes no sense, please let me know.

Also there are still three occurrences of `TODO`, as you can see, but they are all separate tasks that can be in future PRs.

Acentelles (approved):

PR #1071, opened 2021-11-15
Gui for Stack-Yaml Tool

mariari:
This tool can visualize dependencies between our local packages


Acentelles:
Really like this. Can you update the README file in yaml-generator showing how to use it?

Acentelles (approved):
Ack modulo README

Acentelles:
Thanks! I think it’s a good practice to point to the references. E.g. in case of ASDF, point to https://en.wikipedia.org/wiki/Another_System_Definition_Facility, or even better: as the section in the readme is related to dependencies, point to the installation/download page

mariari:
> Thanks! I think it’s a good practice to point to the references. E.g. in case of ASDF, point to https://en.wikipedia.org/wiki/Another_System_Definition_Facility, or even better: as the section in the readme is related to dependencies, point to the installation/download page

I can link so for ASDF and Quicklisp. Though ASDF comes with any modern CL install.

PR #1073, opened 2021-11-15
Rename Translate module to Desugar

Acentelles:
+ Move contextify related files to Context


mariari (changes requested):
I think moving these passes to to the Context modulo is a bad Idea.

The context package just provides the context and operations on it. It does not hold passes of the compiler nor should it.

In the future when the Berlin pipeline is in, the distinction between passes with the context and without will become even more meaningless, and we should really strive to unify passes into less modules, not break them up

Acentelles:
Thanks for the feedback. I'm not entirely sure about this changes either but, 

>  It does not hold passes of the compiler nor should it.

Why shouldn't this package hold context-related passes?

mariari:
> Thanks for the feedback. I'm not entirely sure about this changes either but,
> 
> > It does not hold passes of the compiler nor should it.
> 
> Why shouldn't this package hold context-related passes?

Because the package just holds the infrastructure for the data structure that is the `Context`. But the passes should not be attached to the structure itself, instead they should live with something clearly related to the environment. Esp since all passes are going to be expected to have the Context in their environment, whether they use it or not.

PR #1075, opened 2021-11-17
Implement Berlin Pipeline `eval`

Acentelles:
Implements:
- `CircularList`
- `RecursiveList`
- `eval` function


library/BerlinPipeline/src/Juvix/BerlinPipeline/CircularList.hs
```diff
@@ -1,6 +1,48 @@
...
+firstNested (T (Rec [])) = Nothing
+firstNested (T (Rec (x : _xs))) = firstNested (T x)
+
+removeFirstNested :: T a -> T a
+removeFirstNested (T (Anu a)) = T (Rec [])
+removeFirstNested (T (Rec [])) = T (Rec [])
+removeFirstNested (T (Rec (x : xs))) =
+  case removeFirstNested (T x) of
+    T (Rec []) -> T (Rec xs)
+    T otherwise -> T (Rec (otherwise : xs))
```
mariari, line 33: We should probably move the calls that match into `Rec` into the `Reclist` module itself, as Reclist should have these same functions, and would require less wrapping and unwrapping

library/BerlinPipeline/src/Juvix/BerlinPipeline/CircularList.hs
```diff
@@ -1,6 +1,48 @@
...
+
+instance Semigroup (RecList a) where
+  Rec l <> Rec r = Rec $ l <> r
+  Rec l <> a@(Anu _) = Rec $ l ++ [a]
+  a@(Anu _) <> Rec r = Rec $ a : r
+  l@(Anu _) <> r@(Anu _) = Rec $ l : [r]
```
mariari, line 39: We should move this as well

library/BerlinPipeline/src/Juvix/BerlinPipeline/Env.hs
```diff
@@ -37,23 +39,65 @@ newtype EnvS b = EnvS (State T b)
...
+-- | Register the pipeline function to the environment
 registerStep :: CircularList.T Step.Named -> EnvS ()
-registerStep = notImplemented
+registerStep l = do
+  modify@"registeredPipeline" $ \i -> i <> l
 
+-- | Create a named group of pipeline steps or nested grouping of pipeline steps.
 defPipelineGroup :: NameSymbol.T -> [CircularList.T Step.Named] -> CircularList.T Step.Named
-defPipelineGroup = notImplemented
+defPipelineGroup sym ls = foldl' (<>) (CircularList.init sym) ls
```
mariari, line 49: Maybe we should use `CircularList.empty` instead here?

We should probably change the format of these to have name be had over groups of passes, as currently this is unable to actually add the name

library/BerlinPipeline/src/Juvix/BerlinPipeline/Env.hs
```diff
@@ -37,23 +39,65 @@ newtype EnvS b = EnvS (State T b)
...
+                  let out = Pipeline.metaCIn meta namedInput
+                   in out {Pipeline.languageData = wEnv}
+      Just (CircularList.CircSchema ls) -> notImplemented
+    where
+      shouldStop (Just n) named
+        | n == named = True
+        | otherwise = False
+      shouldStop _ _ = False
+      nextStep = CircularList.firstNested pipeline
+      remainder = CircularList.removeFirstNested pipeline
```
mariari, line 100: We should probably have a few helper functions to facilitate this, so that we don't have to do as much work in this function

library/BerlinPipeline/src/Juvix/BerlinPipeline/Env.hs
```diff
@@ -37,23 +39,65 @@ newtype EnvS b = EnvS (State T b)
...
+      stoppingStep
+    ) = do
+    case nextStep of
+      Nothing -> pure input
+      Just (CircularList.NonCircSchema nStep) -> do
+        let name = Step.name nStep
+        let namedInput = Pipeline.nameCIn name input
+        if shouldStop stoppingStep name
+          then pure input
+          else liftIO $ do
```
mariari, line 73: We probably don't want the `liftIO` here, as it seems overally wide scoping over the actual logic

library/BerlinPipeline/src/Juvix/BerlinPipeline/Env.hs
```diff
@@ -37,23 +39,65 @@ newtype EnvS b = EnvS (State T b)
...
+    where
+      shouldStop (Just n) named
+        | n == named = True
+        | otherwise = False
+      shouldStop _ _ = False
+      nextStep = CircularList.firstNested pipeline
+      remainder = CircularList.removeFirstNested pipeline
 
-extract :: EnvS b -> T
-extract = notImplemented
```
mariari, line 59: We should still have an extract, even if it's in `IO` or gets passed an explicit `Context.T`. As it's a useful function to have

library/BerlinPipeline/src/Juvix/BerlinPipeline/Step.hs
```diff
@@ -10,17 +13,13 @@ import qualified Juvix.BerlinPipeline.Pipeline as Pipeline
 import Juvix.Library
 import qualified Juvix.Library.NameSymbol as NameSymbol
 
-data T = T (Pipeline.ComputationalInput -> Pipeline.ComputationalOutput Pipeline.WorkingEnv)
-
-data Named = Named
-  { nsName :: NameSymbol.T,
-    nsStep :: T
-  }
+data T = T (Pipeline.CIn -> IO (Pipeline.COut Pipeline.WorkingEnv))
```
mariari, line 16: Shouldn't this be some `Monad m` that we get here, I guess that might cause issues higher up. So IO should be fine?

mariari (approved):

PR #1077, opened 2021-11-17
Add LLVM string type and global string variable definition function

rokopt:
This PR adds an LLVM string type which is implemented as an `i8*`, along with a function used to build literal strings by creating a global array of `i8` and then casting it to a pointer with `GEP`.

The code it's generating looks sane to me, but we can't return a `string` type from `main`, so to see it work we'll also need to add support for some imperative print routine such as `puts` or `printf` (or hand-edit the output program to add a call to one!).


rokopt:
Sigh... the CI is failing with an indication that the affected LLVM golden test doesn't match the expected output, but there are no diffs between the expected and actual output (not even whitespace) that I (or `diff`) can find.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -195,6 +198,14 @@ compilePrimApp ty f xs
 -- | Write LLVM code for a primitive.
 -- TODO: implement other primitives.
 mkPrim ::
+  ( HasThrow "err" Types.Errors m,
+    HasState "blocks" (T Name.Name Types.BlockState) m,
```
Acentelles, line 202: Do we need all these constraints?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -646,6 +647,34 @@ printf args = do
...
+        }
+  getElementPtr $
+    Types.Minimal
+      { Types.type' = Types.pointerOf Type.i8,
+        Types.address' =
+          ConstantOperand $
+            C.GlobalReference
+              (Types.pointerOf $ ArrayType (fromIntegral (length str + 1)) Type.i8)
+              name,
+        Types.indincies' = constant32List [0, 0]
```
Acentelles, line 675:
```suggestion
        Types.indices' = constant32List [0, 0]
```

library/Backends/llvm/src/Juvix/Backends/LLVM/Parameterization.hs
```diff
@@ -9,11 +9,24 @@ module Juvix.Backends.LLVM.Parameterization
   )
 where
 
+import qualified Juvix.Backends.LLVM.Codegen.Types as Types
 import Juvix.Backends.LLVM.Primitive
 import qualified Juvix.Core.Base.Types as Core
 import qualified Juvix.Core.IR.Evaluator as IR
 import qualified Juvix.Core.Parameterisation as Param
 import Juvix.Library
+  ( Applicative (pure),
```
Acentelles, line 18: Is this intended?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -220,7 +235,7 @@ typeToLLVM ty@(ErasedAnn.Pi _usage _f _xs) =
   Types.pointerOf
     LLVM.FunctionType
       { LLVM.resultType = typeToLLVM resultType,
-        LLVM.argumentTypes = map typeToLLVM argumentTypes,
+        LLVM.argumentTypes = Juvix.Library.map typeToLLVM argumentTypes,
```
Acentelles, line 238: Can't we just use `map` or `<$>`

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
rokopt: We could certainly make that change, but that spelling is in pre-existing code, so we can't make it without changing the other code which defines and uses it.  I'd suggest a separate PR for that, as it's a separate matter from supporting strings (a cosmetic one).

library/Backends/llvm/src/Juvix/Backends/LLVM/Parameterization.hs
rokopt: It was the pre-commit script that introduced it, so I think that if we didn't, then the script would just try to re-introduce into future PRs.  Maybe I'm misunderstanding something, though.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: The new import of `HashMap` brought in another `map`, which is why I had to make the pre-existing one specify which `map` it was referring to.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: Yes, the combination of the new `generateUniqueName` and the `moduleDefinitions` from `globalString` requires all of those.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
Acentelles: I think it's better to import HashMap qualified

Acentelles (approved):
Thanks for these changes 🎉

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
mariari: We should import it qualified, we should also change the `map` call to `fmap`

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
mariari: does the `Define` effect entail these, or should we make a more limited effect?

rokopt: Ah, good call -- `Define` does entail them all.  I'll substitute that.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: I've now realized that the import was only required by an older version of the code, so I should simply remove it!

PR #1079, opened 2021-11-18
Remove \22 from generated LLVM Code

mariari:
See Golden Tests for differences


library/Backends/llvm/test/Test/Golden.hs
```diff
@@ -158,5 +158,10 @@ toEither (Feedback.Success app a) = Right (app, a)
...
 -- Running by hand example
--- x <- Feedback.runFeedbackT (compile "../../../test/examples/positive/llvm/MainApply.ju")
--- toEither x
+printCompile :: FilePath -> IO ()
+printCompile filePath = do
+  x <- Feedback.runFeedbackT (compile filePath)
+  case toEither x of
+    Right (_, str) ->
+      putStrLn str
+    Left err -> print err
```
andy-morris, line 167: is this part of another branch that got caught up?

mariari: no, I just decided to change the comment to a function to just print the given code

andy-morris (approved):
looks good, thanks!

library/Backends/llvm/test/Test/Golden.hs
mariari: Might as well leave it in!

PR #1080, opened 2021-11-18
Closure Implementation

mariari:
This PR seeks to implement the following:

1. A closure Pass to convert Core forms into a closure Environment Friendly Form
2. A LLVM pass that then can compile closures

The Following Issues will be present after this PR is done:
1. We will leak memory due to how argument allocations of closures is done


library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+--------------------------------------------------------------------------------
+-- New Core Form we will Process Over
+--------------------------------------------------------------------------------
+
+data Annotated term = Ann
+  { usage :: Usage.T,
+    annTy :: ErasedAnn.Type Prim.PrimTy,
+    term :: term
+  }
+  deriving (Show)
```
Acentelles, line 85: What's the reason for using this type instead of the existing
```
data AnnTerm primTy primVal = Ann
  { usage :: Usage.T,
    type' :: Type primTy,
    term :: Term primTy primVal
  }
  deriving (Show, Read, Eq, Generic)
```
from `Erased.Ann`? They look identical

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+newtype Index = Index {num :: Natural} deriving (Show)
+
+----------------------------------------
+-- Storage Location and Indexing Types
+----------------------------------------
+
+data CaptureFrom
+  = -- | @FromAmbiantEnv@ represents the values that we capture having a
+    -- proper name rather then being offsets into the environment.
+    FromAmbiantEnv NameSymbol.T
```
Acentelles, line 51:
```suggestion
    FromAmbientEnv NameSymbol.T
```

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -107,32 +142,34 @@ compileVar sym = do
...
-  | otherwise =
-    throw @"err" (Types.UnsupportedOperation "closures are not supported")
+compileLam ty arguments body = do
+  lamName <- Block.generateUniqueSymbol "lambda"
+  compileFunctionEnv lamName ty arguments $
+    do
+      bod <- compileTerm body
+      Block.ret bod
+
+compileClosure ty captures args body = do
```
Acentelles, line 160: Missing sig?

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
mariari: see `term` itself where the `Erased.Ann` version fixes it to the `Erased.Ann.Term` 

```haskell
term :: Term primTy primVal
```
Likewise I had to make a new `AnnTerm` which holds my own `Term`

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
mariari: Fixed!

mariari:
This PR follows the given strategy of 

https://hackmd.io/IONVOGf3SNe7XVNknTgztQ?view#Closure-Conversion

With the minor change that local arguments are no longer referenced by offset, but rather keep their original name

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -716,9 +721,22 @@ malloc size resultType = do
       callConvention
         CC.Fast
         malloc
-        (emptyArgs [Operand.ConstantOperand (C.Int Types.size_t_int size)])
+        (emptyArgs [Operand.ConstantOperand (C.Int 32 size)])
   bitCast i8Ptr resultType
 
+-- Does the sizeof code actually work!?
```
Acentelles, line 727: I don't understand this comment

Acentelles:
Would it be possible to add more test examples specifically for closures
```
mod Foo where

open Prelude
open LLVM

-- Aliases for LLVM builtins.
sig llvm-add : int8 -> int8 -> int8
let llvm-add = %LLVM.add

sig lam : int -> int
let lam x = x

sig main : int
let main =
  let fi = 3 : int in
  let fo = 5 : int in
  lam (llvm-add fi fo)
```

Acentelles:
Have you thought about testing the outcome of running an llvm program after generating it

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
```diff
@@ -0,0 +1,149 @@
...
+storeFunctionPtr :: Types.RetInstruction m => AST.Operand -> AST.Operand -> m ()
+storeFunctionPtr location value = do
+  fnLocation <- getFunctionPtr location
+  -- cast the value for the caller
+  lamBitCasted <- Block.bitCast value functionPtr
+  Block.store fnLocation lamBitCasted
+
+-- | @storeEnvironmentPtr@ takes the location of a closure and the
+-- environment closure to store
+storeEnvironmentPtr :: Types.RetInstruction m => AST.Operand -> AST.Operand -> m ()
```
Acentelles, line 101: Would it be worth wrapping the `value` of this function into a `newtype` to have more type safety? This function has the same signature as `storeFunctionPtr`

Acentelles (approved):
Ack. Left some comments

mariari:
> Would it be possible to add more test examples specifically for closures
> 
> ```
> mod Foo where
> 
> open Prelude
> open LLVM
> 
> -- Aliases for LLVM builtins.
> sig llvm-add : int8 -> int8 -> int8
> let llvm-add = %LLVM.add
> 
> sig lam : int -> int
> let lam x = x
> 
> sig main : int
> let main =
>   let fi = 3 : int in
>   let fo = 5 : int in
>   lam (llvm-add fi fo)
> ```

Sure, but let us do that in a followup pr, as I'd rather not try to run the golden creation file then replace it right now

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
mariari: shocked over `sizeOf` working, though it generates really ugly code to do its job

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
mariari: For any malloc code, we could make a `newtype` to differentiate the value and the place. Though the sigantures of these two functions would still look the same.

mariari: Though we should have a PR to newytpe all malloc calls

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/ClosureConversion.hs
```diff
@@ -0,0 +1,170 @@
...
+            case HashMap.lookup name tyMap of
+              Just ty -> ty
+              Nothing -> panic "unknown captured argument"
+        }
+
+lamArgumentToClosureArgs :: [NameSymbol.T] -> [Types.ArraySlot]
+lamArgumentToClosureArgs names = zipWith f names (enumFrom (0 :: Natural))
+  where
+    f name idx =
+      Types.Slot {previousName = name, newIndex = Types.Index idx}
```
rokopt, line 132: Would it be clearer to factor out the commonality between this and `lamMCapturesToCapture`?  (I realize that the output types are different, so maybe it would just end up looking confusing or obscure.)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
```diff
@@ -0,0 +1,149 @@
...
+
+sizeOf :: Num a => a
+sizeOf = Types.addressSpace * 2
+
+malloc :: Types.Call m => m AST.Operand
+malloc = Block.malloc sizeOf pointer
+
+mallocEnvironment :: Types.Call m => Integer -> m AST.Operand
+mallocEnvironment length =
+  -- is this the correct offset to get it to compile
```
rokopt, line 68: Is this question now answered in the affirmative?  :)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
```diff
@@ -0,0 +1,149 @@
...
+-- | @loadEnvironmentPtr@ Loads and Dereferences the given closure
+-- environment in the given @location@
+loadEnvironmentPtr :: Types.RetInstruction m => AST.Operand -> m AST.Operand
+loadEnvironmentPtr location =
+  Block.loadElementPtr $
+    Types.Minimal
+      { type' = environmentPtr,
+        address' = location,
+        indincies' = Block.constant32List [0, 1]
+      }
```
rokopt, line 149: The case of `Block.loadElementPtr $ Types.Minimal { ...Block.constant32List...}` looks like it might be common; would it be worth factoring out a utility routine for that?

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -67,7 +67,10 @@ data IndexInto = IndexInto
 -- from the closure array/environment or the closure argument array/environment
 data FunctionEnvironment
   = ClosureEnvironment
-  | ArgumentEnvironemnt
+  | -- Currently unused due to how it's impossible to distinguish
+    -- between functions and closures. We should in time compile with
+    -- this, and will be needed for currying
+    ArgumentEnvironemnt
```
rokopt, line 73: Should we maybe have a `to-fix` test for currying, then?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -694,7 +699,7 @@ printCString str args = do
 defineMalloc :: External m => m ()
 defineMalloc = do
   let name = "malloc"
-  op <- external (Types.pointerOf Type.i8) name [(Types.size_t, "size")]
+  op <- external (Types.pointerOf Type.i8) name [(Type.i32, "size")]
```
rokopt, line 702: What made it necessary to move away from `size_t`?

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -108,8 +109,27 @@ compileTerm (Types.Ann _usage ty t) =
...
 
+-- | @compileTermForApplication@ like @compileTerm@ however it will
+-- promote lambdas to closures to make the calling convention for HOF proper.
+compileTermForApplication ::
+  Types.Define m => Types.Annotated Types.TermClosure -> m LLVM.Operand
+compileTermForApplication ann@(Types.Ann _usage ty t) =
+  case t of
+    Types.LamM as body -> do
+      lam <- compileLam ty as body
+      -- WARNING ∷ Leaking memory
```
rokopt, line 126: It might be worth noting in the "Leaking memory" comments that what will close the leak is to implement GC and replace the `malloc` here with the GC's new allocation routine.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -219,6 +288,117 @@ mkPrim (Types.Prim prim) ty = case prim of
...
+  [NameSymbol.T] ->
+  -- | The body of the lambda abstraction.
+  m a ->
+  m LLVM.Operand
+compileFunctionEnv name ty arguments body = do
+  let (llvmArgtyBeforeClosure, llvmRetty) =
+        functionTypeLLVM ty
+      llvmArgNames =
+        fmap (Block.internName . NameSymbol.toSymbol) arguments
+      -- turn the function pointers to closure for any HOF
```
rokopt, line 377: I'd probably spell out "higher-order functions" here.

rokopt (approved):
Jeremy walked me through this code live this morning, and it makes sense to me.  I've made a few tiny change requests, but nothing semantic, so I'm approving.  Thanks, Jeremy!

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
mariari: the LLVM `size_of` gives us back a `i32` and I didn't want to deal with bit casting it to the right type, so I just moved these to i32's for now

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
mariari: Might be indeed, I just left it as two separate things, as what if you want an i64 index list? (silly, but I left it open for us to remove it later I guess)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
mariari: seems so!

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/ClosureConversion.hs
```diff
@@ -0,0 +1,170 @@
...
+    Just id -> Types.FromClosureEnv id
+
+lamMCapturesToCapture :: [NameSymbol.T] -> Environment -> [Types.Capture]
+lamMCapturesToCapture names Env {ofMap = envMapping, tyMap} =
+  zipWith f names (enumFrom (0 :: Natural))
+  where
+    f name index =
+      Types.Capture
+        { location = nameToCaptured name envMapping,
+          slot = Types.Slot {previousName = name, newIndex = Types.Index index},
```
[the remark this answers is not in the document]
mariari: Sure, feel free to commit that suggestion. The issue is, we've gone through many layers of renaming in that it's just the last name it had, not the original original, we should track that as well(???)

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
mariari: Probably would be good, we need to think hard about how to get currying at not cost while folloiwng the rapier technique laid out by SPJ's paper.

We really should not sacrifice much performance for this.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Closure.hs
mariari: Indeed it might be worth doing that. I didn't to leave open the possibility for other index types, but yeah probably warrented

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
mariari: Yeah, we need to think long and hard how to implement SPJ's rapier technique wtihout sacrificing much performance. Ideally we can think hard about LLVM and what would cost us speed of execution and what the best course of action is.

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/ClosureConversion.hs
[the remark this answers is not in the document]
mariari: The issue with `originalName` is that it's not really the origional program name, as it's undergone many transformation of name transitions. Feel free to make the commit though.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
mariari: Namely because `sizeof` returns an i32, and I didn't feel like trying to bitcast it away, so I just left it as is, and demoted malloc to an i32. we can change it back and note of the hack

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
mariari: Feel free to make the commit

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/ClosureConversion.hs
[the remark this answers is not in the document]
rokopt: That's a good point.  Maybe there's some other choice that would convey that it's a name from a prior phase without potential confusion with a name from earlier in the list.  `priorName`?  `coreName`?

PR #1084, opened 2021-11-25
Correct the usages on the type parameters to categorial combinators

rokopt:
The _type_ parameters to the categorial eliminators for product and coproduct are only for type-checking; they're erased at compile time.  Their usages should be zero, even though the usages for the eliminator terms overall can be non-zero.  This bug was preventing some of the tests in `to-fix/llvm/AlgebraicTypes.ju` from working.  (That test is still "to-fix", though, because other limitations still prevent other parts of it from working:  we need product and sum/coproduct type support in LLVM before that can become a "positive" test.)


andy-morris (approved):
good catch!

PR #1086, opened 2021-12-01
Fix reduction rules in docs for bidi syntax

andy-morris:
As written previously, most of the β rules replaced a variable with a *term*, not an *elimination*, which is invalid syntax. The corrected rules are a bit noisier with the new annotations, but they're just shuffling around data that is already immediately available.


docs/source/compiler/core/core-language.md
```diff
@@ -750,19 +750,37 @@ $$
...
 
 Two terms are convertible $s \equiv t$ if there exists a term $u$ with $s
 \Red u \RedR t$.
 
 - TODO talk about conversion in terms of normalisation, and integrate $\eta$
 
+:::{note}
+The β rules have all those annotations on the right hand sides because it is
+only possible to substitute a variable with another elimination. If the
+annotations are not actually needed, they will be pruned by the υ rule.
```
Acentelles, line 781: What is this υ rule?

Acentelles (approved):

docs/source/compiler/core/core-language.md
andy-morris: the last one in the list, which removes an annotation in a checkable context.

this is safe because given a typing problem like `t : T₁ ⇐ T₂`, we check that `t ⇐ T₁` and `T₁ <: T₂`, which, assuming the subtyping relation makes any sense at all, is equivalent to the simplified problem `t ⇐ T₂`.

PR #1089, opened 2021-12-02
Update LLVM stdlibs

Acentelles:
Add LLVM/Int.ju stdlib and use overloaded functions with qualified names


rokopt (approved):

mariari:
Looks fine, we should probably squash :+1:

mariari (approved):

PR #1091, opened 2021-12-03
Correct type-checking of polymorphic recursive types

rokopt:
This fix is an analogue of #1083 for polymorphic recursive type declarations:  we need to have the type in context already when we check the constructors, in case the type is recursive.


rokopt:
The build of `haskell-src-exts` failed:

```
--  While building package haskell-src-exts-1.23.1 (scroll up to its section to see the error) using:
      /drone/workspace/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_3.2.1.0_ghc-8.10.4 --builddir=.stack-work/dist/x86_64-linux/Cabal-3.2.1.0 build --ghc-options ""
    Process exited with code: ExitFailure 1
```

Scrolling up:

```
haskell-src-exts                 > happy: /drone/workspace/.stack/snapshots/x86_64-linux/c5c301f82d74a7a317f0f5c5b1d8314b7c1c17766e6f7511b861bc67457024c4/8.10.4/share/x86_64-linux-ghc-8.10.4/happy-1.20.0/HappyTemplate-arrays-coerce: openFile: does not exist (No such file or directory)
```

@Fraccaman , is this a result of a drone CI change?

Fraccaman:
Doesn't seems so. The CI didn't commit anything back and from the logs it looks like it just executing things inside the `/drone/workspace/` volume. Could it be a problem with the code?

Fraccaman:
Looks like it's working now :)

mariari (approved):
Looks fine, there are mroe spots to insert into the map

rokopt:
> Looks fine, there are mroe spots to insert into the map

Thanks -- I'll make issue(s) for the others.

PR #1092, opened 2021-12-03
Serialization of Core

mariari:


library/Sexp/src/Juvix/Sexp/Structure/CoreNamed.hs
```diff
@@ -79,6 +79,55 @@ data Let = Let
   }
   deriving (Show)
 
+data CatProduct = CatProduct
+  { catProductFst :: Sexp.T,
+    catProductSnd :: Sexp.T
```
rokopt, line 84: Not a big deal, but as we use `left` and `right` elsewhere, it'd seem more consistent to me to use them here too rather than `fst` and `snd`.

rokopt (approved):

library/Sexp/src/Juvix/Sexp/Structure/CoreNamed.hs
mariari: since this is a product, and since tuples are typically accessed with fst and snd, I think it makes more sense here (co-products are the left and right)

PR #1093, opened 2021-12-07
Paramartized S-expressions

mariari:
Simply gives us a paramartized S-expression.

So far no functions have been updated to take advantage of this fact


library/Sexp/src/Juvix/Sexp/Types.hs
```diff
@@ -16,11 +16,13 @@ import qualified Juvix.Library.LineNum as LineNum
 import qualified Juvix.Library.NameSymbol as NameSymbol
 import Prelude (Show (..), String)
 
+type T = Base ()
+
 -- TODO ∷ make Atom generic, and have it conform to an interface?
 -- This way we can erase information later!
```
rokopt, line 22: I think this PR allows you to remove this comment, right?

rokopt (approved):

library/Sexp/src/Juvix/Sexp/Types.hs
mariari: Yeah we probably could get rid of that comment

mariari:
Will make a followup PR with operations over this new addition

PR #1094, opened 2021-12-08
Reforming the fold functions for Sexp Traversal

mariari:


library/Sexp/src/Juvix/Sexp.hs
```diff
@@ -51,103 +57,105 @@ import Prelude (error)
...
-    Cons cs xs ->
-      Cons (foldPred cs pred f) (foldPred xs pred f)
+    Cons (Atom (A name _)) _
+      | pred name -> mapPredStar (f t) pred f
+    Cons _ _ -> map (\t -> mapPredStar t pred f) t
+    Atom ato -> Atom ato
     Nil -> Nil
-    Atom a -> Atom a
+
+traverse :: Monad m => (B a -> m (B b)) -> B a -> m (B b)
```
Acentelles, line 140: Wouldn't it be better to have instances of `Traversable`, `Foldable`, `Functor`, etc?

Acentelles (approved):

library/Sexp/src/Juvix/Sexp.hs
mariari: These aren't the functions over the generic itself, thus the instance of `Functor` `Traversable` etc would be misleading (The functor over a tree goes on every element, recursing down, where as this nested structure will keep the similar list style traversal).

PR #1069, opened 2021-11-12
Add llvm-hs to the deps to test llvm-9 linkage

mariari:

PR #1096, opened 2021-12-09
Allow execution of generated LLVM modules within HUnit tests

rokopt:
This PR brings in the `llvm-hs` package and uses its `ExecutionEngine` to allow the execution of LLVM code generated by our LLVM backend from HUnit tests.

It therefore exports a new interface which returns an LLVM `Module` rather than a string, so that clients (such as HUnit tests) can get at the `Module` to use it within the `ExecutionEngine`.

The new interface takes a `TermClosure` rather than an `ErasedAnn.Term`, so that tests (maybe we'll find other clients who want to do this as well) can use that extended term type, not only the types of terms that Core can generate.


mariari (commented):
I looked at these before. Overall good changes. However we should wait until #1069 gets merged

@Acentelles tell us if you were able to solve the `nix` issues on your end

rokopt:
This PR has been superseded by #1106 , which includes all of this work and more, so I'm closing this one.

PR #1098, opened 2021-12-13
Nix integration via haskell.nix

SimSaladin:


Acentelles:
Fantastic. I can build it on macOS. Thank you!

Acentelles (approved):

cwgoes:
should we merge this?

mariari:
> should we merge this?

Note quite yet, I've just started to use it.

Well it's still building at this point. I want to make sure it works well with the editors and make files etc. etc.

mariari:
Ended up getting around to playing with this.

It seems I got it built for deployment which is nice, however how do we integrate it to the editor? When I remove the stack file and the corresponding package file, `ghci` has some issues.

Would be nice to get this resolved as I rely on that quite heavily for getting work done

Acentelles (changes requested):
We still need to be able to build submodules (libraries) with nix this way, without stack

SimSaladin:
Note that currently this relies on the `stack.yaml` files - Nix expressions are derived from the stack configuration (it doesn't call `stack` at any point, though), so `stack.yaml` files shouldn't be removed. (It is of course possible to convert to using Cabal instead - just remember that that would mean throwing away Stackage snapshot dependency resolution...)

A nice consequence of this is that the same targets are available via Nix that are available via stack: `nix build .#http:exe:juvix-server` does the same thing as `stack build http:exe:juvix-server`, and so on.

All Nix targets (packages) can be listed with `nix flake show --impure`.

Finally, the development shells provide GHC with dependant packages installed. These can be used to launch GHCi without stack e.g. `nix develop .#http:exe:juvix-server -c ghci`

------

>  how do we integrate it to the editor? 

What is meant by the editor here?

mariari:
> Note that currently this relies on the `stack.yaml` files - Nix expressions are derived from the stack configuration (it doesn't call `stack` at any point, though), so `stack.yaml` files shouldn't be removed. (It is of course possible to convert to using Cabal instead - just remember that that would mean throwing away Stackage snapshot dependency resolution...)

we should probably configure stack to talk to your nix script, or something along those lines #1069 gets that, so when I `stack build` it properly grabs the `llvm-9` package from nix and not the OS.

Also it might be nice to have the sub libraries have the nix integration, though if we have some standard nix extension file that would work as well (#1069 adds all the deps they need to the `stack.yaml` but you can make it run a startup script)

> 
> A nice consequence of this is that the same targets are available via Nix that are available via stack: `nix build .#http:exe:juvix-server` does the same thing as `stack build http:exe:juvix-server`, and so on.
> 
> All Nix targets (packages) can be listed with `nix flake show --impure`.
> 
> Finally, the development shells provide GHC with dependant packages installed. These can be used to launch GHCi without stack e.g. `nix develop .#http:exe:juvix-server -c ghci`

> 
> > how do we integrate it to the editor?
> 
> What is meant by the editor here?

By editor I mostly mean `LSP` mode and any editor REPL that it hooks into. Namely my editor calls `haskell-process-load-file` which I believe dispatches to stack to handle the environment, so I can develop in editor.

PR #1103, opened 2021-12-20
Jan/strings

janmasrovira:
This is a small PR that improves string literals handling.

I separate the changes in 3 categories:
# Tangible changes
String literals are defined as global variables. Before, repeated string literals were defined once for each occurrence, now they are only defined once.

# Code quality changes
I have added some (haskell) auxiliary functions to make code related to strings more readable.

I have added a type called `CString` which encapsulates C-like strings, i.e., strings terminated with `'\00'`.
Conversion from/to Text and CString is seamless thanks to the ConvertString instances. 

# Collateral changes
I added `text` and `hashable` depencencies. Those were already transitive dependencies but I needed to import them to make direct use of them.

I added `ScopedTypeVariables` as a default extension. I have also sorted default extensions alphabetically to improve readability.


library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/CString.hs
```diff
@@ -0,0 +1,41 @@
...
+  show = Prelude.show . toText
+
+instance ConvertText CString Text where
+  toS = toText
+
+instance ConvertText Text CString where
+  toS = fromText
+
+fromText :: Text -> CString
+fromText = CString
```
andy-morris, line 31: might this be a good place to do encoding into utf8? (using a `ByteString` as the contents of a `CString` instead of a `Text`, obviously)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/CString.hs
```diff
@@ -0,0 +1,41 @@
...
+
+toText :: CString -> Text
+toText (CString t) = t <> "\00"
```
Acentelles, line 34: This doesn't seem to pass a roundtrip test

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/CString.hs
janmasrovira: I agree that `CString` is a good candidate to handle utf8 encoding. To me, it makes more sense to keep `Text` as the contents and add a function (`encodeUtf8`) that extracts and encodes the contents in utf8, returning a `ByteString`. Commit 6e66d43 implements that.

Also, the instances of `ConvertText Text CString` `ConvertText CString Text` are not inverses.
Do you think that's a problem?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/CString.hs
janmasrovira: Indeed.
I think that the best solution is to eliminate the instance `ConvertText CString Text` and don't export `toText`, as it is not used outside of the module and it makes more sense to extract the utf8 encoded `ByteString` rather than `Text`. Do you agree?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -163,22 +163,25 @@ module Juvix.Backends.LLVM.Codegen.Block
...
+import qualified LLVM.AST.CallingConvention                as CC
+import qualified LLVM.AST.Constant                         as C
+import qualified LLVM.AST.Global                           as Global
+import qualified LLVM.AST.IntegerPredicate                 as IntPred
+import qualified LLVM.AST.Linkage                          as L
+import qualified LLVM.AST.Name                             as Name
+import qualified LLVM.AST.Operand                          as Operand
+import qualified LLVM.AST.ParameterAttribute               as ParameterAttribute
+import qualified LLVM.AST.Type                             as Type
+import           Prelude                                   (String)
```
Acentelles, line 184: Are you using `stylish-haskell`(https://github.com/haskell/stylish-haskell)? Are these changes intended?

janmasrovira: I am, and yes. I thought that since I was already modifying the import section it would be fine to include the reformating in the PR, as it looks much more readable. Should I not do that?

Acentelles: I fear there might be some conflict with Ormolu, which the CI runs when tests pass. If you have a strong opinion on this, feel free to bring it up in the next meeting

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/CString.hs
Acentelles: Yes, sounds good

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
janmasrovira: I don't have a strong opinion, so I just reverted the changes.

Acentelles (approved):
LGTM

PR #1106, opened 2022-01-14
Introduce LLVM HUnit tests and record types

rokopt:
This PR comprises a first implementation of:

- The use of `llvm-hs` to execute LLVM code within our HUnit tests
- Support within LLVM for record types

For now, I'm boxing all record types -- that is, each record type is a structure of pointers, one pointer per field.  So there's one `malloc` for the record plus one for each of the fields.  The idea is that that will allow recursive structures whenever we're otherwise ready for them, and it also means that we don't currently have to go to the trouble of figuring out which structures are and aren't recursive so that we can calculate the sizes of and inline those that aren't.


rokopt:
(Replying to a comment which I've unintentionally hidden by a forced push)

> This works for now, but what if we have overlapping names in different modules? or maybe even the same

I had been wondering about that myself, and wasn't sure what to do, and thought at first that I'd have to defer figuring it out until we'd decided what Core was going to give us.

However, now that you ask, I think I might have an idea.  For this PR, I decided that the easiest way to allow someone to define a record type by passing in a `TermLLVM` directly -- as the new `HUnit` tests currently do -- was to add that `ScopedRecordDeclM` term which wraps another term with a record declaration.

It's not really scoped, though -- the LLVM type alias is globally declared, and the entry in `recordMap` remains after `compileRecordDecl` returns.

But now that you ask, it occurs to me that I think I could make it actually scoped, and that that would allow us to shadow record names and allow Core to pass us scoped declarations if it wanted to, to handle declarations within modules:  if I were to uniquify the LLVM type alias and store it in `recordMap`, and pop the old value of `recordMap` after returning from `compileRecordDecl`, then record lookups inside the wrapped term would get the right LLVM type, and record lookups _outside_ of it wouldn't be able to see that type, but would see an overlapping one with a larger scope, if there were one.

Does that make sense?

rokopt:
Note:  Jeremy pointed out that there's commonality between `loadField` and `getFieldPtr` that I could factor out.  I'll add a commit to do that, as well as to make properly scoped records as described in the previous comment if it makes sense to do so.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -49,13 +50,24 @@ compileProgram t =
...
+registerPreProcessed t = do
   Closure.register
   Block.defineMalloc
   Block.defineFree
-  mkMain (preProcess t)
+  mkMain t
+
+register ::
+  Types.Define m => ErasedAnn.AnnTerm PrimTy RawPrimVal -> m LLVM.Operand
+register = registerPreProcessed . preProcess
```
mariari, line 64: Good to break it down in this way

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -49,13 +50,24 @@ compileProgram t =
...
+
+termLLVMToModule ::
+  Types.Annotated Types.TermLLVM -> Either Types.Errors LLVM.Module
+termLLVMToModule t =
+  case Block.runEnvState (registerPreProcessed t) mempty of
+    (Right _, state) -> Right $ state |> Types.moduleAST
+    (Left err, _) -> Left err
```
mariari, line 71: I'm guessing this is the new runner for `llvm-hs-9.0.1`, seems you don't run the `preProcess`at all?

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -1,13 +1,37 @@
...
+runIntModule :: AST.Module -> IO Int
+runIntModule mod =
+  LLVMContext.withContext $ \context ->
+    LLVMModule.withModuleFromAST context mod $ \m ->
+      Engine.withMCJIT context Nothing Nothing Nothing Nothing $ \executionEngine ->
+        Engine.withModuleInEngine executionEngine m $ \ee -> do
+          main <- Engine.getFunction ee (AST.Name "main")
+          case main of
+            Just f -> runIntFun f
+            Nothing -> P.error "no main function found"
```
mariari, line 35: Nice this looks like the jit, grabs the main and runs that

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -17,4 +41,12 @@ tests =
   [trivialLLVMCodegenTest]
 
 trivialLLVMCodegenTest :: TestTree
-trivialLLVMCodegenTest = testCase "Trivial LLVM codegen test" $ True @?= True
+trivialLLVMCodegenTest = testCase "Trivial LLVM codegen test" $ do
+  let testVal = 42
+      ty = ErasedAnn.PrimTy $ Primitive.PrimTy $ AST.IntegerType {typeBits = 8}
```
mariari: Don't we have `int8` in one of our aliases or `integer` in our compiled code?

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -17,4 +41,12 @@ tests =
...
+      term = PassTypes.Prim (Primitive.LitInt testVal)
+      compiled =
+        fromRight (P.error "trivial test compilation failed") $
+          Compilation.termLLVMToModule PassTypes.Ann {usage = Usage.SAny, annTy = ty, term = term}
```
mariari, line 60: I guess our code, just make the term main, so it's OK

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -0,0 +1 @@
+module Juvix.Backends.LLVM.Codegen.Record where
```
mariari: Here is where the fun begins

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -1,11 +1,46 @@
...
+  = SymT NameSymbol.T
+  | Star BaseTypes.Universe
+  | PrimTy Prim.PrimTy
+  | Pi Usage.T TypeLLVM TypeLLVM
+  | Sig Usage.T TypeLLVM TypeLLVM
+  | CatProduct TypeLLVM TypeLLVM
+  | CatCoproduct TypeLLVM TypeLLVM
+  | UnitTy
+  | RecordType RecordName
+  deriving (Show, Read, Eq, Generic)
```
mariari, line 31: We should comment that `RecordType` is the deviation from core and the rest is the same, as seen by the next comment.

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -1,11 +1,46 @@
...
+injectErasedTypeIntoLLVM t = case t of
+  ErasedAnn.SymT name -> SymT name
+  ErasedAnn.Star σ -> Star σ
+  ErasedAnn.PrimTy ty -> PrimTy ty
+  ErasedAnn.Pi π a b -> Pi π (injectErasedTypeIntoLLVM a) (injectErasedTypeIntoLLVM b)
+  ErasedAnn.Sig π a b -> Sig π (injectErasedTypeIntoLLVM a) (injectErasedTypeIntoLLVM b)
+  ErasedAnn.CatProduct a b -> CatProduct (injectErasedTypeIntoLLVM a) (injectErasedTypeIntoLLVM b)
+  ErasedAnn.CatCoproduct a b -> CatCoproduct (injectErasedTypeIntoLLVM a) (injectErasedTypeIntoLLVM b)
+  ErasedAnn.UnitTy -> UnitTy
+
```
mariari, line 43: This shows which fields are injected into, the type shoould be a bit more explicit about it

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -1,6 +1,6 @@
 module Juvix.Backends.LLVM.Pass.Types where
 
-import Juvix.Backends.LLVM.Primitive as Prim
+import qualified Juvix.Backends.LLVM.Primitive as Prim
```
mariari, line 3: Nice shocked this compiled, as some of them were annoying with some type signatures

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -164,4 +164,17 @@ data TermLLVM
   | AppM
       (Annotated TermLLVM)
       [Annotated TermLLVM]
+  | ScopedRecordDeclM RecordDecl (Annotated TermLLVM)
+  | FieldM TermFieldSelector
+  | RecordM TermRecordConstructor
```
mariari, line 169: We should comment if these are new.

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
```diff
@@ -164,4 +164,17 @@ data TermLLVM
...
+
+type FieldDecl = (FieldName, TypeLLVM)
+
+type RecordSpec = [FieldDecl]
+
+type RecordDecl = (RecordName, RecordSpec)
+
+type TermFieldSelector = (RecordName, FieldName, Annotated TermLLVM)
+
+type TermRecordConstructor = (RecordName, [Annotated TermLLVM])
```
mariari, line 180: Should we move the new closure code parts to where this is? It seems to me all the extensions should live close together just separated by what they extend. As IIRC the closure extensions were at the top

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
```diff
@@ -36,6 +37,10 @@ type VariantToType = Map.T Symbol SumInfo
 -- numbering to go along with a given name.
 type Names = Map.T Symbol Int
 
+-- | A mapping of record names to LLVM types and lists of (name, type) pairs
+-- | (one for each field).
+type RecordTable = Map.T Symbol (Type, [(Symbol, Type)])
```
mariari, line 42: This has a nice proper type mapping,

Though, for overlap we may eventually want
`[(Type, [(Symbol, Type)])]`

if there is an overlap for any reason, then we can pick the right one if there is a type annotation, and make them match. 

Though maybe we should find a way to disambiguate by core?

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -33,20 +34,62 @@ runIntModule mod =
...
             Nothing -> P.error "no main function found"
 
+llvmIntType :: Word32 -> PassTypes.TypeLLVM
+llvmIntType size = PassTypes.PrimTy $ Primitive.PrimTy $ AST.IntegerType {typeBits = size}
+
+llvmIntVal :: Integer -> PassTypes.TermLLVM
+llvmIntVal = PassTypes.Prim . Primitive.LitInt
+
+llvmAnnotatedTerm :: PassTypes.TypeLLVM -> PassTypes.TermLLVM -> PassTypes.Annotated PassTypes.TermLLVM
+llvmAnnotatedTerm ty term = PassTypes.Ann {PassTypes.usage = Usage.SAny, PassTypes.annTy = ty, PassTypes.term = term}
```
mariari, line 44: Check my `codegen` module, I'm fairly certain I have a function for their `AST.IntegerType` though I may be wrong

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+recordTypeName recordName = Block.internName $ "record-" <> toSymbol recordName
+
+llvmRecordType :: [Type.Type] -> Type.Type
+llvmRecordType llvmFieldTypes =
+  Type.StructureType
+    { isPacked = True,
+      elementTypes =
+        -- We box types, both to allow recursive structures and also to avoid
+        -- having to perform any recursive size calculations for now.
+        map Types.pointerOf llvmFieldTypes
```
mariari, line 45: Smart choice, this will help you a lot in calculations for the sum types. Note there is a function to check for size, but this works quite well.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+  [PassTypes.FieldName] ->
+  [Type.Type] ->
+  m ()
+register recordName fieldNames llvmFieldTypes = do
+  let llvmTypeName = recordTypeName recordName
+  Block.addType llvmTypeName (llvmRecordType llvmFieldTypes)
+  recordTable <- get @"recordTab"
+  let typeRef = Type.NamedTypeReference llvmTypeName
+  let fieldDescs = zip (map toSymbol fieldNames) llvmFieldTypes
+  put @"recordTab" $ Map.insert (toSymbol recordName) (typeRef, fieldDescs) recordTable
```
mariari, line 33: Why not
```haskell
  let typeRef = Type.NamedTypeReference llvmTypeName
  let fieldDescs = zip (map toSymbol fieldNames) llvmFieldTypes
  modify @"recordTab" (Map.insert (toSymbol recordName) (typeRef, fieldDescs))

File: library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs [hunk 16]
```diff
@@ -1 +1,132 @@
...
+        map Types.pointerOf llvmFieldTypes
+    }
+
+-- | Get the names and types of the fields of the record type with the given name.
+recordFields :: Types.Define m => PassTypes.RecordName -> m [(Symbol, Type.Type)]
+recordFields recordName = do
+  recordTable <- get @"recordTab"
+  pure $
+    snd $
+      fromMaybe (P.error "typechecker allowed record of non-existent type") $
```
mariari, line 54: nest to avoid `P.error` we have an error type builtin to the monad. so use that instead please.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+fieldIndex ::
+  Types.Define m =>
+  PassTypes.RecordName ->
+  PassTypes.FieldName ->
+  m Int
+fieldIndex recordName fieldName = do
+  fieldDescs <- recordFields recordName
+  pure $
+    fromMaybe (P.error "typechecker allowed selection of non-existent field") $
+      List.elemIndex (toSymbol fieldName) $ map fst fieldDescs
```
mariari, line 68: Same here, as long as we are in the monad, we can exhibit errors for no cost properly.

mariari, line 68: Also since these are fixed sizes, should we not use an array instead of a list?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+  Type.StructureType
+    { isPacked = True,
+      elementTypes =
+        -- We box types, both to allow recursive structures and also to avoid
+        -- having to perform any recursive size calculations for now.
+        map Types.pointerOf llvmFieldTypes
+    }
+
+-- | Get the names and types of the fields of the record type with the given name.
+recordFields :: Types.Define m => PassTypes.RecordName -> m [(Symbol, Type.Type)]
```
mariari, line 49: Just realized, shouldn't semantically be an array rather than list by how we use it?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+-- | Given the name of a record type and a list of compiled field terms,
+-- | allocate a record and store the given terms in it.
+makeRecord ::
+  Types.Define m =>
+  PassTypes.RecordName ->
+  [AST.Operand] ->
+  m AST.Operand
+makeRecord recordName fieldTerms = do
+  recordTable <- get @"recordTab"
+  let (recordType, fieldDescs) = fromMaybe (P.error "typechecker allowed record of non-existent type") (Map.lookup (toSymbol recordName) recordTable)
```
mariari, line 121: This line seems a bit long.

Also pleasechange the prelude error with the monad one.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+  recordPtr <- Block.mallocType recordType
```
mariari, line 122: we should probably note we leak memory here for now

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
```diff
@@ -1 +1,132 @@
...
+  let fieldNames = map (fromSymbol . fst) fieldDescs
+  let fieldTypes = map snd fieldDescs
+  fieldPtrs <- mapM (flip (getFieldPtr recordName) recordPtr) fieldNames
+  newFields <- mapM Block.mallocType fieldTypes
```
mariari, line 126: This would give us the pointer to the type, almost thought we missed a `poitnerOf` somewhere

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -309,6 +313,46 @@ mkPrim prim ty = case prim of
...
+  Types.Define m =>
+  Monad m =>
+  Types.RecordDecl ->
+  -- | Primitive value.
+  Types.Annotated Types.TermLLVM ->
+  -- | Type of the primitive.
+  m LLVM.Operand
+compileRecordDecl (recordName, fieldDecls) term = do
+  Record.register recordName (map fst fieldDecls) (map (typeToLLVM . snd) fieldDecls)
+  compileTerm term
```
mariari, line 326: beaitiful, deals with the complexity for us.

Should we make a Record type a "top level expression" so it doens't have to take any arguments.... I guess we could have both. This works

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -309,6 +313,46 @@ mkPrim prim ty = case prim of
...
+
+compileRecord ::
+  Types.Define m =>
+  Monad m =>
+  Types.TypeLLVM ->
+  -- | Primitive value.
+  Types.TermRecordConstructor ->
+  -- | Type of the primitive.
+  m LLVM.Operand
+compileRecord (Types.RecordType recordTypeName) (recordName, fieldTerms) | recordTypeName == recordName = do
```
mariari, line 336: Might be best to newline the `|` here

mariari (approved):
Overall looks fine, see my comment for exact changes.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: Right -- if you want to run with the jit, you call that, and then you can execute the module in some way along the lines of what `runIntModule` does (in `library/Backends/llvm/test/Test/Codegen.hs`).

library/Backends/llvm/test/Test/Codegen.hs
rokopt: Oh yes, we have `i8` and `i16`.  I also added a test interface that takes `typeBits` instead -- and, now that I read your comment, I realize that I forgot to use it in a bunch of places!  I added https://github.com/anoma/juvix/pull/1106/commits/2203b0592f7b76d3d4a5238bc79f7d3f997dd1b5 to shorten the test code a lot by using the convenience interfaces.

I also noticed while going over the test code for commonality to factor out that I'd left in the printing of the generated LLVM module, which was useful for debugging but not during automated test execution.  I removed that in https://github.com/anoma/juvix/pull/1106/commits/9e5064fa069109a15ff8a1d56ad2b9d3cf6ac489.

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
rokopt: Good point -- done in https://github.com/anoma/juvix/pull/1106/commits/442dfcd63ac8e61e0e60e28d013edcf25e70fc8f

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
rokopt: I added a comment in https://github.com/anoma/juvix/pull/1106/commits/442dfcd63ac8e61e0e60e28d013edcf25e70fc8f -- is that what you were looking for, or was there a code change you're suggesting here as well?

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
rokopt: Done in https://github.com/anoma/juvix/pull/1106/commits/442dfcd63ac8e61e0e60e28d013edcf25e70fc8f

library/Backends/llvm/src/Juvix/Backends/LLVM/Pass/Types.hs
rokopt: I moved them in https://github.com/anoma/juvix/pull/1106/commits/02cdfb59de1f10f3c369581458d0cbaa4c5bfe39 -- does that seem like a more sensible place?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types/Shared.hs
rokopt: I think what I'll at least try first here is to make the record types truly scoped as I described in https://github.com/anoma/juvix/pull/1106#issuecomment-1013715909 , thus giving Core the chance to use that mechanism to handle declarations in different modules and/or declarations whose names shadow those of other record declarations.  Then the way of having multiple record declarations with the same symbol would be that the `recordTab` would be saved and restored around the term wrapped by a `ScopedRecordDecl`.

library/Backends/llvm/test/Test/Codegen.hs
rokopt: There are interfaces for `IntegerType`s with specific sizes -- `i4`, `i8`, and so on -- but I wanted an interface which produced a `TypeLLVM` and which was parameterized by `typeBits` since I'm constructing `TermLLVM`s manually in test code.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Good call, done in https://github.com/anoma/juvix/pull/1106/commits/fe9d23d65740d4570a9d00a47aae9545cb4d3475

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Thanks, I've switched to the monad throw in https://github.com/anoma/juvix/pull/1106/commits/fc0975b6c7634a2e6496686201045cfe56d7e113 and https://github.com/anoma/juvix/pull/1106/commits/9385ae17c907812daaa88164ad3a2e17fab2f42b.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Also done in https://github.com/anoma/juvix/pull/1106/commits/fc0975b6c7634a2e6496686201045cfe56d7e113 and https://github.com/anoma/juvix/pull/1106/commits/9385ae17c907812daaa88164ad3a2e17fab2f42b.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Switching to the error monad, as I've done, also shortened the line. :)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: True, we could use an array.  `Data.Array` doesn't have as many utility functions as `List`, though; for example, it's not obvious to me that there's a `zip`.  Should we write some of our own for the Juvix library?  Or is it okay to stick with `List`?

Your comment did make me realize that I could add more cross-checking here:  I extended the `makeRecord` interface to allow the `Compilation` and `Record` modules to cooperate to make sure that the fields to a record constructor all had the expected types.  (In particular this makes sure that the lengths match.)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: See my previous comment on `Array`/`List`: https://github.com/anoma/juvix/pull/1106#discussion_r787172995``

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: Do you mean one with a dependent type?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Are you suggesting that I add some kind of comment about one of these lines?  Or do I have some actual semantics wrong?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: Done in https://github.com/anoma/juvix/pull/1106/commits/b7e4f0593262df3ef8a2079355c5ca06bf9bd739

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
rokopt: Done in https://github.com/anoma/juvix/pull/1106/commits/6fcae69dd44e08cbf3b1b650b307b927b9f3a93e

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Record.hs
rokopt: (see https://github.com/anoma/juvix/pull/1106/commits/420b80c9f61e5165d3cbf353f634c1c7fb4fba04 for the added typechecking)

rokopt:
> But now that you ask, it occurs to me that I think I could make it actually scoped, and that that would allow us to shadow record names and allow Core to pass us scoped declarations if it wanted to, to handle declarations within modules: if I were to uniquify the LLVM type alias and store it in `recordMap`, and pop the old value of `recordMap` after returning from `compileRecordDecl`, then record lookups inside the wrapped term would get the right LLVM type, and record lookups _outside_ of it wouldn't be able to see that type, but would see an overlapping one with a larger scope, if there were one.

I've done this in https://github.com/anoma/juvix/pull/1106/commits/ec84c34ce954d8c88a6bf9cac8334a4a6f4b6d73 and https://github.com/anoma/juvix/pull/1106/commits/cb404b4603d7b83561c127c2ac1e39df0ac8ae0d.

PR #1107, opened 2022-01-24
Introduce LLVM sum types

rokopt:
This PR introduces sum types to LLVM, as #1106 does for record types.  It's only a draft right now, for two reasons:

- It's based off #1106 , which isn't merged yet, as it in turn is blocked waiting for the merge of https://github.com/anoma/juvix/tree/samuli/stack-nix .  (The commits specific to this PR begin after https://github.com/anoma/juvix/pull/1107/commits/dd0f5ad006386595d57e77d8395be501a285647d .)
- There is an open unsolved problem:  the code generation for a sum destructor (i.e. a case statement) takes a set of functions, one for each variant, mapping the variant type to the output type.  But I haven't been able to figure out yet how to get the environment at the point of the case statement and pass it to the code that generates the LLVM code for the case statement.  So it only works if the per-variant functions don't use their environment yet.  That's why `makeCase` has this `Block.null Closure.environmentPtr` instead of a potentially-populated environment:  https://github.com/anoma/juvix/blob/6cdac05639fe0df9c4e467c8a96d425cf816e11b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs#L158-L163

### DIFF
--- a/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
+++ b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
@@ -46,9 +46,8 @@
 -- @
 --   compileLam ty captures arguments body
 --    | length captures == 0 = do
---       let (llvmArgty, llvmRetty) =
---             functionTypeLLVM ty
---           llvmArgNames =
+--       (llvmArgty, llvmRetty) <- functionTypeLLVM ty
+--       let llvmArgNames =
 --             fmap (Block.internName . NameSymbol.toSymbol) arguments
 --           llvmArguments =
 --             zip llvmArgty llvmArgNames
@@ -1159,7 +1158,7 @@ callVoid fn args =
 -- --
 -- let -- We should probably get the type from the function itself
 --     -- rather than pass in what it should be here
---     functionType = typeToLLVM returnTy
+--     functionType <- typeToLLVM returnTy
 --     -- ignore attributes for now!
 --     argsAtrributes = zip arguments (repeat [])
 --
--- a/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
+++ b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
@@ -110,6 +110,10 @@ module Juvix.Backends.LLVM.Codegen.Block
     mallocType,
     free,
 
+    -- * Assertions / abnormal termination
+    defineAbort,
+    abort,
+
     -- * Operations
 
     -- ** Integer Operations
@@ -745,7 +749,7 @@ printCString str args = do
 -- Memory management
 --------------------------------------------------------------------------------
 
--- malloc & free need to be defined once and then can be called
+-- malloc, free, and abort need to be defined once and then can be called
 -- normally with `externf`
 
 -- | @defineMalloc@ defines the malloc function
@@ -760,6 +764,13 @@ defineFree :: External m => m ()
 defineFree = do
   let name = "free"
   op <- external voidTy name [(Types.pointerOf Type.i8, "type")]
+  assign (intern name) op
+
+-- | @defineAbort@ defines the abort function
+defineAbort :: External m => m ()
+defineAbort = do
+  let name = "abort"
+  op <- external voidTy name []
   assign (intern name) op
 
 -- | @malloc@ takes the size and the resulting type of the @malloc@. the
@@ -797,6 +808,16 @@ free thing = do
   free <- externf "free"
   casted <- bitCast thing (Types.pointerOf Type.i8)
   unnminstr (callConvention CC.Fast free (emptyArgs [casted]))
+
+-- | @abort@ terminates the program abnormally.
+abort :: (RetInstruction m, Externf m) => m ()
+abort = do
+  abort <- externf "abort"
+  unnminstr $
+    callConvention
+      CC.Fast
+      abort
+      (emptyArgs [])
 
 -- * Operations are functions that are callable on the LLVM backend
 
--- a/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
+++ b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
@@ -176,6 +176,7 @@ import qualified Juvix.Backends.LLVM.Codegen.Types.CString as CString
 import Juvix.Library hiding (Type, local)
 import qualified Juvix.Library.HashMap as Map
 import LLVM.AST
+import Data.List (unzip3)
 import qualified LLVM.AST as AST
 import qualified LLVM.AST.CallingConvention as CC
 import qualified LLVM.AST.Constant as C
@@ -448,6 +449,24 @@ entry = get @"currentBlock"
 -- @
 getBlock :: (HasState "currentBlock" Name m) => m Name
 getBlock = entry
+
+-- | Switches to a block, does something, and returns to the original block.
+withBlock :: (HasState "currentBlock" Name m) => Name -> m a -> m a
+withBlock tmpBlock action = do
+  cur <- getBlock
+  setBlock tmpBlock
+  a <- action
+  setBlock cur
+  return a
+
+-- | Creates a block and switches to a it. Does something, and returns to the
+-- original block.
+withNewBlock :: (HasState "currentBlock" Name m,
+                 NewBlock m)
+             => Symbol -> m a -> m a
+withNewBlock blockSym action = do
+  blockName <- addBlock blockSym
+  withBlock blockName action
 
 -- | @addBlock@ generates a new block. This is used every time a
 -- jump/decision/goto point is needed. For example:
@@ -1054,43 +1073,52 @@ generateIf ty cond tr fl = do
 -- @generateIf@ is a layer over @cbr@.
 --
 -- @
--- Block.generateSwitch ty term cases
+-- Block.generateSwitch ty term [(caseName, caseValue, caseBody)]
 -- @
-generateSwitch ::
+generateSwitch :: forall m.
   ( RetInstruction m,
     HasState "blockCount" Int m,
     HasState "names" Names m,
     HasState "symTab" SymbolTable m
   ) =>
+  -- | Type of caseValue.
   Type ->
+  -- | Operand that we are comparing.
   Operand ->
-  [Symbol] ->
-  [C.Constant] ->
-  [m Operand] ->
+  -- | List of (caseName, caseValue, caseBody).
+  -- * caseName: informative tag.
+  -- * caseValue: like a literal pattern.
+  -- * caseBody: the code to run.
+  [(Symbol, C.Constant, m Operand)] ->
+  -- | Default case. If Nothing, an error is thrown if the default block is
+  -- reached.
+  Maybe (m ()) ->
   m Operand
-generateSwitch ty tag names values cases = do
-  internNames <- mapM addBlock $ map ("switch." <>) names
-  defaultBlock <- addBlock "switch-default"
+generateSwitch ty tag caseInfos mdefault = do
   exitBlock <- addBlock "switch-exit"
-  _ <- switch tag defaultBlock $ zip values internNames
-  phiPairs <-
-    mapM
-      ( \(n, c) -> do
-          setBlock n
-          op <- c
-          _ <- br exitBlock
-          n' <- getBlock
-          pure (op, n')
-      )
-      (zip internNames cases)
-  setBlock defaultBlock
-  abort
-  _unreachable <- br exitBlock
-  let errVal = AST.ConstantOperand $ C.Int {C.integerBits = 8, C.integerValue = 0}
-  castedErr <- bitCast errVal ty
-  defaultBlock <- getBlock
+  defaultPhi@(_, defaultBlock) <- makeDefaultBlock exitBlock
+  phiPairs <- zipWithM (makePhiPair exitBlock) names cases
+  void $ switch tag defaultBlock $ zip values (map snd phiPairs)
   setBlock exitBlock
-  phi ty $ (castedErr, defaultBlock) : phiPairs
+  phi ty $ defaultPhi : phiPairs
+  where
+  errVal = AST.ConstantOperand $ C.Int {C.integerBits = 8, C.integerValue = 0}
+  makeDefaultBlock :: Name -> m (Operand, Name)
+  makeDefaultBlock exitBlock = withNewBlock "switch-default" $ do
+    case mdefault of
+      Nothing -> abort
+      Just op -> op
+    void (br exitBlock)
+    castedErr <- bitCast errVal ty
+    defaultBlock <- getBlock
+    return (castedErr, defaultBlock)
+  makePhiPair :: Name -> Symbol -> m Operand -> m (Operand, Name)
+  makePhiPair exitBlock symbol case_ = withNewBlock ("switch." <> symbol) $ do
+    op <- case_
+    void (br exitBlock)
+    blockName <- getBlock
+    pure (op, blockName)
+  (names, values, cases) = unzip3 caseInfos
 
 --------------------------------------------------------------------------------
 -- Effects
--- a/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
+++ b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
@@ -235,6 +235,7 @@ emptyCodegen =
       Types.typTab = Map.empty,
       Types.varTab = Map.empty,
       Types.recordTab = Map.empty,
+      Types.sumTab = Map.empty,
       Types.count = 0,
       Types.names = Map.empty,
       Types.strings = mempty,
--- a/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
+++ b/library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
@@ -140,6 +140,7 @@ module Juvix.Backends.LLVM.Codegen.Block
     phi,
     switch,
     generateIf,
+    generateSwitch,
 
     -- ** Calling Operations
     emptyArgs,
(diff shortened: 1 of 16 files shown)


rokopt:
> There is an open unsolved problem: the code generation for a sum destructor (i.e. a case statement) takes a set of functions, one for each variant, mapping the variant type to the output type. But I haven't been able to figure out yet how to get the environment at the point of the case statement and pass it to the code that generates the LLVM code for the case statement. So it only works if the per-variant functions don't use their environment yet.

I had an idea about this:

https://github.com/anoma/juvix/pull/1107/commits/541509943a632f493c822c0e530e1bdf9c6e2255
https://github.com/anoma/juvix/pull/1107/commits/b4333ffe0b8ebb717493cc606d6892abb19b4173

I'm not certain that it makes sense in all cases, but I'll at least mark the PR ready for review now.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+  indexPtr <- getIndexPtr term
+  index <- Block.load indexType indexPtr
+  variantPtrLoc <- getVariantPtr term
+  variantPtr <- Block.load variantPtrType variantPtrLoc
+  let appliedCases =
+        map
+          ( \(ty, (caseFunc, environment)) -> do
+              castedPtr <- Block.bitCast variantPtr (Types.pointerOf ty)
+              variant <- Block.load ty castedPtr
+              Block.call outputType caseFunc [(environment, []), (variant, [])]
```
mariari, line 164: is the call arguments correct?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+-- | WARNING:  The memory allocations herein leak memory, pending our
+-- | implementing garbage collection.
+makeSum ::
+  Types.Define m =>
+  PassTypes.SumName ->
+  PassTypes.VariantName ->
+  Type.Type ->
+  Type.Type ->
+  AST.Operand ->
+  m AST.Operand
```
mariari, line 181: We really need to make data types over these functions, it feels like we pass a lot of arguments to a lot of functions

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+makeSum sumName variantName compiledSumType compiledVariantType variantTerm = do
+  (sumType, variantDescs) <- getSumDesc sumName
+  if compiledSumType /= sumType
```
mariari, line 184:
```suggestion
  unless (compiledSumType == sumType) $ do
```

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+          "typechecker allowed mismatched sum types: expected "
+            <> show sumType
+            <> "; got "
+            <> show compiledSumType
+    else pure ()
+  variantIndex <- case List.findIndex (\desc -> fst desc == toSymbol variantName) variantDescs of
+    Just index -> pure index
+    Nothing ->
+      throw @"err" $
+        Types.NonExistentVariant "typechecker allowed selection of non-existent variant"
```
mariari, line 197: Seems we might want a hashtable to hold these values instead of lists? Idk though with other semantics, but the search function suggests that to me

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+  let variantType = snd $ variantDescs P.!! variantIndex
+  if compiledVariantType /= variantType
```
mariari, line 199:
```suggestion
  unless (compiledVariantType == variantType) $ do
```

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
```diff
@@ -101,6 +103,31 @@ data Errors
...
+  | -- | Found a match term whose cases' types did not match the variant
+    -- | types in the sum type declaration (this indicates a bug in the
+    -- | typechecker, if the type of the term did not come directly from a
+    -- | test, or some other module that bypasses the typechecker)
+    MismatchedCaseTypes Text
+  | -- | Found a sum term whose type did not match the one found by
+    -- | looking up the type's name (this indicates a bug in the
+    -- | typechecker, if the type of the term did not come directly from a
+    -- | test, or some other module that bypasses the typechecker)
+    MismatchedSumTypes Text
```
mariari, line 130: Nice use of comments, I wonder if | is needed after every --? It seems like it was like that before, but I wonder what haddock wants.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -217,19 +232,20 @@ compileApp returnTy f@Types.Ann {term} xs =
...
       case Typed.typeOf function of
         -- Closure calling case
         LLVM.PointerType (LLVM.NamedTypeReference {}) _ -> do
+          llvmTypes <- mapM (typeToLLVM . Types.annTy) xs
           let argumentTypes =
                 Closure.environmentPtr :
-                llvmFunctionsToClosure (fmap (typeToLLVM . Types.annTy) xs)
+                llvmFunctionsToClosure llvmTypes
               functionType =
                 LLVM.FunctionType newReturnType argumentTypes False
```
mariari, line 250: If we abstract out `getCompiledEnvironment` can we call it here at all?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+  [Type.Type] ->
+  AST.Operand ->
+  [AST.Operand] ->
+  [AST.Operand] ->
+  m AST.Operand
+makeCase sumName outputType caseTypes term cases environments = do
+  (sumType, variantDescs) <- getSumDesc sumName
+  let values = map tagConstant [0 .. length variantDescs - 1]
+  let variantNames = map fst variantDescs
+  let variantTypes = map snd variantDescs
```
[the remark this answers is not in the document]
mariari: I'm wrong. Ended up writing this for a few functions. But I'm incorrect about all these `second` calls

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
```diff
@@ -353,15 +370,57 @@ compileRecord ty (recordName, _) =
...
+
+compileSumDecl ::
+  Types.Define m =>
+  Types.SumDecl ->
+  -- | Primitive value.
+  Types.Annotated Types.TermLLVM ->
+  -- | Type of the primitive.
+  m LLVM.Operand
+compileSumDecl (sumName, variantDecls) term = do
+  llvmTypes <- mapM (typeToLLVM . snd) variantDecls
```
mariari, line 391: Please prefer `traverse` to `mapM` due to 

```haskell
λ> traverse
traverse
  :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
λ> mapM
mapM :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)
```

Applicative is a smaller interface and thus a bit more general

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -35,15 +35,79 @@ runIntModule mod =
...
             Nothing -> P.error "no main function found"
 
+llvmPrimIntType :: Word32 -> AST.Type
+llvmPrimIntType size = AST.IntegerType {typeBits = size}
+
 llvmIntType :: Word32 -> PassTypes.TypeLLVM
 llvmIntType size = PassTypes.PrimTy $ Primitive.PrimTy $ AST.IntegerType {typeBits = size}
 
+llvmStringType :: PassTypes.TypeLLVM
+llvmStringType = PassTypes.PrimTy $ Primitive.PrimTy $ CodegenTypes.pointerOf $ AST.IntegerType {typeBits = 8}
```
mariari, line 45: Line seems just a bit long?

library/Backends/llvm/test/Test/Codegen.hs
```diff
@@ -157,8 +222,44 @@ llvmRecordNameShadowingTest = testCase "LLVM codegen test with record name shado
...
+      annotatedIntCaseTerm = llvmAnnotatedTerm intTy intCaseTerm
+      stringCaseTerm = PassTypes.MatchM (sumTypeName, annotatedStringSumTerm, [intCase, stringCase])
+      annotatedStringCaseTerm = llvmAnnotatedTerm intTy stringCaseTerm
+      mulOp = annotatedBinOp ASTTypes.i8 Primitive.Mul
+      mulTerm = annotatedBinOpApp intTy mulOp annotatedIntCaseTerm annotatedStringCaseTerm
+      term = PassTypes.ScopedSumDeclM testSumTy mulTerm
+      annotatedTerm = llvmAnnotatedTerm intTy term
+      compiled = case Compilation.termLLVMToModule annotatedTerm of
+        Left err -> P.error $ "LLVM sum types test compilation failed: " <> show err
+        Right t -> t
```
mariari, line 262: I feel like we could maybe abstract this out a bit. Hard to read such tests just reading down it. Not sure how completely to break it up to be more recongizable at a glance

mariari (approved):
Overall looks fine. We discussed alternative strategies we should go for. Namely splitting out match into case and making `compilation.hs` and the code around it even simpler. Moving complexity to the pass.

Note that I was being lazy when I reviewed `compilation.hs` so I probably missed some details.

I think adding more records for the values into the sum type functions would be nice, as the argument list is a bit long and we can find some common bundling

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -1028,6 +1050,48 @@ generateIf ty cond tr fl = do
...
+    HasState "blockCount" Int m,
+    HasState "names" Names m,
+    HasState "symTab" SymbolTable m
+  ) =>
+  Type ->
+  Operand ->
+  [Symbol] ->
+  [C.Constant] ->
+  [m Operand] ->
+  m Operand
```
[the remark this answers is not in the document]
janmasrovira: I've tried to improve the signature in a2868b2. I've also added some documentation

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -1028,6 +1050,48 @@ generateIf ty cond tr fl = do
...
+generateSwitch ty tag names values cases = do
```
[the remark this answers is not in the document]
janmasrovira: I agree. I've merged the arguments in a single list of 3-tuples (a2868b2)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -1028,6 +1050,48 @@ generateIf ty cond tr fl = do
...
+          n' <- getBlock
+          pure (op, n')
+      )
+      (zip internNames cases)
+  setBlock defaultBlock
+  abort
+  _unreachable <- br exitBlock
+  let errVal = AST.ConstantOperand $ C.Int {C.integerBits = 8, C.integerValue = 0}
+  castedErr <- bitCast errVal ty
+  defaultBlock <- getBlock
```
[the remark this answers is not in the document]
janmasrovira: what do you mean more precisely?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Block.hs
```diff
@@ -1028,6 +1050,48 @@ generateIf ty cond tr fl = do
...
+  setBlock exitBlock
```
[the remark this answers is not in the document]
janmasrovira: I've added two functions that will help with this in a2868b2. Namely `withBlock` and `withNewBlock`. I think it would be a good idea to try to use them, when possible, instead of manually switching from/to blocks.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+
+-- | @register@ registers a sum type with the given name and
+-- | variant names and types.  It returns the old sum table so
+-- | that the caller can restore it (with restoreTable) after leaving
+-- | the scope of a sum declaration.
+register ::
+  Types.Define m =>
+  PassTypes.SumName ->
+  [PassTypes.VariantName] ->
+  [Type.Type] ->
```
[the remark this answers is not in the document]
janmasrovira: made those a tuple as well (a2868b2)

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+  sumTable <- get @"sumTab"
+  case Map.lookup (toSymbol sumName) sumTable of
+    Just sumDesc -> pure sumDesc
+    Nothing ->
+      throw @"err" $
+        Types.NonExistentSumType "typechecker allowed sum of non-existent type"
+
+lookupType :: Types.LookupType m => PassTypes.SumName -> m Type.Type
+lookupType sumName = do
+  getSumDesc sumName >>= pure . fst
```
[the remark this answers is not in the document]
janmasrovira: ` fst <$> getSumDesc sumName` seems to more haskell way to handle it

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+      }
+
+makeCase ::
+  Types.Define m =>
+  PassTypes.SumName ->
+  Type.Type ->
+  [Type.Type] ->
+  AST.Operand ->
+  [AST.Operand] ->
+  [AST.Operand] ->
```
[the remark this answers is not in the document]
janmasrovira: done in 1e0672c

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
```diff
@@ -0,0 +1,224 @@
...
+  m AST.Operand
+makeCase sumName outputType caseTypes term cases environments = do
+  (sumType, variantDescs) <- getSumDesc sumName
+  let values = map tagConstant [0 .. length variantDescs - 1]
+  let variantNames = map fst variantDescs
+  let variantTypes = map snd variantDescs
+  let expectedCaseTypes = map (`oneArgFunctionType` outputType) variantTypes
+  if caseTypes /= expectedCaseTypes
```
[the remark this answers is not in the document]
janmasrovira: done in 1e0672c

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
janmasrovira: done in 1e0672c

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
janmasrovira: done in 1e0672c

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
janmasrovira: only the first `|` is needed by haddock. I think @rokopt has used `|` in every line so it looks more uniform.

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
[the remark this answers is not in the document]
mariari: `>>|` is just `<$>` depends if you want pipping ordering or reverse ordering

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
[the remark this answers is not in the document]
mariari: Looks good, do you think it might be good to have an explicit type for it, or is the tuple representation the best in your opinon?

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Sum.hs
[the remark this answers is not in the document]
mariari: Thanks

library/Backends/llvm/src/Juvix/Backends/LLVM/Codegen/Types.hs
mariari: Uniformity is good, though I think since we o gen out haock, oes this affect it at all?

janmasrovira: no, this does not affect the output. I'm fine with or without the extra pipes. However, not having them is slightly more convinient for me since when I do `M-q` in emacs to format the haddock paragraph, it does not add the extra pipes.

library/Backends/llvm/src/Juvix/Backends/LLVM/Compilation.hs
janmasrovira: I personally prefer `mapM` here. I find it easier to read. I think it's fine to not always use the more general interface. I usually use `return` instead of `pure`, I prefer `map` over `fmap` for lists, I prefer `Nothing` over `mzero`, etc.

PR #1109, opened 2022-02-01
add some primitives for integers: eq, neq, le, lt

janmasrovira:
this PR adds some primitives to the llvm backend.
Namely:
- PrimLeq.
- PrimLt.
- PrimEq.
- PrimNeq.

Currently, it is assumed that the arguments are integer types.
More work should be done to adapt these to floats.

### DIFF
--- a/library/Backends/llvm/package.yaml
+++ b/library/Backends/llvm/package.yaml
@@ -34,6 +34,9 @@ dependencies:
   - bytestring
   - data-structures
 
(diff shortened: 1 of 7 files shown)


Acentelles:
Thank you! Very much needed for the validity predicates we want to compile in the next release. Would it be possible to add some tests? (a simple program that uses these primitives that compiles e2e would suffice)

PR #1110, opened 2022-02-03
Validity predicates

Acentelles:
- Includes validity predicate examples
- Creates a standard library


test/examples/vps/SimpleFungibleToken.ju
```diff
@@ -13,28 +16,28 @@ let key = string
...
-let vp token keysChanged verifiers = 
-    let tup = foldl (checkKeys verifiers) (0, false) keysChanged in
+sig vp : address -> list key -> list address -> bool
+let vp token keys-changed verifiers = 
+    let tup = foldl (check-keys verifiers) (0, false) keys-changed in
     let change = tup.first in
-    let allChecked = tup.second in
-    and (change Int.(==) 0) allChecked
+    let all-checked = tup.second in
+    and (change Int.(==) 0) all-checked
```
mariari, line 42: - `and (change Int.(==) 0) all-checked`
 + `and (change Int.== 0) all-checked`

stdlib/LLVM/Data/Ord.ju
```diff
@@ -0,0 +1,3 @@
+mod LLVM.Data.Ord where
+
+type Ordering = LT | EQ | GT
```
mariari, line 3: Probably should use lowercase for the name

mariari (approved):
The golden tests make this look big, and also make me not want to squash. However due to the merge from develop we should squash this.

The standard library will need to be shuffled around and reformed after this, but we'll deal with that later.

PR #1111, opened 2022-02-03
Add a number of new instances to `Sexp.Base`

rokopt:
This PR allows the use of more library functions on `Sexp.Base` and `Sexp.T` by introducing instances for:

- `Ord`
- `Typeable`
- `Read`
- `Hashable` (formerly on only `Sexp.T`, not `Sexp.Base`)
- `Functor`
- `Foldable`
- `Traversable`
- `Recursive` (from `makeBaseFunctor`)
- `Corecursive` (from `makeBaseFunctor`)

The latter two, along with new `Fix`, `Mu`, and `Nu` forms of `Sexp.Base` (derived from the new `BaseF` generated by `makeBaseFunctor`), allow the use of a number of recursion libraries; see:

- https://hackage.haskell.org/package/recursion-schemes
- https://giters.com/vil1/recursion-schemes-cookbook
- https://hackage.haskell.org/package/data-fix
- https://hackage.haskell.org/package/recursion-schemes-5.2.2.2/docs/Data-Functor-Foldable.html
- https://www.cs.ox.ac.uk/ralf.hinze/LN.pdf


library/Sexp/src/Juvix/Sexp/Types.hs
```diff
@@ -132,3 +165,22 @@ showNoParens (Cons car cdr)
...
+
+type instance Foldable.Base (Base a) = BaseF a
+
+-- Define explicit general fixpoints, least fixpoints, and greatest fixpoints
+-- of BaseF (they are all isomorphic to each other and to Base).
+type FixBase a = Fix.Fix (BaseF a)
+
+type MuBase a = Fix.Mu (BaseF a)
+
+type NuBase a = Fix.Nu (BaseF a)
```
mariari, line 186: Nice looking forward to seeing what we can build with these

mariari (approved):
Looks good, looking forward to seeing what we can derive with recursion schemas and the like

PR #1115, opened 2022-02-08
Mariari/context cleanup

mariari:
Adds `Info` table to definition lookup.

This is part of a 2 part initiative to simplify the context interface.

This first phase adds a layer between the `Context.Definition` form, this lets us store information like source code or documentation.

The next phase, will be removing the richness of `Context.Definition` and revert it to a much more basic simplified type.

Part 1 is needed, to incode the data we will be losing in part 2


library/Context/src/Juvix/Context/Types.hs
```diff
@@ -42,6 +39,28 @@ data From b
   | Outside b
   deriving (Show, Functor, Traversable, Foldable, Eq)
 
+data Info term ty sumRep = Info
+  { infoTable :: HashMap.T Symbol Sexp.T,
+    infoDef :: Definition term ty sumRep
+  }
+  deriving (Show, Read, Generic, Eq, NFData)
+
+data InfoRecord term ty sumRep = InfoRecord
```
paulcadman, line 48: Will `InfoRecord` be removed in part 2 of this work? i.e can the `Record` (module) be encapsulated in the `Info` type?

paulcadman (approved):
Looks good - just one comment about `InfoRecord`

library/Context/src/Juvix/Context/Types.hs
mariari: We'll still need it, as as the general `Info` type will still have a `def` which is still a variant.

```haskell
data Info = Term ... | Module ... | CurrentModule
```

thus the same trick is stil needed

PR #1116, opened 2022-02-09
Hack around double-`show` golden test bug

rokopt:
#939 popped up for me again during development on https://github.com/anoma/juvix/tree/terence/geb-1, and as usual, I have no idea why, or what the root cause is.  I haven't found a way to avoid its blocking my testing, so I've tried to find a way to hack around it.  What I've been able to come up with that seems to work is just stripping a bunch of troublesome characters -- sometimes generated by `show`, such as `"` and `,` -- before comparing actual to expected output, and in the case of LLVM compilation tests in particular, even before saving the golden files, because something seems to be mishandling strings, especially strings that contain commas.

I'm not sure what people will think of such a hideous hack, but I'm also not sure how to proceed without it, so I'll submit a PR.  :)


test/examples-golden/positive/llvm/Const/Const.llvm
```diff
@@ -1,19 +1 @@
...
-declare external fastcc  i8* @malloc(i32)    
-
-
-declare external fastcc  void @free(i8*)    
-
-
-define external fastcc  i8 @main()    {
-main:
-  ret i8 10 
-}
```
mariari, line 19: Seems like hte new lines in the llvm backend are removed and replaced with > which makes them harder to read

rokopt: > Seems like hte new lines in the llvm backend are removed and replaced with > which makes them harder to read

Yeah -- that's annoying, but without stripping the whitespace, I still hit the bug.

It is possible to see the real LLVM code, though, by using `compile` from the command line on the golden-test `.ju` file, so maybe this is tolerable?

mariari: yeah as long as I can see the file unaltered as well I'm happy

rokopt:
This PR will be made obsolete by @Acentelles 's removal of the problematic code in #1118 .  Thanks!

PR #1117, opened 2022-02-09
Add missing bitcast in closure code

rokopt:
This bug was revealed by an interaction between strings and closures, and we debugged it this morning.

Unfortunately, my golden test expected-output generation is affected by #939 , so the tests won't pass until this is merged with some variant on #1116 .


mariari (approved):
Thank you, and thank you @Acentelles for finding this bug

rokopt:
> Unfortunately, my golden test expected-output generation is affected by #939 , so the tests won't pass until this is merged with some variant on #1116 .

I forgot that the format is the same as the one produced by `juvix compile`, so that I could work around this for now by manually generating the output.  I've added a commit to do that.

PR #1121, opened 2022-02-12
Mariari/info simplified p2

mariari:


library/Translate/src/Juvix/Contextify/ToContext/Sexp.hs
```diff
@@ -119,67 +129,33 @@ declare (Sexp.List [inf, n, i]) ctx
...
-        meta = lookupInfoTable name ctx
-        addSum con =
-          Context.Sum Nothing name
-            |> Context.SumCon
-            |> Context.Info meta
-            |> Context.add (NameSpace.Pub con)
+        addSum con ctx =
+          Structure.SumCon (pure name)
+            |> Structure.fromSumCon
+            |> injectNewTerm ctx con
```
mariari, line 147: This allows the sum constructors to have infix information, as this does not override such info

library/Translate/src/Juvix/Contextify/ToContext/Sexp.hs
```diff
@@ -119,67 +129,33 @@ declare (Sexp.List [inf, n, i]) ctx
...
-                  Context.add
-                    (NameSpace.Pub atomName)
-                    (Context.Info meta (Context.Information [Context.Prec prec]))
-                    ctx,
-                opensS = [],
-                modsDefinedS = []
-              }
+        newCtx =
+          injectMetaInformation ctx atomName ("precedence", Sexp.serialize prec)
+     in Type.PS newCtx [] [] |> pure
```
mariari, line 134: Here as well, as this just updates the meta information for infix precedence

library/Translate/src/Juvix/Contextify/ToContext/Sexp.hs
```diff
@@ -119,67 +130,33 @@ declare (Sexp.List [inf, n, i]) ctx
...
-                      meta
-                      (Context.TypeDeclar (Sexp.Cons (Sexp.atom "type") t))
-                  )
-                  newCtx,
-              opensS = [],
-              modsDefinedS = []
-            }
+        ctxWithType =
+          injectNewTerm newCtx name (Sexp.Cons (Sexp.atom "type") t)
+     in Type.PS ctxWithType [] [] |> pure
```
mariari, line 152: This is where precedence information is injected

paulcadman (approved):
Nice - I like the simplification of the Context types especially. I'll need to do some work to fix-up the Berlin Pipeline work for the improved types.

library/Context/src/Juvix/Closure.hs
```diff
@@ -30,13 +31,20 @@ data Information = Info
...
     mOpen :: Maybe NameSymbol.T
   }
   deriving (Show, Eq, Generic)
 
+newtype Info
+  = Prec Precedence
+  deriving (Show, Read, Generic, Eq, Data, NFData)
+
+instance Hashable Info where
+  hash (Prec pred) = hash pred
```
Acentelles, line 46: Would `deriving newtype Hashable` work here?

library/Context/src/Juvix/Context/NameSpace.hs
```diff
@@ -111,3 +105,19 @@ fromList List {publicL, privateL} =
 extractValue :: From a -> a
 extractValue (Pub aa) = aa
 extractValue (Priv a) = a
+
+--------------------------------------------------------------------------------
+-- Aeson Instances
+--------------------------------------------------------------------------------
+
+instance (A.ToJSON b) => A.ToJSON (T b) where
```
Acentelles, line 113: I don't think we need them any more

library/Context/src/Juvix/Context/Precedence.hs
```diff
@@ -78,3 +75,18 @@ fixity "<=" = nonAssoc 4
 fixity ">" = nonAssoc 4
 fixity ">=" = nonAssoc 4
 fixity _ = default'
+
+--------------------------------------------------------------------------------
+-- Aeson Instances
+--------------------------------------------------------------------------------
+instance A.ToJSON Associativity where
```
Acentelles, line 82: Same here

Acentelles (commented):
Can you add some golden tests for infixivity working for Sum types?

library/Translate/test/Test/Contextify.hs
```diff
@@ -44,6 +44,22 @@ moduleResolution =
                 "(:lambda-case \
                 \  (() (case TopLevel.A.a ((Cons a (Cons x y)) (add a x y)))))"
         Just expected T.@=? unwrapLookup "fi" t,
+      T.testCase "infix declaration works with declaration" $ do
+        Right t <-
+          contextualizeFoo
+            "type foo = (::) declare infixl (::) 20 let fi = a * 3 :: 10"
```
Acentelles, line 50: How do you know that `(::)` takes any arguments here?

Acentelles: This is not being tested end to end

mariari: Doesn't the LLVM only return int types, and since we can't pattern match, what is there to test with this?

Acentelles: In this case we don't know the arity of `::`. In fact, if anything, this example would suggest that arity is 0, so infixivity does not make sense.

I would expect and end to end test. We don't need pattern matching.

mariari: it says the arity

```haskell
infixl (::) 20 
```

I have another example where I set it to 1, to make sure the infixivity works

mariari: As for coherence, this PR changes nothing about the coherence of infixivity constructors, it simply allows the `declare` to work over them. If they worked before this, then they worked before this, however this only changes the declaration thereof as what was wanted

Acentelles: We want features to work. In this case, we want infixivity for sum constructors to work. What is the value from the user's perspective of infixivity to work partially (only for the parser)?

Acentelles (approved):

PR #1124, opened 2022-02-15
Remove Plonk Backend

Acentelles:
- Remove Plonk library
- Move plonk examples to llvm
- Rename llvm to LLVM folder for consistency with other libraries


mariari (approved):
As long as the tests are proper given forward, which looks like they are(?), then this is fine.

PR #1134, opened 2022-02-22
Initial categorial terms with recursive type support

rokopt:
The work scoped out below is not complete (and much of it is not intended for until after the upcoming release), but we've decided to mark this PR ready for review even though it hasn't reached the point of providing the user-visible features (recursive types) yet.  This should fend off further merge pain and also provide the _internal_ features that are already present, such as the interpreter backend, the composable separate term with monad-transformer interfaces for syntax, typechecking, evaluation, erasure, and code generation, and lots more serialization and showing.

The overall goals of this PR are:

- Introduce a new Core term which is homoiconic and extensible through changes only to its own internal modules, leaving all of its interfaces with the rest of HR, Core, and the backends untouched -- it is mutually-recursive with other Core terms so that Core terms can be embedded within it and vice versa; it's a "term à la carte"
- Use the new term to implement first-order recursive nameless (categorial) datatypes (as well as unit/void and product/coproduct, which will make the existing categorial terms obsolete) and functions to/from them

The recursive datatypes/functions are the target for the impending Juvix release, and this work does not depend on any other work to get them in.  However, it does not provide names -- if we wish, for example, to provide names for the categorial types using `Record` and `Sum` types, then _that_ will require integration.  It also does not provide pattern matching -- rather, it provides recursion through fixpoint (and cofixpoint) operators.  If we want pattern matching, that will also require integration (as well as a lot of separate work in pattern matching implementation itself, which as far as I know we're not planning).

So it's independent of other work in raw functionality, but its own functionality _is_ pretty raw.  We'll have to work out how much syntactic sugar we can add by when.

The extensibility aims to allow for rapid development after release of further features; here's one possible order:

- Higher-order datatypes and functions
- Refinement (equality) types (equalizers/coequalizers)
- Dependent/quotient types (base-change/cobase-change adjunctions)
- Metaprogramming (refined-ADT/arbitrary-category representation adjunction)
- Effects (adjunctions between arbitrary categories)

The infrastructure already added in order to support the new term includes the following:

- A new backend, `Interpreter`, which "compiles" Core terms to Haskell types/functions and executes them -- this allows a separation of testing between compilation from core to a backend and backend codegen (thus making error isolation easier and allowing faster development of core terms without having to wait for codegen for a more complicated backend, which could perhaps in some cases even be done in parallel)
- Monad-transformer interfaces for the modules through which the new term interacts with Core terms:
  - `Syntax`
  - `Typechecker`
  - `Eval`
  - `Erasure`
  - `Codegen`

Next things to do before release:
- Codegen for Core functions embedded in categorial terms
- Translate categorial products/coproducts to LLVM's existing record/sum support
- Translate categorial fixpoints to LLVM recursive functions


rokopt:
Added a couple commits to update to the most recent version of `samuli/fixed-stack-nix-shell`.

PR #1136, opened 2022-02-24
Modify Makefile to allow building for Darwin

Acentelles:


paulcadman (approved):
Thanks @Acentelles! This works for me, builds and passes all tests.

PR #1137, opened 2022-02-24
Update Stack LTS

Acentelles:
Update ghc version required by the latest [haskell-language-server](https://haskell-language-server.readthedocs.io/en/latest/installation.html)

To make it work, I downloaded the source code
```
git clone https://github.com/haskell/haskell-language-server --recurse-submodules
cd haskell-language-server
```
Installed the version that uses ghc 8.10.7
```
stack ./install.hs hls-8.10.7
```

And add the executable `haskell-language-server-8.10.7` to the path for the editor to recognise it


paulcadman:
The changes in the test files are due to regeneration? Is there a reason why they needed to be generated as part of this PR? Did something break due to the library changes?

Acentelles:
I believe that the newer lts has an updated version of the pretty printer of these files

paulcadman (approved):
This builds and runs tests successfully for me 👍

PR #1139, opened 2022-02-24
yaml-gen: Globally disable nix when run on macOS

paulcadman:
We cannot get the llvm-hs dependency to build with nix on macOS
and so we disable it in all projects on this platform.


mariari (approved):
Seems fine, we should probably move the `nix-yaml` logic out into a function, but this seems fine

Acentelles (approved):

PR #1108, opened 2022-01-25
BerlinPipeline Automation

mariari:
Automation work

PR #1141, opened 2022-02-25
Adds Berlin pipeline implementation and integrates with frontend

paulcadman:
This is a rebased version of https://github.com/anoma/juvix/pull/1108

This PR:

1.  Implements @mariari 's [Berlin pipeline proposal](https://hackmd.io/iG-uj5t_T3i82dxBRb_PSA). See the modules within [BerlinPipeline:Juvix.BerlinPipeline](https://github.com/anoma/juvix/tree/mariari/berlin-automation-2/library/BerlinPipeline/src/Juvix/BerlinPipeline).
2. Translates existing desugar / contextify passes to use the Berlin pipeline. See [translate:Juvix.BerlinPasses](https://github.com/anoma/juvix/blob/mariari/berlin-automation-2/library/Translate/src/Juvix/BerlinPasses.hs).
3. Integrates the Berlin pipeline into the frontend pipeline. See [pipeline:Juvix.Pipeline](https://github.com/anoma/juvix/blob/46c9900adaff0bce81dd3d4a66af2483f4189a7b/library/Pipeline/src/Juvix/Pipeline.hs#L148)
4. Updates the Easy module to [use the Berlin pipeline passes](https://github.com/anoma/juvix/blob/46c9900adaff0bce81dd3d4a66af2483f4189a7b/library/Playground/Easy/src/Easy.hs#L194)

This PR does not provide an interface to the user to enable tracing - which is one of the aims for the Berlin pipeline. Subsequent work will add the option to trace through the desugar and contextify compiler passes using the CLI.

In the meantime here's a sketch on how tracing can be enabled using the API. This will run desugar/contextify passes on `(:defun foo (x) (+ x 1))` (an expression which is obtained from the frontend syntax by the parser.

The `example` function emits a trace of the compiler passes with label  `"Desugar.cond-runner"`, `"Desugar.condTransform"` as specified in the pipeline setup.

```haskell
import qualified Juvix.BerlinPipeline.Meta as Meta
import qualified Juvix.BerlinPipeline.Pipeline as Pipeline
import qualified Juvix.BerlinPipeline.Env as Pipeline.Env
import qualified Juvix.Context as Context

Right sexp = Sexp.parse "(:defun foo (x) (+ x 1))"

startingEnv :: IO Pipeline.WorkingEnv
startingEnv =
  Context.empty "JU-USER"
    >>| Pipeline.WorkingEnv [Pipeline.Sexp sexp]

exampleMeta :: IO Meta.T
exampleMeta = do
  Pipeline.CIn _ surroudning <-
    startingEnv
      >>= Pipeline.Env.run eval
        . Pipeline.modifyTraceCIn
          (`Trace.enable` ["Desugar.cond-runner", "Desugar.condTransform"])
        . Pipeline.emptyInput
  surroudning ^. Pipeline.metaInfo |> pure

example :: IO ()
example = do
  exampleMeta >>= Meta.info
```


library/Translate/src/Juvix/BerlinPasses/Contextify.hs
```diff
@@ -0,0 +1,347 @@
...
+    noOpJob = Automation.noOpJob ctx (pa ^. current)
+    updateJob name info =
+      Pipeline.UpdateJob
+        { newContext = Context.addGlobal name info ctx,
+          process =
+            Pipeline.Process
+              { _current = Pipeline.InContext name,
+                _newForms = []
+              }
+        }
```
mariari, line 191: I think once we merge this, we can make this pass a `simplified`. As it looks like we've made more work for ourselves by just binding over the pass that is one (figureRecord just sets the def and doesn't return any additional forms.)

mariari (approved):
Looks good, glad to see it in

library/Translate/src/Juvix/BerlinPasses/Contextify.hs
paulcadman: You're right - I'll change this.

PR #1144, opened 2022-03-01
Adds a pretty printer for S-expressions

paulcadman:
The pretty printer for S-expressions will help in the generation of readable tracing from the Berlin pipeline output.

The pretty printer will automatically nest s-expressions when the display becomes too large.

```haskell
import qualified Juvix.Sexp.Pretty.Base as Sexp
import Prettyprinter
import Prettyprinter.Util (putDocW)

-- helpers
mkAtom n = Sexp.Atom $ Sexp.N n Nothing
mkCons = foldr Sexp.Cons
mkList = mkCons Sexp.Nil

-- normal 80 char width
putDocW 80 $ ppSexp $ mkList [mkAtom n | n <- [0..4]]

-- artificially narrow width
putDocW 5 $ ppSexp $ mkList [mkAtom n | n <- [0..4]]
```

In the first case the s-expression is printed like:

```
(0 1 2 3 4)
```

In the second case the s-expression is printed like:

```
(0 1
   2
   3
   4)
```

This way of formatting s-expressions is common because the first entry is usually a function name, the rest are arguments.


mariari (approved):
Looks fine, not too much in here, looks straight forward

PR #1146, opened 2022-03-22
Fix `doc/` -> `docs/` links in several files

M4444:
This should also be fixed on the `How to contribute` button at https://juvix.org/community/, and there's also [one last file](https://github.com/anoma/juvix/blob/develop/experimental/ML%20Conversion/Conversion/ML.hs) that contains `doc/` but it liked to `doc/Aricheture/Sexp.org` which appears to have been deleted in [6cee9c52](https://github.com/anoma/juvix/commit/6cee9c52779c7fb487cae8bfa1d4a0373f4c0720) so I'm not sure how you would like to deal with this case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ test-compile: build update-local-stdlibs
 	rm test/examples/positive/llvm/*.ll
 
 bench:
-	$(STACK) bench --benchmark-arguments="--output ./doc/Code/bench.html"
+	$(STACK) bench --benchmark-arguments="--output ./docs/Code/bench.html"
 
 repl-lib:
 	$(STACK) ghci juvix:lib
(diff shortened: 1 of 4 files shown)


mariari:
> This should also be fixed on the `How to contribute` button at https://juvix.org/community/, and there's also [one last file](https://github.com/anoma/juvix/blob/develop/experimental/ML%20Conversion/Conversion/ML.hs) that contains `doc/` but it liked to `doc/Aricheture/Sexp.org` which appears to have been deleted in [6cee9c52](https://github.com/anoma/juvix/commit/6cee9c52779c7fb487cae8bfa1d4a0373f4c0720) so I'm not sure how you would like to deal with this case.

Thank you for updating the documentaiton. I'll personally update that file with the new location of that, which I believe is here

https://heliaxdev.github.io/juvix-docs/docs/design/s-expression/

mariari (approved):

PR #1377, opened 2022-07-14
Check all the type parameter names are different when declaring an inductive type

jonaprieto:


src/Juvix/Syntax/Concrete/Scoped/Error.hs
```diff
@@ -32,6 +32,7 @@ data ScoperError
   | ErrMultipleCompileBlockSameName MultipleCompileBlockSameName
   | ErrMultipleCompileRuleSameBackend MultipleCompileRuleSameBackend
   | ErrWrongKindExpressionCompileBlock WrongKindExpressionCompileBlock
+  | ErrWrongInductiveParameterName WrongInductiveParameterName
```
janmasrovira, line 35: I would suggest `Dup` or `Duplicate` instead of `Wrong`

src/Juvix/Syntax/Concrete/Scoped/Error/Types.hs
```diff
@@ -522,3 +522,27 @@ ambiguousMessage n es =
...
+    GenericError
+      { _genericErrorLoc = i,
+        _genericErrorMessage = prettyError msg,
+        _genericErrorIntervals = [i]
+      }
+    where
+      param = e ^. wrongInductiveParameterName
+      i = getLoc param
+      msg =
+        "The second occurrence of the parameter name"
```
janmasrovira, line 544: I would say "Invalid name X. Inductive parameter names cannot be repeated". I feel it is more concise and easier to understand

janmasrovira (approved):

PR #1379, opened 2022-07-15
Refine hole in type signature to function type

### LINKED ISSUE #1331: Refine hole in type signature to function type
Currently typechecking the following fails. 

```agda
module Tmp;

inductive Bool {
  false : Bool;
  true : Bool;
};

not : _;
not false ≔ true;
not true ≔ false;

end;
```
`impossible` is thrown because there are more patterns than arguments in the type of the signature.

janmasrovira:
Fixes #1331.
Depends on #1378 and needs to be rebased after that is merged


src/Juvix/Syntax/MicroJuvix/Error/Types.hs
```diff
@@ -9,8 +9,8 @@ import Juvix.Syntax.MicroJuvix.Language
 -- not match the type of the inductive being matched
 data WrongConstructorType = WrongConstructorType
   { _wrongCtorTypeName :: Name,
-    _wrongCtorTypeExpected :: Name,
-    _wrongCtorTypeActual :: Name,
+    _wrongCtorTypeExpected :: InductiveName,
```
jonaprieto, line 12: Test!

jonaprieto: Hello Jonathan!

jonaprieto: Hi

PR #1392, opened 2022-07-18
Keep qualified names

### LINKED ISSUE #1344: Debugging: Add qualification in error message.

janmasrovira:
Fixes #1344 

Adds a new field to Abstract Name called `_namePretty :: Text`.
This field contains the full textual representation of the name, including qualification.


src/Juvix/Syntax/Abstract/Name.hs
```diff
@@ -16,6 +16,7 @@ data Name = Name
   { _nameText :: Text,
     _nameId :: NameId,
     _nameKind :: NameKind,
+    _namePretty :: Text, -- How to print this name in error messages
```
jonaprieto, line 19: is this really neccessary? I'd expect a function (lens) to reconstruct such a (pretty) name on the fly. Not to carry with that info everywhere.

jonaprieto: Can we remove it and make a function to get the same result?

janmasrovira: I don't see a way to do that. `_nameText` has to contain the unqualified name since it is used for compilation. Can you implement the function that you are referencing?

jonaprieto: Maybe. It was a question. ;)

jonaprieto (dismissed):

jonaprieto (approved):

Issue #1413, opened 2022-07-25
Implement an html documentation generator similar to Haddock


janmasrovira:
Current goals:
- It will be part of the compiler, not an external tool (at least for now).
- It should be possible to write _Haddock comments_, or _Judoc comments_ (as we shall call them for now), to document different parts of a program. On the first iteration we will only support functions. Later we want to add
  1. Inductive types
  2. Constructors
  3. Modules
  4. Individual function arguments (to be discussed)
-  Will support a very basic language described below. In the nearish future we may want to discuss if we want to support a popular markup language, such as Markdown, and rely on an existing parser.
- The Judoc comments need to be scoped and type checked.

The initial syntax of Judoc comments is very simple and will change in the future.
A Judoc comment spans multiple lines. Each line must start with `--- ` (note the trailing space).
A Judoc comment can have text and inlined expressions (surrounded by `;`). Having `;` as a delimiter is convenient because it is already a reserved symbol of the language (needed for parsing).

Example:
```
module Example;

--- The identity function.
--- Example:
--- ;id Type; is the same as ;Type;
id : {A : Type} -> A -> A;
id a := a;

end;
```

Related:
- #1359

PR #1416, opened 2022-07-26
Implement an html documentation generator similar to haddock (#1413)

janmasrovira:
The 1st iteration of the design of juvix documentation is almost a copy of haddock with the linuwial theme. I've changed some colors following the palette that juvix.org uses in the landing page
Currently the generated documentation consists of:
1. index with the tree of all modules reachable from the main module (which modules show up should be refined in the future)
2. A page for each module with its signatures documented and linked identifiers
3. Pretty printed source code of each module with linked identifiers

To test it
```
cd examples/milestone/MiniTicTacToe
juvix internal doc --open MiniTicTacToe.juvix
```
Note that the `open` flag will not work on Mac since it uses `xdg-open` internally. This will need to be fixed in a later PR

Next I think it would be a good idea to replace the existing `html` command with the new `doc` command


jonaprieto:
Great work!

jonaprieto (approved):

PR #1421, opened 2022-07-28
JuvixCore

lukaszcz:
JuvixCore framework and evaluator.

# Description

JuvixCore is the Core internal representation for Juvix. It is based on a simple extended lambda-calculus, stripped down enough to enable easy compilation, but high-level enough to enable easy translation from the frontend.

# Checklist:

- [x] Node datatype
- [x] Info datatype
- [x] Recursors on Node
- [x] Evaluator
- [x] Typeclass instances for Node:
  - [x] Equality (need to ignore Info)
  - [x] Pretty printing
- [x] Errors
- [x] Parser for Node
- [x] Evaluator CLI
- [x] Evaluator tests
- [x] Strings
- [x] IO monad builtins
- [x] Transformations:
  - [x] Eta-expansion of builtin and constructor applications


src/Juvix/Core/Evaluator.hs
```diff
@@ -0,0 +1,135 @@
...
+        --
+        -- eval env (App l r) =
+        --   case eval env l of
+        --     Closure env' b ->
+        --       let !v = eval env r in eval (v : env') b
+        --
+        -- To do this more efficently for multi-argument functions (without
+        -- creating closures for each intermediate function and matching each
+        -- twice) we gather all application arguments, evaluate them from left
+        -- to right and push the results onto the environment.
```
lukaszcz, line 57: Probably this doesn't make much difference since I changed Builtin and Constructor to BuiltinApp and ConstructorApp.

lukaszcz:
Regarding Data.TMap, it has very fast lookup because it's based on arrays, but insertion and deletion are O(n) where n is the number of keys. This should be OK for our purposes, because we have only a small constant number of different infos at any given node, and we'll be doing lookup more often than insertion or deletion.

I still think we should change our Info type currently based on HashMaps to TMap, but we should be aware of the O(n) insertion/deletion issue, in case the above assumptions no longer hold in the future.

lukaszcz:
The typerep-map package (with TMap) requires base < 4.15 and ghc-prim < 0.7, we have base = 4.16 and ghc-prim = 0.8. So I will keep the HashMap-based implementation for now. We can decide what to do with this later.

paulcadman (approved):
These changes do not affect the rest of the compiler (yet), are very well tested and @lukaszcz explained the main ideas to us in person. I propose we merge this PR now so we can begin work on integration with the rest of the compiler.

jonaprieto (approved):

Issue #1419, opened 2022-07-27
 Support Wasm import and exported symbol names in C backend


paulcadman:
### Background

See #1355 

To support compilation of WASM binaries that can be executed in Anoma we need to be able to choose names for symbols we generate in the C backend.

Specifically Anoma expects validity predicates to have a particular name and signature in the binary WASM file, this is visible in the Anoma source as: https://github.com/anoma/anoma/blob/5831077bf1f1ff917f97264471b0ddcae224e739/macros/src/lib.rs#L75

This feature is also required to support our example web application https://github.com/anoma/juvix/issues/1352.

### How to export a name 

To export a symbol it should be in scope in the entry point module at compilation time.

Any function with a name consisting only of alpha numeric plus `_` characters will be exported.

### How to import a name 

Any axiom without a compile block with a name consisting only of alpha numeric plus `_` characters will be marked as imported.

### Code generation

The compiler can use the `exports` map from the configuration to name the corresponding symbols in the WASM binary. We can do this with the [`export_name` attribute](https://clang.llvm.org/docs/AttributeReference.html#export-name), and [`import_name` attribute](https://clang.llvm.org/docs/AttributeReference.html#import-name)

The `export_name` attribute sets the name for the exported symbol and ensures that the symbol is exported in the binary:

```c
__attribute__((export_name("exportedName1")))
int juvixName1_1234(int fa0)
```

The `import_name` attribute adds a function and signature to the import table of the Wasm binary.

```c
__attribute__((import_name("importedName1")))
int juvixName1_1234(int fa0)
```

janmasrovira:
I think what you suggest is reasonable

paulcadman:
I changed the issue to remove yaml configuration since we will export / import all symbols with alpha numeric names, this is sufficient for now.

paulcadman:
This was implemented in https://github.com/anoma/juvix/pull/1425, https://github.com/anoma/juvix/pull/1426 and https://github.com/anoma/juvix/pull/1449

PR #1425, opened 2022-07-29
Export all functions with alpha numeric names from entrypoint module

paulcadman:
This is part of the implementation for https://github.com/anoma/juvix/issues/1419

This PR adds `__attribute__((export_name(<name>))` declarations to every type signature corresponding to every symbol in the entry point module that has an alpha numeric name (so that it can safely be present in the output C file). This is an intermediate step to adding the export information to the `juvix.yaml` file as described in #1419.

This causes these functions to be present in the export table of the resulting WASM module.

To test this we need to be able to invoke the exported functions using the WASM module. This raises the issue of how we will execute these functions in the Anoma VM and (possibly) in the browser. For this use case it is not appropriate to use the WASM-WASI runtime as these IO functions (providing `fd_write`, `fd_read`, `proc_exit` etc.) are not available in the Anoma VM or in the browser.

So this PR also adds a new standalone runtime for the compiler, in addition to the WASI standalone and the WASI libc runtimes. This runtime is appropriate for compiling WASM libraries which export functions that can be invoked without the WASI imports (and that do not, for example, have a main entry point). This is the runtime we'll use for compiling validity predicates for example.

To try this runtime out:

```shell
cd tests/positive/MiniC/ExportName
juvix compile --runtime standalone Input.juvix
```

Then use wasmer to inspect the output WASM file, see it exports all the functions in scope in the module:

```shell
wasmer inspect Input.wasm
Type: wasm
Size: 169 B
Imports:
  Functions:
  Memories:
  Tables:
  Globals:
Exports:
  Functions:
    "one": [] -> [I32]
    "two": [] -> [I32]
    "three": [] -> [I32]
    "four": [] -> [I32]
    "five": [] -> [I32]
    "six": [] -> [I32]
    "fun": [] -> [I32]
  Memories:
    "memory": not shared (2 pages..)
  Tables:
  Globals:
```

We can then invoke the `fun` function for example:

```shell
wasmer run Input.wasm --invoke fun
6
```


jonaprieto (approved):
🚢

Issue #1355, opened 2022-06-28
Support the compilation of validity predicates that can be deployed to Anoma


paulcadman:
Currently `minijuvix compile` produces WASI binaries that can be executed using a local WASM runtime like wasmer.

This issue is to add support for compilation of pure WASM binaries that export a function that can be executed on Anoma.

(See https://github.com/anoma/anoma/blob/5831077bf1f1ff917f97264471b0ddcae224e739/macros/src/lib.rs#L75 for the corresponding Rust validity predicate macro that produces the WASM signature via C FFI).

paulcadman:
After https://github.com/anoma/juvix/pull/1449 we've demonstrated the ability to run a validity predicate compiled from Juvix in Anoma. 

See:

https://github.com/anoma/juvix/blob/7bf337217d3fdcc1f8f5d5b976577d2e76555caa/tests/positive/MiniC/AlwaysValidVP/Input.juvix

We do not have support for deserialising the arguments (Address struct, transaction data, set of keys changed and set of verifiers) that are passed to the validity predicate.

For this we'll need to support decoding a pointer to a Juvix type using the Borsh spec https://borsh.io

PR #1426, opened 2022-08-01
WASM import all non-compile axioms with alphanum names in entrypoint

paulcadman:
> ⚠️ The tests now depend on `node` in order to test injecting WASM import functions into a Juvix program.

This is part of the implementation for https://github.com/anoma/juvix/issues/1355

This PR adds `__attribute__((import_name(<name>))` declarations to every type signature corresponding to axioms with no-compile block that has an alpha numeric name in the entry point module (so that it can safely be present in the output C file). 
This causes these functions to be present in the import table of the resulting WASM module.

To try this out:

```shell
cd tests/positive/MiniC/ImportExportName
juvix compile --runtime standalone Input.juvix
```

To see the imports and exports of the module:

```shell
wasmer inspect Input.wasm
Type: wasm
Size: 172 B
Imports:
  Functions:
    "env"."hostDisplayString": [I32] -> []
  Memories:
  Tables:
  Globals:
Exports:
  Functions:
    "juvixRender": [] -> []
  Memories:
    "memory": not shared (2 pages..)
  Tables:
  Globals:
```

You can then inject an implementation of `hostDisplayString` from nodeJS, which is called in Juvix and invoke the `juvixRender` Juvix function from the nodeJS file with:

```shell
node input.js
Hello World from Juvix!
```

### DIFF
--- a/docs/org/SUMMARY.org
+++ b/docs/org/SUMMARY.org
@@ -6,6 +6,8 @@
   - [[./getting-started/quick-start.md][Quick start]]
   - [[./getting-started/dependencies.md][Installing dependencies]]
 - [[./examples/README.md][Juvix Examples]]
+- [[./tutorials/README.md][Tutorials]]
+  - [[./tutorials/nodejs-interop.md][NodeJS Interop]]
 
 - [[./language-reference/README.md][Language reference]]
   - [[./language-reference/comments.md][Comments]]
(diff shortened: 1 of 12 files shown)


jonaprieto:
Awesome! Could you please add one entry in the documentation about this particular example, the Hello World? I'd possibly put it in the section Tutorials.

jonaprieto (approved):

PR #1442, opened 2022-08-05
Add initial support for examples in Html documentation

### LINKED ISSUE #1441: Support examples in juvix documentation

janmasrovira:
Closes #1441.

This PR adds the possibility to add _examples_ in judoc comments.
An example has the following syntax:
```
--- >>> expr;
```
The `---` is used to prefix judoc comments as usual. The `>>>` syntax is borrowed from haddock (we are free to change it if we want). The trailing `;` is convenient to stop the parsing of the expression.

Examples are passed down the pipeline and normalized during type checking. The normalized expression is showed in the html generated documentation.

Note that currently we only support normalization for a small subset of expressions (see https://github.com/anoma/juvix/pull/1404).

A concrete example.
Given the following definitions
```
--- A row in a board implemented as ;List Square;
---
--- >>> Row;
Row : Type;
Row ≔ List Square;
```
The output looks like this:
image

### DIFF
--- a/examples/milestone/TicTacToe/Logic/Game.juvix
+++ b/examples/milestone/TicTacToe/Logic/Game.juvix
@@ -69,6 +69,16 @@ inductive Square {
   occupied : Symbol → Square;
 };
 
+--- A row in a board implemented as ;List Square;
+---
+--- >>> Row;
+Row : Type;
+Row ≔ List Square;
+
+-- A column in a board implemented as ;List Square;
+-- Column : Type;
+-- Column ≔ Row;
+
 --- Equality for ;Square;s
 ==Square : Square → Square → Bool;
 ==Square (empty m) (empty n) ≔ m == n;
@@ -79,14 +89,6 @@ inductive Square {
 showSquare : Square → String;
 showSquare (empty n) ≔ " " ++str natToStr n ++str " ";
 showSquare (occupied s) ≔ " " ++str showSymbol s ++str " ";
-
--- A row in a board implemented as ;List Square;
--- Row : Type;
--- Row ≔ List Square;
-
--- A column in a board implemented as ;List Square;
--- Column : Type;
--- Column ≔ Row;
 
 -- TODO use a type alias?
 --- A grid of squares
--- a/examples/milestone/TicTacToe/Logic/Game.juvix
+++ b/examples/milestone/TicTacToe/Logic/Game.juvix
@@ -69,11 +69,11 @@ inductive Square {
   occupied : Symbol → Square;
 };
 
---- A row in a board implemented as ;List Square;
----
---- >>> Row;
-Row : Type;
-Row ≔ List Square;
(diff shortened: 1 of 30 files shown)


examples/milestone/TicTacToe/Logic/Game.juvix
```diff
@@ -69,6 +69,16 @@ inductive Square {
   occupied : Symbol → Square;
 };
 
+--- A row in a board implemented as ;List Square;
+---
+--- >>> Row;
+Row : Type;
```
paulcadman, line 75: This should be commented out?

If I use the `Row` type alias in the program and compile it I get:

```
juvix: TypeUniverse Micro to C: not yet supported
  error, called at src/Juvix/Prelude/Base.hs:242:9 in juvix-0.2.2-Bjv4Lkpbnpb5ZXqkLcn1KX:Juvix.Prelude.Base
  error, called at src/Juvix/Compiler/Backend/C/Data/Base.hs:23:19 in juvix-0.2.2-Bjv4Lkpbnpb5ZXqkLcn1KX:Juvix.Compiler.Backend.C.Data.Base
```
(limitation of the current C backend).

paulcadman (commented):
I'm trying to add an example to:

https://github.com/anoma/juvix/blob/9543b06cf50d94f82b9529ea0a816c89f6c60a31/examples/milestone/TicTacToe/Logic/Game.juvix#L23

The comment is wrong - so illustrates why the example feature is useful.

If I write:

```
>>> concat ("a" ∷ "b" ∷ nil);
```

I expect the output to be `"ab"` but I get:

```
concat (∷ {String} ("a" {String}) (∷ {String} ("b" {String}) (nil {String})))
```

This might be a limitation of the magic literals?

paulcadman:
The above doesn't work because currently only type alias expressions are supported?

janmasrovira:
> I'm trying to add an example to:
> 
> https://github.com/anoma/juvix/blob/9543b06cf50d94f82b9529ea0a816c89f6c60a31/examples/milestone/TicTacToe/Logic/Game.juvix#L23
> 
> The comment is wrong - so illustrates why the example feature is useful.
> 
> If I write:
> 
> ```
> >>> concat ("a" ∷ "b" ∷ nil);
> ```
> 
> I expect the output to be `"ab"` but I get:
> 
> ```
> concat (∷ {String} ("a" {String}) (∷ {String} ("b" {String}) (nil {String})))
> ```
> 
> This might be a limitation of the magic literals?

This is due to the limitations of the current algorithm for normalization as it is only supposed to work for trivial definitions: See https://github.com/anoma/juvix/pull/1404

examples/milestone/TicTacToe/Logic/Game.juvix
janmasrovira: done

paulcadman (approved):
🥇

PR #1443, opened 2022-08-08
Refactor pretty to reduce duplication

janmasrovira:
This PR unifies the type `Ann` for all languages and its related auxiliary functions.

Future work:
We could discuss the possibility to unify all (or most) of the `PrettyCode` classes.


jonaprieto (approved):
Nice! less code

paulcadman (approved):

PR #1446, opened 2022-08-10
Add Towers of Hanoi and Pascal triangle examples

paulcadman:
These are a couple of examples we can have as exercises or demos in the retreat workshop.


janmasrovira (approved):
Awesome!

PR #1449, opened 2022-08-11
Allow _ in Wasm exported names to support Anoma signature

paulcadman:
The Anoma validity predicate Wasm signature is:

```
"_validate_tx": [I64, I64, I64, I64, I64, I64, I64, I64] -> [I64]
```

So we need to allow exported names containing '_'.

Anoma gets the exported function by name and type at:

https://github.com/anoma/anoma/blob/50b5e77f04a9afc036656353335bd232fcdba8a7/shared/src/vm/wasm/run.rs#L258

This PR adds a Juvix module that exports a function with this signature
and a test that the resulting Wasm function can be called.


janmasrovira (approved):
:checkered_flag:

jonaprieto (approved):

PR #1453, opened 2022-08-12
Give a proper type to literal natural numbers

janmasrovira:
Before this PR, literal natural numbers were assigned the type `{A : Type} -> A`.
Now, when the type checker finds a literal number, it checks that the builtin natural numbers have been defined and it gives it the proper type. This change was easy. Most of the changed lines are about fixing tests.


paulcadman (approved):
Good job!

PR #1454, opened 2022-08-15
Fixes TicTacToe Web example

paulcadman:
Now that integer literals have the builtin int type we need an actual
IOUnit value to use in the IO sequence implementation.

This commit also adds the TicTacToe web example to the test suite. It is
a typecheck / C generation only test because it uses the Wasm browser APIs.


janmasrovira (approved):
Great!

PR #1457, opened 2022-08-18
Add a native compile target for demos

paulcadman:
Add an option to compile a Juvix program to a native executable:

```
$ cd examples/milestone/HelloWorld
$ juvix compile -t native HelloWorld.juvix
$ ./HelloWorld
hello world!
```

This option is useful for demos because it does not depend on the Wasmer / Wasi environment.


jonaprieto (approved):
Tested and discussed in person.

PR #1458, opened 2022-08-19
Fix typechecker

### LINKED ISSUE #1455: Typechecking of incorrect polymorphic self-application generates wrong error message
Typechecking
```agda
module identity;

id : {A : Type} -> A -> A;
id a := a;

id' : {A : Type} -> A -> A;
id' {A} := id {(A -> A) -> A -> A} id;

end;
```
results in
```
The expression id {(A → A) → A → A} (id {_}) has type:
  A → A
but is expected to have type:
  A
```
Unless I misunderstood the syntax, this is wrong. The body of `id'` is `id id` and it should have type `A -> A`.

The type of the first `id` is `((A -> A) -> A -> A) -> (A -> A) -> A -> A` and the type of `id id` is `(A -> A) -> A -> A`. So the program is correctly rejected, but with a wrong error message.

janmasrovira:
Fixes #1455. It also fixes matching of functions where the impliciteness of the left type did not match.


jonaprieto (approved):
Discussed in person.

PR #1459, opened 2022-08-19
Add --stdin flag

janmasrovira:
Adds the `--stdin` flag. When active, the main file is read from stdin. Useful for VScode and probably for emacs as well.


jonaprieto (dismissed):
Tested and discussed in person.

jonaprieto (approved):

PR #1461, opened 2022-08-19
Use WASI-SDK for clang

paulcadman:
WASI-SDK already has wasm setup, this makes installation easier.


paulcadman:
We can revisit this at some point - closing for now.

PR #1467, opened 2022-08-20
example/tic-tac-toe: clear screen in between prints

tzemanovic:
I was playing around with this example - there are some issues with my changes and I expect they'll also break some tests, so I wouldn't merge this, but I just wanted to share some observations. I'm not sure if it's relevant for now, as @lukaszcz mentioned that the `IO` will be refactored and `>>` is temp, but here goes:
- on e.g. line 48, I had to use parens around `clear >> putStr (prompt x)`, because otherwise `,` operator takes precedence over `>>` and fails with:
  ```
  The expression , {_} {_} (putStr (++str msg (prompt (state b p noError)))) (state b p noError) has type:
    × _ _
  but is expected to have type:
    IO
  ```
- I'm not sure what's going with this, but I tried to do something like:
  ```
  clearBefore : IO -> IO;
  clearBefore x ≔ clear >> x;
  ```
  When I replaced the calls to `clear >>` with `clearBefore $`, it seems that the order of these effects is reversed and the screen is cleared after the prints

### DIFF
--- a/examples/milestone/TicTacToe/CLI/TicTacToe.juvix
+++ b/examples/milestone/TicTacToe/CLI/TicTacToe.juvix
@@ -22,6 +22,11 @@ compile parsePositiveInt {
   c ↦ "parsePositiveInt";
 };
 
+axiom clear : IO;
+compile clear {
+  c ↦ "system(\"clear\")";
+};
+
 -- | Reads a ;ℕ; from stdin
 getMove : Maybe ℕ;
 getMove ≔ validMove (parsePositiveInt (readline));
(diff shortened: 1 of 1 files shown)


paulcadman:
Thanks for trying this @tzemanovic, the observations are useful.

The reason the tests fail is because the `system` C function is not implemented in the minimal Juvix wasm runtime because it does not have access to libc functions.

jonaprieto:
Regarding the operator precedence, we will probably open an issue in the anoma/juvix-stdlib repo to discuss it. Thank you @tzemanovic for the input.

PR #1472, opened 2022-08-21
Properly type check patterns that need normalization

### LINKED ISSUE #1466: Cannot pattern match synonym of inductive type
spotted by @memasdeligeorgakis 
```
module M;

inductive ℕ {
  z : ℕ;
  s : ℕ → ℕ;
};

nat : Type;
nat := ℕ;

infixl 1 +;
+ : nat → nat → nat;
+ z b := b;
+ (s a) b := s (a + b);

end;
```

The problem is that when type checking a pattern, the type needs to be normalized, which is currently not beig done. Hence `nat` is not detected as an inductive type.

janmasrovira:
Fixes #1466


jonaprieto (approved):

jonaprieto:
Great!

PR #1486, opened 2022-08-29
Add an option to show name ids in errors

### LINKED ISSUE #1215: Add an option to show name ids in errors
As the title says, for example, instead of:

```
The expression plus-nat has type:
  Nat -> Nat -> Nat
but is expected to have type:
  Nat -> Nat
```

The implemented flag `--show-name-ids` is not working at the moment for
error messages.

```
The expression plus-nat@18 has type:
  Nat@15 -> Nat@15 -> Nat@15
but is expected to have type:
  Nat@15 -> Nat@15
```

lukaszcz:
# Description

* `Reader GenericOptions` effect added to the function `genericError` in Data/Error/GenericError.hs.
* GlobalOptions converted into GenericOptions and stored in EntryPoint in Main/getEntryPoint
* `ppCode` in Typechecking/Error/Pretty.hs still has the GenericOptions argument and I resist removing it. I don't see much benefit in changing this function to use effects instead of an options argument, because it is used only in the error messages in Analysis/*. Changing it makes sense only if we rewrite all error message creation to use monadic style, and I don't see the benefit of it. It's more readable to create an error message as an expression than do it piecewise with do-notation, and then one needs exactly the ppCode function with an GenericOptions argument. NB this ppCode just calls Micro.ppCode and runReader.
* Fixes #1215.

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -10,8 +10,8 @@ import System.Console.ANSI qualified as Ansi
 
 data App m a where
   ExitMsg :: ExitCode -> Text -> App m ()
-  ExitJuvixError :: GenericOptions -> JuvixError -> App m a
-  PrintJuvixError :: GenericOptions -> JuvixError -> App m ()
+  ExitJuvixError :: JuvixError -> App m a
+  PrintJuvixError :: JuvixError -> App m ()
   ReadGlobalOptions :: App m GlobalOptions
   RenderStdOut :: (HasAnsiBackend a, HasTextBackend a) => a -> App m ()
   RunPipelineEither :: Sem PipelineEff a -> App m (Either JuvixError a)
@@ -32,22 +32,22 @@ runAppIO g = interpret $ \case
   Say t
     | g ^. globalOnlyErrors -> return ()
     | otherwise -> embed (putStrLn t)
-  PrintJuvixError opts e -> do
-    printErr opts e
-  ExitJuvixError opts e -> do
-    printErr opts e
+  PrintJuvixError e -> do
+    printErr e
+  ExitJuvixError e -> do
+    printErr e
     embed exitFailure
   ExitMsg exitCode t -> embed (putStrLn t >> exitWith exitCode)
   Raw b -> embed (ByteString.putStr b)
   where
-    printErr opts e =
-      embed $ hPutStrLn stderr $ run $ runReader opts $ Error.render (not (g ^. globalNoColors)) (g ^. globalOnlyErrors) e
+    printErr e =
+      embed $ hPutStrLn stderr $ run $ runReader (genericFromGlobalOptions g) $ Error.render (not (g ^. globalNoColors)) (g ^. globalOnlyErrors) e
 
-runPipeline :: Member App r => GenericOptions -> Sem PipelineEff a -> Sem r a
-runPipeline opts p = do
+runPipeline :: Member App r => Sem PipelineEff a -> Sem r a
+runPipeline p = do
   r <- runPipelineEither p
   case r of
-    Left err -> exitJuvixError opts err
+    Left err -> exitJuvixError err
     Right res -> return res
 
 newline :: Member App r => Sem r ()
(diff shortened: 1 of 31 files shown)


src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping/Error/Pretty.hs
```diff
@@ -10,11 +10,11 @@ import Juvix.Data.PPOutput
...
 import Text.EditDistance
 
-ppCode :: Scoped.PrettyCode c => c -> Doc Ann
-ppCode = runPP . Scoped.ppCode
+ppCode :: Scoped.PrettyCode c => Scoped.Options -> c -> Doc Ann
+ppCode opts = runPP opts . Scoped.ppCode
 
-runPP :: Sem '[Reader Scoped.Options] (Doc Scoped.Ann) -> Doc Ann
-runPP = code . run . runReader Scoped.defaultOptions
+runPP :: Scoped.Options -> Sem '[Reader Scoped.Options] (Doc Scoped.Ann) -> Doc Ann
```
jonaprieto, line 16: This should be removed. You have the Scoped.Options in the reader effect.

app/Commands/Dev/Core.hs
```diff
@@ -11,8 +11,9 @@ newtype CoreReplOptions = CoreReplOptions
   { _coreReplShowDeBruijn :: Bool
   }
 
-newtype CoreEvalOptions = CoreEvalOptions
-  { _coreEvalNoIO :: Bool
+data CoreEvalOptions = CoreEvalOptions
+  { _coreEvalShowDeBruijn :: Bool,
+    _coreEvalNoIO :: Bool
```
jonaprieto, line 16: The changes in this file are not related to this PR. In the future, please separate them in a different PR.

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping/Error/Pretty.hs
lukaszcz: The whole point of the runPP function is that runs the effect, so it needs concrete Scoped.Options to do this. Earlier it just used Scoped.defaultOptions

app/Commands/Dev/Core.hs
lukaszcz: Yes, you're right. I should've made a small PR to add `--show-de-bruijn` to `core eval`.

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping/Error/Pretty.hs
jonaprieto: Oh, that's right.

jonaprieto (approved):

app/App.hs
```diff
@@ -10,8 +10,8 @@ import System.Console.ANSI qualified as Ansi
 
 data App m a where
   ExitMsg :: ExitCode -> Text -> App m ()
-  ExitJuvixError :: JuvixError -> App m a
-  PrintJuvixError :: JuvixError -> App m ()
+  ExitJuvixError :: GenericOptions -> JuvixError -> App m a
```
janmasrovira, line 13: The GenericOptions argument is not needed

janmasrovira (changes requested):

app/App.hs
lukaszcz: Ah, I didn't know that. I'll change this.

app/Commands/Dev/Core.hs
lukaszcz: Since I'm changing things, it won't be that much of a hassle to separate it into another PR. I'll try doing this.

lukaszcz: Done

app/App.hs
lukaszcz: Done

janmasrovira (approved):
Looks good 🚢

jonaprieto (approved):

PR #1487, opened 2022-08-30
Fix developBeta in Core/Extra.hs

lukaszcz:
# Description

Fix a mistake in the example "developBeta" function from Core/Extra.hs. This function is not currently used anywhere and I didn't notice that I forgot to decrease free variable indices in the lambda body.


paulcadman (approved):
👍

PR #1489, opened 2022-08-30
Restructure recursors and add some lens interfaces

janmasrovira:
This PR restructures the recursors in two categories: Applicative and Monadic.
In Applicative we find folds and maps to the leaves. In Monadic we find maps (top-down and bottom-up).
The advantage of having Applicative folds is that they can play well with some lens interface (e.g. `Traversal`, `SimpleFold`).
As an illustrative example, this PR implements a `SimpleFold` over free variables inside a node, and a `Traversal` over all the identifiers. One can then easily use the existing tools from the lens ecosystem to define a function to determine whether a node is closed, as shown in this PR.

Having a lens interface is not always possible, however, I think that for the cases where it is, we should try to have it.

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Recursors/Collector.hs
+++ b/src/Juvix/Compiler/Core/Extra/Recursors/Collector.hs
@@ -1,7 +1,5 @@
 module Juvix.Compiler.Core.Extra.Recursors.Collector where
 
--- import Juvix.Compiler.Core.Extra.Recursors.Base
-
 import Juvix.Compiler.Core.Data.BinderList (BinderList)
 import Juvix.Compiler.Core.Data.BinderList qualified as BL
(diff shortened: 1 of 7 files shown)


src/Juvix/Compiler/Core/Extra/Recursors/Collector.hs
```diff
@@ -0,0 +1,29 @@
+module Juvix.Compiler.Core.Extra.Recursors.Collector where
+
+-- import Juvix.Compiler.Core.Extra.Recursors.Base
```
lukaszcz, line 3: What is this comment for?

janmasrovira: was unnecessary. I have removed it

lukaszcz (approved):

PR #1492, opened 2022-08-30
Add --allow-different-user to workflow stack command

paulcadman:
Adds the `--allow-different-user` to GitHub workflow stack command.

I've tested this in my [fork of juvix](https://github.com/paulcadman/juvix/blob/6b54fd25df17f79053d788c5f6b5a383d595d127/.github/workflows/linux-static-binary.yaml#L35)  and it successfully builds and uploads the resulting linux binary as a build artefact, see https://github.com/paulcadman/juvix/actions/runs/2956659468


janmasrovira (approved):
:crossed_fingers:

PR #1494, opened 2022-08-30
Implement lambda lifting

### LINKED ISSUE #1477: JuvixCore transformation: Lambda-lifting
Depends on PR #1421

Notes
---------------------
* Lambda-lifting a single node can be performed with dmapM and the InfoTableBuilder effect.
* With the current Node data structure where only explicit lambdas need to be lifted, a "straightforward" bottom-up lambda-lifting transformation should be possible in O(n k) where:
  - n is the size of the Node (in practice, n should be treated as proportional to input program size),
  - k is the maximum number of distinct free variables in any given subnode, equal to the maximum number of binders on any path in the program tree (in practice, k may be treated as a non-negligible constant; k is roughly the maximum number of lexically available variables at any given point in the source program -- this is in practice much smaller than the total number of variables and usually bounded by a constant)
* Each lambda-expression `\y1 .. \yn A` should be replaced with `f x1 .. xm` where:
  - $FV(\lambda y_1 ... \lambda y_n A) = \\{x_1, ..., x_m\\}$,
  - `f` is a fresh function symbol,
  - `f := \x1 ... \xm \y1 .. \yn A'` is added to the IdentContext in the InfoTable, with `A'` the result of recursively lambda-lifting `A`.
* We need to avoid the situation where in `\x \y A` first `\y A` is lifted out separately. So we need to go top-down (dmapM), use `unfoldLambdas` from `Core/Extra/Base.hs` when noticing a lambda, and manually recursively call lambda-lifting on the body `A` which does not begin with a lambda-abstraction anymore.
* Literature:
  - [Danvy & Schultz, Lambda-lifting in quadratic time](https://scholar.lib.vt.edu/ejournals/JFLP/jflp-mirror/articles/2004/A2004-01/JFLP-A2004-01.pdf)
  - [Peyton Jones & Lester, Implementing functional languages: a tutorial, Chapter 6: Lambda-lifting](https://www.researchgate.net/profile/David-Lester-4/publication/228377970_Implementing_functional_languages_a_tutorial/links/02bfe51041cede4142000000/Implementing-functional-languages-a-tutorial.pdf)
  - Peyton Jones, The implementation of functional programming languages (1987 book), Chapters 13, 14
  - [Haskell Wiki: Lambda-lifting](https://wiki.haskell.org/Lambda_lifting)

Starting point
------------------
* Core/Transformation/Base.hs defines the transformation framework.
* Core/Data/InfoTableBuilder.hs may be needed to modify the InfoTable.
* Core/Info/FreeVarsInfo.hs computes free variables for each subnode in O(n k) total time (untested).
* Core/Transformation/LambdaLifting.hs contains a stub for the lambda-lifting transformation.

Justification
------------
The problem with closure conversion is that it turns every local function into a heap-allocated closure with a heap-allocated environment. For functions that can easily be compiled to use the stack, one should compile them to use the stack.

Note that because of partial application, we need closures even with lambda-lifting. But we do not want to be dynamically allocating closures on the heap if we don't need to. There are typically much fewer closure-allocating functions in an eager language than in a lazy one.

Once we get to more sophisticated optimisations, lambda-lifting probably should be done selectively. But I think it's a reasonable default. It turns out well for functions that can be compiled to use the stack, but may result in slower code for ones that need to allocate closures anyway. See: [Peyton Jones & Graf, Selective Lambda Lifting](https://pp.ipd.kit.edu/uploads/publikationen/graf19sll.pdf).

janmasrovira:
Closes #1477.

Implement lambda lifting for non nodes without let rec.

### DIFF
--- a/src/Juvix/Compiler/Core/Extra.hs
+++ b/src/Juvix/Compiler/Core/Extra.hs
@@ -30,7 +30,7 @@ freeVars f = ufoldNA reassemble go
   where
     go k = \case
       NVar var@(Var _ idx)
-        | idx >= k -> NVar <$> f var
+        | idx >= k -> NVar <$> f (shiftVar (-k) var)
       n -> pure n
 
 getIdents :: Node -> HashSet Ident
@@ -85,13 +85,13 @@ captureFreeVars fv
     s :: HashMap Index Index
     s = HashMap.fromList (zip indices [0 ..])
     mapFreeVars :: Node -> Node
-    mapFreeVars = dmapNR go
+    mapFreeVars = dmapN go
       where
-        go :: Index -> Node -> Recur
+        go :: Index -> Node -> Node
         go k = \case
           NVar (Var i u)
-            | Just v <- s ^. at (u - k) -> End (NVar (Var i (v + k)))
-          m -> Recur m
+            | Just v <- s ^. at (u - k) -> NVar (Var i (v + k))
+          m -> m
 
 -- | substitute a term t for the free variable with de Bruijn index 0, avoiding
 -- variable capture; shifts all free variabes with de Bruijn index > 0 by -1 (as
--- a/src/Juvix/Compiler/Core/Extra.hs
+++ b/src/Juvix/Compiler/Core/Extra.hs
@@ -9,8 +9,8 @@ module Juvix.Compiler.Core.Extra
   )
 where
 
-import Data.HashSet qualified as HashSet
 import Data.HashMap.Strict qualified as HashMap
+import Data.HashSet qualified as HashSet
 import Juvix.Compiler.Core.Extra.Base
 import Juvix.Compiler.Core.Extra.Equality
 import Juvix.Compiler.Core.Extra.Info
@@ -50,20 +50,24 @@ countFreeVarOccurrences idx = gatherN go 0
       NVar (Var _ idx') | idx' == idx + k -> acc + 1
       _ -> acc
 
+shiftVar :: Index -> Var -> Var
+shiftVar m = over varIndex (+ m)
+
 -- | increase all free variable indices by a given value
 shift :: Index -> Node -> Node
 shift 0 = id
 shift m = umapN go
   where
-    go k n = case n of
-      NVar (Var i idx) | idx >= k -> mkVar i (idx + m)
-      _ -> n
+    go k = \case
+      NVar v
+        | v ^. varIndex >= k -> NVar (shiftVar m v)
+      n -> n
 
 -- | Prism for NLam
 _NLam :: SimpleFold Node Lambda
 _NLam f = \case
-    NLam l -> NLam <$> f l
-    n -> pure n
+  NLam l -> NLam <$> f l
+  n -> pure n
 
 -- | Fold over all of the transitive descendants of a Node, including itself.
 cosmos :: SimpleFold Node Node
--- a/src/Juvix/Compiler/Core/Extra.hs
+++ b/src/Juvix/Compiler/Core/Extra.hs
@@ -10,6 +10,7 @@ module Juvix.Compiler.Core.Extra
 where
 
 import Data.HashSet qualified as HashSet
+import Data.HashMap.Strict qualified as HashMap
 import Juvix.Compiler.Core.Extra.Base
 import Juvix.Compiler.Core.Extra.Equality
 import Juvix.Compiler.Core.Extra.Info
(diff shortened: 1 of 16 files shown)


src/Juvix/Compiler/Core/Extra.hs
```diff
@@ -53,6 +56,24 @@ shift m = umapN go
...
+    s = HashMap.fromList (zip indices [0 ..])
+    mapFreeVars :: Node -> Node
+    mapFreeVars = dmapN go
+      where
+        go :: Index -> Node -> Maybe Node
+        go k = \case
+          Var i u
+            | Just v <- s ^. at (u - k) -> Just (Var i (v + k))
+          _ -> Nothing
+
```
lukaszcz, line 76: I don't get the point of all the conversions going on here. First, when invoking `captureFreeVars` in `lambdaLiftNode` we convert BinderList into a HashMap by first converting it to a list of pairs, and then in `captureFreeVars` we immediately convert this HashMap to a list of pairs, and then take the indices and convert them to a HashMap. Couldn't this be streamlined somehow, reducing the number of conversions? If I understand correctly, `indices` is the same as `freevars` in `lambdaLiftNode` modulo order. Infos for them can be fetched directly from BinderList, e.g. by `map (\idx -> BL.lookup idx bi) freevars`

lukaszcz:
Please add one or two tests for this. Lambda-lifting will be an essential part of the compilation pipeline, so it's good to have some tests for it in the test suite.

lukaszcz:
I think we ended up with O(n^2) after all because of the adjusting of indices in the body. I forgot one needs to do this adjustment when removing some lambdas. Let's leave it for now -- this won't bite us until we have much larger programs with nested local definitions.

Ultimately, lambda-lifting should probably be separated from a general optimisation removing unused function arguments. But we can keep this together until we start working on optimisations in earnest.

lukaszcz (changes requested):
Please add one or two tests for this. Lambda-lifting will be an essential part of the compilation pipeline, so it's good to have some tests for it in the test suite.

lukaszcz:
> I think we ended up with O(n^2) after all because of the adjusting of indices in the body. I forgot one needs to do this adjustment when removing some lambdas. Let's leave it for now -- this won't bite us until we have much larger programs with nested local definitions.
> 
> Ultimately, lambda-lifting should probably be separated from a general optimisation removing unused function arguments. But we can keep this together until we start working on optimisations in earnest.

No, actually it's fine. It's O(nk) because we first lambda-lift recursively, and only walk through the already lambda-lifted body to adjust the indices, so we won't be walking through it again once we lift it out.

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: The point was to guarantee from the type signature that there are no repeated indices in the argument of captureFreeVars. But probably that's not the most reasonable here. I have updated the code to use a list of tuples instead

janmasrovira:
> Please add one or two tests for this. Lambda-lifting will be an essential part of the compilation pipeline, so it's good to have some tests for it in the test suite.

I added three simple tests

src/Juvix/Compiler/Core/Transformation/Eta.hs
```diff
@@ -33,13 +33,13 @@ etaExpandConstrs argsNum = umap go
...
   where
-    go :: Node -> Node
+    go :: Node -> Maybe Node
     go n =
       let (l, args) = unfoldApp n
        in case l of
-            Constr i tag args' -> Constr i tag (args' ++ args)
-            BuiltinApp i op args' -> BuiltinApp i op (args' ++ args)
-            _ -> n
+            Constr i tag args' -> Just (Constr i tag (map squashApps (args' ++ args)))
```
lukaszcz, line 40: This is wrong, because it stops on the first Constr it sees top-down. It should recursively go into the arguments of Constr not to miss some Constr there.

Please restore the old dmap and its variants, and name the new ones dmap', dmapM', etc.

lukaszcz (changes requested):
I think old dmap variants needs to be restored. See the comment around squashApps.

src/Juvix/Compiler/Core/Transformation/Eta.hs
lukaszcz: Oh, sorry I didn't notice the recursive invocation of squashApps

lukaszcz: But I still think old dmap variants should be restored. The point is that the *N and *B variants manage the binder information for you, but you need to do this manually when you do the recursive call yourself.

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
```diff
@@ -4,13 +4,36 @@ module Juvix.Compiler.Core.Transformation.LambdaLifting
...
+    go :: BinderList Info -> Node -> Sem r Recur
+    go bl = \case
+      l@Lambda {} -> do
+        l' <- lambdaLiftNode bl l
+        let freevars = toList (getFreeVars l')
+            freevarsAssocs = [(i, BL.lookup i bl) | i <- freevars]
+            fBody' = captureFreeVars freevarsAssocs l'
+        f <- freshSymbol
+        traceM "register"
+        registerIdentNode f fBody'
```
lukaszcz, line 29: I forgot, but we should also be creating an IdentifierInfo for each new function. This can be done with registerIdent from InfoTableBuilder effect. Probably we should change the name fields in IdentifierInfo, InductiveInfo, etc., to Maybe types. The `Type` fields should be filled in as faithfully as possible, but one can always use Dynamic so Maybe is never necessary.

lukaszcz (changes requested):

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
lukaszcz: Do you think it's better to add fetching the type information in a separate PR? We could also do it that way, but it's important to have it ultimately.

For now, one could safely assume that the Info in the BinderList for a given variable contains TypeInfo which gives the type of the variable (or use Dynamic if the TypeInfo is not there). So at this stage one could easily add type information for function arguments. For the result, it could be decided how to do it after the translation Internal -> Core is complete.

janmasrovira: Unless I'm missing something, that's what I'm doing in `typeFromArgs`

lukaszcz: Sorry, I somehow missed that. I didn't look too closely. Since you didn't reply to the original comment I assumed it wasn't fixed.

lukaszcz: I think it's fine now.

lukaszcz (dismissed):

lukaszcz (approved):

PR #1496, opened 2022-08-31
Remove Haskell support

jonaprieto:
This PR removes completely any support for Haskell.

PR #1497, opened 2022-08-31
Remove mono

jonaprieto:
Merge this PR after #1496.


paulcadman (commented):
We can probably remove the documentation on monomorphization as part of this?

```
./docs/org/notes/monomorphization.org
./docs/org/SUMMARY.org:30:  - [[./notes/monomorphization.md][Monomorphization]]
```

jonaprieto:
> We can probably remove the documentation on monomorphization as part of this?
> 
> ```
> ./docs/org/notes/monomorphization.org
> ./docs/org/SUMMARY.org:30:  - [[./notes/monomorphization.md][Monomorphization]]
> ```

What if we keep it for the record/wiki, and we only remove it from itSUMMARY?

paulcadman:
> > We can probably remove the documentation on monomorphization as part of this?
> > ```
> > ./docs/org/notes/monomorphization.org
> > ./docs/org/SUMMARY.org:30:  - [[./notes/monomorphization.md][Monomorphization]]
> > ```
> 
> What if we keep it for the record/wiki, and we only remove it from the SUMMARY?

Yes - I think that's fine. I suggest removing it from the summary so it doesn't seem like we have monomorphization in the current compiler but it seems OK to keep the contents.

PR #1503, opened 2022-09-01
Refactor BinderInfo

### LINKED ISSUE #1498: Refactor BinderInfo to separate its fields
The fields of BinderInfo are independent in the sense that each may be independently available. For example, when eta-expanding we might want to add info on binder type but there is no sensible name. If we parse JVC files we have the names but not the types. There might be more independent fields in the future.

The solution is to change BinderInfo to store another Info object which contains the infos (NameInfo, TypeInfo) associated with the binder.

lukaszcz:
# Description

Refactors BinderInfo to separate its fields.

Fixes #1498

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Recursors/Monadic.hs
+++ b/src/Juvix/Compiler/Core/Extra/Recursors/Monadic.hs
@@ -8,7 +8,7 @@ import Juvix.Compiler.Core.Extra.Recursors.Base
 umapG ::
   forall c m.
   Monad m =>
-  Collector (Int, Maybe [BinderInfo]) c ->
+  Collector (Int, [Info]) c ->
   (c -> Node -> m Node) ->
   Node ->
   m Node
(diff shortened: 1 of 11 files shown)


src/Juvix/Compiler/Core/Extra/Recursors/Monadic.hs
```diff
@@ -8,7 +8,7 @@ import Juvix.Compiler.Core.Extra.Recursors.Base
 umapG ::
   forall c m.
   Monad m =>
-  Collector (Int, Maybe [BinderInfo]) c ->
+  Collector (Int, [Info]) c ->
```
janmasrovira, line 11: why do we have `Info` instead of `BinderInfo`?

lukaszcz: Because now the binder's info is stored in another Info. I thought about making a type alias e.g. `type BinderInfo = Info`, but BinderInfo is already used as the wrapper and I didn't come up with a good name. Or one can return the wrapper, but then one needs to unwrap it every time.

janmasrovira: What's the purpose/meaning of the wrapper `BinderInfo`?

lukaszcz: It's technical. You can't just store something in an Info without wrapping it. It has to have a unique type whose representation is used as a key.

lukaszcz: I think now it's nice because I got rid of Maybe. If there's no BinderInfo then we'll just get Info.empty

janmasrovira (approved):

PR #1507, opened 2022-09-02
LetRec in Core

lukaszcz:
Add blocks of local mutually recursive definitions (LetRec) to Core.


src/Juvix/Compiler/Core/Evaluator.hs
```diff
@@ -65,6 +66,10 @@ eval !ctx !env0 = convertRuntimeNodes . eval' env0
       NCtr (Constr i tag args) -> mkConstr i tag (map' (eval' env) args)
       NLam l@Lambda {} -> Closure env l
       NLet (Let _ v b) -> let !v' = eval' env v in eval' (v' : env) b
+      NRec (LetRec _ vs b) ->
+        let !vs' = map (eval' env') (toList vs)
+            !env' = revAppend vs' env
+         in foldr GHC.seq (eval' env' b) vs'
```
paulcadman, line 72: Does this code rely on the evaluation order of `GHC.seq` ? The docs say that:

```
the expression seq a b does not guarantee that a will be evaluated before b
```

 https://hackage.haskell.org/package/ghc-prim-0.9.0/docs/GHC-Prim.html#v:seq

I'm not sure, but I thought I'd point this out just in case.

lukaszcz: Thanks for pointing this out! Indeed, I thought `seq` specifies the evaluation order. I'll switch to `pseq`.

lukaszcz: The evaluation order for function/constructor arguments and letrec blocks should actually be left unspecified in the language semantics, because it's a pain to preserve it with optimisations. However, it's convenient to have a well-defined order in "debug mode" when optimisations are turned off. That's why I'm trying to make it always deterministic in the evaluator.

paulcadman: minor: Maybe we could call `pseq` directly from `GHC.Conc` which is in `base` https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Conc.html#v:pseq to avoid adding the dependency on the `parallel` package?

The `pseq` in the `parallel` package appears to call `pseq` in `GHC.Conc`: https://hackage.haskell.org/package/parallel-3.2.2.0/docs/src/Control.Parallel.html#pseq

lukaszcz: Good idea! I'll change it.

paulcadman (approved):

PR #1513, opened 2022-09-05
Eager evaluation of Constr arguments

lukaszcz:
# Description

Constructor arguments were evaluated lazily by mistake. The PR fixes this, switching to the indended eager evaluation.


src/Juvix/Prelude/Base.hs
```diff
@@ -223,8 +223,15 @@ tableNestedInsert k1 k2 = tableInsert (HashMap.singleton k2) (HashMap.insert k2)
 --------------------------------------------------------------------------------
 
 revAppend :: [a] -> [a] -> [a]
-revAppend [] ys = ys
-revAppend (x : xs) ys = revAppend xs (x : ys)
+revAppend [] !ys = ys
+revAppend (x : xs) !ys = revAppend xs (x : ys)
+
+map' :: (a -> b) -> [a] -> [b]
```
paulcadman, line 235: minor: Rename to `strictMap` ?

src/Juvix/Prelude/Base.hs
```diff
@@ -223,8 +223,15 @@ tableNestedInsert k1 k2 = tableInsert (HashMap.singleton k2) (HashMap.insert k2)
...
+map' _ [] = []
+map' f (h : t) =
+  let !v = f h
+   in let !vs = map' f t
```
paulcadman, line 239: minor: combine lets:

```
let !v = f h
    !vs = map' f t
in v : vs
```

paulcadman (approved):
I made a couple of minor suggestions, feel free to ignore.

src/Juvix/Prelude/Base.hs
lukaszcz: No! This changes the semantics. GHC is free to first evaluate `vs` and then `v` if you combine them (and this is actually what happens with my GHC).

paulcadman: oh I didn't know that - thanks 👍

paulcadman: Good idea ! Someone could subtly refactor this

paulcadman (approved):

PR #1514, opened 2022-09-05
Back recursor types with type families

janmasrovira:
In this PR we explore the use of type families to improve the reliability of all the map variations for the Node type.

After discussing the topic we agreed that we would keep type families as part of the implementation but we would also expose a list of functions with concrete types and mnemonic names.

Also, I have renamed `Monadic` -> `Map` and `Applicative` -> `Fold`. I think it makes more sense like this.


src/Juvix/Compiler/Core/Extra/Recursors/Map/Named.hs
```diff
@@ -0,0 +1,95 @@
+module Juvix.Compiler.Core.Extra.Recursors.Map.Named where
+
+import Juvix.Compiler.Core.Extra.Recursors.Base
+import Juvix.Compiler.Core.Extra.Recursors.Map
+import Juvix.Compiler.Core.Extra.Recursors.Parameters
+
+dmapLRM :: Monad m => (BinderList Info -> Node -> m Recur) -> Node -> m Node
```
lukaszcz, line 7: I see you changed *B to *L and changed the order, so now M is always at the end if present. That's fine, but the suffixes in Fold/Named.hs should be made consistent with this convention.

src/Juvix/Compiler/Core/Extra/Recursors/Fold/Named.hs
```diff
@@ -0,0 +1,41 @@
+module Juvix.Compiler.Core.Extra.Recursors.Fold.Named where
+
+import Data.Functor.Identity
+import Juvix.Compiler.Core.Extra.Recursors.Base
+import Juvix.Compiler.Core.Extra.Recursors.Fold
+
+ufoldA :: Applicative f => (a -> [a] -> a) -> (Node -> f a) -> Node -> f a
+ufoldA uplus f = ufoldG unitCollector uplus (const f)
+
+ufoldAB :: Applicative f => (a -> [a] -> a) -> (BinderList Info -> Node -> f a) -> Node -> f a
```
lukaszcz, line 10: Should be ufoldLA to make it consistent with the convention for maps. Analogously with all other names in this module.

lukaszcz (changes requested):

lukaszcz (dismissed):

lukaszcz (approved):

PR #1515, opened 2022-09-06
Implement some instances for BinderList

janmasrovira:
Having these instances is convenient

### DIFF
--- a/src/Juvix/Compiler/Core/Data/BinderList.hs
+++ b/src/Juvix/Compiler/Core/Data/BinderList.hs
@@ -1,7 +1,6 @@
 module Juvix.Compiler.Core.Data.BinderList where
 
 import Data.HashMap.Strict qualified as HashMap
-import Data.List qualified as List
 import Juvix.Compiler.Core.Language.Base
 
 data BinderList a = BinderList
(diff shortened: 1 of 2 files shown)


src/Juvix/Compiler/Core/Data/BinderList.hs
```diff
@@ -12,34 +11,59 @@ data BinderList a = BinderList
...
-length :: BinderList a -> Int
-length = (^. blLength)
+  length :: BinderList a -> Int
+  length = (^. blLength)
+
+  toList :: BinderList a -> [a]
+  toList bl =
+    map snd $
+      sortBy (compare `on` fst) $
+        HashMap.toList (bl ^. blMap)
```
lukaszcz, line 30: I don't understand the extra indentation here. `null`, `length` and `toList` are not in the Foldable instance anymore?

janmasrovira: they are. Maybe the diff is confusing. I think it may be easier to look at the file directly

lukaszcz: Okay then, I thought only foldr is in the instance. Resolve the conflicts and I'll approve and merge.

lukaszcz (approved):

PR #1517, opened 2022-09-06
Add `internal core read` command

janmasrovira:
This PR adds the `internal core read` command. This command allows the user to read a jvc file, apply some transformations on it and print the result.


src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
```diff
@@ -4,13 +4,11 @@ module Juvix.Compiler.Core.Transformation.LambdaLifting
   )
 where
 
-import Juvix.Compiler.Core.Data.InfoTableBuilder
+import Juvix.Compiler.Core.Pretty
```
lukaszcz, line 7: Minor: we don't need pretty printing here, beyond debugging.

lukaszcz (approved):

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
janmasrovira: I wonder why ghc didn't complain about that... Anyway, it is fixed now

lukaszcz (approved):

PR #1518, opened 2022-09-06
Stripped version of Core Node datatype

lukaszcz:
The Stripped.Node datatype specifies precisely the format of Nodes needed for compilation to JuvixAsm.

Ultimately, I didn't use type families for two reasons.
1. Most importantly, they don't work with the UNPACK pragma. The compiler gives an error.
2. They don't seem to be very useful in this particular case. I implemented pretty printing for Stripped.Node, and anything that could be done with type families could also be done in a simpler way with parametric polymorphism by abstracting one or two extra arguments to a printing function. Beyond pretty printing, we won't be having many more general functions that need to work both on Node and Stripped.Node.

The reason why type families didn't turn out very useful might be that Node and Stripped.Node have different constructors, so the top-level case needs to be handled separately anyway.

### DIFF
--- a/src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
+++ b/src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
@@ -5,16 +5,16 @@ import Juvix.Compiler.Core.Language.Stripped
 data InfoTable = InfoTable
   { _infoMain :: Maybe Symbol,
     _infoFunctions :: HashMap Symbol FunctionInfo,
-    _infoInductives :: HashMap Name InductiveInfo,
+    _infoInductives :: HashMap Symbol InductiveInfo,
     _infoConstructors :: HashMap Tag ConstructorInfo
   }
 
 data FunctionInfo = FunctionInfo
   { _functionName :: Maybe Name,
     _functionSymbol :: Symbol,
-    -- _functionNode has `_functionArgsNum` free variables corresponding to the
+    -- _functionBody has `_functionArgsNum` free variables corresponding to the
     -- function arguments
-    _functionNode :: Node,
+    _functionBody :: Node,
     _functionType :: Type,
     -- a function can have 0 arguments
     _functionArgsNum :: Int,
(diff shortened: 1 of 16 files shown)


src/Juvix/Compiler/Core/Pretty/Base.hs
```diff
@@ -183,6 +224,28 @@ instance PrettyCode Node where
     Closure env l@Lambda {} ->
       ppCode (substEnv env (NLam l))
 
+instance PrettyCode Stripped.Node where
```
janmasrovira, line 227: can we have an embedding from `Stripped.Node` to `Node` and then just pretty print `Node`?

src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
```diff
@@ -0,0 +1,55 @@
...
+  { _argumentName :: Maybe Name,
+    _argumentType :: Type
+  }
+
+data InductiveInfo = InductiveInfo
+  { _inductiveName :: Name,
+    _inductiveKind :: Type,
+    _inductiveConstructors :: [ConstructorInfo],
+    _inductiveParams :: [ParameterInfo],
+    _inductivePositive :: Bool
```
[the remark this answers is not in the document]
lukaszcz: Probably not. I'll remove it. At this point, however, it's not entirely clear whether some fields will be needed or not. But we can add fields later if we see we need them.

src/Juvix/Compiler/Core/Pretty/Base.hs
lukaszcz: The issue with converting the InfoTable back could be that you forget some information and need to provide some defaults which might not make much sense.

janmasrovira: I see, then it makes more sense to leave it as it is

src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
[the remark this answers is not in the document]
lukaszcz: Done. I also changed the InfoTables to index Inductives by Symbols. This makes more sense now that we have TypeConstr as a Node.

janmasrovira (approved):

PR #1523, opened 2022-09-07
Disallow tab characters as spaces

### LINKED ISSUE #1464: VSCode syntax highlighting uses different colors within the same token
The attached screenshot shows example code which induces the Juvix VSCode extension to use different colors for different parts of some of the same words.

janmasrovira:
The use of tab characters as spaces would be too hard to properly handle in the current implementation of semantic highlighting for both emacs and vscode. For this reason, we have decided to forbid tab characters. This PR implements this change.

Closes #1464


jonaprieto (approved):
:) That was quick!

PR #1526, opened 2022-09-09
Add CanonicalProjection

janmasrovira:
This pr adds this class:
```
class CanonicalProjection a b where
  project :: a -> b
```
It is meant to be used for records types that have an obvious projection into another record. E.g.
```
instance CanonicalProjection GlobalOptions Internal.Options where
  project g =
    Internal.Options
      { Internal._optShowNameIds = g ^. globalShowNameIds
      }
```

I have updated the pretty printing functions that expect options to accept anything that has a canonical projection into the options.

The benefit is that the callsites can usually be simplified since you do not need to do an adhoc projection or remember the name of the projection if it exists.

```diff
-                  let ppOpts =
-                        Internal.defaultOptions
-                          { Internal._optShowNameIds = globalOpts ^. globalShowNameIds
-                          }
-                  App.renderStdOut (Internal.ppOut ppOpts micro)
+                  App.renderStdOut (Internal.ppOut globalOpts micro)
```

We have discussed this with @paulcadman and we have a small preference to have this. It'd be nice to know the opinion of @lukaszcz

### DIFF
--- a/app/Commands/Dev/Core.hs
+++ b/app/Commands/Dev/Core.hs
@@ -1,6 +1,7 @@
 module Commands.Dev.Core where
 
 import Juvix.Compiler.Core.Data.TransformationId.Parser
+import Juvix.Compiler.Core.Pretty.Options qualified as Core
 import Juvix.Prelude hiding (Doc)
 import Options.Applicative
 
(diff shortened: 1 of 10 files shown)


lukaszcz:
I'm in favour of implicit coercions, or things that approximate them! 

Unrelated: I think we should have user-declarable implicit coercions in Juvix at some point (e.g. Nat -> Integer). Without implicit coercions in Coq things like MathComp, or any formalisation with sophisticated type structure, would be totally unreadable.

paulcadman (approved):

PR #1527, opened 2022-09-09
Refactor CLI

janmasrovira:
Maintenance refactor of the `app/` subtree.

Changes include:
1. commands are separated into modules to make it less cluttered.
1. Input files are no longer a global option, so completion and help work as intended.
2. Completion for input juvix files only suggest `.juvix` files.
3. Global flags *must* be given before any subcommand.
4. Now `--output` and `--output-dir` work with respect to where `juvix` is called and not the root of the project.
5. `internal` subcommand renamed to `dev`. It now shows in the `utility commands` help section instead of the `compiler commands` section.


lukaszcz (dismissed):

paulcadman (approved):
Excellent - autocompletion of input filenames works again 🥳

PR #1547, opened 2022-09-23
Evaluator minor style refactor

janmasrovira:
I've removed some code duplication. I assume that GHC will automatically inline the auxilliary definitions so performance should not be affected


lukaszcz (approved):
I checked the runtime on benchmarks. It's fine after adding the INLINE pragmas.

PR #1565, opened 2022-09-29
refactor wildcard cases

janmasrovira:
this pr removes the catchall wildcard in two functions. It still uses wildcards, but is safer in the sense that if we add a new type it will generate an exhaustiveness warning


lukaszcz (approved):
Okay, we can do it this way. My point is that with the increasing number of cases / multiple patterns you get diminishing returns. Yes, the compiler would catch a mistake if you forget to add a case, so you save time here, but you need to duplicate all the cases for each constructor, so you lose time there and make the code uglier.

PR #1566, opened 2022-09-29
Keywords refactor

janmasrovira:
The goal of this PR is to refactor lexers (Concrete/Core/Asm) so that they can share the functionality related to keywords.

Also, it removes the use of `notFollowedBy` in identifier/keyword parsing


paulcadman (approved):
🚢

PR #1571, opened 2022-10-03
Eta expansion at the top of each core function definition (#1481)

### LINKED ISSUE #1481: JuvixCore transformation: Eta-expansion at the top of each function definition
Depends on PR #1421

Notes
------
* Eta-expand only at the top of the Nodes stored in InfoTable/IdentContext.
* Use type information associated with function definitions to determine eta-expansion.

Starting point
------------------
* Core/Transformation/Base.hs defines the transformation framework.
* Core/Transformation/Eta.hs already contains eta-expansion for constructors and builtins.

janmasrovira:
Closes #1481.


lukaszcz (changes requested):
Doesn't work when multiple binders need to be created. I added a test.

PR #1572, opened 2022-10-03
1569 rewrite the test for lambda lifting to use evaluation

janmasrovira:
changes in this pr:
- To test lambda lifting we run the transformation and then the evaluator, reusing the existing code.
- I introduced the identity transformation, used to test mapT'.
- Use the axiom color instead of keyword for the core builtins.
- unfoldLambdas has been changed to conform to mkApps. I added unfoldLambdasRev that preserves the old behaviour.
- Some minor style changes.


src/Juvix/Compiler/Core/Translation/FromSource.hs
```diff
@@ -463,7 +463,7 @@ atoms ::
   ParsecS r Node
 atoms varsNum vars = do
   es <- P.some (atom varsNum vars)
-  return $ mkApps' (List.head es) (List.tail es)
+  return $ mkApps' (List.head es) (reverse (List.tail es))
```
lukaszcz, line 466: Now I'm confused. `P.some (atom varsNum vars)` parses `f t1 .. tn` into a list `[f, t1, .., tn]` (confusing expressions with corresponding nodes) so shouldn't we be calling `mkApps' f [t1, .., tn]`, without the reverse? Why did it work before?

test/Core/Eval/Base.hs
```diff
@@ -10,21 +10,30 @@ import Juvix.Compiler.Core.Info qualified as Info
...
+  (String -> IO ()) ->
+  Assertion
+coreEvalAssertion mainFile expectedFile trans testTrans step = do
   step "Parse"
   r <- parseFile mainFile
   case r of
     Left err -> assertFailure (show (pretty err))
     Right (_, Nothing) -> do
-      step "Compare expected and actual program output"
+      step "Compare expected an actual program output"
```
lukaszcz, line 31: Why the change? "and" was fine, now it's not grammatical

src/Juvix/Compiler/Core/Extra/Base.hs
```diff
@@ -128,10 +128,10 @@ unfoldType' ty = case ty of
 {- functions on Node -}
 
 mkApps :: Node -> [(Info, Node)] -> Node
-mkApps = foldl' (\acc (i, n) -> mkApp i acc n)
+mkApps m = foldl' (\acc (i, n) -> mkApp i acc n) m . reverse
```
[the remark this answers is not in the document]
lukaszcz: BTW, why was this wrong? 
`foldl App f [t1, t2, t3] => App (App (App f t1) t2) t3`
which is what is wanted? Now you reverse the argument order, why?

src/Juvix/Compiler/Core/Extra/Base.hs
```diff
@@ -128,10 +128,10 @@ unfoldType' ty = case ty of
...
 
 mkApps' :: Node -> [Node] -> Node
-mkApps' = foldl' mkApp'
+mkApps' n = foldl' mkApp' n . reverse
 
 unfoldApps :: Node -> (Node, [(Info, Node)])
 unfoldApps = go []
```
lukaszcz, line 137: `mkApps` was correct. `unfoldApps` is "wrong" in that it behaves unintuitively: `unfoldApps (App (App f t1) t2) => (f, [t2, t1])`. This should be changed, not `mkApps`.

lukaszcz: No, wait, it's also fine, you get `(f, [t1, t2])` because you save the right branch first. What was exactly the problem you were complaining about? I don't see any unexpected behaviour in either `mkApps` or `unfoldApps`.

lukaszcz (changes requested):
I don't understand the alleged problem with mkApps / unfoldApps. They are fine and work in the expected way. I think your problem must be somewhere else.

src/Juvix/Compiler/Core/Extra.hs
```diff
@@ -116,7 +120,7 @@ developBeta = umap go
 
 etaExpand :: Int -> Node -> Node
 etaExpand 0 n = n
-etaExpand k n = mkLambdas' k (mkApps' (shift k n) (map mkVar' (reverse [0 .. k - 1])))
+etaExpand k n = mkLambdas' k (mkApps' (shift k n) (map mkVar' [0 .. k - 1]))
```
lukaszcz, line 123: The `reverse` was intensional here. You want to eta-expand `f` to `\x \y f x y` not to `\x \y f y x`

lukaszcz (changes requested):

src/Juvix/Compiler/Core/Extra/Base.hs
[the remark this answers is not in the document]
janmasrovira: this wasn't wrong. The change needs to be reverted

test/Core/Eval/Base.hs
janmasrovira: typo. Will fix it

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: this change was a consequence of the (incorrect) reverse in mkApps. Will fix it

src/Juvix/Compiler/Core/Extra/Base.hs
```diff
@@ -128,10 +128,10 @@ unfoldType' ty = case ty of
 {- functions on Node -}
 
 mkApps :: Node -> [(Info, Node)] -> Node
-mkApps = foldl' (\acc (i, n) -> mkApp i acc n)
+mkApps m = foldl' (\acc (i, n) -> mkApp i acc n) m . reverse
 
 mkApps' :: Node -> [Node] -> Node
-mkApps' = foldl' mkApp'
+mkApps' n = foldl' mkApp' n . reverse
```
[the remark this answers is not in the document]
janmasrovira: done

src/Juvix/Compiler/Core/Translation/FromSource.hs
janmasrovira: done

src/Juvix/Compiler/Core/Extra/Base.hs
janmasrovira: done

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: done

janmasrovira:
the source of all the confusion was in `captureFreeVars`. I was missing the reverse in the this line:
```
    s = HashMap.fromList (zip (reverse indices) [0 ..])
```

src/Juvix/Compiler/Core/Extra.hs
lukaszcz: Not done. Where is the reverse?

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
```diff
@@ -24,48 +21,45 @@ lambdaLiftNode aboveBl top =
     typeFromArgs :: [ArgumentInfo] -> Type
     typeFromArgs = \case
       [] -> mkDynamic' -- change this when we have type info about the body
-      (a : as) -> mkPi' argTy (typeFromArgs as)
-        where
-          argTy = fromMaybe mkDynamic' (a ^. argumentType)
+      (a : as) -> mkPi' (a ^. argumentType) (typeFromArgs as)
```
lukaszcz, line 24: Do you remember that Pi introduces a binder? It seems this may be incorrect with (type) variables occurring in types. The target of Pi may need to be shifted to adjust its context. This depends on how we create ArgumentInfo, i.e., what are the contexts of types in it. The initial idea was that it doesn't matter, because initially there were no explicit binders in types. Now we need to decide something, e.g., the context of the first argument is global (closed), the context of the second contains the first variable, etc., then the types in ArgumentInfo correspond to the argument types obtained from unfoldType. I think this is the current decision in JVC parser, but there the function is already top level so its type must be closed. With this decision you don't need shifting here unless the type you create ArgumentInfo from was not closed in the first place.

The proper way to do this is to close also under the free variables occurring in argument types, and recursively so. We may defer this, but it's an issue -- otherwise it won't work with polymorphic functions as polymorphism is explicit now.

lukaszcz (changes requested):

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: now it is

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
janmasrovira: Afaik currently recursors don't look at the types of the nodes, so `getFreeVars` does not return free type variables in the type of some node. This could be fixed by changing the behavriour of the recursor, right?
Once we have a function that returns all free variables (including those appearing in types), it should be enough to close the terms with the sorted variables.I.e. the only required change would be
`captureFreeVars freevarsAssocs l'`
`captureFreeVars (sortByIx freevarsAssocs) l'`
Does this make sense?

janmasrovira: p.s. I think we should make this change in a separate pr, so I propose that we merge this and then tackle this issue

lukaszcz: I think that would work for polymorphism, but not for dependent types. Consider:
```
\(A : Type) \(B : T A) (\x : B) f (\(y : B) h y x)
```
The result of lambda-lifting should be:
```
def lam := \(A : Type) \(B : T A) \(x : B) \(y : B) h y x

\(A : Type) \(B : T A) (\x : B) f (lam A B x)
```
But `A` does not occur anywhere in the expression being lifted. To make it work in general one needs to close recursively until there are no more free variables (but I think one could do a single closure the way you propose).

Yes, we can change it in a separate PR.

lukaszcz (approved):

PR #1574, opened 2022-10-04
Fix symbol numbering bug

lukaszcz:
This should fix the bug with disappearing definitions.

### DIFF
--- a/src/Juvix/Compiler/Core/Data/InfoTable.hs
+++ b/src/Juvix/Compiler/Core/Data/InfoTable.hs
@@ -13,8 +13,8 @@ data InfoTable = InfoTable
     _infoInductives :: HashMap Symbol InductiveInfo,
     _infoConstructors :: HashMap Tag ConstructorInfo,
     _infoAxioms :: HashMap Name AxiomInfo,
-    _infoSymbolsNum :: Word,
-    _infoTagsNum :: Word
+    _infoNextSymbol :: Word,
+    _infoNextTag :: Word
   }
 
 emptyInfoTable :: InfoTable
@@ -27,8 +27,8 @@ emptyInfoTable =
       _infoInductives = mempty,
       _infoConstructors = mempty,
       _infoAxioms = mempty,
-      _infoSymbolsNum = 0,
-      _infoTagsNum = 0
+      _infoNextSymbol = 0,
+      _infoNextTag = 0
     }
 
 data IdentKind = IdentSym Symbol | IdentTag Tag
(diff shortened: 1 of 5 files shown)


src/Juvix/Compiler/Core/Data/InfoTable.hs
```diff
@@ -12,7 +12,9 @@ data InfoTable = InfoTable
     _infoIdentifiers :: HashMap Symbol IdentifierInfo,
     _infoInductives :: HashMap Symbol InductiveInfo,
     _infoConstructors :: HashMap Tag ConstructorInfo,
-    _infoAxioms :: HashMap Name AxiomInfo
+    _infoAxioms :: HashMap Name AxiomInfo,
+    _infoSymbolsNum :: Word,
```
janmasrovira, line 16: I would prefer the name `infoNextSymbol` instead or something similar. It clarifies that this field is only used for fresh names and does never need to be decreased.
Same for tags.

src/Juvix/Compiler/Core/Data/InfoTableBuilder.hs
```diff
@@ -26,52 +26,35 @@ checkSymbolDefined sym = do
...
-        modify' (over stateInfoTable (over identMap (HashMap.insert (ci ^. (constructorName . nameText)) (IdentTag (ci ^. constructorTag)))))
+        modify' (over infoConstructors (HashMap.insert (ci ^. constructorTag) ci))
+        modify' (over identMap (HashMap.insert (ci ^. (constructorName . nameText)) (IdentTag (ci ^. constructorTag))))
       RegisterIdentNode sym node ->
-        modify' (over stateInfoTable (over identContext (HashMap.insert sym node)))
+        modify' (over identContext (HashMap.insert sym node))
       SetIdentArgsInfo sym argsInfo -> do
-        modify' (over stateInfoTable (over infoIdentifiers (HashMap.adjust (set identifierArgsInfo argsInfo) sym)))
-        modify' (over stateInfoTable (over infoIdentifiers (HashMap.adjust (set identifierArgsNum (length argsInfo)) sym)))
+        modify' (over infoIdentifiers (HashMap.adjust (set identifierArgsInfo argsInfo) sym))
```
janmasrovira, line 54: with lenses:
```
        modify' (set (infoIdentifiers . at sym . _Just . identifierArgsInfo) argsInfo)
        modify' (set (infoIdentifiers . at sym . _Just . identifierArgsNum) (length argsInfo))
```

src/Juvix/Compiler/Core/Data/InfoTable.hs
lukaszcz: Done

src/Juvix/Compiler/Core/Data/InfoTableBuilder.hs
lukaszcz: Okay.

janmasrovira (approved):

PR #1576, opened 2022-10-04
As-patterns

### LINKED ISSUE #1529: Syntax for naming subpatterns
Do we have a syntax for naming subpatterns? I don't think so. It's an extremely useful feature and I would advocate including it. 

Then we need a syntax for it, possibly a reserved symbol to indicate the name.

Some options:
```
f : {A : Type} -> List (List A) -> List A;
f (cons x@(cons _ _) nil) := x;
f (cons x#(cons _ _) nil) := x;
```

We don't want to enforce spaces around the "naming" symbol (related to issues #1528, #1495).

ii8:
This closes #1529

### DIFF
--- a/src/Juvix/Compiler/Concrete/Language.hs
+++ b/src/Juvix/Compiler/Concrete/Language.hs
@@ -285,6 +285,7 @@ instance HasFixity PatternPostfixApp where
 
 data PatternArg = PatternArg
   { _patternArgIsImplicit :: IsImplicit,
+    _patternArgName :: Maybe S.Symbol,
     _patternArgPattern :: Pattern
   }
   deriving stock (Show, Eq, Ord)
@@ -322,8 +323,8 @@ data PatternAtom (s :: Stage)
   = PatternAtomIden (PatternAtomIdenType s)
   | PatternAtomWildcard Wildcard
   | PatternAtomEmpty Interval
-  | PatternAtomParens (PatternParensType s)
-  | PatternAtomBraces (PatternParensType s)
+  | PatternAtomParens (Maybe (SymbolType s)) (PatternParensType s)
+  | PatternAtomBraces (Maybe (SymbolType s)) (PatternParensType s)
 
 data PatternAtoms (s :: Stage) = PatternAtoms
   { _patternAtoms :: NonEmpty (PatternAtom s),
@@ -981,7 +982,8 @@ deriving stock instance
     Show (IdentifierType s),
     Show (PatternAtomIdenType s),
     Show (PatternParensType s),
-    Show (PatternType s)
+    Show (PatternType s),
+    Show (SymbolType s)
   ) =>
   Show (PatternAtom s)
 
@@ -990,7 +992,8 @@ deriving stock instance
     Eq (IdentifierType s),
     Eq (PatternAtomIdenType s),
     Eq (PatternParensType s),
-    Eq (PatternType s)
+    Eq (PatternType s),
+    Eq (SymbolType s)
   ) =>
   Eq (PatternAtom s)
 
@@ -999,7 +1002,8 @@ deriving stock instance
     Ord (IdentifierType s),
     Ord (PatternAtomIdenType s),
     Ord (PatternParensType s),
-    Ord (PatternType s)
+    Ord (PatternType s),
+    Ord (SymbolType s)
   ) =>
   Ord (PatternAtom s)
 
@@ -1008,7 +1012,8 @@ deriving stock instance
     Show (IdentifierType s),
     Show (PatternAtomIdenType s),
     Show (PatternParensType s),
-    Show (PatternType s)
+    Show (PatternType s),
+    Show (SymbolType s)
   ) =>
   Show (PatternAtoms s)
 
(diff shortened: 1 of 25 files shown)


lukaszcz:
Thanks for submitting this! You can use `make format` to ensure that your code follows our formatting guidelines (so ormolu won't fail).

ii8:
Struggling to install ormolu on my musl system, fixed it manually in the meantime.

paulcadman:
> Struggling to install ormolu on my musl system, fixed it manually in the meantime.
> (nvm that wasn't it)

You should be able to run `stack install ormolu` from within the Juvix project to install it. If not I can try to push the fix to your branch.

ii8:
> `stack install ormolu`

This reports `Unable to find installation URLs for OS key: linux64-custom-musl-tinfo6`. I did it on a different machine now, I'll figure this out at some point.

paulcadman:
> > `stack install ormolu`
> 
> This reports `Unable to find installation URLs for OS key: linux64-custom-musl-tinfo6`. I did it on a different machine now, I'll figure this out at some point.

I don't think that `stack` works well with musl unfortunately. Please let us know if you need some help.

janmasrovira:
Thanks for submitting this pr @ii8.

I have briefly looked at the code and I'd like to suggest some changes.
Currently you've defined:
```
data PatternBind = PatternBind
  { _patBindName :: S.Symbol,
    _patBindPattern :: PatternArg
  }
```
and made it is possible for the user to write:
```
asas : Nat → Nat;
asas p@p'@_ := p + p';
```
I don't think there is a point in allowing multiple as-patterns on the same pattern. For this reason, I think it would be better to modify `PatternArg` as follows:
```
data PatternArg = PatternArg
  { _patternArgIsImplicit :: IsImplicit,
    _patternArgName :: Maybe S.Symbol,
    _patternArgPattern :: Pattern
  }
```

Also, the name for implicit patterns should be outside of the braces. I.e.
```
im' : {A : Type} → (A × Nat) → ({B : Type} → B → B) → A;
im' {i@A} p@(a, zero) f := (f {A} ∘ fst) p;
```
should become
```
im' : {A : Type} → (A × Nat) → ({B : Type} → B → B) → A;
im' i@{A} p@(a, zero) f := (f {A} ∘ fst) p;
```

ii8:
> I don't think there is a point in allowing multiple as-patterns on the same pattern. For this reason, I think it would be better to modify `PatternArg` as follows:

Yes, I have thought about doing it like that, in fact I had written the exact code you suggested. I did it like this in the end because both agda and haskell allow `a@b@c`.

> Also, the name for implicit patterns should be outside of the braces. I.e.

Same thing here, I followed agda which allows `a@{b}` and `{a@b}`.

I was indecisive myself on both of these, so I just picked the way agda does it and made the PR so I can get your opinion on it.

janmasrovira:
@ii8 In this particular case, we don't see enough reason to follow Haskell and Agda. We prefer to have it in the way that I proposed.

ii8:
There is now a pattern name in `PatternArg`. The `@` can now be used only before an opening parethesis or brace. I've also disallowed aliasing like `A@{B}` where `B` is a variable because it's equally pointless and it would complicate type checking.
With this in place it may be possible to eliminate `PatternVariable` completely from `Pattern` and instead desugar `a` to `a@_` (the latter can't be written in source but I think you know what I mean.)

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping/Error/Types.hs
```diff
@@ -659,6 +659,68 @@ instance ToGenericError DoubleBracesPattern where
...
+          opts' = fromGenericOptions opts
+          name = e ^. aliasBinderPatternName
+          implicit = e ^. aliasBinderPatternIsImplicit
+          pat = e ^. aliasBinderPatternArg
+          i = getLoc pat
+          wrap = case implicit of
+            Implicit -> "braces"
+            Explicit -> "parentheses"
+          msg =
+            "As patterns require a type constructor in the"
```
janmasrovira, line 720: I would use
`As-patterns require a constructor inside the`

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -1151,11 +1151,11 @@ checkPatternAtom ::
   PatternAtom 'Parsed ->
   Sem r (PatternAtom 'Scoped)
 checkPatternAtom p = case p of
+  PatternAtomIden n -> PatternAtomIden <$> checkPatternName n
   PatternAtomWildcard i -> return (PatternAtomWildcard i)
   PatternAtomEmpty i -> return (PatternAtomEmpty i)
-  PatternAtomParens e -> PatternAtomParens <$> (checkPatternAtoms e >>= parsePatternAtoms)
-  PatternAtomIden n -> PatternAtomIden <$> checkPatternName n
-  PatternAtomBraces n -> PatternAtomBraces <$> (checkPatternAtoms n >>= parsePatternAtoms)
+  PatternAtomParens n e -> PatternAtomParens <$> (traverse groupBindLocalVariable n) <*> (checkPatternAtoms e >>= parsePatternAtoms)
```
janmasrovira, line 1157: remove unnecessary parens

src/Juvix/Compiler/Concrete/Pretty/Base.hs
```diff
@@ -570,8 +571,15 @@ instance SingI s => PrettyCode (PatternAtom s) where
       SScoped -> ppCode n
     PatternAtomWildcard {} -> return kwWildcard
     PatternAtomEmpty {} -> return $ parens mempty
-    PatternAtomParens p -> parens <$> ppPatternParenType p
-    PatternAtomBraces p -> braces <$> ppPatternParenType p
+    PatternAtomParens n p -> ppEnclosed parens n p
+    PatternAtomBraces n p -> ppEnclosed braces n p
+    where
+      ppEnclosed :: (Member (Reader Options) r) => (Doc Ann -> Doc Ann) -> Maybe (SymbolType s) -> PatternParensType s -> Sem r (Doc Ann)
```
janmasrovira, line 577: unneeded parens around `Member (Reader Options) r`

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -322,8 +323,8 @@ data PatternAtom (s :: Stage)
   = PatternAtomIden (PatternAtomIdenType s)
   | PatternAtomWildcard Wildcard
   | PatternAtomEmpty Interval
-  | PatternAtomParens (PatternParensType s)
-  | PatternAtomBraces (PatternParensType s)
+  | PatternAtomParens (Maybe (SymbolType s)) (PatternParensType s)
```
janmasrovira, line 326: we try to follow the rule of having at most 1 argument to constructors that are not records. 
It can be refactored later, so you don't need to change it for this pr.

janmasrovira (commented):
Looking good.

I have one last request before we can merge it. Can you please add a negative test for each of the error messages that you have introduced?

janmasrovira:
Another consideration, currently `x@(zero)` is allowed but `x@zero` is not. I think they should both be allowed.
Currently there is not much point in allowing alias for nullary constructors. But, when we have implicit arguments for constructors it may be useful. And of course the parentheses should not be required.

janmasrovira:
@ii8 if you can solve the merge conflicts we'll merge the branch as it is. Otherwise we'll open a new pr

ii8:
@janmasrovira sry, I've been a bit busy, I'm working on it now though. I'll do the minimum and then make a new pr for negative test cases and so on.

janmasrovira (approved):

janmasrovira:
@ii8 Thanks again for this contribution!

PR #1579, opened 2022-10-05
Letrec lifting

### LINKED ISSUE #1516: LetRec-lifting
Lifting of LetRec to the top level. This can be implemented either as part of lambda-lifting or as another transformation performed before lambda-lifting (see the notes).

Depends on PRs: #1507, #1477.

Notes
------
* A mutually recursive definition 
  ```
  letrec {
    f1 := A1;
    f2 := A2;
  } in B
  ```
  is transformed to 
  ```
  let f1 := f1' x1 .. xm in
  let f2 := f2' x1 .. xm in
  B'
  ```
  where:
  - $FV(A1) ∪ FV(A2) = \\{x_1, ..., x_m\\} ∪ \\{f1, f2\\}$,
  - B' is the result of recursively lambda-lifting B,
  - f1' := λx1 ... λxm A1'[f1 := f1' x1 ... xm, f2 := f2' x1 ... xm] is added to the `IdentContext` in the `InfoTable`, with A1' the result of recursively lambda-lifting A1,
  - analogously for f2'.
* If e.g. A1 above is a lambda-abstraction, lifting it out should be combined with lifting out the letrec, so we don't create a separate function for A1 but use f1' instead. This means that if LetRec-lifting is implemented as a separate transformation, it must always be performed before lambda-lifting.
* The above algorithm with LetRec lifting should still work in O(n k) (see PR #1477) and I think it's essentially equivalent to the Danvy-Schultz algorithm with the assumption that each LetRec node defines a single strongly connected component.
  - Optimisation: break up LetRec nodes into true strongly connected components based on a local analysis of the call graph (separate transformation before lambda-lifting).
* Literature:
  - [Danvy & Schultz, Lambda-lifting in quadratic time](https://scholar.lib.vt.edu/ejournals/JFLP/jflp-mirror/articles/2004/A2004-01/JFLP-A2004-01.pdf)

janmasrovira:
closes #1516.
This pr adds letrec lifting to the lambda lifting transformation.

It also adds other features useful for debugging:
- Adds the flags `--eval` and `no-print` to the `juvix dev core read` command.
- Adds a simple scope checker that checks that all top nodes are closed. 
- The symbol number is always printed for symbols that are being defined.


src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
```diff
@@ -51,6 +55,83 @@ lambdaLiftNode aboveBl top =
...
+              letItems =
+                [ mkApps' (mkIdent' s) (map (mkVar' . fst) fv)
+                  | (s, fv) <- zipExact topSyms recItemsFreeVars
+                ]
+          declareTopSyms
+          let -- free variables in the lets and the body need to be shifted
+              -- because we are introducing binders.
+              -- the topmost let is shifted 0
+              -- the lowermost is shifted (len - 1)
+              -- the final body is not shifted
```
lukaszcz, line 122: I think there is something wrong with this shifting. I changed the letrec test and it fails. You maybe forgot that the values had extra n variables that were then removed? But `substs` should take care of that, so I don't know.

test/Core/Eval/Base.hs
```diff
@@ -28,7 +28,7 @@ coreEvalAssertion mainFile expectedFile trans testTrans step = do
   case r of
     Left err -> assertFailure (show (pretty err))
     Right (_, Nothing) -> do
-      step "Compare expected and actual program output"
+      step "Compare expected an actual program output"
```
lukaszcz, line 31: Typo

tests/Core/positive/test040.jvc
```diff
@@ -1,6 +1,6 @@
 -- letrec
 
-def sum := letrec sum := \x if x = 0 then 0 else x + sum (x - 1) in sum;
+def sum := letrec sum' := \x if x = 0 then 0 else x + sum' (x - 1) in sum';
```
lukaszcz, line 3: I know it's better not to have variable shadowing, but the point was to test if it works correctly when you do have it.

lukaszcz (changes requested):

src/Juvix/Compiler/Core/Extra.hs
```diff
@@ -122,7 +137,7 @@ developBeta = umap go
 
 etaExpand :: Int -> Node -> Node
 etaExpand 0 n = n
-etaExpand k n = mkLambdas' k (mkApps' (shift k n) (map mkVar' (reverse [0 .. k - 1])))
+etaExpand k n = mkLambdas' k (mkApps' (shift k n) (map mkVar' [0 .. k - 1]))
```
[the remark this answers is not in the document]
janmasrovira: ouch! I forgot to push

test/Core/Eval/Base.hs
janmasrovira: I must have messed something going back and forth between my two laptops... I already fixed some of this issues. Sorry about that, I'll double check everything again

tests/Core/positive/test040.jvc
janmasrovira: ah, I see. I changed it because it was harder to debug that way. I'll change it back

janmasrovira: done

src/Juvix/Compiler/Core/Evaluator.hs
```diff
@@ -55,7 +55,9 @@ eval !ctx !env0 = convertRuntimeNodes . eval' env0
 
     eval' :: Env -> Node -> Node
     eval' !env !n = case n of
-      NVar (Var _ idx) -> env !! idx
+      NVar (Var _ idx)
+        | idx < length env -> env !! idx
+        | otherwise -> error ("invalid index: " <> show idx <> " (length: " <> show (length env) <> ")")
```
lukaszcz, line 60: No! We don't want to compute `length env` (in O(number of binders)) on every variable access in the evaluator! If you really want to have bounds checking then reimplement `!!`.

src/Juvix/Compiler/Core/Transformation/LambdaLifting.hs
```diff
@@ -51,6 +56,85 @@ lambdaLiftNode aboveBl top =
...
+                substs
+                  ( reverse
+                      [ mkApps' (mkIdent' sym) (map (NVar . fst) recItemsFreeVars)
+                        | sym <- topSyms
+                      ]
+                  )
+          -- NOTE that we are first substituting the calls and then performing
+          -- lambda lifting. This is a tradeoff. We have slower compilation but
+          -- slightly faster execution time, since it minimizes the number of
+          -- free variables that need to be passed around.
```
lukaszcz, line 93: To be precise: because of this the complexity is O(n m k) instead of O(n k) where:
- n is the size of the program,
- m is the *total* number of letrecs in the whole program,
- k is the maximum nesting of binders.

k can in practice be treated as a non-negligible constant; m not necessarily, but we won't notice this until we have programs with many letrecs

lukaszcz (changes requested):
Fix the evaluator. Otherwise seems fine.

src/Juvix/Compiler/Core/Evaluator.hs
janmasrovira: I have updated it to use `drop`. Is it fine now?

janmasrovira: Actually, I think it will be better to use https://hackage.haskell.org/package/strict-0.4.0.1/docs/Data-Strict-Maybe.html, so the `listToMaybe` does not introduce an unnecessary thunk

janmasrovira: let's keep it as it was before and use the Prelude's `!!`. Now that we have the scope checker, this error message is not as important. And we can always add it back temporarily when debugging.

lukaszcz (approved):

Issue #1577, opened 2022-10-05
Properly newline expressions in the pretty printer


mariari:
I was looking at Juvix examples today, and many examples are quite hard to read

```juvix

  --- Given a player attempted move, updates the state accordingly.
  playMove : Maybe Nat → GameState → GameState;
  playMove nothing (state b p _) := state b p (continue "\nInvalid number, try again\n");
  playMove (just k) (state (board s) player e) := if (not (elem (==) k (possibleMoves (flatten s)))) (state (board s) player (continue "\nThe square is already occupied, try again\n")) (checkState (state (board (map (map (replace player k)) s)) (switch player) noError));
```

Namely the text goes off the screen since it is over `270` characters long. Would be good to have a mechanism that can newline at `80` or `90` characters long automatically in the pretty printed. Should be easy to implement and would improve readability quite a bit.

mariari:
It seems the changes make it harder to read, let me explain

```haskell
  playMove : Maybe Nat → GameState → GameState;
  playMove nothing (state b p _) := state b p (continue
    "\nInvalid number, try again\n");
  playMove (just k) (state (board s) player e) := if (not (elem (==) k
    (possibleMoves (flatten s)))) (state (board s) player (continue
    "\nThe square is already occupied, try again\n")) (checkState (state (board
    (map (map (replace player k)) s)) (switch player) noError));
```

the `(` after the if is more forward than the line after it, making watching a sense of nesting hard. This gets worse in  `player (continue` as the same indentation level stays the same, meaning it's impossible to see where expressions end. is `(checkState ...)` apart of continue? I can't tell!

### A plan

I think there still is an easy way to fix this, the library seems heavily inspired from what I currently use in projects

[Make an block (group in the library terms)](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:group), where you align each argument of a function together, that way they either all newline or don't. Probably best to indent things like

```haskell
 if (won (state b p e))
     (state b p (terminate ("Player " ++str (showSymbol (switch p)) ++str " wins!")))
     (if (draw (state b p e))
         (state b p (terminate "It's a draw!"))
         (state b p e));
```

Where the `align`ment starts after the function word (`if` in this case).

Thankfully It seems the library has done most of the work already with [encloseSep](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:encloseSep), and you can likely just invoke it as follows

```haskell
[...]

PR #1581, opened 2022-10-06
Properly newline expressions in the pretty printer

janmasrovira:
follow-up to #1577 

image

I think it still uses too much horizontal space and it should be improved. Also, it only works for regular application. Infix chains are treated as before.


mariari:
> follow-up to #1577
> 
> image

A very easy way is to maybe just newline after the `:=` and indent by 2, so that the space up to the next line isn't 20 spaces off. (A group can be had here to optionally newline if it gets big to preserve the same like `:=` for small expressions

janmasrovira:
> > follow-up to #1577
> > image
> 
> A very easy way is to maybe just newline after the `:=` and indent by 2, so that the space up to the next line isn't 20 spaces off. (A group can be had here to optionally newline if it gets big to preserve the same like `:=` for small expressions

I actually tried to do that but I didn't succeed. I'll need to spend some more time with it

janmasrovira:
This pr now includes:
1. A new data type (`Ape`) that represents abstract expressions to be pretty printed. They should be generic enough to be used as the backend for `Concrete`, `Abstract` and `Internal`. Currently only `Concrete` uses them.
2. A sensible algorithm for pretty printing `Ape`s. This algorithm takes into account chains of applications or infix operators of the same fixity.
3. This new algorithm is not the default but can be disabled using the global `--no-format` option.
4. A new `juvix-mode` command: `juvix-format-buffer` that formats the current buffer.

I started to work on an enchancement to the algorithm that supports cycling nesting levels when the nesting level comes too close to the end of the page. However, it required a bigger refactor and it will be merged (in case we want it) in another pr. Some wip is fond in https://github.com/anoma/juvix/tree/anchor-experiment branch

lukaszcz (approved):

PR #1582, opened 2022-10-13
Add builtin boolean

paulcadman:
```
builtin boolean
inductive MyBool {
  myTrue : Bool;
  myFalse : Bool;
};
```

The first constructor is mapped to primitive true and the second constructor is mapped to primitive false.

This also adds compilation of builtin boolean in the legacy backend as this was trivial to implement.


lukaszcz (approved):
We should also have a builtin for an `if` so that it can be translated directly to a match (and then to a branch/if) instead of going through a function application. For now, the problem is that there is no way to specify an `if` on the Juvix language level (eager `if` is not an `if`).

PR #1585, opened 2022-10-13
Add builtin if

paulcadman:
A builtin if allows a backend to translate if directly, so that only one branch is evaluated.

```
builtin boolean-if
if : {A : Type} → Bool → A → A → A;
if true x _ := x;
if false _ x := x;
```

An example compilation of if is given for the legacy backend for testing.

I think there was some debate on whether to include `if` as a builtin, or wait until we have a laziness mechanism. Perhaps we could keep this builtin as a temporary construct - we'll need this to make use of `if` in Juvix core.


src/Juvix/Extra/Strings.hs
```diff
@@ -107,6 +107,9 @@ naturalPrint = "natural-print"
 naturalPlus :: IsString s => s
 naturalPlus = "natural-plus"
 
+booleanIf :: IsString s => s
+booleanIf = "boolean-if"
```
janmasrovira, line 111: since we define the type as `Bool`, we should consider `bool-if` instead

paulcadman: We use `natural` (and `natural-plus`) for the type we call `Nat` in the standard library. I am using the same pattern for `boolean`. Does that make sense?

janmasrovira (approved):

Issue #1470, opened 2022-08-21
Support go to definition for the standard library


paulcadman:
Go to definition does not work for symbols from the standard library because the standard library is embedded in the Juvix binary.

To make this work we will write the standard library to a directory inside the `.juvix-build/` directory in the root of the Juvix project. Information about definition sources for the stdlib will then be emitted relative to this directory.

We will write the standard library when the `juvix internal highlight` command is run. We will update the standard library whenever the Juvix version is incremented.

PR #1592, opened 2022-10-18
Support go to definition for the standard library

paulcadman:
The fundamental change in this PR is that the standard library is now read from disk instead of being read from an in memory representation.

When the compiler pipeline is run, the standard library that is embedded in the juvix binary is written to `.juvix-build/stdlib` in the Juvix project. When the standard library is used, it is read from this location.

### Implementation

`topModulePathToFilePath'` now obtains the absolute path of a top module from the `Files` effect instead of constructing the path using the root directory of the Juvix project. The `Files` effect knows which modules are in the standard library and so can return the relevant path to the caller.

https://github.com/anoma/juvix/blob/67ca7df84c0c680d13aeb8ab5fd6691ef21ac1c7/src/Juvix/Compiler/Concrete/Data/Name.hs#L105

### `Loc` construction

The `mkLoc` function used to have the root path of the Juvix project passed in to construct the location file path from the megaparsec `SourcePosition`. However the `sourceName` field of the `SourcePosition` is already absolute, so the path construction was no-op and could be removed.

https://github.com/anoma/juvix/pull/1592/files#diff-407b13d649a9b0e8da32f30d91394ce87893373b6b9c3db4bcd76c58897fc14dR40

### Updating `.juvix-build/stdlib`

`.juvix-build/stdlib` is recreated whenever the versionTag (version + commit hash) of the Juvix binary changes. The current version of the standard library is stored in a file `.juvix-build/stdlib/.version`.

### Using a custom stdlib

Instead of using the pre-bundled standard library you can specify `--stdlib-path PATH` on the CLI to use an existing checkout of the standard library instead.

The custom stdlib flag can also be set as an emacs Juvix-mode customisation variable `juvix-stdlib-path`

### New dependencies

Both [path](https://hackage.haskell.org/package/path) and [path-io](https://hackage.haskell.org/package/path-io) dependencies are added to implement the `getFilesRecursive` function:

https://github.com/anoma/juvix/blob/67ca7df84c0c680d13aeb8ab5fd6691ef21ac1c7/src/Juvix/Prelude/Base.hs#L340

I'm not 100% certain but I think the only new transitive dependency this adds is [dlist](https://hackage.haskell.org/package/dlist).

Closes https://github.com/anoma/juvix/issues/1470


package.yaml
```diff
@@ -36,6 +36,8 @@ dependencies:
 - megaparsec == 9.2.*
 - microlens-platform == 0.4.*
 - parser-combinators == 1.3.*
+- path == 0.9.*
```
janmasrovira, line 39: this library looks nice. Maybe we should consider adopting it more widely

src/Juvix/Compiler/Pipeline.hs
```diff
@@ -40,43 +40,43 @@ compile ::
 compile = typecheck >=> C.fromInternal
 
 upToParsing ::
-  Members '[Reader EntryPoint, Files, Error JuvixError, NameIdGen] r =>
+  Members '[Embed IO, Reader EntryPoint, Files, Error JuvixError, NameIdGen] r =>
```
janmasrovira, line 43: Maybe we could add `UpdateStdlib` to the `Files` effect, similarly to what you did for `RegisterStdlib`. This way we don't need to have `Embed IO` in the pipeline

src/Juvix/Data/Effect/Files.hs
```diff
@@ -5,73 +5,107 @@ module Juvix.Data.Effect.Files
...
 
 makeLenses ''FilesState
+makeLenses ''StdlibState
 
 initState :: FilesState
-initState = FilesState mempty
+initState = FilesState Nothing
 
-readStdlibOrFile ::
+stdlibOrFile ::
```
janmasrovira, line 35: can you add a short documentation comment?

src/Juvix/Compiler/Pipeline.hs
paulcadman: good idea!

janmasrovira (approved):
:1st_place_medal:

Issue #1573, opened 2022-10-04
Update language reference to match current state of Juvix


paulcadman:
The Juvix language reference and documentation need to be updated to take account of the latest changes to the stdlib.

e.g `ℕ` to `Nat`. 

```
open import Stdlib.Prelude;

multiplyByTwo : ℕ -> ℕ;
multiplyByTwo n := 2 * n;
```

* Add documentation for lambdas

PR #1594, opened 2022-10-18
Update language reference to match current state of Juvix

paulcadman:
* fixes examples
* adds documentation for emacs keybindings
* adds a standard emacs keybinding for format
* updates CLI documentation

We should wait to document lambdas until they're available to use in the backend.

Closes https://github.com/anoma/juvix/issues/1573

### DIFF
--- a/docs/org/language-reference/compile-blocks.org
+++ b/docs/org/language-reference/compile-blocks.org
@@ -3,7 +3,7 @@
 The compile keyword has two arguments:
 
 - A name of an expression to be compiled.
-- A set of compilation rules using the format (=backend= → =string=)
+- A set of compilation rules using the format (=backend= ↦ =string=)
   where the string is the text we inline.
 
 This is an example:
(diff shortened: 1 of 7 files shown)


docs/org/language-reference/compile-blocks.org
```diff
@@ -3,7 +3,7 @@
 The compile keyword has two arguments:
 
 - A name of an expression to be compiled.
-- A set of compilation rules using the format (=backend= → =string=)
+- A set of compilation rules using the format (=backend= ↦ =string=)
```
jonaprieto, line 6: Better if we use "->" instead of the Unicode symbol here.

paulcadman: The format uses `↦`, do we have a non unicode equivalent for \mapsto?

paulcadman: 😂 thanks - I completely forgot about this!

jonaprieto (approved):

PR #1598, opened 2022-10-25
Binder refactor

janmasrovira:
Important changes:

1.  Remove `BinderInfo` and replace it with an explicit field of type `Binder` in the nodes where it makes sense. Currently `Binder` is only a name  and a type. We may need to extend it or make it dynamically extensible in the future.
2.  `Node'` is now also parameterized by the type in its binders.
4.  Refactor `destruct` to make it easier to understand and less error prone (hopefully).
    Before we had the following fields in `NodeDetails`:
    ```haskell
         -- | 'nodeChildren' are the children, in a fixed order, i.e., the immediate
         -- recursive subnodes
        _nodeChildren :: [Node],
         -- | 'nodeChildBindersNum' is the number of binders introduced for each
         -- child in the parent node. Same length and order as in `nodeChildren`.
         _nodeChildBindersNum :: [Int],
         -- | 'nodeChildBindersInfo' is information about binders for each child, if
         -- present. Same length and order as in `nodeChildren`.
         _nodeChildBindersInfo :: [[Info]],
    ```
    All the three list need to match their length.
    I have introduced a new data type `NodeChild` that has the same information as before but grouped nicely. All the previous fields are now merged in a single list:
    ```haskell
        _nodeChildren :: [NodeChild]
    ```
     where
     ```haskell
     data NodeChild = NodeChild {
     -- | immediate child of some node
     _childNode :: Node,
     -- | Binders introduced by the child
     _childBinders :: [Binder],
     -- | length of `_childBinders`
     _childBindersNum :: Int
     }
    ```
Other changes:
1. Introduce types `LambdaLhs` and `PiLhs`, which are useful for unfolding and folding.
1.  The semicolon after the last item in a letrec is no longer optional for consistency.


src/Juvix/Compiler/Core/Extra/Base.hs
```diff
@@ -77,22 +78,37 @@ mkMatch' :: NonEmpty Node -> [MatchBranch] -> Node
...
+          _caseBranchBinders = [],
+          _caseBranchBindersNum = 0,
+          _caseBranchBody = b1
+        }
 
 mkIf' :: Node -> Node -> Node -> Node
 mkIf' = mkIf Info.empty
 
 {------------------------------------------------------------------------}
 {- Type constructors -}
```
jonaprieto, line 96: Better to be consistent with the comment style.

src/Juvix/Compiler/Core/Translation/FromSource.hs
```diff
@@ -38,10 +37,10 @@ runParser fileName tab input =
...
     (tbl, Right r) -> Right (tbl, r)
 
-binderInfo :: Name -> Type -> Info
-binderInfo name ty =
-  let info = setInfoType ty (Info.singleton (NameInfo name))
-   in Info.singleton (BinderInfo info)
+-- binderInfo :: Name -> Type -> Info
+-- binderInfo name ty =
+--   let info = setInfoType ty (Info.singleton (NameInfo name))
+--    in Info.singleton (BinderInfo info)
```
lukaszcz, line 43: Remove commented out code

lukaszcz (approved):

lukaszcz (approved):

PR #1603, opened 2022-10-31
Improve As-Pattern parsing

ii8:
@janmasrovira is there also a rule against using tuples? I thought there might be since it's effectively the same as multi argument constructors.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Language.hs
+++ b/src/Juvix/Compiler/Concrete/Language.hs
@@ -82,6 +82,11 @@ type family PatternParensType s = res | res -> s where
   PatternParensType 'Parsed = PatternAtoms 'Parsed
   PatternParensType 'Scoped = PatternArg
 
+type PatternAtType :: Stage -> GHC.Type
+type family PatternAtType s = res | res -> s where
+  PatternAtType 'Parsed = PatternBinding
+  PatternAtType 'Scoped = PatternArg
+
 type family ImportType (s :: Stage) :: GHC.Type where
   ImportType 'Parsed = TopModulePath
   ImportType 'Scoped = Module 'Scoped 'ModuleTop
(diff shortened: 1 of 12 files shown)


janmasrovira:
> @janmasrovira is there also a rule against using tuples? I thought there might be since it's effectively the same as multi argument constructors.

yes, we generally avoid tuples as constructor arguments

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -1069,28 +1069,10 @@ instance HasLoc Pattern where
...
-  ( Eq (ExpressionType s),
-    Eq (IdentifierType s),
-    Eq (PatternAtomIdenType s),
-    Eq (PatternParensType s),
-    Eq (PatternType s),
-    Eq (SymbolType s)
-  ) =>
-  Eq (PatternAtoms s)
-  where
+instance Eq (PatternAtom s) => Eq (PatternAtoms s) where
```
janmasrovira, line 1072: well spotted

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -82,6 +82,11 @@ type family PatternParensType s = res | res -> s where
   PatternParensType 'Parsed = PatternAtoms 'Parsed
   PatternParensType 'Scoped = PatternArg
 
+type PatternAtType :: Stage -> GHC.Type
+type family PatternAtType s = res | res -> s where
+  PatternAtType 'Parsed = (Symbol, PatternAtom 'Parsed)
```
janmasrovira, line 87: it'd be better to have a new data type for this instead of a tuple.

janmasrovira (commented):
Thanks for another contribution @ii8! I feel the code is elegantly written.

I'll summarize the changes here:
1. negative tests for errors introduced in #1576.
2. nullary constructors can be `@`'d in patterns without the need of parentheses.
3. other small improvements.

I would like to see the change about replacing the tuple with some record type. If you feel like doing it I'll wait for it, otherwise I'll merge it as it is.

janmasrovira (approved):
:rocket:

janmasrovira (approved):

Issue #1555, opened 2022-09-28
Refactor the JuvixAsm recursor to enable simpler memory information calculation.


lukaszcz:
Ultimately, we will need a few passes on JuvixAsm which don't need full memory information, but only the current heights of the value and temporary stacks. We should refactor the recursor to enable using it without full validation and memory information computation. This would help performance.

lukaszcz:
Closed by #1604

PR #1604, opened 2022-11-03
Compute JuvixAsm stack usage info

### LINKED ISSUE #1553: Compute maximum value stack height and maximum temporary stack height for each JuvixAsm function
This can be easily done with the recursor. Necessary for the translation to JuvixReg.

lukaszcz:
# Description

* Introduce a simplified version of the recursor `recurseS` which only computes stack information without doing extensive input validation. We will need this simpler recursor in a few more passes.
* Using `recurseS`, compute the maximum value stack height and the maximum temporary stack height in each JuvixAsm function. This is necessary for the translation to JuvixReg.

Closes #1553, #1555

### DIFF
--- a/src/Juvix/Compiler/Asm/Extra.hs
+++ b/src/Juvix/Compiler/Asm/Extra.hs
@@ -71,8 +71,8 @@ computeFunctionStackUsage tab fi = do
 
 computeStackUsage :: Member (Error AsmError) r => InfoTable -> Sem r InfoTable
 computeStackUsage tab = do
-  lst <- mapM (secondM (computeFunctionStackUsage tab)) (HashMap.toList (tab ^. infoFunctions))
-  return tab {_infoFunctions = HashMap.fromList lst}
+  fns <- mapM (computeFunctionStackUsage tab) (tab ^. infoFunctions)
+  return tab {_infoFunctions = fns}
 
 computeStackUsage' :: InfoTable -> Either AsmError InfoTable
 computeStackUsage' tab = run $ runError $ computeStackUsage tab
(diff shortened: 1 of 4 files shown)


src/Juvix/Compiler/Asm/Extra.hs
```diff
@@ -30,11 +30,49 @@ validateCode tab args = void . recurse sig args
...
+            return
+              ( max (si ^. stackInfoValueStackHeight) (max (maximum (map (maximum . map fst) cs)) (maybe 0 (maximum . map fst) md)),
+                max (si ^. stackInfoTempStackHeight) (max (maximum (map (maximum . map snd) cs)) (maybe 0 (maximum . map snd) md))
+              )
+        }
+
+computeStackUsage :: Member (Error AsmError) r => InfoTable -> Sem r InfoTable
+computeStackUsage tab = do
+  lst <- mapM (secondM (computeFunctionStackUsage tab)) (HashMap.toList (tab ^. infoFunctions))
+  return tab {_infoFunctions = HashMap.fromList lst}
```
paulcadman, line 75:
```suggestion
  fns <- mapM (computeFunctionStackUsage tab) (tab ^. infoFunctions)
  return tab {_infoFunctions = fns}
```

`HashMap` is Traversable on its values so we don't need to convert to and from list here.

paulcadman (dismissed):
Looks good - just one minor suggestion to consider.

src/Juvix/Compiler/Asm/Extra.hs
lukaszcz: Yes, right, I'll change it.

paulcadman (approved):

PR #1609, opened 2022-11-07
Make lambda lifting correct when free variables occur in the types of binders

### LINKED ISSUE #1584: Make lambda-lifting correct when free variables occur in the types of binders
Lambda-lifting doesn't work with polymorphic functions with nested lambdas which use top-level type variables rigidly, i.e., something like
```
\(A : Type) \(x : A) f (\(y : A) y + x)
```
Or with dependent types.

The inner lambda also needs to be closed under the free variable `A` even though `A` doesn't occur in the body `y + x`. And it's important that the A argument to the lifted lambda comes before the arguments whose types depend on it, i.e.,
```
def lam := \(A : Type) \(x : A) \(y : A) (y + x) 
```
instead of e.g. the incorrect
```
def lam := \(x : A) \(y : A) \(A : Type) (y + x)
```

In general, even if the type of the lambda is closed, the order of variables needs to be preserved. Consider:
```
\(A : Type) \(y : A) y
 ```

With dependent type, it is not enough to close the terms with the sorted variables. Consider:
```
\(A : Type) \(B : T A) (\x : B) f (\(y : B) h y x)
```
The result of lambda-lifting should be:
```
def lam := \(A : Type) \(B : T A) \(x : B) \(y : B) h y x

\(A : Type) \(B : T A) (\x : B) f (lam A B x)
```
But `A` does not occur anywhere in the expression being lifted. To make it work in general one needs to close recursively until there are no more free variables (but I think one could do a single closure the way you propose).

Dependencies
--------------
* Fixing #1532 could help.

### PR DESCRIPTION
**Pull request by janmasrovira**
Closes #1584.

This pr includes:
- Types in binders are handled properly during lambda lifting.
- When closing a term, we also take into account free variables that occur in the binders.
- Binder list is now implemented as a list. Some useful functions have been added

### CODE REVIEW

File: src/Juvix/Compiler/Core/Extra.hs [hunk 1]
```diff
@@ -98,6 +102,44 @@ captureFreeVars fv
...
+captureFreeVarsCtx :: BinderList Binder -> Node -> ([(Var, Binder)], Node)
+captureFreeVarsCtx bl n =
+  let assocs = freeVarsCtx bl n
+   in (assocs, captureFreeVars (map (first (^. varIndex)) assocs) n)
+
+freeVarsCtx' :: BinderList Binder -> Node -> [Var]
+freeVarsCtx' bl = map fst . freeVarsCtx bl
+
+-- | the output list does not contain repeated elements and is sorted by variable index.
+-- TODO it can probably be fore efficient
```
lukaszcz, line 117: It can be more efficient and simpler: you just go through the BinderList and filter out the ones not present in the free variables of the body and previous binders?

Something like this should work, I think (treating BinderList as a list, I hope the shifting is correct but might've bungled it):

freeVarsCtx bl n = go 0 bl (freeVars n) []

go idx [] fvs acc = reverse acc
go idx (b : bs) fvs acc | Set.member idx fvs =
  let b' = shiftVar (idx + 1) b in
  go (idx + 1) bs (Set.union fvs (freeVars (b' ^. binderType))) ((idx, b') : acc)
go idx (_ : bs) fvs acc = go (idx + 1) bs fvs acc

src/Juvix/Compiler/Core/Extra.hs
```diff
@@ -98,6 +102,44 @@ captureFreeVars fv
...
+freeVarsCtx :: BinderList Binder -> Node -> [(Var, Binder)]
```
lukaszcz, line 118: What are the free variables in the returned Binders relative to? It's not apparent from the code.

lukaszcz (changes requested):
Crashes, unfortunately. I added a test which should help in finding the bug.

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: I have added a clarification

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: the error should be fixed. I have also simplified it a bit

src/Juvix/Compiler/Core/Extra.hs
lukaszcz: What I meant is that e.g. if you have the context [x : A$1, y : A$0, A : Type] and you close x then the result list is [x : A$0, A : Type]? At some point you need to shift the index of A in the type of x and I don't see where it happens.

lukaszcz (approved):

src/Juvix/Compiler/Core/Extra.hs
janmasrovira: the shifting happens in `captureFreeVars`

PR #1612, opened 2022-11-08
Add juvix-repl-mode for emacs

paulcadman:
Adds an emacs REPL for Juvix derived from https://www.emacswiki.org/emacs/ComintMode using the `juvix repl` CLI command.

Run `run-juvix-repl` to launch the REPL in a new emacs buffer.

The `juvix-load` (aka `C-c C-l`) will load the current file into the REPL if the REPL is already running. Since we use `juvix-load` for checking/highlighting the Juvix file it seemed like always launching the REPL would be too intrusive. On the other hand, if the REPL is already running then it seems to make sense to load it.

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -28,7 +28,7 @@ runAppIO g root pkg = interpret $ \case
   RenderStdOut t
     | g ^. globalOnlyErrors -> return ()
     | otherwise -> embed $ do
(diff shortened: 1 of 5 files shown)


janmasrovira (approved):
Great!!

I have added the `repl` label to github

PR #1615, opened 2022-11-09
Compute new entrypoint root when loading a file in the REPL

paulcadman:
Previously the REPL would use the `App` effect root (which could be the current directory or the project root of the initially loaded file). Thus files in a different Juvix project could not be loaded. The entrypoint root must be computed each time a new file is `:load`ed.

This PR also adds shelltests for the REPL commands.


app/Commands/Repl.hs
```diff
@@ -73,6 +74,9 @@ noFileLoadedMsg = liftIO (putStrLn "No file loaded. Load a file using the `:load
 welcomeMsg :: MonadIO m => m ()
 welcomeMsg = liftIO (putStrLn [i|Juvix REPL version #{versionTag}: https://juvix.org. Run :help for help|])
 
+preludePath :: FilePath
```
janmasrovira, line 77: I think this can go in `Juvix/Extra/Paths.hs`

janmasrovira (approved):

PR #1617, opened 2022-11-10
Add types to Core functions and constructors when translating from Internal

paulcadman:
This PR adds support for translation of `Function` and `SmallUniverse` expressions to Core.

This means that the type `Expressions` for constructors, functions and axioms can be translated and stored in the corresponding Core {`Identifier`, `Constructor`}`Infos`.


src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -437,28 +490,31 @@ goExpression' = \case
...
+  ([Internal.FunctionParameter], Internal.Expression) ->
+  Sem r Node
+goFunction (params, returnTypeExpr) = do
+  foldr f (goExpression returnTypeExpr) params
+  where
+    f :: Internal.FunctionParameter -> Sem r Node -> Sem r Node
+    f param acc = do
+      paramBinder <- Binder (param ^. Internal.paramName) <$> goExpression (param ^. Internal.paramType)
+      case param ^. Internal.paramName of
+        Nothing -> mkPi mempty paramBinder <$> acc
```
janmasrovira, line 509: we have the shortcut `mkPi'` when the info is empty

paulcadman: `mkPi'` doesn't take a binder so can't be used at this call site.

I thought about adding a new variant of `mkPi` that takes a binder but I decided against it. The pattern we have is that `'` variants allow you to omit the `Info` argument, which is easy to remember. If we have variants for "no Info, Type" and "no Info, Binder" I think I'd always have to look at the definition to figure out which to use.

I suppose an alternative is to make `mkPi'` take a binder and fix the call sites of the existing calls, or to add a new variant called `mkPiBinder'`. What do you think?

janmasrovira: You are right!

> I suppose an alternative is to make mkPi' take a binder and fix the call sites of the existing calls, or to add a new variant called mkPiBinder'. What do you think?
I think it makes more sense to modify mkPi' so that it takes a binder, but it is not a strong preference so I'll leave it up to you

janmasrovira (approved):

src/Juvix/Compiler/Core/Translation/FromInternal.hs
paulcadman: I tried updating `mkPi'` to take a binder but that made the call sites of the existing binderless `mkPi'` calls quite noisy. On balance I think it's best to use the full `mkPi` at the call site instead of changing the others. This could change if we start using the Binder form more.

PR #1621, opened 2022-11-15
Upgrade stack snapshot to use ghc-9.2.5

janmasrovira:


janmasrovira:
According to this: https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html
Version 9.4.x still has limited support for lsp. Consequently, I think we should stick with 9.2.5 for now

paulcadman (dismissed):
Works on macOS aarch64. Maybe we should wait to merge until everyone has tested it?

paulcadman:
To keep the linux-static-binary in sync I've updated the alpine GHC Dockerfile and pushed a new image for alpine GHC 9.2.5 as it doesn't look like the official alpine binary release has been fixed (https://gitlab.haskell.org/ghc/ghc/-/issues/20266).

linux-static-binary also now has a step to build the Juvix runtime.

paulcadman (approved):

PR #1622, opened 2022-11-15
Support basic dependencies

janmasrovira:
This Pr has two goals:
1. Add support for local dependencies.
1. Refactor most of the code that was using `FilePath` to use [`Path`](https://hackage.haskell.org/package/path).

In order to specify the list of dependencies for a package, the field `dependencies` has been added to the `juvix.yaml`.
The `dependencies` field is a list of directories (relative or absolute) that must point to other directories  that also contain a `juvix.yaml`.
As expected, if we add a package to the list of dependencies, we will be able to access its modules through import statements.
Regarding the standard library, the behaviour is as follows:
1. If the `dependencies` field is omitted, the `stdlib` is copied under `.juvix-build/stdlib`.
2. If the `dependencies` field is not omitted and contains the directory `.juvix-build/stdlib`, it is also copied.
3. otherwise the `stdlib` is not copied.

We have discussed some alternatives that are much better than the current behaviour, but we agreed to change that in a separate pr.


jonaprieto:
Testing the discord bot.

jonaprieto:
this

src/Juvix/Prelude/Path.hs
```diff
@@ -1,28 +1,91 @@
...
+destructAbsFile x = (root, dirs, filename x)
+  where
+    (root, dirs) = destructAbsDir (parent x)
+
+isJuvixFile :: Path b File -> Bool
+isJuvixFile = (== Just ".juvix") . fileExtension
+
+isHiddenDirectory :: Path b Dir -> Bool
+isHiddenDirectory p = case toFilePath (dirname p) of
+  "./" -> False
```
paulcadman, line 61: minor: I know we don't care about windows at the moment but consider using https://hackage.haskell.org/package/path-0.9.2/docs/Path-Internal-Posix.html#v:relRootFP instead of "./" literal. [source](https://github.com/commercialhaskell/path/blob/becc65ff039c55f2251480e6230213543b0d2634/src/Path/Internal/Include.hs#L66)

src/Juvix/Prelude/Path.hs
```diff
@@ -1,28 +1,91 @@
...
+    go :: [Path Abs Dir] -> Path Abs Dir -> NonEmpty (Path Abs Dir)
+    go ac p
+      | isRoot p = NonEmpty.reverse (p :| ac)
+      | otherwise = go (p : ac) (parent p)
 
-destructFilePath :: Path b File -> ([Path Rel Dir], Path Rel File)
-destructFilePath p = case nonEmptyUnsnoc (nonEmpty' (splitPath (toFilePath p))) of
-  (ps, f) -> (fmap relDir (maybe [] toList ps), relFile f)
+withTempDir' :: (MonadIO m, MonadMask m) => (Path Abs Dir -> m a) -> m a
+withTempDir' = withTempDir $(mkRelDir ".") "tmp"
```
paulcadman, line 91: I think we should use [withSystemTempDir](https://hackage.haskell.org/package/path-io-1.7.0/docs/Path-IO.html#v:withSystemTempDir) instead of withTempDir to avoid adding the temporary directory to the workspace.

src/Juvix/Prelude/Path.hs
```diff
@@ -1,28 +1,91 @@
...
+isJuvixFile :: Path b File -> Bool
+isJuvixFile = (== Just ".juvix") . fileExtension
+
+isHiddenDirectory :: Path b Dir -> Bool
+isHiddenDirectory p = case toFilePath (dirname p) of
+  "./" -> False
+  '.' : _ -> True
+  _ -> False
+
+parseRelFile' :: FilePath -> Path Rel File
```
paulcadman, line 65: Do we need both `relFile` and `parseRelFile`?

app/App.hs
```diff
@@ -62,24 +62,29 @@ getEntryPoint' opts root pkg inputFile = do
         _entryPointNoTermination = opts ^. globalNoTermination,
         _entryPointNoPositivity = opts ^. globalNoPositivity,
         _entryPointNoStdlib = opts ^. globalNoStdlib,
-        _entryPointStdlibPath = opts ^. globalStdlibPath,
+        _entryPointStdlibPath = someBaseToAbs root <$> opts ^. globalStdlibPath,
         _entryPointPackage = pkg,
-        _entryPointModulePaths = pure (inputFile ^. pathPath),
+        _entryPointModulePaths = pure (someBaseToAbs root (inputFile ^. pathPath)),
```
paulcadman, line 67: This code assumes that `inputFile` is either an absolute path or a relative path inside the Juvix project.

However we previously supported running the juvix commands where `inputFile` is a relative path to a Juvix file but the command is not run from within the juvix project.

e.g say the juvix project root is at `/A/B/C` the Juvix file is `/A/B/C/D.juvix` and the user runs `juvix typecheck B/C/D.juvix` (the `PWD` is `/A`).

There's a shellcheck test that catches this.

app/CommonOptions.hs
```diff
@@ -14,86 +14,98 @@ import System.Process
...
           <> m
       )
-  pure Path {_pathIsInput = False, ..}
+  pure AppPath {_pathIsInput = False, ..}
+
+someFileOpt :: ReadM (SomeBase File)
+someFileOpt = eitherReader aux
+  where
+    aux :: String -> Either String (SomeBase File)
+    aux s = maybe (Left $ s <> " is not a file path") Right (parseSomeFile s)
```
paulcadman, line 102: Relative paths containing `../` components will fail, which we used to support:

```
cd tests/positive/Internal/Positivity
juvix typecheck ../Simple.juvix
../Simple.juvix is not a file path
```

app/CommonOptions.hs
```diff
@@ -14,86 +14,98 @@ import System.Process
...
+
+someDirOpt :: ReadM (SomeBase Dir)
+someDirOpt = eitherReader aux
+  where
+    aux :: String -> Either String (SomeBase Dir)
+    aux s = maybe (Left $ s <> " is not a directory path") Right (parseSomeDir s)
```
paulcadman, line 108: Relative paths containing `../` components will fail, which we used to support:

```
$ cd tests/positive/Internal
$ juvix  --stdlib-path ../../../juvix-stdlib/ typecheck Simple.juvix
option --stdlib-path: ../../../juvix-stdlib/ is not a directory path
```

src/Juvix/Data/Effect/Files/IO.hs
```diff
@@ -5,89 +5,34 @@ module Juvix.Data.Effect.Files.IO
...
-    | otherwise -> return pAbsPath
-    where
-      pAbsPath :: FilePath
-      pAbsPath = rootPath </> p
-
-      isConflict :: Sem r Bool
-      isConflict = do
-        cRootPath <- embed (canonicalizePath rootPath)
-        cStdlibPath <- embed (canonicalizePath (s ^. stdlibRoot))
-        andM [return (cRootPath /= cStdlibPath), embed (doesFileExist pAbsPath)]
```
paulcadman, line 49: Previously we did not consider file lookups inside the stdlib to be file conflicts (`cRootPath /= cStdlibPath`). This facilitates stdlib development and the REPL depends on this behaviour (caught in the shellcheck failtures).

```
$ cd tests/positive/Internal
$ juvix repl
/Users/paul/heliax/juvix/tests/positive/Internal/.juvix-build/stdlib/Stdlib/Prelude.juvix:1:8-22: error:
The top module Stdlib.Prelude is defined in the file:
/Users/paul/heliax/juvix/tests/positive/Internal/.juvix-build/stdlib/Stdlib/Prelude.juvix
But it should be in the file:
/Users/paul/heliax/juvix/tests/positive/Internal/Stdlib/Prelude.juvix
```

src/Juvix/Prelude/Path.hs
janmasrovira: done in 2a249828259abd40a5a85a10f57d5da460ea5582

src/Juvix/Prelude/Path.hs
janmasrovira: done in 2a249828259abd40a5a85a10f57d5da460ea5582

src/Juvix/Prelude/Path.hs
janmasrovira: removed in 2a249828259abd40a5a85a10f57d5da460ea5582

app/CommonOptions.hs
janmasrovira: is it ok if we solve this in a separate pr? none of the current tests depend on this

app/CommonOptions.hs
janmasrovira: idem

src/Juvix/Data/Effect/Files/IO.hs
janmasrovira: this has been addressed

PR #1652, opened 2022-11-30
Files pure refactor

janmasrovira:
This PR refactors the Files effect pure implementation. 
Previously, we only had a hashmap from `FilePath` to `Text`. If one wants to implement recursors this flat implementation is inconvenient. This PR adds a tree representation of a file system and implements a generic recursor in the spirit of [`walkDirRel`](https://hackage.haskell.org/package/path-io-1.7.0/docs/Path-IO.html#v:walkDirRel) from the `path-io` library.

This refactor is important for #1622 

Some efforts to support a general recursor directly in the files effect were made. However, it seems that this is not possible with the interface that polysemy offers for interpretting higher order effects.


src/Juvix/Data/Effect/Files/Pure.hs
```diff
@@ -0,0 +1,141 @@
...
+              helper :: Maybe FSNode -> FSNode
+              helper = maybe (helper (Just emptyNode)) (go ds)
+
+relFile :: FilePath -> Path Rel File
+relFile = fromJust . parseRelFile
+
+relDir :: FilePath -> Path Rel Dir
+relDir = fromJust . parseRelDir
+
+destructPath :: Path b Dir -> [Path Rel Dir]
```
paulcadman, line 57: suggestion: Add `destructPath` and `destructFilePath` to Juvix.Prelude

paulcadman (approved):
Added a suggestion, feel free to ignore.

src/Juvix/Data/Effect/Files/Pure.hs
janmasrovira: I thought about that but I wasn't sure since we are still using Filepath in most of the codebase.

I think I'll create a `Juvix.Prelude.Path` file and move these functions there. For now `Juvix.Prelude` will not export that

paulcadman: I agree with this reasoning. Putting these functions in `Juvix.Prelude.Path` means we're more likely to remember that they exist if we ever move over to `path`.

paulcadman (approved):

PR #1659, opened 2022-12-08
Move applications inside Lets and Cases

### LINKED ISSUE #1654: Move applications inside Lets and Cases
The translation from Core to Core.Stripped requires that only variables or function symbols be present at the head of an application.

We need to transform
```agda
(let x := M in N) Q
```
to
```agda
let x := M in N Q
```
and e.g.
```agda
(case M of { c1 x := N1; c2 x y := N2 }) Q
```
to
```agda
case M of { c1 x := N1 Q; c2 x y := N2 Q }
```

lukaszcz:
Closes #1654.


paulcadman (approved):
🚢

PR #1662, opened 2022-12-09
Replace stack with cabal as the project build tool (experiment)

paulcadman:
The multi-cradle hie configuration for cabal seems to work more smootly with LSP than stack

I did the project translation using https://github.com/hasufell/stack2cabal which generates a `cabal.project.freeze` file from the stackage resolver.

To make the file-embed library work, the files you want to embed need to be specified in the `extra-source-files` configuration: https://github.com/anoma/juvix/blob/d7489d90651d8058d03aa82650a6b30b6027c0ae/juvix.cabal#L34

The cache in the GitHub workflow is disabled for now because there is limited cache space (10GB) so
building with cabal could interfer with the existing stack builds.


paulcadman:
Closing in favour of https://github.com/anoma/juvix/pull/1667

PR #1670, opened 2022-12-15
Pin mdbook to version 0.4.22 in docs build

paulcadman:
mdbook homebrew version (used by the github action to install mdbook) was bumped to 0.4.23
https://github.com/Homebrew/homebrew-core/commit/eb69fcddbab62862737bb76c5d4aff0b241083ce which seems to be incompatible with Ubuntu 20.04 glibc

The mdbook program works with this change, see this successful job: https://github.com/anoma/juvix/actions/runs/3703796868/jobs/6275684983

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,7 +278,7 @@ jobs:
 
   docs:
     needs: build
-    # if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
     strategy:
       matrix:
         os: [ubuntu-20.04]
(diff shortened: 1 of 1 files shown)


jonaprieto (commented):
Consider to change the `uses` field to `peaceiris/actions-mdbook@v1.2.0`.

paulcadman:
> Consider to change the `uses` field to `peaceiris/actions-mdbook@v1.2.0`.

let's try it to see what breaks

paulcadman:
> > Consider to change the `uses` field to `peaceiris/actions-mdbook@v1.2.0`.
> 
> let's try it to see what breaks

Actually - the v1 tag already points to v1.2.0, see https://github.com/peaceiris/actions-mdbook/releases/tag/v1 - it's recommended that users of actions point to the `v1` tag instead of a particular v1 release tag, see https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management

jonaprieto (approved):

PR #1673, opened 2022-12-20
Benchmarks

janmasrovira:


bench/Variants.hs
```diff
@@ -0,0 +1,170 @@
...
+commonOptions args ext = toFilePath (args ^. buildSrc) : outOptions args ext
+
+ocamlExe :: Variant
+ocamlExe =
+  Variant
+    { _variantTitle = Nothing,
+      _variantLanguage = Ocaml,
+      _variantExtensions = ext,
+      _variantBuild = \args ->
+        command_ [] "ocamlc" (commonOptions args ext)
```
lukaszcz, line 38: `ocamlc` creates an interpreted bytecode executable, not an optimized native one

bench/Variants.hs
```diff
@@ -0,0 +1,170 @@
...
+    ext = defaultExt
+
+ocamlByteExe :: Variant
+ocamlByteExe =
+  Variant
+    { _variantTitle = Just "byte",
+      _variantLanguage = Ocaml,
+      _variantExtensions = ext,
+      _variantBuild = \args ->
+        command_ [] "ocamlopt" ("-O2" : commonOptions args ext)
```
lukaszcz, line 51: `ocamlopt` creates a native executable not bytecode

tests/benchmark/mergesort/c/mergesort.c
```diff
@@ -44,7 +44,7 @@ static bool sorted(int *tab, int size) {
     return true;
 }
 
-#define N 2000000
+#define N 100000
```
lukaszcz, line 47: Changing this number to a smaller one significantly changes the character of the benchmark. With 100k elements in the list one needs at most a few 64k pages, so memory management / allocation impact is no longer tested here.

tests/benchmark/mergesort/core/mergesort.jvc
```diff
@@ -51,4 +51,4 @@ def sorted := \xs
 
 def gen := \n \acc if n = 0 then acc else gen (n - 1) (cons n acc);
 
-sorted (sort (gen 2000000 nil))
+sorted (sort (gen 200000 nil))
```
lukaszcz, line 54: Inconsistent change

tests/benchmark/mergesort/c/mergesort.c
janmasrovira: I'll change it back once the pr is ready. It is very useful for debugging purposes to be able to run the benchmarks faster

bench/Compile.hs
```diff
@@ -0,0 +1,99 @@
...
+    liftIO (runMode DefaultMode (config s) [] [fromSuite s])
+
+runExe :: Path Abs File -> IO ()
+runExe p = void (readProcess (toFilePath p) [] "")
+
+fromSuite :: Suite -> Benchmark
+fromSuite b = bgroup (b ^. suiteTitle) (map go (b ^. suiteVariants))
+  where
+    go :: Variant -> Benchmark
+    go v = bench title (nfIO (runExe exe))
```
lukaszcz, line 79: Are you sure this all works as expected with laziness? So e.g. compilation time is not counted into the benchmarks? Seems not, but I'm never entirely sure what gets executed when in Haskell.

In any case, something is wrong. The results for "native runtime" for "fibonacci" should be comparable (probably slightly better) than the results for "native ocaml".

bench/Variants.hs
janmasrovira: swapped

bench/Variants.hs
janmasrovira: swapped

bench/Compile.hs
lukaszcz: Maybe try to run the programs manually, measuring them with `/usr/bin/time` and see if there is a discrepancy there.

janmasrovira: The path was wrong and it was running the wrong executable. Now the results are what you expected

lukaszcz:
Regarding the increase in speed, I think fiddling with the confidence intervals is a good idea. But maybe it's possible to change the confidence intervals individually for different benchmarks? It's fine to have a 20% standard deviation if the measured value is 10 times bigger than all other values, but less so when values are close. And the benchmarks taking most time are probably the outliers, so it might be useful to have much larger intervals for them.

bench/Suites.hs
```diff
@@ -0,0 +1,30 @@
...
+    <> [ Suite "fold" (allVariantsExcept [C] [CoreEval]),
+         Suite "mapfold" (allVariantsExcept [C] [CoreEval])
+       ]
+
+-- "mapfun" -- juvix crashes: Address boundary error
+-- "ackermann", -- juvix crashes
+-- "combinations", -- juvix crashes:  call stack exhausted
+-- "cps", juvix: call stack exhausted
+-- "prime" -- juvix: Address boundary error
+
```
lukaszcz, line 24: Why aren't those benchmarks included? That the current implementation of juvix crashes is a problem with the old backend, but the benchmarks themselves test reasonable things. The crash should just be noted on the plot / in the results.

lukaszcz:
I think there is a point in benchmarking the runtime against other languages even if Juvix crashes.

bench/Suites.hs
janmasrovira: I have included them now. The native and wasm Juvix variants have been excluded since it is not possible to benchmark functions that may crash using the criterion library.

janmasrovira:
I have included all Suites but excluded the variants that crash. I've removed the explicit time limit and put a confidence interval of 90 for every suite. We may want to try different values globally and for each suite, but I propose that  we postone this for a later pr.

lukaszcz (approved):

Issue #1354, opened 2022-06-28
Add full support for where blocks


jonaprieto:

lukaszcz:
Closed by #1674

PR #1674, opened 2022-12-22
Remove where syntax

### LINKED ISSUE #1640: Remove where syntax
We're removing this as Juvix, in its current state is strict, and where-blocks are presented in lazy languages.

jonaprieto:
Closes #1640 #1354


janmasrovira (approved):
:wave: `where`

PR #1676, opened 2022-12-22
Stop ignoring .vscode folder

jonaprieto:


.vscode/settings.json
```diff
@@ -0,0 +1,47 @@
+// Recall to run the following command before even run vscode.
```
paulcadman, line 1: Does this work? JSON doesn't have comments syntax - so this file is not valid JSON. (It's possible that the vscode JSON parser allows this, but good to check).

paulcadman: Looked it up - vscode configuration files use "JSON with comments" so this is fine. https://code.visualstudio.com/docs/languages/json#_json-with-comments

.vscode/settings.json
```diff
@@ -0,0 +1,47 @@
...
+// VScode will use cabal for building and one can seperately run
+// stack without any problem. The converse workflow also works.
+
+// $ cd juvix
+// $ cp cabal.hie.yaml hie.yaml
```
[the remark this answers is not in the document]
jonaprieto: Oh, I've forgotten about this. Thanks! Let me add this.

lukaszcz:
Why do you want to stop ignoring .vscode? For example, the local C extension configuration is saved there, so one won't be able to customize it locally anymore.

.vscode/settings.json
```diff
@@ -0,0 +1,47 @@
...
+
+{
+  "editor.tabSize": 2,
+  "editor.renderWhitespace": "trailing",
+  "haskell.formattingProvider": "ormolu",
+  "haskell.openSourceInHackage": false,
+  "haskell.openDocumentationInHackage": false,
+  "haskell.checkProject": true,
+  "haskell.manageHLS": "GHCup",
+  "haskell.serverExecutablePath": "${HOME}/.local/bin/haskell-language-server",
```
lukaszcz, line 18: I'm not sure if this is a good idea at all. You're forcing people to use GHCup and a particular hard-coded path to the HLS. It's not possible to change this locally because the local configuration files are version-controlled with git.

lukaszcz:
I think .vscode should just be ignored and a different way of achieving whatever you want to achieve needs to be found. Versioning .vscode is like versioning the .emacs file in the Juvix compiler project.

Issue #1696, opened 2023-01-04
Compiling programs with IO to native code no longer works


lukaszcz:
When compiling
```agda
module prog;

open import Stdlib.Prelude;

main : IO;
main := printNatLn zero;

end;
```
with `juvix compile` I get
```
/home/heliax/Documents/progs/juvix/.juvix-build/prog.c:4:16: warning: unknown attribute 'export_name' ignored [-Wunknown-attributes]
__attribute__((export_name("printNatLn")))
               ^~~~~~~~~~~~~~~~~~~~~~~~~
/home/heliax/Documents/progs/juvix/.juvix-build/prog.c:6:16: warning: unknown attribute 'export_name' ignored [-Wunknown-attributes]
__attribute__((export_name("main")))
               ^~~~~~~~~~~~~~~~~~~
/home/heliax/Documents/progs/juvix/.juvix-build/prog.c:11:46: warning: implicit declaration of function 'prim_printString' is invalid in C99 [-Wimplicit-function-declaration]
                                   (prim_io) prim_printString((prim_string) "\n"));
                                             ^
3 warnings generated.
/usr/bin/ld: /tmp/prog-676c26.o: in function `printNatLn_268':
prog.c:(.text+0x1d5): undefined reference to `prim_printString'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
This will be no longer relevant with the new backend, but fixing it for the old backend earlier would be useful for testing.

PR #1698, opened 2023-01-04
Add printString and printBool support to legacy C backend

paulcadman:
This PR implements `printString` and `printBool` builtins for the legacy C backend. Previously IO for strings was done using compile blocks with included C code.

Fixes https://github.com/anoma/juvix/issues/1696


lukaszcz (approved):
Don't we have the same problem with printBool? It's declared as a builtin axiom in the standard library, but I don't see it anywhere in c-runtime.

paulcadman:
> Don't we have the same problem with printBool? It's declared as a builtin axiom in the standard library, but I don't see it anywhere in c-runtime.

You're right, I missed it. I will add it to this PR

lukaszcz (approved):

PR #1707, opened 2023-01-09
Replace --output-dir flag by --internal-build-dir

jonaprieto:


app/CommonOptions.hs
```diff
@@ -82,6 +82,19 @@ parseGenericOutputFile = do
       )
   pure AppPath {_pathIsInput = False, ..}
 
+parseBuildDir :: Mod OptionFields (SomeBase Dir) -> Parser (AppPath Dir)
```
janmasrovira, line 85: I don't see that this belongs in CommonOptions. This is a concrete option. Imo this should be in app/GlobalOptions.hs

app/GlobalOptions.hs
```diff
@@ -110,3 +110,16 @@ parseGlobalFlags = do
...
   return GlobalOptions {..}
+
+parseBuildDir :: Mod OptionFields (SomeBase Dir) -> Parser (AppPath Dir)
+parseBuildDir m = do
+  _pathPath <-
+    option
+      someDirOpt
+      ( long "internal-build-dir"
+          <> metavar "BUILD_DIR"
+          <> help "Path to output directory"
```
janmasrovira, line 121: help "Directory for compiler internal output"

app/GlobalOptions.hs
```diff
@@ -73,10 +73,10 @@ parseGlobalFlags = do
       )
   _globalBuildDir <-
     optional
-      ( parseGenericOutputDir
+      ( parseBuildDir
           ( value (Rel relBuildDir)
               <> showDefault
-              <> help "directory for compiler output"
+              <> help "directory for compiler internal output"
```
janmasrovira, line 79: after the change below this help will not be needed

janmasrovira (approved):

PR #1711, opened 2023-01-09
1704 type synonyms are not always unfolded during type checking

janmasrovira:


lukaszcz (commented):
This doesn't entirely solve the problem. The synonyms are still sometimes not unfolded, e.g., when type-checking a function application.

Checking the module:
```agda
module church;

open import Stdlib.Prelude;

Num : Type;
Num := {A : Type} → (A → A) → A → A;

czero : Num;
czero {_} f x := x;

csuc : Num → Num;
csuc n {_} f := f ∘ n f;

end;
```
results in an error:
```
Type error near church.juvix:12:21-25.
In the expression:
  n
the expression n is expected to have a function type but has type:
  Num
```

janmasrovira:
The above is now solved, however, you need to provide the implicit argument of `n` explicitly:
```
csuc : Num → Num;
csuc n {_} f := f ∘ n {_} f;
```

lukaszcz (approved):
Yes, seems to be working now. I can do Church numerals with type synonyms.

PR #1715, opened 2023-01-10
Improve error message for confusing ':=' with '='

### LINKED ISSUE #1483: Better error message for confusing ":=" with "=".
If I write
```
id : (A : Type) → A → A;
id _ a = a;
```
I get the error message
```
   |
23 | id _ a = a;
   |           ^^
unexpected ";<newline>"
expecting ":=", '(', '.', '_', '{', or '≔'
```
which is pretty bad, because it doesn't explain what really is wrong ("=" instead of ":=") and doesn't point at the source of the problem but at the end of the line.

The error message should be improved. This issue comes up very often for novices, and one occasionally gets confused even after some time using Juvix.

lukaszcz:
Closes #1483 

The error now points to the offending `=`, works correctly with multi-line clauses, and explains exactly what's wrong. E.g. for
```agda
f : Nat → Nat → Nat;
f zero x = x;
```
we get
```
  |
6 | f zero x = x;
  |           ^
expected ":=" instead of "="
```
A minor disadvantage of the proposed solution is that now it's impossible to use `=` without parentheses as a top variable name in a pattern, e.g.
```
f zero = := =;
```
gives an error.

However, if one really wants to name a variable `=`, it is enough just to enclose it in parentheses:
```agda
f : Nat → Nat → Nat;
f zero (=) := =;
f (suc n) (=) := f n =;
```

I believe this slight non-uniformity is well worth the increased usability due to a better error message. Confusing `:=` with `=` is very common. Using `=` as a variable name in a top pattern is rare.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Translation/FromSource.hs
+++ b/src/Juvix/Compiler/Concrete/Translation/FromSource.hs
@@ -564,11 +564,11 @@ parsePatternAtoms = do
 
 functionClause :: Members '[InfoTableBuilder, JudocStash, NameIdGen] r => Symbol -> ParsecS r (FunctionClause 'Parsed)
 functionClause _clauseOwnerFunction = do
-  _clausePatterns <-functionClausePatterns
+  _clausePatterns <- functionClausePatterns
   _clauseBody <- parseExpressionAtoms
   return FunctionClause {..}
 
-functionClausePatterns :: forall r . Members '[InfoTableBuilder, JudocStash, NameIdGen] r => ParsecS r [PatternAtom 'Parsed]
+functionClausePatterns :: forall r. Members '[InfoTableBuilder, JudocStash, NameIdGen] r => ParsecS r [PatternAtom 'Parsed]
 functionClausePatterns =
   P.manyTill patternAtom (kw kwAssign <|> wrongEq)
   where
--- a/src/Juvix/Compiler/Concrete/Translation/FromSource.hs
+++ b/src/Juvix/Compiler/Concrete/Translation/FromSource.hs
@@ -574,7 +574,9 @@ functionClausePatterns =
   where
     wrongEq :: ParsecS r ()
     wrongEq = do
-      P.chunk "="
+      P.try $ do
+        w <- morpheme
+        unless (w == "=") (P.failure Nothing mempty)
       off <- P.getOffset
       parseFailure off "expected \":=\" instead of \"=\""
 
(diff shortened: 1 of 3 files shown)


src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -564,11 +564,20 @@ parsePatternAtoms = do
...
   _clauseBody <- parseExpressionAtoms
   return FunctionClause {..}
 
+functionClausePatterns :: forall r. Members '[InfoTableBuilder, JudocStash, NameIdGen] r => ParsecS r [PatternAtom 'Parsed]
+functionClausePatterns =
+  P.manyTill patternAtom (kw kwAssign <|> wrongEq)
+  where
+    wrongEq :: ParsecS r ()
+    wrongEq = do
+      P.chunk "="
```
janmasrovira, line 577: this will also rule out `==` or any identifier starting with `=`. Was this intended?

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -564,11 +564,20 @@ parsePatternAtoms = do
...
+      off <- P.getOffset
+      parseFailure off "expected \":=\" instead of \"=\""
```
janmasrovira, line 579: this error message does not seem to surface. 
I am getting this:
image


To fix it you can do this:
```diff
-  P.manyTill patternAtom (kw kwAssign <|> wrongEq)
+  P.manyTill patternAtom (wrongEq <|> kw kwAssign)
   where
     wrongEq :: ParsecS r ()
     wrongEq = do
-      P.chunk "="
-      off <- P.getOffset
-      parseFailure off "expected \":=\" instead of \"=\""
+      void (P.single '=')
+      P.failure (Just (P.Tokens (pure '='))) (Set.singleton (P.Tokens (':' :| "=")))
```

janmasrovira: you'll need `import Data.Set qualified as Set`

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
lukaszcz: I fixed this. Now only '=' exactly causes an error.

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
lukaszcz: No longer relevant since the fix above.

janmasrovira (dismissed):

lukaszcz:
There is a small problem that the pretty-printer removes the parentheses around `=` and then the scoping tests fail. I'll try to fix this now.

PR #1717, opened 2023-01-11
make = rule work for every non-nested pattern

janmasrovira:
In the original pr the rules only works for function clauses. I think it is better to modify how patterns are parsed so it works for lambda functions as well.
I also add a test.


lukaszcz (approved):
Yes, this is a better solution.

PR #1724, opened 2023-01-13
Remove wildcard patterns from Internal

janmasrovira:
This pr removes wildcard patterns from the Internal language.
Wildcards are translated to variables with generated names.
This should resolve the issue of having unbound names after inference

### DIFF
--- a/src/Juvix/Compiler/Abstract/Data/Name.hs
+++ b/src/Juvix/Compiler/Abstract/Data/Name.hs
@@ -23,6 +23,15 @@ data Name = Name
 
 makeLenses ''Name
 
+varFromWildcard :: Members '[NameIdGen] r => Wildcard -> Sem r VarName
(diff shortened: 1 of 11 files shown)


src/Juvix/Compiler/Abstract/Data/Name.hs
```diff
@@ -23,6 +23,15 @@ data Name = Name
 
 makeLenses ''Name
 
+varFromWildcard :: Members '[NameIdGen] r => Wildcard -> Sem r VarName
+varFromWildcard w = do
+  _nameId <- freshNameId
+  let _nameText :: Text = "_ω" <> prettyText _nameId
```
paulcadman, line 29: `_ω` 🥇

paulcadman (approved):
makes sense - thanks!

PR #1726, opened 2023-01-15
Fix inference loop

### LINKED ISSUE #1700: Type checking self-application hangs the type-checker
Type checking the definition
```agda
n : Nat;
n := λ{x := x x} id zero;
```
hangs the type-checker.

Note that the following type-checks:
```agda
g : {B : Type} → ({A : Type} → A → A) → B → B;
g x := x x;
```

In general, I think the type-checking bugs are going to crop up until we specify the type system more precisely. I don't know what the type system of Juvix is exactly. It is neither a subset nor a superset of Hindley-Milner. The definition of g is not HM.

janmasrovira:
- Fixes #1700.

I've added a check that prevents cycles when refining holes.

### DIFF
--- a/tests/negative/issue1700/SelfApplication.juvix
+++ b/tests/negative/issue1700/SelfApplication.juvix
@@ -1,5 +1,5 @@
 -- self-application
-module test019;
+module SelfApplication;
 
 open import Stdlib.Prelude;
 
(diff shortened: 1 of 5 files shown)


tests/negative/issue1700/SelfApplication.juvix
```diff
@@ -1,5 +1,5 @@
 -- self-application
-module test019;
```
lukaszcz, line 2: I would also leave this in Compilation/positive/test019.juvix, because we want to later check compilation of this down to WASM.

lukaszcz: I restored the test.

lukaszcz (approved):

Issue #1714, opened 2023-01-10
The translation from Internal to Core doesn't work with type synonyms


lukaszcz:
Trying to evaluate the following via core
```agda
module church;

open import Stdlib.Prelude;

Num : Type;
Num := {A : Type} → (A → A) → A → A;

czero : Num;
czero {_} f x := x;

csuc : Num → Num;
csuc n {_} f := f ∘ n {_} f;

toNat : Num → Nat;
toNat n := n {_} ((+) 1) 0;

main : IO;
main :=
  printNatLn (toNat (csuc (czero)));

end;
```
gives the error
```
juvix: impossible
CallStack (from HasCallStack):
  error, called at src/Juvix/Prelude/Base.hs:315:14 in juvix-0.2.8-DyFVJXJWeol4IAnBe15XnM:Juvix.Prelude.Base
  impossible, called at src/Juvix/Compiler/Core/Translation/FromInternal.hs:526:34 in juvix-0.2.8-DyFVJXJWeol4IAnBe15XnM:Juvix.Compiler.Core.Translation.FromInternal
```

PR #1728, opened 2023-01-16
Do not filter implicit args in internal to core translation

paulcadman:
The internal to core translation was removing implicit arguments from function definitions and applications. This is incorrect as the implicit bindings are required when translating the following (in `csuc`, the binding of the implicit argument is required in an application on the rhs):

```
Num : Type;
Num := {A : Type} → (A → A) → A → A;

csuc : Num → Num;
csuc n {_} f := f ∘ n {_} f;
```

Apart from removing this filter from function and application translation, this required the following changes:

ConstructorInfo:
The _constructorArgsNum field must include the number of type parameters of its inductive type.

PatternConstructorApp:
The pattern arguments must include wildcards for the implicit type parameters passed to the constructor.

BuiltinIf:
The BuiltinIf expression is passed an implicit type argument that must be removed when translating to Core if.

LitString:
A literal string is a function with an implcit type argument. So this must be a translated to a lambda where the type argument is ignored.

Fixes https://github.com/anoma/juvix/issues/1714

### DIFF
--- a/src/Juvix/Compiler/Core/Translation/FromInternal.hs
+++ b/src/Juvix/Compiler/Core/Translation/FromInternal.hs
@@ -276,19 +276,18 @@ goFunctionDef (f, sym) = do
   forM_ mbody (registerIdentNode sym)
   where
     mkBody :: Sem r Node
-    mkBody =
-      if
-          | nPatterns == 0 -> runReader initIndexTable (goExpression (f ^. Internal.funDefClauses . _head1 . Internal.clauseBody))
-          | otherwise ->
-              ( do
-                  let values :: [Node]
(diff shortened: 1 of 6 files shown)


src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -270,23 +278,23 @@ goFunctionDef (f, sym) = do
     mkBody :: Sem r Node
     mkBody =
       if
```
janmasrovira, line 280: use guards instead of `if`

janmasrovira (dismissed):

Issue #1720, opened 2023-01-12
Evaluating strings in the REPL doesn't work


lukaszcz:
On input:
```
Stdlib.Prelude> "Hello world!"
```
one gets:
```
.:1:1-15: error:
Unable to infer the hole
```

PR #1730, opened 2023-01-16
Give a proper type to literal Strings

paulcadman:
Previously literal strings were assigned the magic type `{A : Type} -> A`. Now we have the `BuiltinString` type we can use that to type literal strings instead.

It also means we can now evaluate string literals in the REPL:

```
Juvix REPL version 0.2.8-ff748b9: https://juvix.org. Run :help for help
OK loaded: /Users/paul/heliax/juvix-2023/.juvix-build/stdlib/Stdlib/Prelude.juvix
Stdlib.Prelude> :t "a"
String
Stdlib.Prelude> "a"
"a"
```

Fixes https://github.com/anoma/juvix/issues/1720


janmasrovira (approved):
Awesome, thank you

PR #1732, opened 2023-01-17
Script to count LOC

lukaszcz:


lukaszcz:
Yes, I tried that. But `wc -l` was just easier to parse with `bash` and I
didn't want to spend too much time on this. We can change it later.

lukaszcz:
`cloc` would give better absolute numbers, but it shouldn't change the proportions between front/back/runtime

PR #1735, opened 2023-01-17
Update CI to install Smoke, Github actions, and Makefile fixes

jonaprieto:
This PR adds some maintenance at different levels to the CI config, the Make file, and formatting.

- Most of the actions used by the CI related to haskell, ormolu, hlint and pre-commit have been updated because Github requires NodeJS 16. This change removes all the old warnings related to nodeJs.
In the case of ormolu, the new version makes us format some files that were not formatted before, similarly with hlint.
- The CI has been updated to use the latest version of the Smoke testing framework, which introduced installation of the dependencies for Linux (libicu66) and macOS (icu4c) in the CI. In the case of macOS, the CI uses a binary for smoke. For Linux, we use stack to build smoke from the source. The source here is in a fork of [the official Smoke repo](https://github.com/SamirTalwar/smoke). Such includes some features/changes that are not yet in the official repo.

- The Makefile runs the ormolu and hlint targets using as a path for the binaries the environment variables ORMOLU and HLINT. Thus, export those variables in your environment before running `make check,` `make format` or `make hlint`. Otherwise, the Makefile will use the binaries provided
by `stack`.


paulcadman (dismissed):

paulcadman:
We should document / communicate which versions of hlint and ormolu are now being used by the CI so we can make sure we have the correct versions installed locally. The documentation can be done in a separate PR.

paulcadman (dismissed):
A Herculean labour, good job!

paulcadman (approved):

PR #1737, opened 2023-01-18
Print comments when pretty printing concrete syntax

### LINKED ISSUE #1593: Code formatting removes comment strings
Format a file in emacs containing a comment, the comment is removed.

The emacs formatter uses `juvix dev scope`, you can see the issue in this example:

```
$ cd examples/milestone/HelloWorld
$ cat HelloWorld.juvix
-- HelloWorld.juvix
module HelloWorld;

open import Stdlib.Prelude;

main : IO;
main := putStrLn "hello world!";

end;
$ juvix dev scope HelloWorld.juvix
module HelloWorld;
  open import Stdlib.Prelude;

  main : IO;
  main := putStrLn "hello world!";
end;
```

If we intent to use the scoper output as a formatter, we should preserve the comment strings.

### LINKED ISSUE #1623: Change the name of the Juvix mode format command

janmasrovira:
This pr adds a new pretty printing algorithm for scoped concrete syntax.
This new algorithm works as the old one but will also print the normal comments in the appropriate places.
This is a work in pogress. What remains to be done is printing comments that are in the middle of an expression. Currently these comments will be printed but just before the expression. Also, it should support parsed (pre-scoping) concrete syntax.

To run the new algorithm the flag `--with-comments` must be passed to the `dev scope` command.
```
juvix dev scope M.juvix --with-comments
```
When we finish the pending tasks, we may want to make this the default algorithm and have a `--with-no-comments` flag instead.

I have updated the `juvix-format-buffer` to use the `--with-comments` flag. Now it should be safe (sans the expression quirk mentioned above) to run the formatter within emacs.

I have adapted the scoping tests to also be executed with the new algorithm. These tests make sure that the pretty printed code is sound but do not check that the printed comments are in the appropriate places. For that, I have done some manual testing. In a future pr we should find a way to test it automatically.

- Fixes #1593 
- Closes #1623


juvix-mode/juvix-mode.el
```diff
@@ -69,7 +69,7 @@
         )
     (with-temp-buffer
       (let ((
-             cmd-str (concat "juvix " (if juvix-disable-embedded-stdlib "--no-stdlib " "") "dev scope "
+             cmd-str (concat "juvix " (if juvix-disable-embedded-stdlib "--no-stdlib " "") "dev scope --with-comments "
```
jonaprieto, line 72: Please make this change in https://github.com/anoma/juvix-mode, and remove it from this PR.

src/Juvix/Compiler/Concrete/Data/Highlight/Input.hs
```diff
@@ -23,4 +23,4 @@ filterInput absPth HighlightInput {..} =
     }
 
 filterByLoc :: HasLoc p => Path Abs File -> [p] -> [p]
-filterByLoc p = filter ((== toFilePath p) . (^. intervalFile) . getLoc)
+filterByLoc p = filter ((== p) . (^. intervalFile) . getLoc)
```
jonaprieto, line 26: Why do we have this function here? it doesn't seem the right place to be. It was here before I can see. Maybe we should think to put this in Juvix.Data.Loc, or in Juvix.Extra, somewhere.

janmasrovira: I've moved it to `Juvix.Data.Loc`

juvix-mode/juvix-mode.el
janmasrovira: I'll do it after this gets merged

jonaprieto (approved):

PR #1760, opened 2023-01-24
Translate Nat builtins to the correct Core Ops

paulcadman:
Test.juvix
```
module Test;

open import Stdlib.Prelude;
t : Nat -> Nat -> Nat;
t x y := x + (2 * y);

main : IO;
main := printNatLn (t 1 3);

end;
```

Before this fix:

```
juvix dev internal core-eval -t nat-to-int Test.juvix
```

prints `6` instead of `7` as the `*` is translated to `OpIntAdd` instead of `OpIntMul`.

We can properly test this once the new pipeline is complete.


lukaszcz (approved):
Yeah, I copy-pasted and forgot to change the names.

PR #1762, opened 2023-01-24
Improve arity inference for repl expressions

### LINKED ISSUE #1721: REPL always tries to infer all holes instead of displaying polymorphic types
Typing
```
Stdlib.Prelude> :type null
```
gives the error
```
.:1:1-5: error:
Unable to infer the hole
```
It should display the polymorphic type of `null`.

janmasrovira:
Closes #1721.

Now the arity checker does not add an unnecessary hole and so the repl displays the type properly:
image


paulcadman (dismissed):
wonderful - thank you!

paulcadman (dismissed):

paulcadman (approved):

PR #1764, opened 2023-01-24
Run the new Juvix formatter for all the Juvix examples

jonaprieto:
Juvix now provides an intial and functional formatting tool by calling the command `juvix dev scope file --with-comments` .

This PR adds a new Makefile target `juvix-format` to format all the Juvix programs we showcase in the documentation as Juvix-projects, i.e., files from the `examples` directory. Note the corresponding target in the Makefile also calls the typechecker to ensure the programs do not have type errors introduced by the formatter, considering also that all the Juvix files in the `examples` directory type-checked before this PR. Thus, we should preserve that state. Finally, I included `juvix-format` as part of the `check` target, so we widen the testing of the compiler.

The formatter is not perfect yet, so we need to fix some formatting issues manually.
For example, the end of the line is modified by the formatting. We can fix this by calling after
`make pre-commit`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,8 +154,8 @@ format: clang-format
 clang-format:
 	@cd runtime && ${MAKE} format
 
-TOFORMATFILES = ./examples
-TOFORMAT = $(shell find ${TOFORMATFILES} -name "*.juvix" -print)
+TOFORMATJUVIXFILES = ./examples
+TOFORMAT = $(shell find ${TOFORMATJUVIXFILES} -name "*.juvix" -print)
 
 .PHONY: $(TOFORMAT)
 juvix-format: $(TOFORMAT)
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,18 @@ format: clang-format
 clang-format:
 	@cd runtime && ${MAKE} format
 
+TOFORMATFILES = ./examples
+TOFORMAT = $(shell find ${TOFORMATFILES} -name "*.juvix" -print)
+
+.PHONY: $(TOFORMAT)
+juvix-format: $(TOFORMAT)
+$(TOFORMAT): %:
+	@echo "Formatting $@"
+	@juvix dev scope $@ --with-comments > $@.tmp
+	@mv $@.tmp $@
+	@echo "Typechecking formatted $@"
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -154,6 +154,18 @@ format: clang-format
 clang-format:
 	@cd runtime && ${MAKE} format
 
+TOFORMATFILES = ./examples
+TOFORMAT = $(shell find ${TOFORMATFILES} -name "*.juvix" -print)
```
janmasrovira, line 158: I would prefer `TOFORMATJUVIXFILES` to not get confused

examples/milestone/TicTacToe/CLI/TicTacToe.juvix
```diff
@@ -1,50 +1,62 @@
...
+    c ↦ "readline()";
+  };
+
+  axiom parsePositiveInt : String → Nat;
+
+  compile parsePositiveInt {
+    c ↦ "parsePositiveInt";
+  };
+
+  -- | Reads a ;Nat; from stdin
```
janmasrovira, line 25: this was probably intended to be a judoc comment. Let's replace `-- |` with `---`

examples/milestone/TicTacToe/CLI/TicTacToe.juvix
```diff
@@ -1,50 +1,62 @@
...
+    getMove : Maybe Nat;
+  getMove := validMove (parsePositiveInt (readline));
+
+  do : IO × GameState → GameState;
+  do (_ , s) := playMove getMove s;
+
+  -- | A ;String; that prompts the user for their input
```
janmasrovira, line 32: idem

examples/milestone/TicTacToe/CLI/TicTacToe.juvix
```diff
@@ -1,50 +1,62 @@
...
+    prompt : GameState → String;
+  prompt x := "\n"
+    ++str showGameState x
+    ++str "\nPlayer "
+    ++str showSymbol (player x)
+    ++str ": ";
+
+  -- | Main loop
```
janmasrovira, line 40: idem

janmasrovira (approved):

PR #1765, opened 2023-01-25
Install wasmer binary from Github releases

jonaprieto:
The current Github action responsible for installing Wasmer fails from time to time, and it also is outdated, not following the new NodeJS 16 requirement by Github.

We could use https://github.com/jaxxstorm/action-install-gh-release instead, but unfortunately, it does not have the proper support to expose the binaries contained in an inner folder, as in the case with the Wasmer release. In the meantime, let's use my [fork](https://github.com/jonaprieto/action-install-gh-release) while I open PR to the main repository.


paulcadman (approved):
🚢

PR #1766, opened 2023-01-25
Keep regular comments in html output

### LINKED ISSUE #1723: Non-judoc comments are removed when generating HTML output
This makes it impossible to e.g. write a comment above an import statement.

It also makes it hard to write a tutorial as a Juvix file and generate nice html output for it.

janmasrovira:
- Fixes #1723 
- It refactors parsing/scoping so that the scoper does not need to read files or parse any module. Instead, the parser takes care of parsing all the imported modules transitively.


jonaprieto (approved):
Great work here :)

PR #1771, opened 2023-01-26
Add debugging builtin functions `trace` and `fail`

### LINKED ISSUE #1731: Add debugging builtin functions
In Core there are OpTrace and OpFail builtins (Juvix.Compiler.Core.Language.Builtins; in *.jvc files they are referred to by `trace` and `fail`). We should expose them in the front-end, e.g., by adding builtin functions `trace` and `fail` and defining them as axioms in the `Stdlib.Debug` module.

jonaprieto:
- Closes #1731

### DIFF
--- a/src/Juvix/Compiler/Core/Translation/FromInternal.hs
+++ b/src/Juvix/Compiler/Core/Translation/FromInternal.hs
@@ -365,6 +365,8 @@ goAxiomInductive a = whenJust (a ^. Internal.axiomBuiltin) builtinInductive
       Internal.BuiltinIOSequence -> return ()
       Internal.BuiltinString -> registerInductiveAxiom
       Internal.BuiltinIO -> registerInductiveAxiom
+      Internal.BuiltinTrace -> return ()
+      Internal.BuiltinFail -> return ()
 
     registerInductiveAxiom :: Sem r ()
     registerInductiveAxiom = do
@@ -423,6 +425,9 @@ goAxiomDef a = case a ^. Internal.axiomBuiltin >>= builtinBody of
           )
       Internal.BuiltinString -> Nothing
       Internal.BuiltinIO -> Nothing
+      Internal.BuiltinTrace -> Nothing
+      Internal.BuiltinFail ->
+        Just (mkLambda' (mkLambda' (mkBuiltinApp' OpFail [mkVar' 0])))
 
     axiomType' :: Sem r Type
     axiomType' = runReader initIndexTable (goExpression (a ^. Internal.axiomType))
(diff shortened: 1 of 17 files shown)


src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -491,7 +491,7 @@ lambda = do
 -- Data type construction declaration
 -------------------------------------------------------------------------------
 
-inductiveDef :: Members '[InfoTableBuilder, JudocStash, NameIdGen] r => Maybe (WithLoc BuiltinInductive) -> ParsecS r (InductiveDef 'Parsed)
+inductiveDef :: (Members '[InfoTableBuilder, JudocStash, NameIdGen] r) => Maybe (WithLoc BuiltinInductive) -> ParsecS r (InductiveDef 'Parsed)
```
lukaszcz, line 494: Why do these brackets suddenly appear? Can't we have a separate PR to adapt to the new ormolu version?

src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -423,6 +425,20 @@ goAxiomDef a = case a ^. Internal.axiomBuiltin >>= builtinBody of
...
+                  ( mkLambda'
+                      ( mkLambda'
+                          ( mkBuiltinApp' OpTrace [mkVar' 1, mkVar' 0]
+                          )
+                      )
+                  )
+              )
+          )
+      Internal.BuiltinFail ->
+        Just (mkLambda' (mkLambda' (mkBuiltinApp' OpFail [mkVar' 0])))
```
lukaszcz, line 441: This is wrong: `fail` takes one argument not two.

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
jonaprieto: Hum, I don't know. I run `make check`, which calls `make format`.

src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -423,6 +425,20 @@ goAxiomDef a = case a ^. Internal.axiomBuiltin >>= builtinBody of
           )
       Internal.BuiltinString -> Nothing
       Internal.BuiltinIO -> Nothing
+      Internal.BuiltinTrace ->
+        Just
+          ( mkLambda'
+              ( mkLambda'
+                  ( mkLambda'
+                      ( mkLambda'
+                          ( mkBuiltinApp' OpTrace [mkVar' 1, mkVar' 0]
```
lukaszcz, line 434: Why 4 lambda here? I don't understand this.

Besides, translating `trace` this way is bad for evaluation order. Then `trace 3 (trace 4 unit)` will print "4\n3\n" instead of "3\n\4". I think it's better to translate it the way `if` is translated, with the arguments immediately supplied to `mkBuiltinApp'` -- then the laziness of `trace` with respect to the second argument is preserved.

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
lukaszcz: I run `make format` with ormolu version 0.5.2 and it doesn't add these braces, but also doesn't remove them. Strange.

src/Juvix/Compiler/Core/Translation/FromInternal.hs
jonaprieto: Why not: {a:U} -> String -> a?

lukaszcz: Sorry, OK, this is for type arguments. Makes sense

src/Juvix/Compiler/Core/Translation/FromInternal.hs
jonaprieto: You need for lambdas, because the type I have in mind for trace is {A : Type} -> {B : Type} -> A -> B -> B

lukaszcz (changes requested):
`trace` should be lazy in its second argument. Otherwise, the order of the trace messages is mangled.

src/Juvix/Compiler/Core/Translation/FromInternal.hs
lukaszcz: Yes, sorry, I realised this about the type arguments.

But `trace` should be lazy.

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
jonaprieto: Strange, also because the CI didn't complain about the changes.

janmasrovira: I have the same behaviour as Lukasz. We should try to find a solution for this. Otherwise Jonathan's prs will often include unrelated formatting changes

lukaszcz (approved):

PR #1748, opened 2023-01-23
Basic Geb integration

lukaszcz:
This PR:

- Closes #1647 

It gives compilation errors for language features that require more substantial support (recursion, polymorphism). The additional features are to be implemented in future separate PRs.
* Adds a new target `geb` to the CLI command `juvix dev core compile`, which produces a `*.geb` output file in the `.juvix-build` directory.
* Adds a few tests. These are not yet checked automatically because there is no GEB evaluator; checking the `*.geb` output would be too brittle.

PR #1778, opened 2023-01-27
Support integers in the GEB backend

### LINKED ISSUE #1753: Support integers in the GEB backend
Ultimately, GEB should have a built-in special type of integers. We should just add integers to our Geb datatype and handle them in the evaluator.

lukaszcz:
* Depends on PR #1748 
* Closes #1753


jonaprieto (commented):
There are a few new tests here: test005.x and test006.x.
Where are the actual test for these files, from the Haskell test suite?

lukaszcz:
> There are a few new tests here: test005.x and test006.x. Where are the actual test for these files, from the Haskell test suite?

There aren't any yet, because without the GEB evaluator there is no reliable way to automate the tests! We should add all GEB tests (and write some more) to the Haskell test suite once we have the evaluator.

lukaszcz:
For now, I just checked manually that the GEB output is reasonable.

jonaprieto (approved):

PR #1779, opened 2023-01-27
No run jobs after pr is closed/merged, docs only on pushes to main

jonaprieto:


.github/workflows/ci.yml
```diff
@@ -309,7 +308,7 @@ jobs:
 
   docs:
     needs: build
-    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    if: contains(fromJson('["refs/heads/main"]'), github.ref) || github.event_name == 'workflow_dispatch'
```
paulcadman, line 311: Since the list of branches has only one element I think we can replace this with:

```
if: github.ref == 'refs/heads/main'
```

jonaprieto: Oh true, the example was including a ref to master. Thanks.

paulcadman (approved):

PR #1780, opened 2023-01-27
Fix minor issue with ==% for type equality

jonaprieto:


src/Juvix/Compiler/Abstract/Extra/Functions.hs
```diff
@@ -285,18 +285,21 @@ infix 4 ===
 (===) :: (IsExpression a, IsExpression b) => a -> b -> Bool
 a === b = (toExpression a ==% toExpression b) mempty
 
-infix 4 ==%
-
 -- TODO: make it symmetric
```
janmasrovira, line 288: let's remove this TODO

janmasrovira (approved):

PR #1783, opened 2023-01-27
Use restore/save github action to speed up the CI testing

### LINKED ISSUE #1776: Make it faster the haskell building process for the task job in the CI
The CI test task is redoing the job done by the build task, building the Haskell dependencies, including the codebase in `src`. We should find a way to reuse the work done by `build`. 

One option seems to share artifacts between jobs. 

- https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

jonaprieto:
This PR addresses a caching issue in our CI by streamlining each operating system's build and test processes, reducing CI time. 🤞  Also, our caching strategy has been updated with the new restore/save actions. For example, we aim to cache the .stack folder, and if the stack build is successful, the .stack-build. The building documentation job continues depending on the Linux build. Upon merging this PR, we get back to the point where the CI maintain a cache for each OS to be shared among all PRs, significantly reducing CI testing time. The expected scenario is as follows. The CI can take, on average, 35' in Linux to build and test everything. Using caching, that time is reduced to less than 10'. macOS is a different story. It can easily take one hour, and even more, the first time to build and test the project. After that, it might take an average of 20'.

- Caching strategies [descriptions](https://github.com/actions/cache/blob/main/caching-strategies.md#saving-cache-even-if-the-build-fails)
- Closes #1776

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -344,7 +344,7 @@ jobs:
         run: |
           echo "WASI_SYSROOT_PATH=$GITHUB_WORKSPACE/wasi-sysroot" >> $GITHUB_ENV
       - name: Install the latest Wasmer version
-        uses: jonaprieto/action-install-gh-release@1
+        uses: jonaprieto/action-install-gh-release@2
         with:
           repo: wasmerio/wasmer
           tag: latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -163,6 +163,14 @@ jobs:
           cd main
           make build
 
+      - name: Archive Stack artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: stack-artifacts
+          path: |
+            ~/.stack~
```
paulcadman, line 171:
```suggestion
            ~/.stack
```

paulcadman (dismissed):

paulcadman (approved):

Issue #1742, opened 2023-01-19
Adapt the examples, tests and the standard library to work with the new pipeline


lukaszcz:
This involves removing the `compile` and `foreign` blocks and using builtin functions instead. Perhaps a few more builtins need to be added and the standard library extended by a few functions.

PR #1784, opened 2023-01-30
String builtins

lukaszcz:
- Progress for #1742 
* Adds builtin primitives for operations on strings and removes the corresponding foreign & compile blocks.


janmasrovira (commented):
There seems to be a lot of C format changes unrelated to the pr. Could we separate those in another pr?

lukaszcz:
> There seems to be a lot of C format changes unrelated to the pr. Could we separate those in another pr?

Okay, but we'll be removing the old C runtime code anyway (that's where the reformatting is, because I have "format on save" in VSCode).

runtime/src/juvix/api.h
```diff
@@ -76,6 +76,21 @@
...
+#define JUVIX_STR_CONCAT(var0, var1, var2) CONCAT_CSTRINGS(var0, var1, var2)
+
+#define JUVIX_STR_TO_INT(var0, var1) \
+    (var0 = make_smallint(strtoint(get_cstring(var1))))
+
+#define JUVIX_SHOW(var0, var1)                                     \
+    do {                                                           \
+        char *juvix_buf = print(var1);                             \
+        if (strlen(juvix_buf) / sizeof(word_t) + 1 > MAX_FIELDS) { \
+            error_exit_msg("show: string too long");               \
```
janmasrovira, line 88: how long would a string need to be to cause this crash?

janmasrovira (commented):
great

runtime/src/juvix/object/cstring.c
```diff
@@ -27,3 +33,31 @@ word_t alloc_cstring(const char *str) {
...
+    while (*str == '-' || *str == '+') {
+        if (*str == '-') {
+            puiss = puiss * -1;
+        }
+        str++;
+    }
+    while (*str >= '0' && *str <= '9') {
+        result = result * 10 + (*str - '0');
+        str++;
+        if (result > 107374180) {
```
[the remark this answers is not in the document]
lukaszcz: Yes, you're right. I fixed this.

runtime/src/juvix/api.h
lukaszcz: Not very long, something around 1000 characters. We will need to redo the string implementation later anyway. Now it's just some basic thing that works.

src/Juvix/Compiler/Core/Translation/FromSource.hs
```diff
@@ -511,6 +511,9 @@ builtinAppExpr varsNum vars = do
       <|> (kw kwMinus >> return OpIntSub)
       <|> (kw kwDiv >> return OpIntDiv)
       <|> (kw kwMul >> return OpIntMul)
+      <|> (kw kwShow >> return OpShow)
```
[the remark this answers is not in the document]
lukaszcz: You meant $>, I guess? I changed it, but in my opinion this became less readable after the change. I don't remember all those funny operators, except the basic ones. Always need to look them up. In my opinion, Haskell would be better without them.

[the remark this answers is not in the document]
janmasrovira: yes, I meant `$>`.
I agree that we must be careful with using too many fancy operators. For me `$>` is acceptable, but we can discuss together if we prefer `$>` or `>> return`

runtime/src/juvix/limits.h
```diff
@@ -52,6 +52,12 @@
 
 #define MAX_CLOSURE_SIZE (MAX_FUNCTION_ARGS + CLOSURE_HEAD_SIZE)
 
+#ifdef BITS32
+#define MAX_SMALLINT 1073741823L
```
janmasrovira, line 56: Is it possible to replace the literal by `1<<30 - 1`?

runtime/src/juvix/limits.h
```diff
@@ -52,6 +52,12 @@
...
+#else
+#define MAX_SMALLINT 4611686018427387903LL
```
janmasrovira, line 58: idem

runtime/src/juvix/limits.h
lukaszcz: Then one can't use it in a 'case', but it's probably fine for this one.

lukaszcz: I'll change it

lukaszcz: Done

runtime/src/juvix/limits.h
lukaszcz: Done

janmasrovira (approved):

janmasrovira (approved):

PR #1785, opened 2023-01-30
Allow type signatures to have a body

### LINKED ISSUE #1637: Typing for variables in let expressions
Instead of:

```haskell
let { x : A;
      x := y } in z;
```
We want to type:

```haskell
let x : A := y in z;
```

janmasrovira:
- Closes #1637.

A function type signature is now allowed to have a body. This is valid for both top level and let definitions.
```
not : Bool -> Bool := λ {
 | true := false
 | false := true
};
```


paulcadman (approved):
Looks good 🌟

PR #1787, opened 2023-01-30
Fix termination with as-patterns

### LINKED ISSUE #1702: Termination checking fails with as-patterns
The following fails the termination check:
```haskell
type Tree := leaf : Tree | node : Tree → Tree → Tree;

gen : Nat → Tree;
gen zero := leaf;
gen (suc zero) := node leaf leaf;
gen (suc n@(suc n')) := node (gen n') (gen n);
```
The following is accepted:
```haskell
type Tree := leaf : Tree | node : Tree → Tree → Tree;

gen : Nat → Tree;
gen zero := leaf;
gen (suc zero) := node leaf leaf;
gen (suc (suc n')) := node (gen n') (gen (suc n'));
```

janmasrovira:
- Fixes #1702.
Now, if we have a pattern like `suc @n(suc n')`, then `n` is detected to be a proper subpattern

### DIFF
--- a/src/Juvix/Compiler/Abstract/Extra/Functions.hs
+++ b/src/Juvix/Compiler/Abstract/Extra/Functions.hs
@@ -45,6 +45,20 @@ patternCosmos f p = case p of
       args' <- traverse (traverseOf patternArgPattern (patternCosmos f)) args
       pure (PatternConstructorApp (ConstructorApp r args'))
 
+patternArgNameFold :: SimpleFold (Maybe Name) Pattern
+patternArgNameFold f = \case
+  Nothing -> mempty
+  Just n -> Const (getConst (f (PatternVariable n)))
+
+-- | A fold over all transitive children, including self
+patternArgCosmos :: SimpleFold PatternArg Pattern
+patternArgCosmos f p = do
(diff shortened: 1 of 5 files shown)


src/Juvix/Compiler/Abstract/Extra/Functions.hs
```diff
@@ -45,14 +45,28 @@ patternCosmos f p = case p of
       args' <- traverse (traverseOf patternArgPattern (patternCosmos f)) args
       pure (PatternConstructorApp (ConstructorApp r args'))
 
+patternArgNameFold :: SimpleFold (Maybe Name) Pattern
+patternArgNameFold f = \case
+  Nothing -> mempty
+  Just n -> Const (getConst (f (PatternVariable n)))
```
paulcadman, line 51: I probably don't understand but isn't `Const . getConst` the identity? Or are there different `Const`s here?

janmasrovira: it is needed to get the types right. It differs from identity in the type:
`const . getConst :: Const r a -> Const r b`

paulcadman: Thank you

PR #1790, opened 2023-01-31
Remove braces from let expressions

### LINKED ISSUE #1786: Remove braces for let expressions
Currently the syntax for let expression is as follows:
```
let {
  stmt1;
  stmt2;
 ...
} in body;
```
Since both `let` and `in` are reserved keywords, the braces are not necessary for parsing, thus, it would make sense to remove the braces.

janmasrovira:
- Closes #1786

### DIFF
--- a/tests/positive/StdlibList/Data/List.juvix
+++ b/tests/positive/StdlibList/Data/List.juvix
@@ -12,21 +12,21 @@ match : {A : Type} → {B : Type} → A → (A → B) → B;
 match x f := f x;
 
 foldr : (a : Type) → (b : Type) → (a → b → b) → b → List a → b;
-foldr _ _ _ z (nil _) := z;
-foldr a b f z (:: _ h hs) := f h (foldr a b f z hs);
+foldr _ _ _ z nil := z;
+foldr a b f z (:: h hs) := f h (foldr a b f z hs);
 
 foldl : (a : Type) → (b : Type) → (b → a → b) → b → List a → b;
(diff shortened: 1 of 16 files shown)


tests/positive/StdlibList/Data/List.juvix
```diff
@@ -12,90 +12,88 @@ match : {A : Type} → {B : Type} → A → (A → B) → B;
...
+splitAt a (suc zero) (:: x xs) := , (:: x nil) xs;
+splitAt a (suc (suc m)) (:: x xs) := match (splitAt a m xs) λ {
+    (, xs' xs'') := , (:: x xs') xs''
   };
 
-merge : (a : Type) → (a → a → Ordering) → List a → List a → List a;
-merge a cmp (:: _ x xs) (:: _ y ys) := match (cmp x y) λ {
-    | LT := :: a x (merge a cmp xs (:: a y ys))
-    | _ := :: a y (merge a cmp (:: a x xs) ys)
+terminating merge : (a : Type) → (a → a → Ordering) → List a → List a → List a;
```
paulcadman, line 70: Why do we need `terminating` here now? Was the test not checking termination before?

janmasrovira: all scoping tests are not checked for termination nor type checked. Hence, the changes in this file were unnecessary, but since I was changing the file anyway for the let syntax, I thought it was worth to make it type check.

paulcadman (approved):
🚢

PR #1795, opened 2023-02-01
Adapt benchmarks to the new pipeline

### LINKED ISSUE #1742: Adapt the examples, tests and the standard library to work with the new pipeline

lukaszcz:
Closes #1742


paulcadman (approved):
Looks fine - we will see any issues when the benchmarks run on the nightly build.

lukaszcz:
> Looks fine - we will see any issues when the benchmarks run on the nightly build.

Should be fine. I checked the Juvix programs compile and give the same results as before.

PR #1797, opened 2023-02-01
CI Haskell maintenance

jonaprieto:
This PR does CI maintenance consisting of the following:

- Use https://github.com/freckle/stack-action to build and test the Haskell codebase
- Update Gh-release action: https://github.com/jaxxstorm/action-install-gh-release/pull/39 is merged into the main branch, so we return to the original GitHub action.
- Simplifies pre-commit action. Use the `SKIP` environmental variable to avoid running ormolu in the CI. 
- The ormolu Github action now supports version input since v11. Fixed to use `Ormolu v0.5.2`.
  - https://github.com/mrkkrp/ormolu-action/issues/26),

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,16 +251,6 @@ jobs:
           brew install icu4c
           brew link icu4c --force
 
-      - name: Make runtime
-        run: |
-          cd main
-          make runtime
-
-      - name: Stack setup
-        uses: freckle/stack-action@v4
-        with:
-          working-directory: main
-          test: false
 
       - name: Download and extract wasi-sysroot
         run: >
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,20 +21,19 @@ concurrency:
   group: '${{ github.workflow }}-${{ github.head_ref || github.run_id }}'
   cancel-in-progress: true
 env:
-  STACKFLAGS: '--pedantic --stack-root ${{ github.workspace }}/.stack'
+  SKIP: ormolu
```
paulcadman, line 24: Will the `--pedantic` flag still be passed to the Juvix build? I think this was useful to have.

paulcadman (commented):
I just noticed

.github/workflows/ci.yml
jonaprieto: It's true by default using the frakle action.
https://github.com/freckle/stack-action

paulcadman: Perfect!

jonaprieto:
> How is the `SKIP` environment variable set for a build?

Locally? You do nothing. Here, we need to set it as a list of hook ids, the list is separated by commas. So, pre-commit skips those hooks. For example, 


```
SKIP=ormolu make pre-commit
```

PR #1802, opened 2023-02-02
Short syntax for sequences of function and datatype parameters

### LINKED ISSUE #1692: It should be possible to specify multiple implicit type arguments at once
For example, it would be convenient to write
```
S : {A B C : Type} → (A → B → C) → (A → B) → A → C;
```
Now one needs to write the more verbose
```
S : {A : Type} → {B : Type} → {C : Type} → (A → B → C) → (A → B) → A → C;
```
I think it is inconvenient for serious use, to spell this out every time. There should be a shorter form for implicit type arguments.

### LINKED ISSUE #1801: Remove the usage annotation syntax
I just discovered that one can write e.g.
```
f : {A :0 Type} -> Nat;
```
and this is accepted by the parser, only later the typechecker raises an exception.

I think we should remove the usage annotations, since there are no plans to support them anytime soon. I think the code base shouldn't contain some partial implementations of features we decided long ago we might want to have, but didn't really go through with them.

lukaszcz:
* Closes #1692 
* Closes #1801

### DIFF
--- a/src/Juvix/Compiler/Abstract/Extra/Functions.hs
+++ b/src/Juvix/Compiler/Abstract/Extra/Functions.hs
@@ -122,16 +122,13 @@ matchFunctionParameter ::
   Sem r ()
 matchFunctionParameter pa pb = do
   goParamName (pa ^. paramName) (pb ^. paramName)
-  goParamUsage (pa ^. paramUsage) (pb ^. paramUsage)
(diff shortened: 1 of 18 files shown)


lukaszcz:
This will be implemented in another pull request with a different concept -- modifying the `Function` type in the concrete syntax.

PR #1812, opened 2023-02-07
Make '>>' lazy

### LINKED ISSUE #1773: Add a lazy IO sequencing function  (#1772)
It would be beneficial in some cases to consider a lazy version of the sequencing operator (>>) to not lose some output/precomputations in case the program crashes. For example, assuming there exists a "fail" function (PR #1771 ). Currently, we only get the error message "Something's wrong", where one could have been interested in seeing the output "Answer" as well.

```agda
module builtinFail;
  open import Stdlib.Prelude;

  builtin fail axiom fail : {A : Type} → String → A;

  main : IO;
  main := printStringLn "Answer"
    >> fail "Something's wrong";
end;
```

### LINKED ISSUE #1817: Repl: undeclared identifier
Specific to #1812

Running `main` in the repl of the following module results in an error.
```
module bug;

open import Stdlib.Prelude public;

p : IO × IO;
p := printNatLn 1 , printNatLn 3;

main : IO;
main := uncurry (>>) p;
-- main := printNatLn 1 >> printNatLn 2;

end;
```

The error reads as
```
internal to core: undeclared identifier: >>
  error, called at src/Juvix/Prelude/Base.hs:307:9 in juvix-0.2.9-3NSf0VpZgBTIYxwwS1qUzW:Juvix.Prelude.Base
  error, called at src/Juvix/Compiler/Core/Translation/FromInternal.hs:657:20 in juvix-0.2.9-3NSf0VpZgBTIYxwwS1qUzW:Juvix.Compiler.Core.Translation.FromInternal
```

If the commented line of `main` is used, then no error occurs.

lukaszcz:
* Closes #1773 
* Closes #1817


src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -633,6 +624,12 @@ goExpression' = \case
       varsNum <- asks (^. indexTableVarsNum)
       return (mkVar (setInfoLocation (n ^. nameLoc) (Info.singleton (NameInfo (n ^. nameText)))) (varsNum - k - 1))
     Internal.IdenFunction n -> do
+      funInfoBuiltin <- getFunctionBuiltinInfo n
+      case funInfoBuiltin of
+        Just Internal.BuiltinBoolIf -> error "if must be called with 3 arguments"
```
janmasrovira, line 629: These should not be developer errors. We should throw a `JuvixError` before or at typechecking.
If you prefer it, we can address it in a separate pr

src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -771,3 +778,17 @@ goLiteral intToNat l = case l ^. withLocParam of
   where
     mkLitConst :: ConstantValue -> Node
     mkLitConst = mkConstant (Info.singleton (LocationInfo (l ^. withLocInt)))
+
+getAxiomBuiltinInfo :: Member (Reader Internal.InfoTable) r => Name -> Sem r (Maybe BuiltinAxiom)
+getAxiomBuiltinInfo n = do
+  maybeAxiomInfo <- HashMap.lookup n <$> asks (^. Internal.infoAxioms)
+  case maybeAxiomInfo of
```
janmasrovira, line 785: lift out `return`

src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -771,3 +778,17 @@ goLiteral intToNat l = case l ^. withLocParam of
...
+    Just axiomInfo -> return $ axiomInfo ^. Internal.axiomInfoBuiltin
+    Nothing -> return Nothing
+
+getFunctionBuiltinInfo :: Member (Reader Internal.InfoTable) r => Name -> Sem r (Maybe BuiltinFunction)
+getFunctionBuiltinInfo n = do
+  maybeFunInfo <- HashMap.lookup n <$> asks (^. Internal.infoFunctions)
+  case maybeFunInfo of
+    Just funInfo -> return $ funInfo ^. Internal.functionInfoDef . Internal.funDefBuiltin
```
janmasrovira, line 793: idem

tests/smoke/Commands/repl.smoke.yaml
```diff
@@ -233,13 +233,14 @@ tests:
...
   - name: repl-fail
     command:
-      - juvix 
+      - juvix
       - repl
     args:
       - positive/issue1731/builtinFail.juvix
     stdin: main
     stdout:
-      contains: builtinFail> builtinFail>
```
janmasrovira, line 242: why this change?

src/Juvix/Compiler/Core/Translation/FromInternal.hs
```diff
@@ -719,14 +721,19 @@ goApplication a = do
...
         Just Internal.BuiltinBoolPrint -> app
         Just Internal.BuiltinString -> app
         Just Internal.BuiltinIO -> app
-        Just Internal.BuiltinIOSequence -> app
+        Just Internal.BuiltinIOSequence -> do
+          as <- exprArgs
+          case as of
+            (arg1 : arg2 : xs) ->
+              return (mkApps' (mkConstr' (BuiltinTag TagBind) [arg1, mkLambda' (shift 1 arg2)]) xs)
+            _ -> error ">> must be called with 2 arguments"
```
janmasrovira, line 736: iirc `io-sequence` is not necessarily called `>>`. We should use the proper registered string.

The same goes for `||`, `&&`, etc.

janmasrovira (changes requested):

src/Juvix/Compiler/Core/Translation/FromInternal.hs
lukaszcz: Yes, I was intending to make a separate issue for this. I think these should be checked earlier - during type checking or arity checking. I did it this way because in the internal-to-core translation there was already a check like this for `if`, etc. at different places.

src/Juvix/Compiler/Core/Translation/FromInternal.hs
lukaszcz: Done

src/Juvix/Compiler/Core/Translation/FromInternal.hs
lukaszcz: Done

src/Juvix/Compiler/Core/Translation/FromInternal.hs
lukaszcz: Let's address this when we do proper error checking earlier in the pipeline? I think these errors here should ultimately only be internal - the correct application arity should've been checked earlier.

janmasrovira (approved):

PR #1816, opened 2023-02-07
Allow match branches to have different number of arguments

### LINKED ISSUE #1706: Functions with clauses with differing number of patterns type-check but cannot be compiled or evaluated
The following module type checks:
```agda
module bug;

open import Stdlib.Prelude public;

f : Nat → Nat → Nat;
f zero x := x;
f (suc n) := λ{x := x};

end;
```
But trying to evaluate e.g. `f 1 2` in the REPL gives an error:
```
evaluation error: the number of patterns doesn't match the number of arguments: match suc (suc zero), suc zero with {
  x, zero ↦ x;
  suc n ↦ λ? match suc zero with {
    x ↦ x
  }
}
```
Trying to compile the file gives an error:
```
juvix: impossible
  error, called at src/Juvix/Prelude/Base.hs:313:14 in juvix-0.2.8-DrlqdRy1M7F1IgVIET4zbg:Juvix.Prelude.Base
  impossible, called at src/Juvix/Compiler/Backend/C/Translation/FromInternal.hs:386:32 in juvix-0.2.8-DrlqdRy1M7F1IgVIET4zbg:Juvix.Compiler.Backend.C.Translation.FromInternal
```

janmasrovira:
- Closes #1706 

This pr drops the assumption in the Core evaluator that the number of patterns in each of the branches of a match node should match the number of arguments of the match. Instead, each of the branches is allowed to have less patterns than the number of arguments provided.


test/Compilation/Positive.hs
```diff
@@ -238,5 +238,10 @@ tests =
       "Applications with lets and cases in function position"
       $(mkRelDir ".")
       $(mkRelFile "test037.juvix")
-      $(mkRelFile "out/test037.out")
+      $(mkRelFile "out/test037.out"),
+    posTest
+      "Match with more arguments than patterns"
+      $(mkRelDir ".")
```
paulcadman, line 244: The evaluator doesn't currently run on these tests. (The match to case compilation PR will enable them). Did you test this manually?

janmasrovira: I tested it in the repl

tests/Compilation/positive/issue1706.juvix
```diff
@@ -0,0 +1,10 @@
+module issue1706;
+  open import Stdlib.Prelude public;
+
+  f : Nat → Nat → Nat;
```
paulcadman, line 4: Hypothesis: This will break if the `f (suc n)` appears first because the internal to core translation uses the head clause to determine the number of arguments (similar issue for lambda clauses). 

So add a test where the first clause has fewer arguments that a subsequent clause.

lukaszcz:
I don't think changing the semantics of the evaluator solves the problem, because the "proper" pattern compilation algorithm assumes the number of patterns in each branch to be the same. Either this should be fixed during the Internal-to-Core translation or as a separate transformation in Core which ensures that this assumption holds by eta-expanding or something.

PR #1833, opened 2023-02-09
Add `dev core from-source` command

janmasrovira:
This pr adds a new command that will be useful to debug the translation from source to core.

The available flags are:
```
 -t,--transforms [Transform]
                           hint: use autocomplete
  --show-de-bruijn         Show variable de Bruijn indices
  --eval                   Evaluate the main function
  --no-io                  Don't interpret the IO effects
  JUVIX_FILE               Path to a .juvix file
  -s,--symbol-name NAME    Print/eval a specific function identifier (default
                           for eval: main)
```

by default (i.e. if the `--eval` flag is not given), it will print the core node of all functions in the input file. If the `--eval` flag is given it will evaluate the `main` function.
The flag `-s` can be used if you only want to print or eval a particular function different than `main`.


paulcadman:
We should also remove `dev internal core-eval` at the same time as this command provides a superset of its features. And add/replace the relevant smoke tests.

janmasrovira:
> We should also remove `dev internal core-eval` at the same time as this command provides a superset of its features. And add/replace the relevant smoke tests.

done

app/Commands/Dev/Core/FromSource.hs
```diff
@@ -0,0 +1,72 @@
...
+import Juvix.Compiler.Core.Transformation qualified as Core
+import Juvix.Compiler.Core.Translation
+
+runCommand :: forall r. Members '[Embed IO, App] r => CoreFromSourceOptions -> Sem r ()
+runCommand localOpts = do
+  tab <- (^. coreResultTable) <$> runPipeline (localOpts ^. coreFromSourceInputFile) upToCore
+  path :: Path Abs File <- someBaseToAbs' (localOpts ^. coreFromSourceInputFile . pathPath)
+  let tab' :: InfoTable = Core.applyTransformations (project localOpts ^. coreFromSourceTransformations) tab
+
+      inInputModule :: IdentifierInfo -> Bool
```
paulcadman, line 17: good! - I had missed this check in the previous implementation.

paulcadman:
I'll approve this after the name change.

janmasrovira:
> I'll approve this after the name change.

done

paulcadman (approved):

PR #1834, opened 2023-02-09
The formatter respects the ascii function arrow 

### LINKED ISSUE #1827: Formatter should not transform ASCII symbols to unicode by default
Formatting the following:

```
f : Nat -> Nat;
f _ := 1;
```

yields:

```
f : Nat → Nat;
f _ := 1;
```

Specifically the ASCII `->` symbol has been transformed to the unicode `→`.

The user should be able to choose if they want these kind of non-whitespace transformations applied. By default they should not be applied.

janmasrovira:
- Closes #1827


paulcadman (approved):
I tried this locally - it works 🎉

PR #1835, opened 2023-02-10
Fix `juvix init`

### LINKED ISSUE #1831: `juvix init` generates an invalid Juvix package file
Steps to reproduce:

```
$ mkdir foo
$ cd foo
$ juvix init
...
creating juvix.yaml
you are all set
```

The `juvix.yaml` package file contains:

```
_packageDependencies: []
_packageName: foo
_packageVersion: 0.0.0
```

The expected `juvix.yaml` file should be:

```
dependencies: []
name: foo
version: 0.0.0
```

Also - do we want to set `dependencies` to empty by default?

janmasrovira:
- Closes #1831 

My guess is that the behaviour of `aeson` changed at some point.
Before, providing `toEncoding` was working as expected. Now I needed to implement `toJSON` as well.

### DIFF
--- a/app/Commands/Init.hs
+++ b/app/Commands/Init.hs
@@ -39,16 +39,17 @@ checkNotInProject =
       say "You are already in a Juvix project"
       embed exitFailure
 
-getPackage :: forall r. (Members '[Embed IO] r) => Sem r Package
+getPackage :: forall r. Members '[Embed IO] r => Sem r Package
 getPackage = do
   tproj <- getProjName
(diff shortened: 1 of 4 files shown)


jonaprieto:
Can we have some form of testing for this, possibly reusing the example given in:
- https://github.com/anoma/juvix/issues/1831

janmasrovira:
> Can we have some form of testing for this, possibly reusing the example given in:
> 
> * [`juvix init` generates an invalid Juvix package file #1831](https://github.com/anoma/juvix/issues/1831)

I added a smoke test

PR #1852, opened 2023-02-17
Use APE mechanism to format Function expressions

### LINKED ISSUE #1850: Avoid line breaks in applications within a type signature
The result of formatting:

```
exampleFunction : {A : Type} -> List A -> List A -> List A -> List A -> List A -> List A -> List A -> Nat;
```

is currently:

```
exampleFunction : {A : Type} -> List A -> List A -> List
      A -> List A -> List A -> List A -> List A -> Nat;
```

This type signature would be clearer if there was no line break between `List` and `A`.

paulcadman:
I paired with @janmasrovira on this work.

Before this change - long type signatures were formatted to contain line breaks within applications:

```
  exampleFunction : {A : Type} -> List A -> List A -> List
        A -> List A -> List A -> List A -> List A -> Nat;
```

After this change we treat `->` and an infix operator and format like other infix applications:

```
  exampleFunction :
    {A : Type}
      -> List A
      -> List A
      -> List A
      -> List A
      -> List A
      -> List A
      -> List A
      -> Nat;
```

* Fixes #1850


jonaprieto:
Great!

janmasrovira (approved):

PR #1862, opened 2023-02-23
Remove the old C backend

### LINKED ISSUE #1799: Remove the old C backend
* Depends on PR #1556
* `foreign` and `compile` blocks should be removed from the language. They cannot be easily adapted to the new pipeline. Ultimately, we want to have an FFI, but it requires more thought and discussion on how to do it.
* There seem to be many test files that aren't used even now, or haven't been updated. For example, the brittle tests for lambda-lifting are still in the repository, and some tests still have `compile` blocks for the `ghc` backend. Removing the old C backend is a good opportunity to clean up the test files.
* The examples not (yet) usable with the new pipline should be removed (web application, Anoma).
* The Validity Predicate example won't work with the new pipeline until 0.5 when we have proper CosmWasm / Anoma / Namada integration. It should be removed and brought back in 0.5 with a version for each supported WASM backend.
* The documentation should be updated to reflect the changes.

lukaszcz:
* Depends on PR #1832 
* Closes #1799 
* Removes Backend.C.Translation.FromInternal
* Removes `foreign` and `compile` blocks
* Removes unused test files
* Removes the old C runtime
* Removes other dead code


janmasrovira (approved):
:wave:

paulcadman (approved):

PR #1864, opened 2023-02-23
Jan compilation pipeline

janmasrovira:
Fixes test049
The indices in the binders of the lambdas introduced by lambda lifting are now handled properly.


lukaszcz (changes requested):
More tests fail than before, so it's not fixing things.

Run: `stack test --fast --ta '-p "JuvixCore tests"'`

The tests on "Dependent lambda-abstractions" and "Lifting and polymorphism" should all succeed after fixing the issues with lambda-lifting.

Manually checking lambda-lifting on "Lifting and polymorphism" (test049.jvc) confirms that it's not correct.

lukaszcz (approved):

PR #1866, opened 2023-02-27
Add k offset when computing binder offset when capturing free variables

janmasrovira:
The test included in this pr fail during lambda lifting. This pr includes the fix for that.


lukaszcz:
I don't understand what this PR fixes exactly. The name of the test is "free variable capture with polymorphic binders", but the test doesn't have any variable capture.

Could you explain this in a comment? Or maybe it's possible to have a test without the "match", so that all errors are entirely due to lambda-lifting and it's more visible what's happening?

lukaszcz:
> I don't understand what this PR fixes exactly. The name of the test is "free variable capture with polymorphic binders", but the test doesn't have any variable capture.
> 

Ah, ok, you meant the function `captureFreeVars`, not variable shadowing in the source code. I'll change the name and merge, but I still don't completely understand the problem.

PR #1868, opened 2023-02-28
Add transformation group synonyms in `core read -t`

janmasrovira:
This pr adds the option to pass the arguments `pipeline-eval`, `pipeline-geb`,  `pipeline-stripped` to `core read -t`. These will desugar as expected to the corresponding sequences of transformations.

This pr also removes the default `pipeline-eval` if the user does not provide an explicit transformation. Instead, no transformations will be applied by default. To have the same behavour as now, the option `-t pipeline-eval` will need to be passed

### DIFF
--- a/app/CommonOptions.hs
+++ b/app/CommonOptions.hs
@@ -8,7 +8,6 @@ where
 
 import Control.Exception qualified as GHC
 import Juvix.Compiler.Core.Data.TransformationId.Parser
-import Juvix.Compiler.Core.Pipeline qualified as Core
 import Juvix.Prelude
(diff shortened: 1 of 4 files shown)


paulcadman (approved):
Very useful - thanks! This will also apply to `core from-concrete -t` because it uses the same transformation parser? If not I will add it.

lukaszcz (approved):

janmasrovira:
> Very useful - thanks! This will also apply to `core from-concrete -t` because it uses the same transformation parser? If not I will add it.

Yes! it also applies there, I forgot about it

PR #1875, opened 2023-03-02
Print JuvixCore correctly

### LINKED ISSUE #1841: Print JuvixCore programs in such a way that they can be parsed back
It would be helpful for debugging to be able to print the result of translating a Juvix program to JuvixCore and then separately load and evaluate it. Right now, the printing of JuvixCore programs uses a format which cannot be parsed back by the Core parser (missing identifier names, printing of builtin inductive types, symbol ids printed in identifiers definitions but not references).

lukaszcz:
Print JuvixCore InfoTable in such a way that it can be parsed back by the JuvixCore parser.

* Depends on PR #1832 
* Depends on PR #1862 
* Closes #1841 
* Adds "JuvixCore print" tests which read the files from Core/positive/*.jvc, print them, read them back and check if the evaluation results are preserved.


src/Juvix/Compiler/Core/Extra/Utils/Base.hs
janmasrovira, line 1: See pr #1885

janmasrovira (approved):

janmasrovira (approved):

PR #1877, opened 2023-03-04
Print the whole table when no specific node is given

janmasrovira:
Currently `juvix dev core from-concrete` only prints functions and ignores the types when no node name is passed with the `-s` flag. This pr makes it so the whole info table is printed.


lukaszcz:
I cherry-picked your commits into PR #1875

janmasrovira:
Closing as this was included in https://github.com/anoma/juvix/pull/1875

PR #1883, opened 2023-03-14
Make keyword `end` optional for top modules

janmasrovira:
Changes in this PR:
1. The keyword `end` is now optional for top level modules. 
2. When pretty printing, `end` is never printed for top level modules.
3. When pretty printing, top level modules do not have indentation.

These changes only affect `Concrete` syntax.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Print/Base.hs
+++ b/src/Juvix/Compiler/Concrete/Print/Base.hs
@@ -69,6 +69,7 @@ instance SingI t => PrettyPrint (Module 'Scoped t) where
         <> line
         <> topSpace
         <> moduleBody'
+        <> line
         <>? ending
     where
       topSpace :: Sem r ()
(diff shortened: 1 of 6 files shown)


src/Juvix/Compiler/Concrete/Print/Base.hs
```diff
@@ -59,23 +59,33 @@ ppModulePathType x = case sing :: SStage s of
...
+      topSpace = case sing :: SModuleIsTop t of
+        SModuleLocal -> mempty
+        SModuleTop -> line
+
+      localIndent :: Sem r () -> Sem r ()
+      localIndent = case sing :: SModuleIsTop t of
+        SModuleLocal -> indent
+        SModuleTop -> id
+
+      ending :: Maybe (Sem r ())
```
paulcadman, line 85: You don't need `Maybe` here? Or are you trying to keep it the same as The `Pretty/Base.hs` version?

paulcadman (approved):
I think this is a good improvement! 1 minor comment

src/Juvix/Compiler/Concrete/Print/Base.hs
janmasrovira: Fixed. I forgot to remove the `Maybe`, thanks!

paulcadman (approved):

Issue #1777, opened 2023-01-27
Add a format command


jonaprieto:
Let's make `format` the command out of `dev scope` with a few extra options.

- `--check` which only checks if the file is formatted
- `--in-place` to replace the file contents with the juvix formatting output

We can then use this in the Juvix stdlib CI to bring more uniformity to that codebase.

janmasrovira:
Once this is done, I think we should ensure that all juvix test files are formatted

paulcadman:
Recommended safe behaviour of `--in-place` (from `gofmt`) to avoid data loss:

* Write to the file only if the file's formatting is different from `juvix format`'s version.
* If an error occurs during overwriting, the original file is restored from an automatic backup.

PR #1886, opened 2023-03-14
Add `juvix format` command

paulcadman:
This PR adds `juvix format` that can be used to format either a single Juvix file or all files in a Juvix project. 

## Usage

```
$ juvix format --help
Usage: juvix format JUVIX_FILE_OR_PROJECT [--check] [--in-place]

  Format a Juvix file or Juvix project

  When the command is run with an unformatted file it prints the reformatted source to standard output.
  When the command is run with a project directory it prints a list of unformatted files in the project.

Available options:
  JUVIX_FILE_OR_PROJECT    Path to a .juvix file or to a directory containing a
                           Juvix project.
  --check                  Do not print reformatted sources or unformatted file
                           paths to standard output.
  --in-place               Do not print reformatted sources to standard output.
                           Overwrite the target's contents with the formatted
                           version if the formatted version differs from the
                           original content.
  -h,--help                Show this help text
```

## Location of main implementation

The implementation is split into two components:
* The src API: `format` and `formatProject` https://github.com/anoma/juvix/blob/73952ba15c90f33919c75426cf62ae7e83fd3225/src/Juvix/Formatter.hs
* The CLI interface:  
https://github.com/anoma/juvix/blob/73952ba15c90f33919c75426cf62ae7e83fd3225/app/Commands/Format.hs

## in-place uses polysemy Resource effect

The `--in-place` option makes a backup of the target file and restores it if there's an error during processing to avoid data loss. The implementation of this uses the polysemy [Resource effect](https://hackage.haskell.org/package/polysemy-1.9.0.0/docs/Polysemy-Resource.html). The recommended way to interpret the resource effect is to use `resourceToIOFinal` which makes it necessary to change the effects interpretation in main to use `Final IO`: https://github.com/anoma/juvix/blob/73952ba15c90f33919c75426cf62ae7e83fd3225/app/Main.hs#L15

## Format input is `FilePath`

The format options uses `FilePath` instead of `AppFile f` for the input file/directory used by other commands. This is because we cannot determine if the input string is a file or directory in the CLI parser (we require IO). I discussed some ideas with @janmasrovira on how to improve this in a way that would also solve other issues with CLI input file/parsing but I want to defer this to a separate PR as this one is already quite large.

One consequence of Format using `FilePath` as the input option is that the code that changes the working directory to the root of the project containing the CLI input file is changed to work with `FilePath`:

https://github.com/anoma/juvix/blob/f715ef6a531f63c40ac3f629dd9cfea7e867507a/app/TopCommand/Options.hs#L33

## New dependencies

This PR adds new dependencies on `temporary` and `polysemy-zoo`.

`temporary` is used for `emptySystemTempFile` in the implementation of the TempFile interpreter for IO:

https://github.com/anoma/juvix/blob/73952ba15c90f33919c75426cf62ae7e83fd3225/src/Juvix/Data/Effect/Files/IO.hs#L49

`polysemy-zoo` is used for the `Fresh` effect and `absorbMonadThrow` in the implementation of the pure TempFile interpreter:
 https://github.com/anoma/juvix/blob/73952ba15c90f33919c75426cf62ae7e83fd3225/src/Juvix/Data/Effect/Files/Pure.hs#L91

NB: The pure TempFile interpreter is not used, but it seemed a good idea to include it while it's fresh in my mind.

* Closes https://github.com/anoma/juvix/issues/1777

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,11 @@ name: The Juvix compiler CI
       - reopened
       - synchronize
       - ready_for_review
+
 concurrency:
   group: "${{ github.workflow }}-${{ github.head_ref || github.run_id }}"
   cancel-in-progress: true
+
 env:
   SKIP: ormolu,format-juvix-examples
 
@@ -116,7 +118,8 @@ jobs:
         if: ${{ success() }}
         run: |
           cd main
-          make -s juvix-format
+          make -s check-format-juvix-examples
+          make -s typecheck-juvix-examples
 
       - name: Add ~/.local/bin to PATH
         run: |
@@ -313,7 +316,8 @@ jobs:
         if: ${{ success() }}
         run: |
           cd main
-          make -s juvix-format
+          make -s check-format-juvix-examples
+          make -s typecheck-juvix-examples
 
       - name: Install Smoke
         uses: jaxxstorm/action-install-gh-release@v1.10.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  SKIP: ormolu,format-juvix-examples
+  SKIP: ormolu,format-juvix-examples,typecheck-juvix-examples
 
 jobs:
   pre-commit:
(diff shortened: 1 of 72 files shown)


jonaprieto (approved):
Good~!

PR #1890, opened 2023-03-15
Remove missing Juvix examples and webapp example from docs build

paulcadman:
This PR removes the TicTacToe WebApp, we can no longer compile it because the compiler does not support the standalone WASM target.

The docs build passed in https://github.com/anoma/juvix/actions/runs/4425543266


janmasrovira (approved):
:wave: :cry:

PR #1892, opened 2023-03-15
Prepare Release 0.3.0

paulcadman:
* Bump version in package.yaml
* Update version smoke test
* Updates installing doc (though link will not work until we've produced a linux binary after release)
* Updates changelog.org


janmasrovira (approved):
:fireworks:

PR #1893, opened 2023-03-15
The `rmap` recursor

lukaszcz:
The `rmap` recursor allows to specify changes in binders while going downward through a Core node. This should help in implementing transformations on Core which need to add/remove/change binders.

* Depends on PR #1875 
* Adds unit tests for `rmap`
* Changes the `NatToInt` transformation to use `rmap`

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
+++ b/src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
@@ -1,5 +1,6 @@
 module Juvix.Compiler.Core.Extra.Recursors.RMap where
 
+import Data.Functor.Identity
 import Juvix.Compiler.Core.Data.BinderList qualified as BL
 import Juvix.Compiler.Core.Extra.Base
 import Juvix.Compiler.Core.Extra.Recursors.Base
@@ -26,7 +27,7 @@ rmapG ::
   forall c m.
   (Monad m) =>
   Collector (Int, [Binder]) c ->
-  (([BinderChange] -> c -> Node -> m Node) -> c -> Node -> m Node) ->
+  ((c -> [BinderChange] -> Node -> m Node) -> c -> Node -> m Node) ->
   Node ->
   m Node
 rmapG coll f = go mempty 0 (coll ^. cEmpty)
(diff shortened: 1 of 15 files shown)


src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
```diff
@@ -0,0 +1,111 @@
...
+    BCAdd Int
+  | -- | `BCKeep b` -- keep the binder `b`
+    BCKeep Binder
+  | -- | `BCRemove (BinderRemove b n)` -- remove the binder `b` and replace the
+    -- occurrences of the bound variable with the node `n`; the de Bruijn
+    -- indices of `n` are with respect to the result
+    BCRemove BinderRemove
+
+bindersFromBinderChange :: [BinderChange] -> [Binder]
+bindersFromBinderChange =
```
janmasrovira, line 27: I'd say is more idiomatic to write:
```
bindersFromBinderChange :: [BinderChange] -> [Binder]
bindersFromBinderChange = concatMap helper
  where
  helper = \case
     BCAdd {} -> []
     BCKeep b -> [b]
     BCRemove (BinderRemove b _) -> [b]
```

src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
```diff
@@ -0,0 +1,111 @@
...
+  reverse
+    . foldl'
+      ( \bs chg -> case chg of
+          BCAdd _ -> bs
+          BCKeep b -> b : bs
+          BCRemove (BinderRemove b _) -> b : bs
+      )
+      []
+
+bindersNumFromBinderChange :: [BinderChange] -> Int
```
janmasrovira, line 37: I'd prefer:
```
bindersNumFromBinderChange :: [BinderChange] -> Int
bindersNumFromBinderChange = length . bindersFromBinderChange
```

src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
```diff
@@ -0,0 +1,111 @@
...
+bindersNumFromBinderChange =
```
janmasrovira, line 38: It is a bit unintuitive to me that the number of binders in `BCAdd` is ignored. Could it be clarified in a comment why we want that?

lukaszcz: Because these are the binders in the *original* node, not in the result. I'll add a comment. I'll also add similar helper functions for the binders in the result node.

src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
lukaszcz: Done

src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
lukaszcz: Done

src/Juvix/Compiler/Core/Extra/Recursors/RMap.hs
lukaszcz: Done

janmasrovira (approved):

PR #1894, opened 2023-03-15
Automatically detect and split mutually recursive blocks in let expressions

### LINKED ISSUE #1677: Break up LetRec nodes in Core into strongly connected components.
Break up LetRec nodes into true strongly connected components based on a local analysis of the call graph. This may have a significant impact on performance if each let-block in the front-end is translated naively into a single LetRec node in Core.

janmasrovira:
- Closes #1677


src/Juvix/Compiler/Internal/Pretty/Base.hs
```diff
@@ -91,8 +91,17 @@ instance PrettyCode Let where
...
   ppCode = \case
     LetFunDef f -> ppCode f
+    LetMutualBlock b -> ppMutual b
+    where
+      ppMutual :: MutualBlock -> Sem r (Doc Ann)
+      ppMutual m@(MutualBlock b)
+        | [_] <- toList b = ppCode b
+        | otherwise = do
+            b' <- ppCode m
+            return (kwMutual <+> braces (line <> indent' b' <> line))
```
paulcadman, line 104: Where should / can we write the `mutual` keyword? Could you add a test that demonstrates its use?

janmasrovira: we cannot write the `mutual` keyword, but I thought it would be useful to print it in `Internal` syntax as a means to easily see for debugging purposes what are the actual mutual blocks

paulcadman (approved):

PR #1897, opened 2023-03-17
Normalize types in repl

janmasrovira:
When using `:type` in the repl, the types must be normalized to ensure that a whole is not printed to the user.
Before this pr, giving this to the repl:
```
:type let even : _; odd : _; odd zero := false; odd (suc n) := not (even n); even zero := true; even (suc n) := not (odd n) in even 10
```
would result in `_`.


paulcadman (approved):
🚀

PR #1903, opened 2023-03-20
Fix JuvixAsm validation

lukaszcz:


src/Juvix/Compiler/Core/Evaluator.hs
```diff
@@ -149,8 +149,8 @@ eval !ctx !env0 = convertRuntimeNodes . eval' env0
                 evalError "string to integer: not an integer" n
           _ ->
             evalError "string conversion: argument not a string" n
-      OpFail -> unary $ \msg -> Exception.throw (EvalError (fromString ("failure: " ++ printNode (eval' env msg))) Nothing)
-      OpTrace -> binary $ \msg x -> Debug.trace (printNode (eval' env msg)) (eval' env x)
+      OpFail -> unary $ \msg -> Exception.throw (EvalError ("failure: " <> printNode (eval' env msg)) Nothing)
+      OpTrace -> binary $ \msg x -> unsafePerformIO (hPutStrLn herr (printNode (eval' env msg)) >> return (eval' env x))
```
paulcadman, line 153: According to the GHC docs https://hackage.haskell.org/package/base-4.17.0.0/docs/Debug-Trace.html#g:1 `trace` should output to stderr already, but I suppose you found that the traces were appearing on stdout?

lukaszcz:
They appear on stdout when you compile it to WASM/native, because the
runtime outputs them to stdout.

PR #1905, opened 2023-03-21
CI pre-commit maintenance 

jonaprieto:
This PR goes after:
- #1797 

Included in this PR:

- Add Clang formatting (v.15) as part of the pre-commits.
- Add new pre-commits:
   - forbid binary files 
   - check toml files for mdbook
   - detect-private-key
   - format-juvix-examples: check all the Juvix programs in the `examples` folder type-check for local runs.
- Remove .pre-commit-hooks and add ormolu for local pre-commit runs instead.


paulcadman (approved):
🚀

Issue #1907, opened 2023-03-21
Support recursive datatypes in the GEB backend


lukaszcz:
Implement recursion unrolling for inductive datatypes.

Depends on: https://github.com/anoma/geb/issues/62

paulcadman:
The geb backend is no longer supported.

PR #1909, opened 2023-03-21
Check for recursive inductive types in the GEB pipeline

lukaszcz:
Currently, only recursion in functions is checked. This PR also adds some utilities (TypeDependencyInfo) that will be useful for issue #1907.

### DIFF
--- a/src/Juvix/Compiler/Core/Transformation/CheckGeb.hs
+++ b/src/Juvix/Compiler/Core/Transformation/CheckGeb.hs
@@ -70,28 +70,24 @@ checkGeb tab =
           _ -> return node
 
     checkNoRecursion :: Sem r ()
-    checkNoRecursion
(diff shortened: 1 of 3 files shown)


src/Juvix/Compiler/Core/Transformation/CheckGeb.hs
```diff
@@ -79,6 +81,18 @@ checkGeb tab =
       | otherwise =
           return ()
 
+    checkNoRecursiveTypes :: Sem r ()
+    checkNoRecursiveTypes
+      | isCyclic (createTypeDependencyInfo tab) =
```
janmasrovira, line 86: better to have:
```
checkNoRecursiveTypes = 
   when (isCyclic (createTypeDependencyInfo tab))
       $ throw .....
```

janmasrovira (approved):

janmasrovira (approved):

PR #1911, opened 2023-03-21
Fix REPL state to include enough information to rerun the pipeline

janmasrovira:
Previously we were:
* discarding the types table 
* discarding the name ids state
after processing an expression in the REPL.

For example evaluating:
```
let even : _; odd : _; odd zero := false; odd (suc n) := not (even n); even zero := true; even (suc n) := not (odd n) in even 10
```
would loop in the REPL.

We noticed that the `n` in `suc n` was being given type `Type` instead of `Nat`. This was because the name id given to n was incorrect, the REPL started using name ids from 0 again.

We fixed this issue by storing information, including the types table and name ids state in the Artifacts data structure that is returned when we run the pipeline for the first time. This information is then used when we call functions to compile / type check REPL expressions.


src/Juvix/Compiler/Pipeline.hs
```diff
@@ -199,33 +203,109 @@ coreToGeb spec = Core.toGeb >=> return . uncurry (Geb.toResult spec) . Geb.fromC
 -- Run pipeline
 --------------------------------------------------------------------------------
 
-runIOEither :: forall a. BuiltinsState -> EntryPoint -> Sem PipelineEff a -> IO (Either JuvixError (Artifacts, a))
-runIOEither builtinsState entry =
+-- | It returns `ResolverState` so that we can retrieve the `juvix.yaml` files,
+-- which we require for `Sope` tests.
```
lukaszcz, line 207: typo in comment: Sope -> Scope

lukaszcz (dismissed):

src/Juvix/Compiler/Pipeline.hs
janmasrovira: fixed https://github.com/anoma/juvix/pull/1911/commits/342cc1b22c78a7bfadc7cb46c1fe24c99429d2b0

PR #1918, opened 2023-03-23
Add --numeric-version flag

jonaprieto:
As the title says. 

```
❯ juvix --numeric-version
0.3.0
```


tests/smoke/Commands/version-help-doctor.smoke.yaml
```diff
@@ -9,6 +9,14 @@ tests:
         regex: |-
           ^Juvix version 0.3.0-([a-f0-9]{7}).*
 
+  - name: cli-numeric-version
+    command:
+      - juvix
+      - --numeric-version
+    exit-status: 0
+    stdout: |
+      0.3.0
```
janmasrovira, line 18: have you considred a regular expression so we don't need to update the test every version?

stack.hie.yaml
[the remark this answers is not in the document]
jonaprieto: Oh, I didn't see this. Thanks.

tests/smoke/Commands/version-help-doctor.smoke.yaml
jonaprieto: We can do that, sure ;)

paulcadman (dismissed):

Issue #1919, opened 2023-03-23
Runtime juvix assertion error in Hanoi example 


paulcadman:
I compiled the Hanoi example with the `--debug` flag for the native target on macOS aarch64:

https://github.com/anoma/juvix/blob/dbe9ff61d794637a3315068c65bfd296512f07dc/examples/milestone/Hanoi/Hanoi.juvix

```
juvix compile -g Hanoi.juvix
```

When running the program you get the following error:

```
./Hanoi
examples/milestone/Hanoi/.juvix-build/Hanoi.c:216: Juvix assertion failed. Please report.
```

NB: The program runs without error and with the expected output when I use the `wasm32-wasi` target, or when I use the `native` target without the `--debug` flag .

After doing some debugging the trace looks like this:

JUVIX_STR_CONCAT  - https://github.com/anoma/juvix/blob/dbe9ff61d794637a3315068c65bfd296512f07dc/runtime/src/juvix/api.h#L79
CONCAT_CSTRINGS - https://github.com/anoma/juvix/blob/dbe9ff61d794637a3315068c65bfd296512f07dc/runtime/src/juvix/object/cstring.h#L43
ALLOC - https://github.com/anoma/juvix/blob/dbe9ff61d794637a3315068c65bfd296512f07dc/runtime/src/juvix/mem/alloc.h#L50

and the failing assertion is caused by an `is_same_page` check:

https://github.com/anoma/juvix/blob/dbe9ff61d794637a3315068c65bfd296512f07dc/runtime/src/juvix/mem/alloc.h#L51

lukaszcz:
Probably forgot to include the memory allocated for string concatenation in the memory usage count. I'll look into it.

PR #1920, opened 2023-03-23
Add a test suite for milestone examples

paulcadman:
In this PR I will add tests for the example programs in `examples/milestone`.

There's currently an runtime assertion error generated by the Hanoi example https://github.com/anoma/juvix/issues/1919, so it'd be good to test these programs in the future.


test/Examples.hs
```diff
@@ -0,0 +1,7 @@
+module Examples where
+
+import Base
+import Examples.Positive qualified as P
+
+allTests :: TestTree
+allTests = testGroup "Juvix compilation pipeline tests" [P.allTests]
```
lukaszcz, line 7: There should be a different test group name, e.g., "Juvix milestone examples compilation tests"

lukaszcz (approved):

PR #1921, opened 2023-03-23
Let folding

### LINKED ISSUE #1899: Let-folding
Implement an optimizing transformation that folds lets whose values are immediate, i.e., they don't require evaluation (variables or constants).

For example, transform
```
let x := y in let z := x + x in x + z
```
to
```
let z := y + y in y + z
```

This is easy to implement using the `rmap` recursor and will remove the spurious lets created by the pattern matching compiler (both the naive and the new proper one).

lukaszcz:
* Closes #1899

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Utils.hs
+++ b/src/Juvix/Compiler/Core/Extra/Utils.hs
@@ -39,6 +39,13 @@ isTypeConstr tab ty = case typeTarget ty of
     isTypeConstr tab (fromJust $ HashMap.lookup _identSymbol (tab ^. identContext))
   _ -> False
 
+isImmediate :: Node -> Bool
+isImmediate = \case
+  NVar {} -> True
+  NIdt {} -> True
+  NCst {} -> True
+  _ -> False
+
 freeVarsSorted :: Node -> Set Var
 freeVarsSorted n = Set.fromList (n ^.. freeVars)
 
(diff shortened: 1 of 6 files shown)


src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -39,6 +39,13 @@ isTypeConstr tab ty = case typeTarget ty of
     isTypeConstr tab (fromJust $ HashMap.lookup _identSymbol (tab ^. identContext))
   _ -> False
 
+isImmediate :: Node -> Bool
```
janmasrovira, line 42: add a short description

src/Juvix/Compiler/Core/Transformation/Optimize/LetFolding.hs
```diff
@@ -0,0 +1,20 @@
+module Juvix.Compiler.Core.Transformation.Optimize.LetFolding where
```
janmasrovira, line 1: add a short description of the transformation at the top of the module

src/Juvix/Compiler/Core/Extra/Utils.hs
lukaszcz: done

src/Juvix/Compiler/Core/Transformation/Optimize/LetFolding.hs
lukaszcz: done

paulcadman (approved):

PR #1929, opened 2023-03-24
Let-fold lambdas

### LINKED ISSUE #1926: out of memory errors when running WASM compiled bench examples
The runtime crash can be reproduced with the merge sort and fibonacci bench tests:

https://github.com/anoma/juvix/blob/8d7e669f7499b4122d9b1a9baf42828d1c163c7a/tests/benchmark/mergesort/juvix/mergesort.juvix
https://github.com/anoma/juvix/blob/8d7e669f7499b4122d9b1a9baf42828d1c163c7a/tests/benchmark/fibonacci/juvix/fibonacci.juvix

```
$ cd tests/benchmark/mergesort/juvix
$ juvix compile -t wasm32-wasi mergesort.juvix
$ wasmer mergesort.wasm
error: out of memory
error: failed to run `mergesort.wasm`
│   1: RuntimeError: unreachable
╰─▶ 2: unreachable
```

When running with wasmtime instead of wasmer there's a bit more information:

```
$ wasmtime mergesort.wasm
error: out of memory
Error: failed to run main module `mergesort.wasm`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
           0: 0x2795 - <unknown>!<wasm function 14>
           1: 0x20c3 - <unknown>!<wasm function 11>
           2: 0x20af - <unknown>!<wasm function 10>
           3: 0x1963 - <unknown>!<wasm function 3>
    2: wasm trap: wasm `unreachable` instruction executed
```

There are no errors when running the same example after compiling with the native target.

lukaszcz:
* Closes #1926

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Utils.hs
+++ b/src/Juvix/Compiler/Core/Extra/Utils.hs
@@ -46,6 +46,7 @@ isImmediate = \case
   NVar {} -> True
   NIdt {} -> True
   NCst {} -> True
+  NLam {} -> True
   _ -> False
 
 freeVarsSorted :: Node -> Set Var


lukaszcz:
Well, there is a problem with this that it may duplicate the lambdas. What we should really do is to have another let-folding pass after lambda-lifting.

But I propose we merge this and introduce a proper optimization framework sometime around 0.4 (with only let-folding in it initially).

PR #1935, opened 2023-03-27
Add the `--unroll` option

### LINKED ISSUE #1928: Add an option to specify the recursion unrolling depth
An option e.g. `--unroll N` would globally override the default depth 140.

Later, we should have a mechanism to specify unrolling depth on a per-function basis.

lukaszcz:
* Closes #1928

### DIFF
--- a/app/GlobalOptions.hs
+++ b/app/GlobalOptions.hs
@@ -7,7 +7,7 @@ import CommonOptions
 import Juvix.Compiler.Abstract.Pretty.Options qualified as Abstract
 import Juvix.Compiler.Core.Options qualified as Core
 import Juvix.Compiler.Internal.Pretty.Options qualified as Internal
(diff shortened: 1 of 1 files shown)


src/Juvix/Compiler/Pipeline/EntryPoint.hs
```diff
@@ -44,8 +45,9 @@ defaultEntryPoint root mainFile =
       _entryPointStdin = Nothing,
       _entryPointPackage = defaultPackage root buildDir,
       _entryPointGenericOptions = defaultGenericOptions,
-      _entryPointTarget = TargetCNative64,
```
janmasrovira, line 47: was this change intended?

lukaszcz: It shouldn't matter, because if you need a different target you should override it. I tried to make defaults consistent.

janmasrovira (approved):

PR #1940, opened 2023-03-28
Refactor Geb values

lukaszcz:
* Geb value representation changed to match more closely with the morphism representation.
* Implements quoting Geb values


jonaprieto (approved):
Great!

Issue #1914, opened 2023-03-23
Add `juvix dev repl` and remove `-t` flag from `juvix repl`


paulcadman:
Currently `juvix repl` supports the `-t,--transforms` flag. This flag is useful for debugging the compiler but it has no use for developers who are just developing Juvix programs. For normal users this should always be set to the `pipeline-eval` meta-flag.

So I propose that we:
* remove the `-t` flag from `juvix repl` (with the behaviour that it works as if `-t pipeline-eval` was set.
* add `juvix dev repl` which exposes the `-t` flag. The default setting should be `pipeline-eval`.

For `juvix dev repl`:
* The `disambiguate-names` transform is always applied, regardless of the `-t` setting (this matches the behaviour of `juvix dev core from-concrete`).
* Add a `--no-disambiguate`, similar to `juvix dev core from-concrete` that disables the `disambiguate-names` transform.
* Only those transforms specified by the `-t` flag should be applied if it is specified, except for `disambiguate-names`.

The reason for always applying `disambiguate-names` is that if it is not set, then the JuvixCore parser may not be able to parse back the output of the REPL. Unless you're debugging `disambiguate-names` itself (for which you can use the `--no-disambiguate` flag) you always want to be able to parse back the output when debugging the other transforms.

PR #1941, opened 2023-03-28
Add juvix dev repl command

paulcadman:
The new `juvix dev repl` command is a copy of the `juvix repl` with the addition of `--no-disambiguate` flag that is present on the `juvix dev core from-concrete` command.

The `juvix repl` command now does not have the `--transforms`, `--show-de-bruijn` flags as these are only relevant for compiler developers. The eval transforms are always applied.

By default `juvix dev repl` uses the eval transforms. You can override this by specifying the `-t` flag.

Also we now run `disambiguateNames` transform on the info table in the `dev repl` (unless the `--no-disambiguate-names` flag is set). This is so the output of the `juvix dev repl` will match that of `juvix dev core from-concrete` and also so the output can be parsed by back the core parser.

* Closes https://github.com/anoma/juvix/issues/1914

### DIFF
--- a/app/Commands/Dev.hs
+++ b/app/Commands/Dev.hs
@@ -16,6 +16,7 @@ import Commands.Dev.Parse qualified as Parse
 import Commands.Dev.Runtime qualified as Runtime
 import Commands.Dev.Scope qualified as Scope
 import Commands.Dev.Termination qualified as Termination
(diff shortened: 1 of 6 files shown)


janmasrovira (approved):
Great!

janmasrovira (approved):

PR #1946, opened 2023-03-29
Option `--show-args-num`

lukaszcz:
* Adds an options to display the `_identifierArgsNum` field when printing Core definitions.

### DIFF
--- a/src/Juvix/Compiler/Core/Pretty/Options.hs
+++ b/src/Juvix/Compiler/Core/Pretty/Options.hs
@@ -23,7 +23,7 @@ traceOptions =
   Options
     { _optShowIdentIds = False,
       _optShowDeBruijnIndices = True,
-      _optShowArgsNum = False
+      _optShowArgsNum = True
     }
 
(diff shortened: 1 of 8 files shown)


src/Juvix/Compiler/Core/Pretty/Options.hs
```diff
@@ -13,14 +14,16 @@ defaultOptions :: Options
...
+      _optShowArgsNum = False
     }
 
 traceOptions :: Options
 traceOptions =
   Options
     { _optShowIdentIds = False,
-      _optShowDeBruijnIndices = True
+      _optShowDeBruijnIndices = True,
+      _optShowArgsNum = False
```
janmasrovira, line 26: I'd say this should be `True`

lukaszcz: done

janmasrovira (approved):

janmasrovira (approved):

PR #1962, opened 2023-03-31
Documentation for projects and modules

janmasrovira:
This pr adds the reference documentation for modules and projects.

I've temporarily moved some of the old documentation to `modules_old.md`. The contents of that file are more understanding focused and thus they do not fit the language reference. We may want to reuse that for a tutorial in the future.


docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
-import A;
-open A hiding {a};
-
-axiom
-a : A;
+open Data.List;
+```
 
+Now we can simply write `sort`. It is important to remember that when we
+open a module, that module must be in scope.
```
paulcadman, line 55: Use the term "imported" as we have already introduced this concept, and we haven't introduced the concept of scope.

```suggestion
open a module, that module must already be imported.
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
 
 ```juvix
-import Prelude;
-open Prelude;
+import Data.List;
+open Data.List;
 ```
 
-We simplify it by the expression:
+When opening a module, if we want to open an explicit subset of its definition,
```
paulcadman, line 71:
```suggestion
When opening a module, if we want to open an explicit subset of its definitions,
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
--- B.juvix
-module A;
-open import A hiding {a};
+open Data.List hiding {head; tail}
+```
 
-axiom
-a : A;
+Normally, all opened definitions are available under the current module, but
+they are not exported. Meaning that if another module imports the current
```
paulcadman, line 86: I think this is clearer, but feel free to dismiss.

```suggestion
All opened definitions are available under the current module, but
they are not exported by default. Meaning that if another module imports the current
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
+module, it will only be able to access the definitions defined there but not
+those which have been opened. If we want opened definitions to be exported, we
+will use the `public` keyword thus:
```
paulcadman, line 89:
```suggestion
must use the `public` keyword thus:
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
+we will use the `using` keyword thus:
 
 ```juvix
-open import Prelude;
+open Data.List using {List; sort; reverse}
 ```
 
-The `hiding` keyword can be used within an `open-import` statement.
+If we want to open all definitions of a module minus a subset, we
+will use the `hiding` keyword thus:
```
paulcadman, line 79:
```suggestion
must use the `hiding` keyword thus:
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
 ```juvix
-import Prelude;
-open Prelude;
+import Data.List;
+open Data.List;
 ```
 
-We simplify it by the expression:
+When opening a module, if we want to open an explicit subset of its definition,
+we will use the `using` keyword thus:
```
paulcadman, line 72:
```suggestion
we must use the `using` keyword thus:
```

docs/reference/language/modules.md
```diff
@@ -1,189 +1,139 @@
...
+apply, and they probably follow your intuition:
+
+1. Opening or defining a symbol shadows inherited instances of that symbol.
+2. Opening a symbol does _not_ shadow a defined instanance of that symbol in the
+   current module.
+3. Conversely, defining a symbol in the current module does _not_ shadow an
+   opened instance of that symbol.
+
+As a consequence of 2 and 3, using a symbol that is both defined and opened
+locally will result in an ambiguousity error.
```
paulcadman, line 139:
```suggestion
locally will result in an ambiguity error.
```

docs/reference/language/modules_old.md
```diff
@@ -0,0 +1,166 @@
+# Module system
+
+## Defining a module
+
+The `module` keyword stars the declaration of a module followed by its
```
paulcadman, line 5:
```suggestion
The `module` keyword starts the declaration of a module followed by its
```

docs/reference/language/modules.md
janmasrovira: > Use the term "imported" as we have already introduced this concept, and we haven't introduced the concept of scope.

I used `in scope` instead of `imported` because we can also open local modules. We should think of a more clear way to include both

paulcadman: You're right. Do you think - "that module must either be imported or defined as a local module" is too verbose?

janmasrovira: I've added your suggestion as a clarification for "in scope".

paulcadman (approved):

paulcadman (approved):

PR #1963, opened 2023-04-01
Store stdlib dependency as a relative path

janmasrovira:


src/Juvix/Data/Effect/Files/Pure.hs
```diff
@@ -84,10 +85,14 @@ re cwd = reinterpret $ \case
   RemoveFile' p -> removeFileHelper p
   RenameFile' p1 p2 -> renameFileHelper p1 p2
   CopyFile' p1 p2 -> copyFileHelper p1 p2
+  JuvixConfigDir -> return juvixConfigDirPure
   where
     cwd' :: FilePath
     cwd' = toFilePath cwd
 
+juvixConfigDirPure :: Path Abs Dir
```
paulcadman, line 93: Should this be `Path Rel Dir` ? Because we expect it to be combined with an absolute root directory?

src/Juvix/Prelude/Path.hs
```diff
@@ -22,6 +22,14 @@ infixr 5 <//>
...
 (<//>) = (Path.</>)
 
+infixr 5 <///>
+
+-- | Appens a relative path to some directory
+(<///>) :: SomeBase Dir -> Path Rel t -> SomeBase t
+(<///>) s r = case s of
+  Abs a -> Abs (a <//> r)
+  Rel a -> Rel (a <//> r)
+
```
paulcadman, line 32: Perhaps use https://hackage.haskell.org/package/path-0.9.2/docs/Path-Posix.html#v:mapSomeBase

```suggestion
(<///>) s r = mapSomeBase (<//> r) s
```

src/Juvix/Data/Effect/Files/Pure.hs
janmasrovira: This directory is only used in the pure implementation of `Files`. It has to be absolute, because that's what the `IO` implementation returns:
```
juvixConfigDirIO :: IO (Path Abs Dir)
juvixConfigDirIO = (<//> versionDir) . absDir <$> getUserConfigDir "juvix"
```

paulcadman: got it - thank you!

paulcadman (approved):
🥇 🚀

paulcadman (approved):

PR #1970, opened 2023-04-03
Check for the executable (WASM/native) pipeline prerequisites

### LINKED ISSUE #1959: C compilation errors when `main` has a function type
Compiling
```agda
module ret;

open import Stdlib.Prelude;

main : Nat -> Nat;
main x := x;
```
results in
```
/Users/heliax/Documents/progs/juvix/.juvix-build/ret.c:31:19: error: variable 'juvix_arg_0' is uninitialized when used here [-Werror,-Wuninitialized]
        STMP(0) = ARG(0);
                  ^~~~~~
```

lukaszcz:
* Closes #1959


src/Juvix/Compiler/Core/Transformation/CheckExec.hs
```diff
@@ -0,0 +1,40 @@
...
+                    _coreErrorLoc = loc
+                  }
+        _ ->
+          return tab
+      where
+        ii = fromJust $ HashMap.lookup sym (tab ^. infoIdentifiers)
+        loc = fromMaybe defaultLoc (ii ^. identifierLocation)
+
+        mockFile :: Path Abs File
+        mockFile = $(mkAbsFile "/core-to-geb")
```
paulcadman, line 37: Suggestion: Does the mockFile name make sense here, given that the transform is not used in the Geb pipeline?

I'm not sure what to suggest instead. Something like `/check-exec` might help identify where the error is coming from.

lukaszcz: Yes, I just copied it and forgot to change the name

runtime/src/juvix/io.c
```diff
@@ -32,6 +32,18 @@ void io_flush() {
     }
 }
 
+static void io_print(word_t x) {
```
paulcadman, line 35: It's odd these changes appear in the diff, is this a GitHub bug? We just merged these into main in https://github.com/anoma/juvix/pull/1969 ?

lukaszcz: Apparently, GitHub believes that the commit for #1969 is a part of this PR. Strange. There is no option to rebase. I don't know what to do with this.

paulcadman (approved):

PR #1975, opened 2023-04-04
Refactor Core datastructures

### LINKED ISSUE #1972: Remove information duplication in Core
Some info is unnecessarily duplicated, e.g., the ConstructorInfos appear both in `inductiveConstrutors` in `InductiveInfo` and separately in `infoConstructors` in the `InfoTable`. It would be enough to store the symbols in `inductiveConstructors`.

Similarly, type information for identifier arguments is duplicated from identifier type.

lukaszcz:
* Closes #1972 
* Introduces lookup functions for the InfoTable to avoid using HashMap explicitly and reduce code duplication


janmasrovira (approved):
great

PR #1978, opened 2023-04-04
Fix: format  juvix files in test/positive

jonaprieto:
This PR fixes a formatting issue that drops blank lines between axiom declarations.

It goes after:

- #1980
- Closes https://github.com/anoma/juvix/issues/1986

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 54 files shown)


janmasrovira (dismissed):
For future PRs, I'd say it is a good idea to separate automated changes from manual changes. That is, automated changes, such as formatting, should be done in prs which do not include any manual changes (if possible). I think that helps with reviews, merging and the possibility to revert the changes

PR #1979, opened 2023-04-05
Adsd the tracing library and the sexp library to the project

mariari:
Was told to just create a PR with them.

This PR adds two libraries, 

The Sexp library has the Sexp derivation code.

The StandardLibrary library adds the tracing code in `Trace.hs` and the `Trace` subfolder. It's also needed for the Sexp code, it can be decoupled from each other with some smart shuffling

### DIFF
--- a/library/Sexp/.gitignore
+++ b/library/Sexp/.gitignore
@@ -1,0 +1,2 @@
+.stack-work/
+*~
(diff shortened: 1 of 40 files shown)


jonaprieto:
**Thanks!** Let us evaluate this PR in the context of our codebase. If you don't mind, we'll edit it to extract what's useful for us, and also to ensure compatibility with our codebase.

mariari:
> **Thanks!** Let us evaluate this PR in the context of our codebase. If you don't mind, we'll edit it to extract what's useful for us, and also to ensure compatibility with our codebase.

Feel free to use the code however you want and actually integrate it

PR #1980, opened 2023-04-05
Fix: pprint positive kw for data types

jonaprieto:
As the title says.

- I found this bug while formatting the examples found in the tests folder. 

- In addition to printing the missing positive' keyword for data types, the code also prints certain keyword annotations onto separate lines, the ones that act as attributes to their term. While this is a matter of personal preference, I find that it makes it easier to comment and uncomment individual annotations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,13 @@ METAFILES:=README.md \
 		   CONTRIBUTING.md \
 		   LICENSE.md
 
+STACKFLAGS?=--jobs $(THREADS)
+STACKTESTFLAGS?=--ta --hide-successes --ta --ansi-tricks=false
+SMOKEFLAGS?=--color --diff=git
+STACK?=stack
+
+JUVIXBIN?=juvix
+
 ifeq ($(UNAME), Darwin)
 	THREADS := $(shell sysctl -n hw.logicalcpu)
 else ifeq ($(UNAME), Linux)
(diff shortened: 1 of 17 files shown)


paulcadman:
Printing keyword annotations onto separate lines probably requires more discussion.

My opinion is that it's inconsistent with other keywords. For example we don't format `module`, ` infix` or `type` like:

```
module
Foo;
```

```
infix 6
+;
```

```
type
Foo := | foo;
```

jonaprieto:
Sure! open discussion

> it's inconsistent with other keywords

My preference for this formatting is based on the fact that it can be helpful when refactoring or testing code. When each annotation is commented on its line, it becomes easier to modify them without affecting the rest of its terms. Like minimizing the diff mark. Also, in each case you mentioned, if you were to comment or uncomment the corresponding keyword line, it would effectively render the statement invalid. Annotations like terminating, builtins, or even axioms act as attributes to the term, and they may be removed later.

paulcadman:
OK - let's do it 🚀 !

Suggestion: Update the language reference formatting: https://github.com/anoma/juvix/blob/046e82bf062483dfd1d27b333cbcfc6e01f31a8e/docs/reference/language/axioms.md?plain=1#L10

jonaprieto:
Done, I updated the excerpts in the documentation I saw. Likely, I forgot some, so we need to review the format later, so I'll leave this for another pr if it's ok with you.

paulcadman (approved):

PR #1981, opened 2023-04-05
Fix: Stop 'make check' at first failure

jonaprieto:


Makefile
```diff
@@ -180,20 +180,24 @@ SMOKEFLAGS?=--color --diff=git
...
+		&& export SKIP=ormolu,format-juvix-examples,typecheck-juvix-examples \
+		&& ${MAKE} pre-commit
 
 .PHONY: check
 check: clean
 	@${MAKE} check-only
 
+.PHONY : bench
+bench: runtime submodules
+	@stack bench
```
paulcadman, line 199: Do we want to append one of the stack flags variables to this command?

jonaprieto: We should yes, also use ${STACK} similar to the use for make.

paulcadman (approved):

PR #1982, opened 2023-04-05
Add: the 'seq' builtin (>>>)

### LINKED ISSUE #1913: Better syntax for `trace` and the `seq` operator
In Haskell
```
trace "X" $ expr
```
behaves as expected, first printing "X" before evaluating `expr`. In an eager language `expr` is evaluated first, which is typically not what one wants with `trace` and it differs from the behaviour of `trace "X" (expr)`. Plus it's very inconvenient to put `expr` in parens every time.

A solution is to add a lazy version of `$` as a builtin (e.g. called `$~` or `$&` or `$$`):
```
trace "X" $~ expr
```

### LINKED ISSUE #1976: `trace` is not compiled lazily
Compiling and running:
```
module tr;

open import Stdlib.Prelude;
open import Stdlib.Debug.Trace;

main : Nat;
main := trace "A" (trace "B" 3);
```
gives
```
B
A
3
```
Evaluation in Core gives the correct
```
A
B
3
```

lukaszcz:
* Closes #1913 
* Closes #1976


src/Juvix/Compiler/Asm/Translation/FromCore.hs
```diff
@@ -135,10 +135,22 @@ genCode infoTable fi =
 
     goBuiltinApp :: Bool -> Int -> BinderList Value -> Core.BuiltinApp -> Code'
     goBuiltinApp isTail tempSize refs (Core.BuiltinApp {..}) =
-      snocReturn isTail $
-        DL.append
-          (DL.concat (map (go False tempSize refs) (reverse _builtinAppArgs)))
-          (genOp _builtinAppOp)
+      case _builtinAppOp of
+        OpSeq ->
```
janmasrovira, line 139: let's have `goSeq`

src/Juvix/Compiler/Builtins/Control.hs
```diff
@@ -0,0 +1,27 @@
+module Juvix.Compiler.Builtins.Control where
+
+import Juvix.Compiler.Abstract.Extra
+import Juvix.Compiler.Builtins.Effect
+import Juvix.Prelude
+
+registerSeq :: (Members '[Builtins, NameIdGen] r) => FunctionDef -> Sem r ()
+registerSeq f = do
+  let u = ExpressionUniverse (Universe {_universeLevel = Nothing, _universeLoc = error "Universe with no location"})
```
janmasrovira, line 9: Maybe we should have the following in `Juvix.Data.Universe`
```
smallUniverseNoLoc :: Universe
smallUniverseNoLoc = smallUniverse (error "Universe with no location")
```
At some point we'll need to figure out how to handle these non-existing locations

lukaszcz: done

janmasrovira (approved):

PR #1983, opened 2023-04-05
Fix: add supported targets as option for compile commands

### LINKED ISSUE #1882: Remove the 'geb' and 'core' targets from the 'asm' CLI
Currently, the 'asm' command uses the general 'compile' infrastructure, so it allows the 'geb' and 'core' targets, but then gives an error if they are actually specified.

There is a similar issue with the 'runtime' command.

jonaprieto:
- Closes #1882

### DIFF
--- a/app/Commands/Dev/Core/Compile/Options.hs
+++ b/app/Commands/Dev/Core/Compile/Options.hs
@@ -11,7 +11,12 @@ import Data.List.NonEmpty qualified as NonEmpty
 type CoreCompileOptions = CompileOptions
 
 coreSupportedTargets :: NonEmpty CompileTarget
-coreSupportedTargets = NonEmpty.fromList allTargets
+coreSupportedTargets = NonEmpty.fromList
+  [ TargetWasm32Wasi
+  , TargetNative64
+  , TargetGeb
+  , TargetAsm
+  ]
 
 parseCoreCompileOptions :: Parser CoreCompileOptions
 parseCoreCompileOptions =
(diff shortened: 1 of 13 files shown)


docs/reference/language/modules.md
```diff
@@ -130,7 +130,7 @@ We simplify it by the expression:
 open import Prelude;
 ```
 
-The `hiding` keyword can be used within an `open-import` statement.
+The `hiding` keyword can be used within a `open-import` statement.
```
lukaszcz, line 133: @paulcadman what is the correct article here? Because "open" begins with an "o", so "an" would seem appropriate phonetically. But I'm not sure about the exact rules.

app/Commands/Dev/Core/Compile/Options.hs
```diff
@@ -1,6 +1,20 @@
...
 where
 
 import Commands.Extra.Compile.Options
+import CommonOptions
+import Data.List.NonEmpty qualified as NonEmpty
+
+type CoreCompileOptions = CompileOptions
+
+coreSupportedTargets :: NonEmpty CompileTarget
+coreSupportedTargets = NonEmpty.fromList allTargets
```
lukaszcz, line 14: I think we should exclude `core` target here. It is just a silent no-op, which is confusing.

jonaprieto: Agree, I'll make the list explicit here.

docs/reference/language/modules.md
jonaprieto: Fixed

app/Commands/Dev/Core/Compile/Options.hs
jonaprieto: Done.

lukaszcz (approved):

PR #1988, opened 2023-04-06
Pretty print JuvixCore values consistently with Juvix syntax

### LINKED ISSUE #1957: Preserve fixity information for constructors when translating Internal to Core
This would allow to e.g. print the patterns with the right fixity of the constructors. Now they are all printed in prefix notation, e.g., `:: 1 nil`.

Relatedly, preserving the information about which constructor arguments are implicit would be more robust and correct than the current ad-hoc approach (which just ignores all parameters in the print-out).

### LINKED ISSUE #1985: EtaExpandApps should not delay argument evaluation
Replacing
```
c (f x)
```
with
```
\y c (f x) y
```
delays the evaluation of `f x`. Instead, we should replace it with:
```
(\x \y c x y) (f x)
```

lukaszcz:
This PR implements pretty printing of evaluation results consistently with Juvix syntax. The printed values do not necessarily originate directly from the source code. All functions/lambdas are printed as `<fun>`.

The same mechanism is used to implement pretty printing of unmatched pattern examples. 

Juvix REPL now uses the new value printing mechanism to display evaluation results. Typing `nil` in the REPL will now just display `nil`. The command `juvix dev repl` still prints raw JuvixCore terms.

* Closes #1957 
* Closes #1985

### DIFF
--- a/app/Commands/Repl.hs
+++ b/app/Commands/Repl.hs
@@ -154,12 +154,11 @@ runCommand opts = do
               Left err -> printError err
               Right n
                 | Info.member Info.kNoDisplayInfo (Core.getInfo n) -> return ()
-              Right n ->
-                if
-                    | opts ^. replPrintValues ->
-                        renderOut (Core.ppOut opts (toValue tab n))
-                    | otherwise ->
-                        renderOut (Core.ppOut opts n)
+              Right n
+                | opts ^. replPrintValues ->
+                    renderOut (Core.ppOut opts (toValue tab n))
+                | otherwise ->
+                    renderOut (Core.ppOut opts n)
           Nothing -> noFileLoadedMsg
         where
           defaultLoc :: Interval
--- a/app/Commands/Repl.hs
+++ b/app/Commands/Repl.hs
@@ -11,6 +11,7 @@ import Evaluator
 import Juvix.Compiler.Concrete.Data.Scope (scopePath)
 import Juvix.Compiler.Concrete.Data.ScopedName (absTopModulePath)
 import Juvix.Compiler.Core qualified as Core
+import Juvix.Compiler.Core.Extra.Value
 import Juvix.Compiler.Core.Info qualified as Info
 import Juvix.Compiler.Core.Info.NoDisplayInfo qualified as Info
 import Juvix.Compiler.Core.Pretty qualified as Core
@@ -147,12 +148,18 @@ runCommand opts = do
         ctx <- State.gets (^. replStateContext)
         case ctx of
           Just ctx' -> do
+            let tab = ctx' ^. replContextArtifacts . artifactCoreTable
             evalRes <- compileThenEval ctx' input
             case evalRes of
               Left err -> printError err
               Right n
                 | Info.member Info.kNoDisplayInfo (Core.getInfo n) -> return ()
-              Right n -> renderOut (Core.ppOut opts n)
+              Right n ->
+                if
+                    | opts ^. replPrintValues ->
+                        renderOut (Core.ppOut opts (toValue tab n))
+                    | otherwise ->
+                        renderOut (Core.ppOut opts n)
           Nothing -> noFileLoadedMsg
         where
           defaultLoc :: Interval
--- a/src/Juvix/Compiler/Core/Extra/Value.hs
+++ b/src/Juvix/Compiler/Core/Extra/Value.hs
@@ -10,27 +10,33 @@ import Juvix.Compiler.Core.Language.Value
 toValue :: InfoTable -> Node -> Value
 toValue tab = \case
   NCst Constant {..} -> ValueConstant _constantValue
-  NCtr Constr {..} ->
-    ValueConstrApp
-      ConstrApp
-        { _constrAppName = ci ^. constructorName,
-          _constrAppFixity = ci ^. constructorFixity,
-          _constrAppArgs = map (toValue tab) (drop paramsNum _constrArgs)
-        }
-    where
-      ci = lookupConstructorInfo tab _constrTag
-      ii = lookupInductiveInfo tab (ci ^. constructorInductive)
-      paramsNum = length (ii ^. inductiveParams)
-  node@NLam {} ->
-    let (lams, body) = unfoldLambdas node
-        n = length $ takeWhile (Info.member kExpansionInfo . (^. lambdaLhsInfo)) lams
-     in if
-            | n < length lams ->
-                ValueFun
-            | otherwise ->
-                case body of
-                  NCtr c ->
-                    toValue tab (NCtr (over constrArgs (dropEnd n) c))
-                  _ ->
-                    ValueFun
+  NCtr c -> goConstr c
+  NLam lam -> goLambda lam
   _ -> impossible
+  where
+    goConstr :: Constr -> Value
+    goConstr Constr {..} =
+      ValueConstrApp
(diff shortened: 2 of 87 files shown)


app/Commands/Repl.hs
```diff
@@ -147,12 +148,18 @@ runCommand opts = do
...
           Just ctx' -> do
+            let tab = ctx' ^. replContextArtifacts . artifactCoreTable
             evalRes <- compileThenEval ctx' input
             case evalRes of
               Left err -> printError err
               Right n
                 | Info.member Info.kNoDisplayInfo (Core.getInfo n) -> return ()
-              Right n -> renderOut (Core.ppOut opts n)
+              Right n ->
+                if
```
janmasrovira, line 158: `if` -> guards

src/Juvix/Compiler/Internal/Extra.hs
```diff
@@ -161,7 +161,7 @@ instance HasExpressions FunctionDef where
         }
 
 instance HasExpressions InductiveParameter where
-  leafExpressions _ param@(InductiveParameter _) = do
+  leafExpressions _ param@(InductiveParameter {}) = do
```
janmasrovira, line 164: remove parens

src/Juvix/Compiler/Internal/Pretty/Base.hs
```diff
@@ -157,7 +157,7 @@ ppBlock ::
 ppBlock items = vsep . toList <$> mapM ppCode items
 
 instance PrettyCode InductiveParameter where
-  ppCode (InductiveParameter v) = do
+  ppCode (InductiveParameter v _) = do
```
janmasrovira, line 160: let's use lenses or `{..}`

src/Juvix/Compiler/Internal/Language.hs
```diff
@@ -342,7 +343,7 @@ instance HasAtomicity Pattern where
     PatternVariable {} -> Atom
 
 instance HasLoc InductiveParameter where
-  getLoc (InductiveParameter n) = getLoc n
+  getLoc (InductiveParameter n _) = getLoc n
```
janmasrovira, line 346: let's use lenses or `{..}`

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Positivity/Checker.hs
```diff
@@ -142,7 +142,7 @@ checkStrictlyPositiveOccurrences ty ctorName name recLimit ref =
 
         helperInductiveApp :: InductiveDef -> [(InductiveParameter, Expression)] -> Sem r ()
         helperInductiveApp typ = \case
-          ((InductiveParameter pName, arg) : ps) -> do
+          ((InductiveParameter pName _, arg) : ps) -> do
```
janmasrovira, line 145: let's use lenses or `{..}`

src/Juvix/Compiler/Core/Pretty/Base.hs
```diff
@@ -536,8 +541,59 @@ instance (PrettyCode a) => PrettyCode [a] where
...
+instance PrettyCode ConstrApp where
+  ppCode ConstrApp {..} = do
+    n <- ppName KNameConstructor _constrAppName
+    case _constrAppFixity of
+      Nothing -> do
+        args <- mapM (ppRightExpression appFixity) _constrAppArgs
+        return $ hsep (n : args)
+      Just fixity
+        | isBinary fixity ->
+            case _constrAppArgs of
```
janmasrovira, line 557: i'd prefer to have `goBinary`, `goUnary` defined in the where block to improve readability

src/Juvix/Compiler/Core/Extra/Value.hs
```diff
@@ -0,0 +1,36 @@
...
+import Juvix.Compiler.Core.Extra.Base
+import Juvix.Compiler.Core.Info qualified as Info
+import Juvix.Compiler.Core.Info.ExpansionInfo (kExpansionInfo)
+import Juvix.Compiler.Core.Language
+import Juvix.Compiler.Core.Language.Value
+
+toValue :: InfoTable -> Node -> Value
+toValue tab = \case
+  NCst Constant {..} -> ValueConstant _constantValue
+  NCtr Constr {..} ->
```
janmasrovira, line 13: I'd prefer to have `goConstr`, `goLam` in the where block to improve readability

src/Juvix/Compiler/Internal/Language.hs
```diff
@@ -230,8 +230,9 @@ data Pattern
 
 instance Hashable Pattern
 
-newtype InductiveParameter = InductiveParameter
-  { _inductiveParamName :: VarName
+data InductiveParameter = InductiveParameter
+  { _inductiveParamName :: VarName,
+    _inductiveParamImplicit :: IsImplicit
```
janmasrovira, line 235: Why do we need this? Isn't it the case that all inductive parameters must be explicit?

lukaszcz: I think I've been overzealous here. There is some indication if a parameter is implicit/explicit in Internal or earlier, and I just propagated it down to Core. Ultimately, it's not really used. Do you think I should revert the propagation of implicity information for parameters?

janmasrovira: I think you are referring to the `Abstract` language, where we `Abstract.FunctionParameter` (which has `_paramImplicit`) instead of defining a separate `InductiveParameter` data type without the `isImplicit` field. 
Since we are not going to have implicit parameters for inductive types in the near future, I'd say it is better to revert the change. And since we plan on removing `Abstract`, the anomaly there will also be taken care of

lukaszcz: Okay, I'll do that then

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Positivity/Checker.hs
lukaszcz: reverted

src/Juvix/Compiler/Internal/Language.hs
lukaszcz: reverted

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Positivity/Checker.hs
lukaszcz: reverted

src/Juvix/Compiler/Internal/Pretty/Base.hs
lukaszcz: reverted

src/Juvix/Compiler/Core/Pretty/Base.hs
lukaszcz: done

src/Juvix/Compiler/Core/Extra/Value.hs
lukaszcz: done

src/Juvix/Compiler/Internal/Language.hs
lukaszcz: done

janmasrovira (approved):

Issue #1998, opened 2023-04-13
`juvix compile` output file is not written to the invoke directory


paulcadman:
To reproduce, say `$JUVIX_ROOT` is the root of the juvix repository

```
cd /tmp
juvix compile $JUVIX_ROOT/examples/milestone/HelloWorld.juvix
```

### Expected behaviour

The output file is written to `/tmp/HelloWorld`

### Actual behaviour

The output file is written to `$JUVIX_ROOT/examples/milestone/HelloWorld`, i.e in the same directory as the source file.

PR #1999, opened 2023-04-13
Write compile output file to invoke dir by default

paulcadman:
This PR fixes the issue below and also adds smoke tests for the target option of the compile command.

* Closes https://github.com/anoma/juvix/issues/1998


janmasrovira (approved):
thanks

PR #2003, opened 2023-04-13
Fix:  broken links of Example programs 

### LINKED ISSUE #2001: Example programs links are not working
Example links are not working on the [Example programs](https://docs.juvix.org/reference/examples.html) page.

E.g. this link from the sources of the page: https://docs.juvix.org/examples/html/HelloWorld/HelloWorld.html

My assumption would be that the generated examples html artifacts are not uploaded during the CI deployment. I tried to check that in the pipeline, but not sure what is wrong in there. 

Locally `make html-examples` works for me and creates all the necessary html files under the `_docs/examples/html/` folder.

jonaprieto:
This PR fixes the broken links that refer to Juvix Examples in the documentation. Ideally, this wouldn't occur since we utilize a link checker, but this tool only works well for relative links, which was not the case for the links of the examples.  Additionally, I slightly modified the CI workflow by generating the HTML for the examples first, followed by the entire book.

- Closes  #2001
- Check the output using this: https://jonaprieto.github.io/juvix/


vrom911 (approved):
Thanks so much!

PR #2004, opened 2023-04-13
Qualified imports

janmasrovira:
This pr extends the `import` statement to support optional renaming. The syntax is as follows:
```
import Stdlib.Data.Nat as Nat;
import Stdlib.Data.Nat as Qual.Nat;
```

It is allowed to give the same name to multiple modules:
```
import Stdlib.Data.Nat as X;
import Stdlib.Data.Int as X;

axiom a : X.Nat;
axiom b : X.Int;

open X; -- This will cause an ambiguous name error
```


paulcadman (approved):
Very nice - I tried it in my containers and quickcheck projects and it works well.

PR #2009, opened 2023-04-16
Ide improvements

janmasrovira:
This pr includes the necessary machinery to have type and judoc information for every identifier. This information is showed in juvix-mode as a child-frame on the top right corner as showed in the following screenshot.
image


paulcadman (approved):
This works really well!

Issue #2008, opened 2023-04-14
Add support to read from stdin for the `juvix format` command


jonaprieto:
At present, the `juvix format` requires a file path or a directory as input for formatting Juvix code. The primary motivation behind this option is to substitute the `dev scope` call in the VSCode extension, enabling us to invoke the `format` command more flexibly and read directly from stdin. Also, notice that this flag `--stdin` is offered as global, so one really expects this works.

PR #2019, opened 2023-04-20
Add format stdin smoke tests

vrom911:
* Exploration of #2008 

As I were trying to fix the behaviour of `format` command with the `--stdin` global option, I noticed that actually it is already behaves the same was as `dev scope` command (which was the desire at least at this point).

So, no actual behaviour of the command is channged, only a few tests added to showcase some edge cases of the command with the `--stdin` option together.

Here are the highlights:

* The filename is **not** optional for both `dev scope` and `format` even if with `stdin` the content of the file comes from the input.
* `format` command expects the stdin input to be a proper module
* the name of the stdin module should be aligned with the provided filename even though its sources are not used

I think that some of the above behaviours we can consider to change. Let me know what do you think about any of that.

Otherwise, the `dev scope` command can be replaced with `format`.

### DIFF
--- a/tests/smoke/Commands/format.smoke.yaml
+++ b/tests/smoke/Commands/format.smoke.yaml
@@ -128,3 +128,44 @@ tests:
     stderr:
       contains: "error"
     exit-status: 1
+
+  - name: format-stdin
+    command:
+      - juvix
+      - --stdin
+      - format
+      - positive/Format.juvix
+    stdin: "module Format; open import Stdlib.Prelude; main : Nat; main := 5; "
+    stdout:
+      contains: |
+        module Format;
+
+        open import Stdlib.Prelude;
+
+        main : Nat;
+        main := 5;
+    exit-status: 1
+
+  - name: format-stdin-file-does-not-exist
+    command:
+      - juvix
+      - --stdin
+      - format
+      - positive/NonExistingFormat.juvix
+    stdin: "module Format; open import Stdlib.Prelude; main : Nat; main := 5; "
+    stderr:
+      contains: |
+        positive/NonExistingFormat.juvix: openFile: does not exist (No such file or directory)
+    exit-status: 1
+
+  - name: format-stdin-module-name-not-file-name
+    command:
+      - juvix
+      - --stdin
+      - format
+      - positive/Format.juvix
+    stdin: "module OtherFormat; open import Stdlib.Prelude; main : Nat; main := 5; "
+    stderr:
+      contains: |-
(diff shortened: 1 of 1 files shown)


tests/positive/issue1731/builtinTrace.juvix
```diff
@@ -8,7 +8,7 @@ axiom trace : {A : Type} → A → A;
 
 terminating
 f : Nat → Nat → Nat;
-f x y := if (x == 0) y (trace x >>> (f (sub x 1) y));
+f x y := if (x == 0) y (trace x >>> f (sub x 1) y);
```
jonaprieto, line 11: Just curiosity, did you change this manually? or it was the formatter?

vrom911: That is after `make pre-commit` command work, not sure how it was not caught earlier

jonaprieto: Exactly. pre-commit locally runs the Juvix formatter and the CI as well. I'm supprised, a new mystery to my stack.

jonaprieto:
Is this ready to review or is it a draft? I just rebased it because the CI failed to build the project.

vrom911:
@jonaprieto ye, this PR only adds tests. I am working on what we discussed today, which is handling the command without filepath but with stdin better.

Regarding failing tests, locally, my tests are passing. It seem to be different behaviour possibly due to the Linux machine run? Not sure that this is the only reason for failure

vrom911:
As figured out, I somehow used the older version of the `juvix` locally, the one of a week before, that is why the previous difference with local vs CI testing output.

Can confirm, that currently the `format` command for sure doesn't fail or throw `ErrWrongTopModuleName` anymore. This solves the inconvenience with the `--stdin` command, however, still it is better to not require the filename when intended to use with `--stdin`. Currently working on that.

Also, I wonder what the priority should be if you specify both `--stdin` and the filename ideally?
I would say that reading file should be prioritised over the stdin input in that cases, however, I can see the other way around.
For now, I will keep the `--stdin` priority as it is, but let me know if we want to change that, and I can create the issue to tackle this problem.

jonaprieto:
> better to not require the filename when intended to use with `--stdin`. Currently working on that.

:+1: 

> I wonder what the priority should be if you specify both `--stdin` and the filename ideally?

Based on the usage case, that is the vscode extension, I'd keep giving priority to --stdin input.

jonaprieto (approved):

PR #2028, opened 2023-04-25
Make format command's filepath optional

### LINKED ISSUE #2008: Add support to read from stdin for the `juvix format` command

vrom911:
# Description

No the filepath in the `juvix forma` command is n=made optional. However, in that case, the `--stdin` command is required.

### Implementation details

~For now, as a quick solution, I have introduce the "fake" path that is used for `fomat` command with stdin option.~
I needed to do a couple of big changes:
 * `format` command FILE is now optional, howvere, I check that in case of `Nothing` `--stdin` option should be present, otherwise it will fail
 * `entryPointModulePaths` is now `[]` instead of `NonEmpty`
 * `ScopeEff` now has `ScopeStdin` constructor as well, which would take the input from stdin instead of having path passed around
 * `RunPipelineNoFileEither` is added to the `App` with the bunch of `*Stdin` functions that doesn't require filepath argument to be passed

Fixes #2008

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works:
  - [x] smoke tests


src/Juvix/Parser/Error.hs
```diff
@@ -12,13 +13,15 @@ data ParserError
   = ErrMegaparsec MegaparsecError
   | ErrTopModulePath TopModulePathError
   | ErrWrongTopModuleName WrongTopModuleName
+  | ErrStrinOrFile StdinOrFileError
```
paulcadman, line 16: typo
```suggestion
  | ErrStdinOrFile StdinOrFileError
```

src/Juvix/Parser/Error.hs
```diff
@@ -12,13 +13,15 @@ data ParserError
...
   deriving stock (Show)
 
 instance ToGenericError ParserError where
   genericError = \case
     ErrMegaparsec e -> genericError e
     ErrTopModulePath e -> genericError e
     ErrWrongTopModuleName e -> genericError e
+    ErrStrinOrFile e -> genericError e
```
paulcadman, line 24:
```suggestion
    ErrStdinOrFile e -> genericError e
```

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -28,10 +34,22 @@ fromSource ::
...
+  (_resultTable, _resultModules) <- runParserInfoTableBuilder (runReader e getParsedModuleTops)
   let _resultEntry = e
   return ParserResult {..}
   where
+    getParsedModuleTops ::
+      forall r.
+      (Members '[PathResolver, Files, Error ParserError, InfoTableBuilder, NameIdGen] r) =>
+      Sem r (NonEmpty (Module 'Parsed 'ModuleTop))
+    getParsedModuleTops = case (e ^. entryPointStdin, e ^. entryPointModulePaths) of
+      (Nothing, []) -> throw $ ErrStrinOrFile StdinOrFileError
```
paulcadman, line 46:
```suggestion
      (Nothing, []) -> throw $ ErrStdinOrFile StdinOrFileError
```

app/Commands/Format.hs
```diff
@@ -17,17 +18,32 @@ data FormatRenderMode
...
-      Right p -> formatProject p
+      Just (Left p) -> format p
+      Just (Right p) -> formatProject p
+      Nothing ->
+        if isStdin
+          then formatStdin
+          else do
+            printFailureExit $
+              Text.unlines
+                [ "juvix format error: either 'JUVIX_FILE_OR_PROJECT' or 'stdin' option should be specified",
```
paulcadman, line 43:
```suggestion
                [ "juvix format error: either 'JUVIX_FILE_OR_PROJECT' or '--stdin' option must be specified",
```

app/Commands/Format.hs
```diff
@@ -17,17 +18,32 @@ data FormatRenderMode
...
+                  "Use --help to learn more usage information."
```
paulcadman, line 44:
```suggestion
                  "Use the --help option to display more usage information."
```

app/Commands/Format.hs
```diff
@@ -17,17 +18,32 @@ data FormatRenderMode
...
+        Just (Right {}) -> TargetDir
+        Nothing -> TargetStdin
   runOutputSem (renderFormattedOutput target opts) $ runScopeFileApp $ do
     res <- case f of
-      Left p -> format p
-      Right p -> formatProject p
+      Just (Left p) -> format p
+      Just (Right p) -> formatProject p
+      Nothing ->
+        if isStdin
```
paulcadman, line 38: Style: We prefer to use multiway-if rather than `if ... then ... else`.

i.e change to

```
if 
   | isStdin -> ...
   | otherwise -> ...
```

vrom911: Good to know! Changed to that 🙂 
Do we have a style guide on all these things I can check?

paulcadman: Unfortunately we do not have anything written down. I think we should, as long as we make it clear that we always expect exceptions tot he guidelines. Let's discuss with the whole team.

vrom911: Sounds great!

paulcadman (approved):
Nicely done 🌟

janmasrovira (commented):
I have some suggestions

src/Juvix/Formatter.hs
```diff
@@ -98,6 +93,38 @@ formatProject p = do
...
+
+formatResultFromContents ::
+  forall r.
+  Members '[Output FormattedFileInfo] r =>
+  Text ->
+  NonEmpty AnsiText ->
+  Path Abs File ->
+  Sem r FormatResult
+formatResultFromContents originalContents formattedContents filepath
+  | originalContents /= (ansiPlainText formattedContents) = do
```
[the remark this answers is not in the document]
vrom911: Yess!
Do we run `hlint` somewhere?
Maybe it would make sense to add ti to `pre-commit`?

src/Juvix/Prelude/Lens.hs
```diff
@@ -6,6 +6,19 @@ import Juvix.Prelude.Base
 _head1 :: Lens' (NonEmpty a) a
 _head1 = singular each
 
+-- | points to the first element of a the list.
```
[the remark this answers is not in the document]
vrom911: True, I can use `_head` lens, but for implementin the lens like:
```haskell
mainModulePath :: Lens' EntryPoint (Maybe (Path Abs File))
mainModulePath = entryPointModulePaths . _headMaybe
```

will still need to write it manually.
Would you suggest still to use `^?` and `_head` everywhere else?

app/App.hs
```diff
@@ -6,7 +6,9 @@ import GlobalOptions
 import Juvix.Compiler.Concrete.Translation.FromParsed.Analysis.PathResolver
 import Juvix.Compiler.Pipeline
 import Juvix.Data.Error qualified as Error
-import Juvix.Prelude.Pretty hiding (Doc)
+import Juvix.Prelude.Pretty hiding
+  ( Doc,
+  )
```
jonaprieto, line 11: revert this change. less diff as much as we can.

app/Commands/Dev/Geb/Repl.hs
```diff
@@ -2,7 +2,9 @@
 
 module Commands.Dev.Geb.Repl where
 
-import Commands.Base hiding (command)
+import Commands.Base hiding
+  ( command,
+  )
```
jonaprieto, line 7: Idem.

app/App.hs
vrom911: I am pretty sure this is ormolu, but let me double check

vrom911: Yes, this is the result of the `make ormolu` command for me

app/GlobalOptions.hs
```diff
@@ -150,6 +150,9 @@ entryPointFromGlobalOptionsPre roots premainFile opts = do
   mainFile <- prepathToAbsFile (roots ^. rootsInvokeDir) premainFile
   entryPointFromGlobalOptions roots mainFile opts
 
+entryPointFromGlobalOptionsNoFilePre :: Roots -> GlobalOptions -> IO EntryPoint
+entryPointFromGlobalOptionsNoFilePre roots opts = entryPointFromGlobalOptionsNoFile roots opts
```
jonaprieto, line 154: Do you really need to define this?

app/Commands/Dev/Geb/Repl.hs
```diff
@@ -60,17 +62,21 @@ loadEntryPoint ep = do
...
-  let evalRes =
-        Geb.runEval $
-          Geb.RunEvalArgs
-            { _runEvalArgsContent = content,
-              _runEvalArgsInputFile = epPath,
-              _runEvalArgsEvaluatorOptions = Geb.defaultEvaluatorOptions
-            }
-  printEvalResult evalRes
+  let epPath :: Maybe (Path Abs File) = ep ^. entryPointModulePaths . _headMaybe
+  case epPath of
```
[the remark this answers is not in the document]
vrom911: Done 👍🏼

src/Juvix/Prelude/Lens.hs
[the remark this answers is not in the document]
janmasrovira: That's true, I think we should have a `Traversal'` instead. 
```
mainModulePath :: Traversal' EntryPoint (Path Abs File)
mainModulePath = entryPointModulePaths . _head
```

Then in the call sites we will need to do `entry ^? mainModulePath`

Ideally we would use a `Prism` but microlens doesn't have them

src/Juvix/Compiler/Pipeline.hs
```diff
@@ -192,6 +192,8 @@ upToEval =
 -- Internal workflows
 --------------------------------------------------------------------------------
 
+instance Member Parser.PragmasStash '[]
```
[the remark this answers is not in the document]
vrom911: Thanks! This makes sense 👍🏼

app/App.hs
jonaprieto: Which version of ormolu are you running? The CI runs "ormolu 0.5.2.0 HEAD 810eb7839cbf4e358821447e8f4603955999e62d using ghc-lib-parser 9.4.2.20220822"

vrom911: I'm using `0.5.2.0` too:
Screenshot 2023-04-27 at 12 44 19

jonaprieto (approved):

jonaprieto:
Good job

PR #2036, opened 2023-04-26
Inlining

### LINKED ISSUE #1989: Inlining
Since we now focus on the GEB pipeline, it makes sense to implement inlining on the Core level before tackling more low-level optimisations in the executable pipeline. Optimisations on JuvixCore would also affect the GEB pipeline. With the infrastructure we've built up and the `rmap` recursor, it is actually easy to implement inlining once we know which functions to inline. At first, these should just be explicitly specified by the user (issue #1965).

Without inlining, using small higher-order functions from the standard library, e.g., composition, `flip`, `maybe`, incurs a significant overhead.

The main work is actually to implement the parsing of pragmas first: issue #1965.

lukaszcz:
* Closes #1989 
* Depends on #1997 

Work in progress


paulcadman (dismissed):

jonaprieto (approved):
Already approved

PR #2038, opened 2023-04-26
Normalization by Evaluation

### LINKED ISSUE #2032: Normalization by evaluation
Implement normalization by evaluation with the sharing of common normal subexpressions of type Int via lets.

lukaszcz:
* Closes #2032 

Work in progress


janmasrovira:
Could we use `yaml` instead of `json` for the tests?
I feel like `yaml` is much more human friendly.

lukaszcz:
> Could we use `yaml` instead of `json` for the tests?
> I feel like `yaml` is much more human friendly.

The files are in the format required for VampIR (https://github.com/anoma/VampIR-Book/blob/main/src/section_3_6.md). The idea is that these tests will later be used with VampIR, i.e., you compile to VampIR, run it and see if everything is fine.

src/Juvix/Compiler/Core/Normalizer.hs
```diff
@@ -0,0 +1,177 @@
+module Juvix.Compiler.Core.Normalizer where
```
jonaprieto, line 1: What do you think if we better place this file under the transformation directory?

lukaszcz: We can, depends how you look at it. I think of it as more analogous to evaluation than an intermediate transformation on Core.

jonaprieto (changes requested):
The PR looks good to me overall. One thing. It currently lacks tests for potential user interactions, particularly for the new `:n` commands and other associated flags. Therefore, I recommend the following:

- [x] Update `tests/smoke/Commands/dev/core.smoke.yaml` to include:
   - [x] `:n` smoke test. Refer to examples in `geb.smoke.yaml` or `repl.smoke.yaml` for guidance.
   - [x] Test for the `--normalize` flag with the `juvix dev core read` command.
   - [ ] Test for the `--normalize` flag with the `juvix dev core from-concrete` command.

src/Juvix/Compiler/Core/Normalizer.hs
jonaprieto: Okey.

lukaszcz:
> The PR looks good to me overall. One thing. It currently lacks tests for potential user interactions, particularly for the new `:n` commands and other associated flags. Therefore, I recommend the following:
> 
> * [ ]  Update `tests/smoke/Commands/dev/core.smoke.yaml` to include:
>   
>   * [ ]  `:n` smoke test. Refer to examples in `geb.smoke.yaml` or `repl.smoke.yaml` for guidance.
>   * [ ]  Test for the `--normalize` flag with the `juvix dev core read` command.
>   * [ ]  Test for the `--normalize` flag with the `juvix dev core from-concrete` command.

It lacks tests for potential user interactions, because the only expected users for those interactions are the developers. But I will the smoke tests.

jonaprieto (approved):

PR #2040, opened 2023-04-27
Fix ordering of statements in Abstract -> Internal

### LINKED ISSUE #2039: Type synonyms inside local modules are not handled properly
The following example fails to typecheck:
```
module LocalSynonym;

open import Stdlib.Prelude;

bool : Type;
bool := Bool;

module L;
  haha : bool;
  haha := false;
end;
```
with the error:
```
The expression false has type:
  Bool
but is expected to have type:
  bool
```
This behaviour only happens inside the local module. So I suspect it can be related to effect shadowing

### LINKED ISSUE #2055: Type aliases don't normalise correctly
The following T1 is not strictly positive, but it type checks atm (v0.3.2).
Notice the alias converts to T0 putting T1 in a negative position.

```agda
-- tests/negative/Internal/Positivity/E10.juvix
module E10;
type T0 (A : Type) :=
  | t : (A -> T0 A) -> T0 A;

alias : Type -> Type;
alias := T0;

type T1 :=
  | c : alias T1 -> T1;
```agda

This might be solved by:

- https://github.com/anoma/juvix/pull/2040

### PR DESCRIPTION
**Pull request by janmasrovira**
- Closes #2039 
- Closes #2055 
- Depends on #2053 

Changes in this pr:
- Local modules are removed (flattened) in the translation abstract -> internal.
- In the translation abstract -> internal we group definitions in mutually recursive blocks. These blocks can contain function definitions and type definitions. Previously we only handled functions.
- The translation of Internal has been enhanced to handle these mutually recursive blocks.
- Some improvements the pretty printer for internal (e.g. we now print builtin tags properly).
- A "hack" that puts the builtin bool definition at the beginning of a module if present. This was the easiest way to handle the implicit dependency of the builtin stringToNat with bool in the internal-to-core translation.
- A moderately sized test defining a simple lambda calculus involving and an evaluator for it. This example showcases mutually recursive types in juvix.

### CODE REVIEW

File: tests/positive/MutualType.juvix [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+type List (a : Type) :=
+  | --- The empty list
+    nil : List a
+  | --- An element followed by a list
+    :: : a → List a → List a;
+
+Forest : Type -> Type;
+
+--- N-Ary tree.
+positive
```
janmasrovira, line 14: remove positive

jonaprieto (approved):

PR #2043, opened 2023-05-01
Fix fromSource behaviour when both stdin and filename

vrom911:
# Description

While working on highlight command, I noticed that after `format` command changes the behaviour of the `highlight` command when both `--stdin` option specified and FILENAME is provided is changed.
Instead on returning the result for the `stdin` input, it started to return empty result.

This PR fixes that behaviour, it will now act exactly as before introducing the changes that came with the `format` command fixes.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works:
  - [x] smoke test


janmasrovira (commented):
Since we are relying on this behaviour for `juvix --stdin typecheck FileName.juvix` in the emacs mode as well, it'd be nice to add a smoke test for that. Where we check that the file name is indeed part of the error

janmasrovira (approved):

Issue #2077, opened 2023-05-10
error when format run on a project containing a subproject


paulcadman:
Say you have a juvix project like:

```
.
├── Format.juvix
├── juvix.yaml
└── sub
    ├── SubFormat.juvix
    └── juvix.yaml
```

i.e an outer juvix project containing an inner subproject.

The file `Format.juvix` should be processed when running `juvix format .` from the root of the outer project and `SubFormat.juvix` should not be processed as it is a member of the inner project. The current behaviour is that the subproject `sub` is traversed. This causes an error because the file `SubProject.juvix` is scoped as a member of the outer project:

```
 error:
The top module SubProject is defined in the file:
/sub/SubProject.juvix
But it should be in the file:
/SubProject.juvix
```

The format command should not recurse into directories which contain a `juvix.yaml` file.

PR #2078, opened 2023-05-10
Fix format on projects that contain subprojects

paulcadman:
This PR fixes the behaviour of the `format` command when run on a project that contains subprojects. Files in subprojects must not be processed by the formatter.

The format issue was caused by a bug in the `walkDirRel` function that is used to traverse a file system tree:

https://github.com/anoma/juvix/blob/9a64184253eb9129d2ac36d538d0c5f55cc46d35/src/Juvix/Data/Effect/Files.hs#L36

In this function, the passed handler can return a function that determines if a candidate subdirectory should be traversed. The first argument of this function indicates if the candidate subdirectory contains a juvix.yaml file. In the formatter and the path resolver we use this argument to exclude such subdirectories from the traversal.

Previously the first argument was calculated from the files in the current directory instead of the candidate subdirectory - which was the source of the bug.

The callers of walkDirRel are also fixed to match the updated behaviour.

* Closes https://github.com/anoma/juvix/issues/2077


janmasrovira (approved):
great!

PR #2086, opened 2023-05-12
Direct translation from normalized JuvixCore to VampIR

### LINKED ISSUE #2034: Direct translation of JuvixCore to VampIR
Implement a direct translation of JuvixCore to VampIR.

It can be assumed that we're translating a single main function of type `Int -> ... -> Int -> Int` which has the form `\x1 ... \xn -> b` where the body `b` can contain only:
* the variables `x1`,...,`xn`,
* integer constants,
* fully applied arithmetic operations and comparisons (`OpInt*` and `OpEq` from `BuiltinOp` in Core),
* if-then-else, i.e., case expressions on booleans,
* boolean constructors (with tags `TagTrue` and `TagFalse`),
* fully applied failure nodes (`OpFail`),
* `let` bindings of type `Int` at the top only (these should be translated to VampIR local definitions; see also #2033).

In particular, `b` doesn't contain lambdas, identifiers or inductive types other than booleans.

Ultimately, we should support arbitrary finite first-order data types in the type of the main function, with some fixed encoding to tuples of integers/field elements. It is easy to support `Bool` in addition to `Int`.

A quick and generally incorrect way of translating the failure nodes is to replace them with 0. The resulting VampIR programs will compile because Int is the only type. They will work correctly as long as the recursion limit is not reached. Perhaps this should be done at first. A proper way is to encode them with pairs (see B.7 in the Juvix-to-VampIR pipeline report).

Implementing booleans in VampIR is explained in the VampIR book (section [Boolean logic](https://github.com/anoma/VampIR-Book/blob/main/src/section_4_2.md)). If-then-else with both branches integers can be implemented as:
```
def if b x y = b * x + (1 - b) * y;
```
Equality checking can be implemented as:
```
def isZero x = {
  def xi = fresh (1 | x);
  x * (1 - xi * x) = 0;
  1 - xi * x
};
def equal x y = isZero (x - y);
```
I'm not sure how we should implement less-than, but we can ignore it for now.

References
-----------
* [VampIR book](https://github.com/anoma/VampIR-Book/blob/main/src/SUMMARY.md)
* [Juvix to VampIR pipeline report](https://heliax.slack.com/files/U03MUDTCR3L/F055JR1B0D6/vampir-pipeline-report.pdf)

lukaszcz:
* Closes #2034.
* Adds the `vampir` target to the `compile` command.
* Adds two tests which are not yet enabled because `vamp-ir` is not available in the CI (these and more tests will be enabled in #2103).

### DIFF
--- a/src/Juvix/Compiler/Backend/Html/Translation/FromTyped.hs
+++ b/src/Juvix/Compiler/Backend/Html/Translation/FromTyped.hs
@@ -356,11 +356,19 @@ goJudocMay :: (Members '[Reader HtmlOptions, Reader NormalizedTable] r) => Maybe
 goJudocMay = maybe (return mempty) goJudoc
 
 goJudoc :: forall r. (Members '[Reader HtmlOptions, Reader NormalizedTable] r) => Judoc 'Scoped -> Sem r Html
-goJudoc (Judoc bs) = mconcatMapM goBlock bs
+goJudoc (Judoc bs) = mconcatMapM goGroup bs
   where
+    goGroup :: JudocGroup 'Scoped -> Sem r Html
+    goGroup = \case
+      JudocGroupLines l -> mconcatMapM goBlock l
+      JudocGroupBlock l -> goGroupBlock l
+
+    goGroupBlock :: JudocBlockParagraph 'Scoped -> Sem r Html
+    goGroupBlock = mconcatMap goBlock . (^. judocBlockParagraphBlocks)
+
     goBlock :: JudocBlock 'Scoped -> Sem r Html
     goBlock = \case
-      JudocParagraph ls -> Html.p . concatWith (\l r -> l <> " " <> r) <$> mapM goLine (toList ls)
+      JudocParagraphLines ls -> Html.p . concatWith (\l r -> l <> " " <> r) <$> mapM goLine (toList ls)
       JudocExample e -> goExample e
 
     goLine :: JudocParagraphLine 'Scoped -> Sem r Html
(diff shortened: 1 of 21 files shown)


src/Juvix/Compiler/Backend/VampIR/Translation/FromCore.hs
```diff
@@ -0,0 +1,82 @@
...
+          _ -> impossible
+      NCtr Constr {..} -> case _constrTag of
+        BuiltinTag TagTrue -> ExpressionConstant 1
+        BuiltinTag TagFalse -> ExpressionConstant 0
+        _ -> impossible
+      NCase c -> translateCase translateIf impossible c
+        where
+          translateIf :: Node -> Node -> Node -> Expression
+          translateIf val br1 br2 = ExpressionIfThenElse $ IfThenElse (convertExpr val) (convertExpr br1) (convertExpr br2)
+      _ -> impossible
```
janmasrovira, line 82: I'd be happier if we avoided having long `case` expressions with big right hand sides. I suggest to follow the `go*` pattern to ease readability

src/Juvix/Compiler/Backend/VampIR/Pretty/Keywords.hs
```diff
@@ -0,0 +1,38 @@
+module Juvix.Compiler.Backend.VampIR.Pretty.Keywords where
+
+import Juvix.Compiler.Backend.VampIR.Language
+import Juvix.Data.CodeAnn
+import Juvix.Extra.Strings qualified as Str
+
+kwDef :: Doc Ann
```
janmasrovira, line 7: Since we introduced `Data.Keyword`, it would make sense to use it here. It will make these definitions more reusable (e.g. in a parser).
Feel free to ignore this if you think otherwise

lukaszcz: I adapted this from analogous code for GEB. Maybe it would make sense to refactor the GEB code, because there we have both keywords based on Data.Keyword and ones for printing implemented like here.

But for VampIR I think using Data.Keyword is an overkill. We will not be parsing VampIR input, only generating it, and most fields in Keyword are redundant. We essentially only need a name for a keyword string.

jonaprieto (approved):

PR #2096, opened 2023-05-15
Add VampIR to the CI

jonaprieto:
This PR:

- Makes `vamp-ir` available in the CI (pre-release 0.1.2)
- [Use a setup-wasmer action to install `wasmer`](https://github.com/marketplace/actions/setup-wasmer)
- Fixes cache option value for `jaxxstorm/action-install-gh-release`'s usages

Adds support for:

- #2103 

Related: 

- https://github.com/anoma/vamp-ir/issues/90

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,11 +145,22 @@ jobs:
         with:
           working-directory: smoke
 
-      - name: Install Smoke
-        shell: bash
-        run: |
-          cd smoke
-          stack install --stack-root ${{ steps.stack.outputs.stack-root }}
+      # - name: Install Smoke
+      #   shell: bash
+      #   run: |
+      #     cd smoke
+      #     stack install --stack-root ${{ steps.stack.outputs.stack-root }}
+
+      - name: Install Smoke for testing
+        uses: jaxxstorm/action-install-gh-release@v1.10.0
+        with:
+          repo: jonaprieto/smoke
+          platform: linux
+          tag: v2.3.2
+          chmod: 0755
+          rename-to: smoke
+          extension-matching: disable
+          cache: enable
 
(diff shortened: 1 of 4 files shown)


paulcadman:
We stopped using the setup-wasmer action because it was unreliable https://github.com/anoma/juvix/pull/1765 - is there any reason to change back to it now?

jonaprieto:
> We stopped using the setup-wasmer action because it was unreliable #1765 - is there any reason to change back to it now?
> 
> I remember that https://get.wasmer.io would periodically go down, blocking the CI.

I got an error that wasmer was not found in the PATH: https://github.com/anoma/juvix/actions/runs/5023502428/jobs/9008167955#step:14:30
This was just after changing the corresponding value for `cache` from `true` to `enable`. Weird. So I remember setup-wasmer, and went to check the repo. I found it has been receiving a couple of updates lately.

paulcadman:
> So I remember setup-wasmer, and went to check the repo. I found it has been receiving a couple of updates lately.

I'm not sure, it's still using https://get.wasmer.io/ in the action. 

Looking at the action failures, it's restored something it thinks is the wasmer binary from the cache. But something went wrong. This is perhaps a bug with the install-gh-release action?

jonaprieto:
> > So I remember setup-wasmer, and went to check the repo. I found it has been receiving a couple of updates lately.
> 
> I'm not sure, it's still using https://get.wasmer.io/ in the action.
> 
> Looking at the action failures, it's restored something it thinks is the wasmer binary from the cache. But something went wrong. This is perhaps a bug with the install-gh-release action?
Yes, perhaps.
I suggest we merge this and possibly investigate this issue later.

paulcadman (approved):

PR #2098, opened 2023-05-15
Modify `open import` syntax

### LINKED ISSUE #2013: Qualified import ignored when using `open import` variant
| ⚠️ The following example is available in a branch at https://github.com/anoma/juvix/tree/open-qualified-import, specifically at https://github.com/anoma/juvix/blob/2c08ab08363bbc524db4e010f47b52a7a4283a46/tests/positive/QualifiedImports/Main.juvix

When qualifying an `open import`, the qualified name cannot be used.

For example:

```
module Foo;

open import Stdlib.Data.Ord as O;

axiom e : Ordering;

axiom f : O.Ordering;
```

The current behaviour is that `O.Ordering` is a scope error (`Qualified symbol not in scope`). My expectation is that `O.Ordering` would be in scope.

I can't find an argument why we shouldn't support this. Perhaps qualified `open import` is never what one would want?

### LINKED ISSUE #2088: Move the `open` keyword in `open import` statements
- As previously discussed in #2013, we will change the syntax of `open import` statements to be more intuitive.

The new syntax will follow this pattern:
```
<import statement> <open statement without module name>;
```

The plan is to support parsing both the new and old syntax for a short while and use the formatter to refactor existing code.

janmasrovira:
- Closes #2088 
- Closes #2013 

This pr implements the syntax changes proposed in #2088.
It also makes it possible to combine `import as` and `open` in the same statement. Note that this was unsupported until now. E.g. 
```
import Stdlib.List as List open;
```
Which is equivalent to:
```
import Stdlib.List as List;
open List;
```

The old syntax is still supported by the parser, but the formatter will always print code in the new syntax.
Since this leaves the parser in a temporary state, I didn't think it was worth to put a lot of effort on it, so I have left a temporary `P.try (StatementOpenModule <$> newOpenSyntax)` which tries to parse the new syntax. This `try` will need to be removed when we remove the old syntax. Also, we will need to update the `HasLoc` instance for `open` statements to reflect the syntax changes.

I have formatted only the formatting tests. The rest of the juvix files should be formatted in an isolated pr.

### DIFF
--- a/test/Scope/Positive.hs
+++ b/test/Scope/Positive.hs
@@ -226,5 +226,9 @@ tests =
     PosTest
       "Pragmas"
       $(mkRelDir ".")
-      $(mkRelFile "Pragmas.juvix")
+      $(mkRelFile "Pragmas.juvix"),
+    PosTest
+      "Import as open"
+      $(mkRelDir "ImportAsOpen")
+      $(mkRelFile "Main.juvix")
   ]
--- a/examples/milestone/TicTacToe/CLI/TicTacToe.juvix
+++ b/examples/milestone/TicTacToe/CLI/TicTacToe.juvix
@@ -5,9 +5,9 @@
 --- The module Logic.Game contains the game logic.
 module CLI.TicTacToe;
 
-open import Stdlib.Data.Nat.Ord;
-open import Stdlib.Prelude;
-open import Logic.Game;
+import Stdlib.Data.Nat.Ord open;
+import Stdlib.Prelude open;
+import Logic.Game open;
 
 --- A ;String; that prompts the user for their input
 prompt : GameState → String;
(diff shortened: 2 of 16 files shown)


paulcadman:
suggestion: raise an issue in https://github.com/anoma/juvix-docs so we remember to update the language reference

test/Scope/Positive.hs
```diff
@@ -226,5 +226,9 @@ tests =
     PosTest
       "Pragmas"
       $(mkRelDir ".")
-      $(mkRelFile "Pragmas.juvix")
+      $(mkRelFile "Pragmas.juvix"),
+    PosTest
```
paulcadman, line 230: suggestion: Add a test that covers `using` and `hiding` syntax (unless we already have one).

paulcadman (dismissed):

test/Scope/Positive.hs
janmasrovira: > suggestion: raise an issue in https://github.com/anoma/juvix-docs so we remember to update the language reference

done (https://github.com/anoma/juvix-docs/issues/15)

paulcadman (approved):

paulcadman (approved):

Issue #2104, opened 2023-05-18
Add flag to set C compiler optimization level and default to `-O0`


paulcadman:
The `juvix compile -t native` command transforms a Juvix program to C and then compiles it using `clang`. We set the clang optimization level to `-O3`:
 https://github.com/anoma/juvix/blob/a164083b70476d284eb65abae4fa85ad6a14d0c3/app/Commands/Extra/Compile.hs#L175

The C compilation step takes the majority of the time when compiling the [taiga simulator](https://github.com/anoma/taiga-simulator) project with the native code target.

The following shows the elapsed time for some different scenarios when run from the root of the taiga-simulator. I ran the commands several times to guard against outlier results.

```
# Juvix pipeline + generate C code:
time juvix compile -C Test/AppsTest.juvix
1.54s
# Unmodified Juvix, with -O3 clang option:
time juvix compile Test/AppsTest.juvix
7.99s
# Modified Juvix, with -O0 clang option:
time juvix compile Test/AppsTest.juvix
2.02s
# Modified Juvix, with -O1 clang option:
time juvix compile Test/AppsTest.juvix
3.78s
# Modified Juvix, with -O2 clang option:
time juvix compile Test/AppsTest.juvix
4.20s
``` 

From this we see that using the `-O3` clang option has a high cost.

We should consider making the default optimization option `-O0` and making this flag configurable from `juvix compile` so that users can make the choice if they want to trade off compile time for clang optimisations.

PR #2106, opened 2023-05-19
Pass through compile optimization flag to C compiler and disable optimization for --debug

paulcadman:
In this PR we pass through the `juvix compile` optimization flag to the C compiler in the native compilation.

NB: Clang supports -On for any positive n. -O4 and higher is equivalent to -O3

Also we disable optimizations when the `-g` / `--debug` option is specified.

* Closes https://github.com/anoma/juvix/issues/2104

### DIFF
--- a/app/Commands/Dev/Core/Compile/Base.hs
+++ b/app/Commands/Dev/Core/Compile/Base.hs
@@ -25,10 +25,7 @@ getEntry PipelineArg {..} = do
     ep
       { _entryPointTarget = getTarget (_pipelineArgOptions ^. compileTarget),
         _entryPointDebug = _pipelineArgOptions ^. compileDebug,
-        _entryPointOptimizationLevel =
-          if
-              | _pipelineArgOptions ^. compileDebug -> 0
-              | otherwise -> _pipelineArgOptions ^. compileOptimizationLevel,
+        _entryPointOptimizationLevel = fromMaybe defaultOptLevel (_pipelineArgOptions ^. compileOptimizationLevel),
         _entryPointInliningDepth = _pipelineArgOptions ^. compileInliningDepth
       }
   where
@@ -40,6 +37,11 @@ getEntry PipelineArg {..} = do
       TargetVampIR -> Backend.TargetVampIR
       TargetCore -> Backend.TargetCore
       TargetAsm -> Backend.TargetAsm
+
+    defaultOptLevel :: Int
+    defaultOptLevel
+      | _pipelineArgOptions ^. compileDebug = 0
+      | otherwise = defaultOptimizationLevel
 
 runCPipeline ::
   forall r.
--- a/app/Commands/Extra/Compile.hs
+++ b/app/Commands/Extra/Compile.hs
@@ -176,7 +176,7 @@ native64Args buildDir o outfile inputFile =
     <> [ "-DARCH_NATIVE64",
          "-DAPI_LIBC",
          "-m64",
-         "-O" <> show optValue,
+         "-O" <> show (maybe defaultOptLevel (max 1) (o ^. compileOptimizationLevel)),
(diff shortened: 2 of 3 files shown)


lukaszcz:
I'm thinking maybe we should always add the `-finline-functions` or `-finline-hint-functions` option to `clang` regardless of the overall level? Would it slow things down noticeably (I don't think it should)? The point is that if `clang` doesn't inline the functions marked for inlining in the runtime, this results in a massive overhead, because they are used to implement single JuvixAsm instructions (or even only parts of them). And I can't find a definitive answer in the documentation if `-O1` enables inlining in general.

paulcadman:
Setting `"-finline-functions"` doesn't change the compilation time for the taiga-simulator, even at `-O0`. So there's no issue with adding this.

janmasrovira (approved):

lukaszcz:
I can't get inlining to work with -O0. Maybe we should use at least -O1 for clang?

lukaszcz:
The flag `-finline-functions` seems to have no visible effect. Things are inline at -O1 already, but -O0 seems to prevent it regardless of other flags passed.

app/Commands/Extra/Compile.hs
```diff
@@ -172,14 +172,20 @@ native64Args buildDir o outfile inputFile =
...
        ]
     <> ( if
              | not (o ^. compilePreprocess || o ^. compileAssembly) ->
                  ["-ljuvix"]
              | otherwise -> []
        )
+  where
+    optValue :: Int
+    optValue
+      | o ^. compileDebug = 0
```
lukaszcz, line 187: This doesn't disable optimization in general, only clang optimizations

lukaszcz:
> I can't get inlining to work with -O0. Maybe we should use at least -O1 for clang?

I pushed a change that does that. I think we should use -O1 with clang. The difference wasn't that big? If someone wants fast compilation with (very) slow execution they can just use `juvix eval`.

paulcadman:
> I pushed a change that does that. I think we should use at least -O1 with clang. The difference between -O1 and -O0 in compilation time wasn't that big? If someone wants fast compilation with (very) slow execution they can just use `juvix eval`.

The taiga simulator compiles in ~2s with clang -O0 , and ~4s with clang -O1

app/Commands/Dev/Core/Compile/Base.hs
```diff
@@ -24,7 +24,10 @@ getEntry PipelineArg {..} = do
     ep
       { _entryPointTarget = getTarget (_pipelineArgOptions ^. compileTarget),
         _entryPointDebug = _pipelineArgOptions ^. compileDebug,
-        _entryPointOptimizationLevel = _pipelineArgOptions ^. compileOptimizationLevel,
+        _entryPointOptimizationLevel =
+          if
+              | _pipelineArgOptions ^. compileDebug -> 0
```
lukaszcz, line 29: This has the disadvantage that one cannot override the optimization level with `-On` when compiling in debug mode. The `compileOptimizationLevel` should probably be a `Maybe`, defaulting to 0 or 1 if it's `Nothing` depending on whether `--debug` is enabled.

app/Commands/Extra/Compile.hs
```diff
@@ -172,14 +172,19 @@ native64Args buildDir o outfile inputFile =
...
        ]
     <> ( if
              | not (o ^. compilePreprocess || o ^. compileAssembly) ->
                  ["-ljuvix"]
              | otherwise -> []
        )
+  where
+    optValue :: Int
+    optValue
+      | o ^. compileDebug = 1
```
lukaszcz, line 186: If somebody has overridden the optimization level by explicitly passing `-On` then shouldn't it be taken into account?

PR #2108, opened 2023-05-19
Scope check symbols inside `using {..}` and `hiding {..}`

### LINKED ISSUE #2089: Enhance Def. Location Output to Also Provide Location for Symbols in `using`/`hiding` List
Currently, our compiler provides location information for function definitions and other symbols, facilitating easy navigation to their definitions using Emacs mode or VSCode.
However, it lacks this feature for symbols enclosed in the `using`/`hiding` list.
This behaviour is unexpected.

image

janmasrovira:
- Closes #2089 

Now the symbols inside `using {..}` and `hiding {..}` are properly scoped, which means that they will be properly colored and will have goto information. If the referenced module does not contain a symbol in the list, an error will be thrown.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Language.hs
+++ b/src/Juvix/Compiler/Concrete/Language.hs
@@ -488,7 +488,7 @@ deriving stock instance
 
 data UsingHiding (s :: Stage)
   = Using (NonEmpty (UsingItem s))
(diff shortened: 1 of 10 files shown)


jonaprieto:
Tested locally. +1

jonaprieto (approved):

PR #2112, opened 2023-05-22
Add Bottom node

### LINKED ISSUE #2056: Unexpected error internal to core error with type aliases: undeclared identifier
Related to:
- #2055 
- https://github.com/anoma/juvix/pull/2040
- #2005

```agda
module E10;

type T0 (A : Type) :=
  | t : (A -> T0 A) -> T0 A;

axiom alias : Type -> Type;

type T1 :=
  | c : alias T1 -> T1;
```

I'm getting this error:

```
juvix: internal to core: undeclared identifier: alias
  error, called at src/Juvix/Prelude/Base.hs:342:9 in juvix-0.3.2-LpHIe68kEcQ6DZIdmT4nNv:Juvix.Prelude.Base
  error, called at src/Juvix/Compiler/Core/Translation/FromInternal.hs:799:20 in juvix-0.3.2-LpHIe68kEcQ6DZIdmT4nNv:Juvix.Compiler.Core.Translation.FromInternal 
```

janmasrovira:
- Closes #2056 
- Depends on #2103 

I am not sure about the implementation of `isType` for `NBot`. (solved).

The `Eq` instance returns `True` for every two `Bottom` terms, regardless of their type.


src/Juvix/Compiler/Core/Transformation/Check/Base.hs
```diff
@@ -1,22 +1,36 @@
...
       _coreErrorLoc = fromMaybe defaultLoc loc
     }
 
+axiomError :: Members '[Error CoreError, InfoTableBuilder] r => Symbol -> Maybe Location -> Sem r a
+axiomError sym loc = do
+  tbl <- getInfoTable
+  let nameTxt = identName tbl sym
+  throw
+    CoreError
+      { _coreErrorMsg = ppOutput ("The symbol" <+> annotate (AnnKind KNameAxiom) (pretty nameTxt) <> " is defined as an axiom and thus it cannot be compiled"),
```
jonaprieto, line 29: Isn't some function for `annotate (AnnKind KNameAxiom)..`? If not, let us define them,
e.g. `annAxiom` and similarly for other kinds.

src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -58,6 +63,27 @@ filterOutTypeSynonyms tab = pruneInfoTable tab'
...
     idents' = HashMap.filter (\ii -> not (isTypeConstr tab (ii ^. identifierType))) (tab ^. infoIdentifiers)
 
+isType :: Node -> Bool
+isType = \case
+  NPi {} -> True
+  NUniv {} -> True
+  NPrim {} -> True
+  NTyp {} -> True
+  NDyn {} -> True
+  NBot (Bottom i) -> isType (Info.getInfoType i)
```
lukaszcz, line 73: Change this to `False`. This is not entirely correct, but there really isn't a reliable way to check if the type of Bottom is a type constructor (and that's not even what the code here tries to check). There is no type info available, unless you always store it when creating Bottom nodes (otherwise it cannot be computed in general, because where would you get it from?).

src/Juvix/Compiler/Core/Extra/Base.hs
```diff
@@ -192,6 +192,12 @@ mkDynamic i = NDyn (Dynamic i)
 mkDynamic' :: Type
 mkDynamic' = mkDynamic Info.empty
 
+mkBottom :: Info -> Type
+mkBottom i = NBot (Bottom i)
+
+mkBottom' :: Type
+mkBottom' = NBot (Bottom mempty)
```
jonaprieto, line 199:
```suggestion
mkBottom' :: Type
mkBottom' = mkBottom mempty
```
However, I think you should remove this.

src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -38,6 +40,9 @@ substEnvInBranch env br = over caseBranchBody (substEnv env') br
 isClosed :: Node -> Bool
 isClosed = not . has freeVars
 
+mkAxiom :: Interval -> Type -> Node
+mkAxiom loc ty = mkBottom (Info.setInfoLocation loc (Info.setInfoType ty mempty))
```
lukaszcz, line 44: I think the general question is whether we want those bottoms to always have types or not. If they are intrinsically typed (they can always have type Dynamic), then the type should be a separate field, not in the info, much like it is a separate field in `Binder`.

jonaprieto, line 44: We should probably consider defining functions to shorten this kind of pattern. Perhaps, `setInfoByLocType` or something shorter.

jonaprieto (changes requested):

src/Juvix/Compiler/Core/Extra/Utils.hs
lukaszcz: In any case, storing Nodes in Infos can only be temporary, because the recursors don't descend into stuff stored in Info (by design). The type infos are used basically only by ComputeTypeInfo, but you need to run it just before using it, without some intervening transformations that could change the types (especially binders in them).

src/Juvix/Compiler/Core/Extra/Utils.hs
janmasrovira: done in #e8e41f41c

src/Juvix/Compiler/Core/Extra/Utils.hs
janmasrovira: I'll add a field

src/Juvix/Compiler/Core/Transformation/Check/Base.hs
janmasrovira: I don't think we need an extra function for each Kind annotation. But, if we really want them, we should refactor that in a separate PR, as this would entail a lot of changes not related to this

src/Juvix/Compiler/Core/Extra/Base.hs
janmasrovira: I've applied your suggestion.

why do you think I should remove it?
we are following the same pattern for every other `Node`.

src/Juvix/Compiler/Core/Extra/Utils.hs
janmasrovira: done

janmasrovira: I've added a `_bottomType` field to `Bottom`, so now I don't set the type info

src/Juvix/Compiler/Core/Transformation/Check/Base.hs
jonaprieto: Sure, I didn't intend to include those changes here.

src/Juvix/Compiler/Core/Extra/Base.hs
jonaprieto: Reason: the last time I searched for references to `mkBottom` there were none.

lukaszcz (approved):

src/Juvix/Compiler/Core/Extra/Base.hs
lukaszcz: It might make sense to keep it though, given that it's the "universal" interface for nodes. One would naturally expect the primed version to be available, which provides the default empty info.

I'm not sure if all other primed versions are used, but they're there in case someone needs them.

jonaprieto (approved):

PR #2118, opened 2023-05-23
Add `main` field to juvix.yaml

### LINKED ISSUE #2067: Enhance `compile` subcommand functionality for projects
Let us enhance the compile subcommand to work seamlessly with the main field in `juvix.yaml` as requested here,

- #2065

but exhibiting a behavior similar to that described for the `typecheck` subcommand.

- #2066

Expected behavior:

- If no arguments are given, it should compile the entry file specified in the `main` field in
  `juvix.yaml` (fail if there is no `main` field in `juvix.yaml`  with a *useful* error message like
  "No entry file specified in `juvix.yaml`" and if `juvix.yaml` does not exist, saying "No
  `juvix.yaml` found in current directory, maybe you want to run `juvix init`?")

```
$ juvix compile
```

- If a path is provided, it should compile the project at that location (similarly to above)

```
$ juvix compile ./my-project
```

Note that one might tend to use `.` as the path to compile the current directory.

- If a filepath is given, it should compile the specified file. (Nothing should be done here, 
as this is already implemented.)

```
$ juvix compile ./my-project/src/Main.juvix
```

janmasrovira:
- Closes #2067

This pr adds the field `main` to `juvix.yaml`. This field is optional and should contain a path to a juvix file that is meant to be used for the `compile` (and `dev compile`) command when no file is given as an argument in the CLI. This makes it possible to simply run `juvix compile` if the `main` is specified in the `jvuix.yaml`.

I have updated the `juvix.yaml` of the milestone examples.

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -80,16 +80,26 @@ runAppIO args@RunAppIOArgs {..} =
     ExitJuvixError e -> do
       printErr e
       embed exitFailure
-    ExitMsg exitCode t -> embed (putStrLn t >> hFlush stdout >> exitWith exitCode)
+    ExitMsg exitCode t -> exitMsg' exitCode t
     SayRaw b -> embed (ByteString.putStr b)
   where
+    exitMsg' :: ExitCode -> Text -> Sem r x
+    exitMsg' exitCode t = embed (putStrLn t >> hFlush stdout >> exitWith exitCode)
     getMainFile' :: Maybe (AppPath File) -> Sem r (Path Abs File)
     getMainFile' m = case m of
       Just p -> embed (prepathToAbsFile invDir (p ^. pathPath))
       Nothing -> case pkg ^. packageMain of
         -- TODO do something special if it is the global package?
         Just p -> embed (prepathToAbsFile invDir p)
-        Nothing -> error ("A path to the main file must be given in the CLI or specified in `main` field of the " <> pack (toFilePath juvixYamlFile) <> " file")
+        Nothing -> missingMainErr
+    missingMainErr :: Sem r x
+    missingMainErr =
+      exitMsg'
+        (ExitFailure 1)
+        ( "A path to the main file must be given in the CLI or specified in the `main` field of the "
+            <> pack (toFilePath juvixYamlFile)
+            <> " file"
+        )
     invDir = _runAppIOArgsRoots ^. rootsInvokeDir
     pkg :: Package
     pkg = _runAppIOArgsRoots ^. rootsPackage
--- a/app/App.hs
+++ b/app/App.hs
@@ -10,6 +10,7 @@ import Juvix.Prelude.Pretty hiding
   ( Doc,
   )
 import System.Console.ANSI qualified as Ansi
+import Juvix.Extra.Paths.Base
 
 data App m a where
   ExitMsg :: ExitCode -> Text -> App m a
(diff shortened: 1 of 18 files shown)


app/App.hs
```diff
@@ -77,10 +80,29 @@ runAppIO args@RunAppIOArgs {..} =
...
       printErr e
       embed exitFailure
-    ExitMsg exitCode t -> embed (putStrLn t >> hFlush stdout >> exitWith exitCode)
+    ExitMsg exitCode t -> exitMsg' exitCode t
     SayRaw b -> embed (ByteString.putStr b)
   where
+    exitMsg' :: ExitCode -> Text -> Sem r x
+    exitMsg' exitCode t = embed (putStrLn t >> hFlush stdout >> exitWith exitCode)
+    getMainFile' :: Maybe (AppPath File) -> Sem r (Path Abs File)
+    getMainFile' m = case m of
```
jonaprieto, line 89:
```suggestion
    getMainFile' = \case
```

app/App.hs
```diff
@@ -77,10 +80,29 @@ runAppIO args@RunAppIOArgs {..} =
...
+      Just p -> embed (prepathToAbsFile invDir (p ^. pathPath))
+      Nothing -> case pkg ^. packageMain of
+        -- TODO do something special if it is the global package?
```
jonaprieto, line 92: What do you mean by "special"?

app/App.hs
```diff
@@ -77,10 +80,29 @@ runAppIO args@RunAppIOArgs {..} =
...
+        Just p -> embed (prepathToAbsFile invDir p)
+        Nothing -> missingMainErr
+    missingMainErr :: Sem r x
+    missingMainErr =
+      exitMsg'
+        (ExitFailure 1)
+        ( "A path to the main file must be given in the CLI or specified in the `main` field of the "
```
jonaprieto, line 99:
```suggestion
        ( "File path to the main file is not provided. Ensure to specify it in the `main` field of the "
```

tests/smoke/Commands/compile.smoke.yaml
```diff
@@ -11,6 +11,18 @@ tests:
...
+      shell:
+        - bash
+      script: |
+        cd ./examples/milestone/HelloWorld
+        juvix compile
+        ./HelloWorld
+    exit-status: 0
+    stdout: |
+      hello world!
+
```
jonaprieto, line 25: Can you please add a test to see the error in the absence fo the main field in the juvix. yaml?

jonaprieto (changes requested):

app/App.hs
janmasrovira: I was thinking that maybe we want to explicitly forbid the `main` file in the global project.
I don't think there is ever a usecase for running `juvix compile` outside of a project.
By default, the global project doesn't have a `main` file by default, so maybe we leave it as it is and remove the `TODO` comment

app/App.hs
janmasrovira: done

tests/smoke/Commands/compile.smoke.yaml
janmasrovira: done

app/App.hs
janmasrovira: I can't see that this improves the error message. Moreover, it is incorrect, because it does not mention the CLI

PR #2119, opened 2023-05-23
Add `:def` command to the repl

janmasrovira:
This pr adds a new command, `:def` to the repl. This command expects a single identifier that must be in scope and then prints its definition. For constructors, the whole type definition is printed.

It also applies some refactors to the code for repl command.
1. Before there was a mega `where` block of definitions. I have hoisted most of the definitions there to the top level. I feel like now it is easier to navigate and read.
2. Use `ExceptT` instead of local `case` expressions for errors.
3. Use forks of `haskeline` and `repline`. These forks are necessary because these libraries do not export the constructors `HaskelineT` and `InputT` respectively, thus, making it impossible to catch errors in their underlying monad.

### DIFF
--- a/tests/smoke/Commands/repl.smoke.yaml
+++ b/tests/smoke/Commands/repl.smoke.yaml
@@ -1,6 +1,26 @@
 working-directory: ./../../../tests/
 
 tests:
+  - name: repl-def
+    command:
+      - juvix
+      - repl
+      - ../examples/milestone/HelloWorld/HelloWorld.juvix
+    stdin: ":def Nat"
+    stdout:
+      contains: "zero"
+    exit-status: 0
+
+  - name: repl-def-error
+    command:
+      - juvix
+      - repl
+      - ../examples/milestone/HelloWorld/HelloWorld.juvix
+    stdin: ":def 1"
+    stdout:
+      contains: ":def expects a single identifier, but got:"
+    exit-status: 0
+
   - name: open
     command:
       - juvix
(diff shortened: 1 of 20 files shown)


tests/smoke/Commands/repl.smoke.yaml
```diff
@@ -1,6 +1,26 @@
...
 tests:
+  - name: repl-def
+    command:
+      - juvix
+      - repl
+      - ../examples/milestone/HelloWorld/HelloWorld.juvix
+    stdin: ":def Nat"
+    stdout:
+      contains: "zero"
+    exit-status: 0
```
jonaprieto, line 12: If I run this one, I get:

```
HelloWorld> :def Nat
builtin nat type Nat =
  zero
  | suc Nat
```
I'd expect the actual def. in concrete syntax. 

```
HelloWorld> :def Nat
builtin nat
type Nat :=
  | zero : Nat
  | suc : Nat → Nat;
```

jonaprieto: In this case, perhaps, it's better to suggest using parenthesis. 

```
HelloWorld> :def +
/repl:1:1-2: error:
Error solving infixities
  +
```

janmasrovira: > In this case, perhaps, it's better to suggest using parenthesis.
> 
> ```
> HelloWorld> :def +
> /repl:1:1-2: error:
> Error solving infixities
>   +
> ```

Currently, I am parsing expressions, but I should probably parse identifiers. That way, the parentheses won't be needed

janmasrovira: both problems have now been addressed

PR #2120, opened 2023-05-23
Add `main` field to `juvix.yaml`

### LINKED ISSUE #2067: Enhance `compile` subcommand functionality for projects

janmasrovira:
- Closes #2067

This pr adds the field `main` to `juvix.yaml`. This field is optional and should contain a path to a juvix file that is meant to be used for the `compile` (and `dev compile`) command when no file is given as an argument in the CLI. This makes it possible to simply run `juvix compile` if the `main` is specified in the `jvuix.yaml`.

I have updated the `juvix.yaml` of the milestone examples.

### DIFF
--- a/examples/milestone/TicTacToe/juvix.yaml
+++ b/examples/milestone/TicTacToe/juvix.yaml
@@ -1,2 +1,3 @@
 name: TicTacToe
+main: TicTacToe.juvix
 version: 0.1.0
--- a/examples/milestone/TicTacToe/juvix.yaml
+++ b/examples/milestone/TicTacToe/juvix.yaml
@@ -1,3 +1,3 @@
 name: TicTacToe
-main: TicTacToe.juvix
+main: CLI/TicTacToe.juvix
 version: 0.1.0
--- a/tests/smoke/Commands/compile.smoke.yaml
+++ b/tests/smoke/Commands/compile.smoke.yaml
@@ -23,6 +23,22 @@ tests:
     stdout: |
       hello world!
 
+  - name: hello-world-no-arg-error
+    command:
+      shell:
+        - bash
+      script: |
+        temp=$(mktemp -d)
+        trap 'rm -rf -- "$temp"' EXIT
+        cd ./examples/milestone/
+        cp -r HelloWorld "$temp"
+        cd "$temp/HelloWorld"
+        sed -i '/^main:/d' juvix.yaml
+        juvix compile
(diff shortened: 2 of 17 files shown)


examples/milestone/TicTacToe/juvix.yaml
```diff
@@ -1,2 +1,3 @@
 name: TicTacToe
+main: TicTacToe.juvix
```
paulcadman, line 2: The main function is in `CLI/TicTacToe.juvix`, because we used to have CLI and Web versions. I assume paths of this type will work in this field?

```suggestion
main: CLI/TicTacToe.juvix
```

tests/smoke/Commands/compile.smoke.yaml
```diff
@@ -11,6 +11,40 @@ tests:
...
+    command:
+      shell:
+        - bash
+      script: |
+        temp=$(mktemp -d)
+        trap 'rm -rf -- "$temp"' EXIT
+        cd ./examples/milestone/
+        cp -r HelloWorld "$temp"
+        cd "$temp/HelloWorld"
+        if command -v gsed >/dev/null 2>&1; then
```
paulcadman, line 36: Now that Jonathan has done https://github.com/anoma/juvix/pull/2123 we should remove the gsed workaround.

janmasrovira: I've reverted this change

jonaprieto (approved):
Looks good to me.

PR #2121, opened 2023-05-24
Remove --no-format option

### LINKED ISSUE #2084: Remove `--no-format` global flag
The flag was meant to disable the new pretty printing algorithm.
But we should consider always using it and removing the old one.

vrom911:
Instead, always act as `--no-format` option is set to `False` as previous default.

The change seem to not affect any current formatting, so I assume it passes the checks on testing.

Fixes #2084 

# Checklist:

- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/tests/smoke/Commands/dev/geb.smoke.yaml
+++ b/tests/smoke/Commands/dev/geb.smoke.yaml
@@ -5,23 +5,23 @@ tests:
     command:
       - juvix
       - --no-colors
-      - dev 
+      - dev
       - geb
       - repl
     stdout:
-      contains: "Welcome to the Juvix Geb REPL!"
+      contains: 'Welcome to the Juvix Geb REPL!'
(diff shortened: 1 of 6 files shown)


janmasrovira (approved):

tests/smoke/Commands/dev/geb.smoke.yaml
```diff
@@ -5,23 +5,23 @@ tests:
...
-      - dev 
+      - dev
       - geb
       - repl
     stdout:
-      contains: "Welcome to the Juvix Geb REPL!"
+      contains: 'Welcome to the Juvix Geb REPL!'
     exit-status: 0
-    
+
```
jonaprieto, line 14: We should stop ignoring in the pre-commit task any file in tests/smoke. Atm, we ignore the whole `tests` folder.

Issue #1991, opened 2023-04-07
Optional and named arguments


lukaszcz:
We could consider adding optional arguments in general. This would generalise the accumulator arguments in recursive lambdas (issue #1638). Plus, optional arguments are useful independently.

But for them to fit into our syntax we would need to require the following.
1. Forbid multi-clause definitions with optional arguments: if there are optional arguments then the definition has only one clause. For example:
```haskell
fun : Nat -> Nat -> Nat;
fun (n := 0) m := n + m;
```
Then it makes sense that for a lambda the optional arguments can be declared before the lambda, like with `go` below:
```haskell
go(acc1 := [])(acc2 := 0)@\{
     []        := (acc1, acc2)
   | (x :: xs) := go (acc1 := x :: acc1) (acc2 := acc2 + 1) xs
  };
```
Then inside the lambda we can have multiple pattern matching clauses. This generalises the accumulator arguments to recursive lambdas (issue #1638).
2. The optional arguments can only be provided in an application by explicitly naming them, to avoid ambiguity with partial application, e.g., `go (acc1 := x :: acc1) (acc2 := acc2 + 1) xs`.
3. Perhaps we should somehow indicate the optionality of arguments in the type? Otherwise, the types are confusing. For example:
```haskell
fun : ?(n : Nat) -> Nat -> Nat;
fun (n := 0) m := n + m;
```

janmasrovira:
Would it be correct to think about optional arguments in the following way?
- An optional argument behaves as an implicit argument except that when it is omitted, the default value is inserted instead of a hole.

If so, we could use the following syntax:
```
fun : {n := 0 : Nat} -> Nat -> Nat;
```
This syntax has the advantages:
1. The braces `{}` already hint at the optionality, since we also use them for implicit arguments. So we don't need special `?` syntax.
2. Since the default value is given in the type signature, the function is allowed to have multiple clauses. So we drop restriction 1.

About restriction 2. Is the behavour that you have in mind as follows?
Given a function with named arguments `f : {a1 := 1 : Nat} {a2 : Type} {a3 := 4 : Nat} -> Nat -> Nat`, in the call site, if we write 
```
f (a3 := 10)
```
It will be desugared into
```
f {1} {_} {a3}
```
That is, when we pass an explicit named argument, we assume that arguments that come before it, have been omitted.

Another example, If we have
```
f 7
```
Then, it desugars to
```
f {1} {_} {4}
```

lukaszcz:
[...]

Issue #1642, opened 2022-11-22
Add syntax support for record updates


jonaprieto:
```haskell
type A (B : Type) :=
  constructorName {
    cfield1 : X1;
    cfield2 : X2;
 };
```
An update for `A` is written as follows:

```haskell
f : A B;
f := constructorName { cfield1 := x; cfield2 := y};

g : A B -> A B;
g r := r { cfield1 := x};
```

lukaszcz:
The above record update syntax doesn't play well with the proposed syntax for named arguments:
* #1991.
Named arguments automatically provide record creation syntax:
```
constructorName (cfield1 := x; cfield2 := y)
```
is just
```
constructorName x y
```
But then
```
r {cfield1 := x}
```
would desugar to
```
r {x}
```
and this conflicts with the record update syntax.

I propose to introduce a keyword `with` for record updates:
```
r with {cfield1 := x}
```

lukaszcz:
I think it would be convenient, especially in combination with traits, to also allow `with` for record creation, and allow full top-level declaration syntax including optional signatures inside the braces on the right. Then given a trait:
```haskell
trait
type Encoding (A : Type) :=
  MkEncoding : 
     (encode : A -> Field) -> 
     (decode : Field -> A) ->
     Encoding A;
```
one could write:
```haskell
type E (A : Type) := 
| C1 : A -> E 
| C2 : E;

instance 
encoding : {A : Type} -> {{Encoding A}} -> Encoding (E A);
encoding := MkEncoding with {
  encode : E A -> Field;
  encode (C1 x) := encode x + 1;
  encode C2 := 0;

  decode : Field -> E A;
  decode zero := C2; 
  decode (suc x) := C1 (decode x);
};
```

janmasrovira:
I think adding the `with` keyword to remove ambiguity is reasonable. 
I also like the proposal for inline function definitions in record creation/update.

Issue #1641, opened 2022-11-22
Add record declaration syntax support 


jonaprieto:
Example:

```haskell
type A (B : Type) :=
  constructor1 {
    cfield1 : X1;
    cfield2 : X2;
  }
 | constructor2 {
    cfield3 : Y1;
 };
```

lukaszcz:
After implementing
* #1991 

we might not need or want a separate syntax for record declarations. With the proposed syntax for named arguments a record can be defined as:
```haskell
type R :=
  Rctr : (cfield1 : X1) -> (cfield2 : X2) -> R
```
If there is only one constructor, accessor functions can be automatically generated, maybe with an additional specifier like
```haskell
record type R := Rctr : ...
```

lukaszcz:
But perhaps the named argument syntax becomes inconvenient with a large number of fields. So we may actually want the proposed syntactic sugar:
```haskell
record
type R := 
  Rctr { 
    cfield1 : X1;
    cfield2 : X2;
  };
```

janmasrovira:
I agree with your points.

A couple of extra comments:
1. The `record` keyword is not neccesary for parsing, we could simply disambiguate based on what's written. The problem then becomes that we don't know if we should generate field accessors or not. But I would say that it is ok to always generate them if there is a single constructor.
3. I think it'd be more consistent to use parentheses instead of braces after the constructor name. That way, we would respect the existing juvix convention to use braces for implicit arguments.

lukaszcz:
> 2. I think it'd be more consistent to use parentheses instead of braces after the constructor name. That way, we would respect the existing juvix convention to use braces for implicit arguments.

We use braces often for blocks of declarations, e.g., in lets. The fields of a constructor are a bit like such a block of signature declarations. Maybe we could use `with` here to indicate that we're starting a block of signatures (and to keep consistent with the syntax for updating/creating records)?
```haskell
trait
type Encoding (A : Type) := MkEncoding with {
  encode : A -> Field; 
  decode : Field -> A;
};
```

janmasrovira:
we are not using braces in lets. What blocks of declarations are you referring to?

janmasrovira:
I wouldn't strongly oppose to braces however. I think they have a big pro, which is that they are often used for record-like structures in other languages

lukaszcz:
No? Damn, I'm not keeping up with the syntax.

lukaszcz:
I think I got confused with Core syntax.

[...]

PR #2126, opened 2023-05-24
Iterator syntax

### LINKED ISSUE #1992: Special syntax for folds, maps, etc
Motivation
----------

Folds are unreadable and confusing, but still used very often. We ourselves have been confused a few times about what some of the folds in our codebase do. And it happened for very simple ones (like `foldl' (<+>) x args` or `foldr mkLambda' body binders`). What about less experienced functional programmers?

The fold readability issue has been noticed before. Notably, it inspired Leonardo de Moura to introduce a monadic `for ... in` syntax and "mutable" variables in a do-notation in Lean 4 (https://leanprover.github.io/functional_programming_in_lean/monad-transformers/do.html#loops). My goal is to propose a syntax for "purely" functional non-monadic folds that tackles the readability issue.

In fact, making "loops" more readable was my main motivation behind proposing #1638. Recursive lambdas make writing tail-recursive functions a bit quicker, but they are only moderately useful in terms of improving readability. Folding over a data structure is one of the most common special cases of tail-recursive functions, and most analogous to an imperative for-loop. Hence, it might deserve a special readable syntax.

Writing explicit (tail) recursive functions makes clear how the iteration proceeds but is far from perfect: it has the disadvantage of relying on the implementation details of the data structure you're iterating over. Folds abstract away the details of iteration mechanics, which is a good thing. It would be best to be able to use folds in a more readable syntax which makes it clear how the iteration goes.

The essence of folds
--------------------

The anatomy of a fold (left) is simple.
1. We have some accumulator variables `acc1`, ..., `accn` which we initialise to some values `val1`,...,`valn`.
2. We go through a data structure (list) in some specified order (left-to-right).
3. At each step, we receive the current values of `acc1`,...,`accn` and the current element `x`. From those we need to produce the new values of `acc1`,...,`accn`.
4. After going through all elements, the result of the fold is specified in terms of the final values of `acc1`,...,`accn` (typically the result is the value of one of the accumulators, or several accumulators gathered into a tuple).

Simple for-notation
-------------------

First, the simple case of only one accumulator. This is the most common case which deserves a separate treatment because it allows to simplify the syntax.

The special syntax in the case of one accumulator would be:
```haskell
for (acc := a) (x in d)
  Expr[acc,x]
```
where `Expr[acc,x]` is an expression which can refer to the variables `acc` and `x` bound by the `for`. This would desugar to:
```haskell
foldl \{acc x := Expr[acc,x]} a d
```
where `foldl` comes from the `Foldable` type class. It should be possible to use the `for`-notation with any `d` for which an instance of `Foldable` is available.

The notation shows clearly what the accumulator and its initial value are (`acc := a`), what we're folding over (`x in lst`), and how to produce the next accumulator value (`Expr[acc,x]`).

Examples:

1. Summing a list.
  ```haskell
  for (s := 0) (x in lst)
    s + x
  ```
  desugars to
  ```haskell
  foldl \{s x := s + x} 0 lst
  ```
2. Reversing a list.
  ```haskell
  for (acc := []) (x in lst)
    x :: acc
  ```
  desugars to
  ```haskell
  foldl \{acc x := x :: acc} [] lst
  ```
3. Filter and map.
  ```haskell
  for (acc := []) (x in reverse lst)
    if (p x) (f x :: acc) acc
  ```
  desugars to
  ```haskell
  foldl \{acc x := if (p x) (f x :: acc) acc} [] (reverse lst)
  ```

General for-notation
--------------------

With multiple accumulators one needs to explicitly name the "looping" function and specify what the result of the folding should be in terms of the final values of the accumulators.

The general scheme is:
```
for loop (acc1 := a1) .. (accn := an) (x in d)
  Expr[loop,acc1,..,accn,x]
in
Ret[acc1,..,accn]
```
and it desugars to:
```
let loop acc1 .. accn :=
  foldStep
    \{(acc1, .., accn) x := Expr[loop,acc1,..,accn,x]}
    (acc1, .., accn)
in
case loop a1 .. an
| (acc1, .., accn) := Ret[acc1, .., accn]
```
The `loop` function takes as its first `n` arguments the new accumulator values. It is called in `Expr` to invoke the next step of the iterator.

For lists `foldStep` is defined by:
```agda
foldStep : {A B : Type} -> (B -> A -> List A -> B) -> B -> List A -> B;
foldStep f acc nil := acc;
foldStep f acc (x :: xs) := f acc x xs;
```
Ultimately, `foldStep` should be provided in the `Foldable` type class. The advantage of using `foldStep` instead of `foldl` is that `loop` can then be an ordinary function which captures the continuation of the iteration.

`Expr` evaluates to a function which takes the next substructure to
iterate over and continues the iteration (for lists, the argument is
the tail of the list considered in the current iteration step).

Examples:

1. Complex loop with two accumulators.
  ```haskell
  for loop (acc := 0) (bcc := []) (x in lst)
    let
      a := f x;
      b := g x;
    in
    if (p x)
      (loop (acc + a) (b :: bcc))
      (loop (acc + b) (a :: bcc))
  in
  acc :: bcc
  ```
  desugars to:
  ```haskell
  let loop acc bcc :=
    foldStep
      λ{(acc, bcc) x :=
        let
          a := f x;
          b := g x;
        in
        if (p x)
          (loop (acc + a) (b :: bcc))
          (loop (acc + b) (a :: bcc))}
      (acc, bcc)
  in
  case loop 0 [] lst
  | (acc, bcc) := acc :: bcc
  ```
2. Compute the sum of numbers in a list, each multiplied by their index in the list.
  ```haskell
  for loop (s := 0) (idx := 0) (x in lst)
    loop (s + idx * x) (idx + 1)
  in
  s
  ```
  desugars to:
  ```haskell
  let loop s idx :=
    foldStep
      \{(s, idx) := loop (s + idx * x) (idx + 1)}
      (s, idx)
  in
  case loop 0 0 lst
  | (s, idx) := s
  ```
3. Compute a list elements satisfying `p` and a list of elements not satisfying `p`.
  ```haskell
  for loop (yes := []) (no := []) (x in lst)
    if
      (p x)
      (loop (x :: yes) no)
      (loop yes (x :: no))
  in
  (yes, no)
  ```
  desugars to:
  ```haskell
  let loop yes no :=
    foldStep
      \{(yes, no) :=
          if
            (p x)
            (loop (x :: yes) no)
            (loop yes (x :: no))}
      (yes, no)
  in
  case loop [] [] lst
  | (yes, no) := (yes, no)
  ```
4. Compute the list of suffixes of `lst`, in the order of decreasing length.
  ```haskell
  for loop (acc := []) (x in lst)
    \{tl := loop (tl :: acc) tl}
  in
  lst :: reverse acc
  ```
  desugars to
  ```haskell
  let loop :=
    foldStep
      \{acc x := \{tl := loop (tl :: acc) tl}}
  in
  case loop [] lst
  | acc := lst :: reverse acc
  ```
  For a fold on trees, this would compute a list of subtrees.
5. Map on lists without reversal.
  ```haskell
  for loop (acc := []) (x in lst)
    \{tl := f x : loop [] tl}
  ```
  desugars to
  ```haskell
  let loop :=
    foldStep
      \{acc x := \{tl := f x : loop [] tl}}
  in
  loop [] lst
  ```
  For a fold on trees, this would compute a list of mapped tree elements.

Extensions
----------

One may wish to include `zip` in the notation, e.g., by allowing to pattern match on the loop variable when its inductive type has only one constructor:
```haskell
for (acc := 0) (x, y in zip lst1 lst2)
  acc + x * y
```
This is equivalent to but more concise and readable than:
```haskell
for (acc := 0) (p in zip lst1 lst2)
  case p | (x, y) := acc + x * y
```

Nested loops:
```haskell
for (acc := []) (x in lst1) (y in lst2)
  (x, y) :: acc
```
is equivalent to:
```haskell
for (acc := []) (x in lst1)
  for (acc' := acc) (y in lst2)
    (x, y) :: acc'
```

map- and filter-notations
-------------------------

For syntactic consistency, one may consider introducing the map-notation
```
map (x in d) Expr[x]
```
which would desugar to
```
fmap \{x := Expr[x]} d
```
where `fmap` comes from the `Functor` instance for `d`.

This notation has the advantage that the mapping expression (which describes how to modify `x`) comes at the end. So one does not need to look for the thing mapped over (`d`) several lines below when `Expr` is big.

An analogous filter-notation could be used:
```
filter (x in d) Expr[x]
```
for a `Filterable` type class.

Perhaps it's an overkill to have a special notation just for `filter`. At this point, the special syntax should be generalised into a user-definable language feature.

Generalisation
--------------

We may allow the user to define syntactic recursion schemes of different (predefined?) kinds. The syntax sugar we've considered so far could be classified into three broad schemes: "folding", "mapping" and "complex-folding".

Declaring, e.g.:
```haskell
syntax scheme folding myfor with fld;
```
would declare `myfor` as a folding recursion scheme which desugars to `fld`. Then one could use the syntax
```haskell
myfor (acc := a) (x in d)
  Expr[acc,x]
```
instead of
```haskell
fld \{acc x := Expr[acc,x]} a d
```

For example, to declare a recursion scheme for `foldr` one would use:
```agda
syntax scheme folding rfor with (foldr . flip);
```

Example use - map:
```haskell
rfor (acc := []) (x in lst)
  f x :: acc
```

To declare a recursion scheme for `foldM` one would use:
```agda
syntax scheme folding forM with foldM;
```

Example use:
```haskell
forM (acc := 0) (x in lst) do
  register x
  return (acc + x)
```

Similarly, declaring
```haskell
syntax scheme mapping map with fmap;
syntax scheme mapping mapM with fmapM;
syntax scheme mapping filter with ffilter;
```
would enable the syntax
```haskell
map (x in lst) Expr[x]
mapM (x in lst) Expr[x]
filter (x in lst) Expr[x]
```

Note that the desugaring of recursion schemes is a purely syntactic operation which happens before type-checking.

It is a bit more complicated to generalise the for-notation with multiple accumulators and an explicit loop. Declaring
```haskell
syntax scheme complex-folding fld with fldStep;
```
should enable using
```haskell
fld loop (acc1 := a1) .. (accn := an) (x in d)
  Expr[loop,acc1,..,accn,x]
in
Ret[acc1,..,accn]
```
instead of
```haskell
let loop acc1 .. accn :=
  fldStep
    \{(acc1, .., accn) x := Expr[loop,acc1,..,accn,x]}
    (acc1, .., accn)
in
case loop a1 .. an
| (acc1, .., accn) := Ret[acc1, .., accn]
```
But this should be generalised to work when `acc1`,..,`accn` are monadic, and to the situations when we first recurse and then evaluate the next values of the accumulators.

Digression (potentially contentious subjective opinion)
-------------------------------------------------------

In my opinion, ML is generally more readable than Haskell because people tend to express recursion and control flow directly and explicitly. Folds etc. are used less often. In Haskell, one typically tries to use higher-order functions to express iteration and control flow. Maximally concise "point-free" programming seems to be preferred in the Haskell community: the more you abstract your code by omitting explicit recursion, ifs, cases and lambdas in favour of higher-order functions, the better your coding style is considered to be (at least by `hlint`, which I was initially surprised by, but later got used to). As a result, to understand a piece of code, in addition to understanding the semantics of the language, one needs to understand the behaviour of all the different higher-order functions used to express control flow. For a trivial example, one can understand `case x of Nothing -> c; Just x' -> E[x']` just from the rules of the language, but for `maybe c (\x' -> E[x']) x` one needs to look up `maybe` in the standard library, remember which argument is which, what arguments the higher-order function receives in which order, etc. One needs to use more "brain cycles" to think "higher-order" - this might easily be underestimated by someone who never had much difficulty with this. In most programming language communitites, the first option would be considered better coding style as more straightforward. Even `\a b x -> f b (g a x)` could be considered more readable than `\a b -> f b . g a`, because you immediately see the number of function arguments. One would very rarely write the second in OCaml, but `hlint` suggests to change the first into it. The more exotic higher-order functions you use, the more problematic this becomes. I think this coding style is one of the major reasons for the steep learning curve of Haskell for more "mainstream" programmers.

A big advantage of the "abstract" style is that one can reuse different "control flow" and "iteration" mechanisms captured in higher-order functions (even more so with type classes), and separate them from the logic specific to the particular program (which also makes changing the underlying data structures easier). But I think the disadvantage of unreadable hermetic code should not be underestimated. One remedy might be to have special readable syntax for the common cases, like the do-notation or folds, and prefer to write things more explicitly in the less common cases.

Conclusion
----------

The proposed syntactic sugar may seem rather trivial, especially in the most common simple variants for folding and mapping. I nevertheless believe that such seemingly simple syntactic notations can significantly improve the readability and pragmatic usability of a programming language.

lukaszcz:
* Closes #1992 

A function identifier `fun` can be declared as an iterator with
```
syntax iterator fun;
```
For example:
```haskell
syntax iterator for;
for : {A B : Type} -> (A -> B -> A) -> A -> List B -> List A;
for f acc nil := acc;
for f acc (x :: xs) := for (f acc x) xs;
```
Iterator application syntax allows for a finite number of initializers `acc := a` followed by a finite number of ranges `x in xs`. For example:
```
for (acc := 0) (x in lst) acc + x
```
The number of initializers plus the number of ranges must be non-zero.

An iterator application
```
fun (acc1 := a1; ..; accn := an) (x1 in b1; ..; xk in bk) body
```
gets desugared to
```
fun \{acc1 .. accn x1 .. xk := body} a1 .. an b1 .. bk
```
The `acc1`, ..., `accn`, `x1`, ..., `xk` can be patterns.

The desugaring works on a purely syntactic level. Without further restrictions, it is not checked if the number of initializers/ranges matches the type of the identifier. The restrictions on the number of initializers/ranges can be specified in iterator declaration:
```
syntax iterator fun {init: n, range: k};
syntax iterator for {init: 1, range: 1};
syntax iterator map {init: 0, range: 1};
```
The attributes (`init`, `range`) in between braces are parsed as YAML to avoid inventing and parsing a new attribute language. Both attributes are optional.


janmasrovira (commented):
- See https://github.com/anoma/juvix/pull/2139 for some style changes

We should have a test that shows the syntax for multiple initializers/ranges.
I propose that the syntax for that follows this schema:
```
fun (acc1 := a1; .. ; accn := an) (x1 in b1; .. ; xk in bk) body
```

janmasrovira:
- @lukaszcz have you considered how this syntax (specially initializers) could interact with the proposed syntax in https://github.com/anoma/juvix/issues/1991?

lukaszcz:
> * @lukaszcz have you considered how this syntax (specially initializers) could interact with the proposed syntax in [Optional and named arguments #1991](https://github.com/anoma/juvix/issues/1991)?

Yes, there is a possibility of ambiguity, I realise this.

My proposal is for the iterators to take priority.

If one really wants to have a function application with named non-implicit arguments at the beginning, then one can put it in parentheses
```
(fun) (arg := a)
```
This would not be parsed as an iterator because an iterator requires an identifier at the front.

This would perhaps be inconvenient if used often, e.g., for the record syntax. But we need something like the `with` notation anyway that I proposed in the comments (for record update and/or construction), because otherwise record update is already ambiguous with named implicit arguments.

lukaszcz:
See:
* #1642 
* #1641

lukaszcz:
With a bit more effort in the parser/scoper, one could also recognise when an identifier is an iterator and prioritise iterator syntax only then. But this would require reworking the current iterator parsing concept (to have separate "initializer" and "range" atoms and build (or not) an iterator application only during scoping).

janmasrovira (approved):

PR #2144, opened 2023-05-30
Fix printing of infix constructor values

lukaszcz:
To print e.g.
```agda
(::) :: nil
```
instead of
```agda
:: :: nil
```


janmasrovira:
Let's add one test

janmasrovira (approved):

PR #2164, opened 2023-06-02
Specialization optimisation

### LINKED ISSUE #2147: Specialisation optimisation
Specialisation is necessary to efficiently compile folds/maps/etc. This becomes important with heavy iterator use, to guarantee that iterator applications are compiled to efficient loops without hidden function calls/closure creation. Since specialisation will be done on the Core level, it will affect both the WebAssembly and GEB/VampIR backends.

lukaszcz:
* Closes #2147 

Adds a `specialize` pragma which allows to specify (explicit) arguments considered for specialization. Whenever a function is applied to a constant known value for the specialized argument, a new version of the function will be created with the argument pasted in. For example, the code

```juvix
{-# specialize: [1] #-}
mymap : {A B : Type} -> (A -> B) -> List A -> List B;
mymap f nil := nil;
mymap f (x :: xs) := f x :: mymap f xs;

main : Nat;
main := length (mymap λ{x := x + 3} (1 :: 2 :: 3 :: 4 :: nil));
```

will be transformed into code equivalent to

```juvix
mymap' : (Nat -> Nat) -> List Nat -> List Nat;
mymap' f nil := nil;
mymap' f (x :: xs) := λ{x := x + 3} x :: mymap' xs;

main : Nat;
main := length (mymap' (1 :: 2 :: 3 :: 4 :: nil));
```


src/Juvix/Compiler/Core/Transformation/Optimize/SpecializeArgs.hs
```diff
@@ -0,0 +1,206 @@
...
+      NIdt {} -> True
+      NLam {} -> True
+      NCst {} -> True
+      NCtr Constr {..} -> all (isSpecializable tab) _constrArgs
+      NApp {} ->
+        let (h, _) = unfoldApps' node
+         in isSpecializable tab h
+      _ -> False
+
+-- Checks if an argument is passed without modification to recursive calls.
```
janmasrovira, line 22:
```suggestion
-- | Checks if an argument is passed without modification to recursive calls.
```

src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -262,23 +258,35 @@ shiftLetItem n l =
...
-  reverse $ snd $ foldl' (\(args', acc) ty -> (mkVar' 0 : map (shift 1) args', substs args' ty : acc)) (reverse args, []) (drop k argtys)
+-- | substitute a term t for the free variable with de Bruijn index idx, avoiding
+-- variable capture; shifts all free variabes with de Bruijn index > idx by -1 (as
+-- if the binder idx was removed)
+substVar :: Index -> Node -> Node -> Node
+substVar idx t = umapN go
   where
-    k = length args
+    go k n = case n of
+      NVar (Var {..})
```
janmasrovira, line 286:
```suggestion
      NVar Var {..}
```

src/Juvix/Compiler/Core/Transformation/Optimize/SpecializeArgs.hs
```diff
@@ -0,0 +1,206 @@
...
+                            node'' <- lambdaLiftNode' True bl node'
+                            return $ End node''
+                  Nothing ->
+                    return $ Recur node
+                where
+                  ii = lookupIdentifierInfo tab _identSymbol
+                  pspec = ii ^. identifierPragmas . pragmasSpecialiseArgs
+                  argsNum = ii ^. identifierArgsNum
+                  (tyargs, tgt) = unfoldPi' (ii ^. identifierType)
+              _ ->
```
janmasrovira, line 125: I think having these dangling `_ -> ...` is bad style, as it makes code harder to read.
Please, refactor the `go` function to improve how it is structured in order to improve readability, or mark this comment as resolved if you think that is not worth it

lukaszcz: Yes, that's true because the body of the main case is big and you don't see what these wildcards are connected to. I can just put the main case in a separate function.

src/Juvix/Compiler/Core/Transformation/Optimize/SpecializeArgs.hs
lukaszcz: Done

janmasrovira (approved):

PR #2169, opened 2023-06-02
Cache includes

janmasrovira:
Avoids translating the same Internal module to Core more than once.


janmasrovira:
A cache has been implemented in #2219

Issue #2159, opened 2023-06-01
REPL crash when evaluating a symbol that is in the scope of the loaded module, but is not exported by it


paulcadman:
To reproduce:

Load a module that includes `import Stdlib.Prelude open`. e.g from the root of the repo:

```
juvix repl examples/milestone/HelloWorld/HelloWorld.juvix
``` 

Then evaluate a symbol that is in-scope in the loaded module, but not exported by it. For example the constructor `just`:

```
HelloWorld> just
internal to core: undeclared identifier: just
CallStack (from HasCallStack):
  error, called at src/Juvix/Prelude/Base.hs:352:9 in juvix-0.3.4-Jeg8jsiIum31kDXB75qzlS:Juvix.Prelude.Base
  error, called at src/Juvix/Compiler/Core/Translation/FromInternal.hs:840:20 in juvix-0.3.4-Jeg8jsiIum31kDXB75qzlS:Juvix.Compiler.Core.Translation.FromInternal
```

I think my expectation is to get `Symbol not in scope: just` error instead.

Full stack trace:

```
CallStack (from HasCallStack):
  error, called at src/Juvix/Prelude/Base.hs:352:9 in juvix-0.3.4-inplace:Juvix.Prelude.Base
  error, called at src/Juvix/Compiler/Core/Translation/FromInternal.hs:840:20 in juvix-0.3.4-inplace:Juvix.Compiler.Core.Translation.FromInternal
CallStack (from -prof):
  Juvix.Prelude.Base.error (src/Juvix/Prelude/Base.hs:352:1-26)
  Juvix.Compiler.Core.Translation.FromInternal.goExpression (src/Juvix/Compiler/Core/Translation/FromInternal.hs:(799,1)-(868,54))
  Juvix.Compiler.Pipeline.Artifacts.runReaderArtifacts (src/Juvix/Compiler/Pipeline/Artifacts.hs:(110,1)-(112,15))
  Juvix.Compiler.Pipeline.Artifacts.readerTypesTableArtifacts (src/Juvix/Compiler/Pipeline/Artifacts.hs:101:1-60)
  Juvix.Compiler.Core.Data.InfoTableBuilder.runInfoTableBuilder (src/Juvix/Compiler/Core/Data/InfoTableBuilder.hs:(70,1)-(130,29))
[...]

PR #2172, opened 2023-06-05
Do not filter unreachable symbols when compiling for REPL

paulcadman:
Say we have a module that import/open the Prelude:

Test.juvix
```
module Test;
import Stdlib.Prelude open;
```

When the module is compiled, we have a step in the compiler pipeline which filters out unreachable symbols. For this module all symbols are filtered because the module contains no definitions.

So if the module is loaded in the REPL, no symbols will be available to process through the evaluator. The REPL is a place to explore the symbols in the module so (like with Haskell's GHCi) it would be useful if all symbols were available in the REPL session. That's what this PR implements.

* Closes https://github.com/anoma/juvix/issues/2159

### DIFF
--- a/src/Juvix/Compiler/Pipeline.hs
+++ b/src/Juvix/Compiler/Pipeline.hs
@@ -126,7 +126,7 @@ upToCoreTypecheck =
 upToEval ::
   (Members '[HighlightBuilder, Reader EntryPoint, Files, NameIdGen, Error JuvixError, Builtins, PathResolver] r) =>
   Sem r Core.CoreResult
-upToEval = do
+upToEval =
   upToCore >>= \r -> Core.toEval (r ^. Core.coreResultTable) >>= \tab -> return r {Core._coreResultTable = tab}
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


src/Juvix/Compiler/Pipeline.hs
```diff
@@ -89,6 +89,14 @@ upToCore ::
 upToCore =
   upToInternalReachability >>= Core.fromInternal
 
+--- | A version of upToCore that does not filter out unreachable symbols.
+--- | Intended for use with the REPL.
+upToCoreRepl ::
```
janmasrovira, line 94: have you considered adding a field to the entrypoint that indicates if we want to skip reachability filtering? it seems to be slightly better since then we don't need to add another pipeline segment specific to the repl.

paulcadman: Good idea - I'll do it this way

paulcadman: Done

paulcadman: I completely misread your comment, I've now added a field to the entrypoint record.

PR #2173, opened 2023-06-05
Always print source of formatted file unless --check is specified

paulcadman:
This PR changes the behaviour of the formatter when run on files that are already formatted. Previously the source of a file that was already formatted was not output by the formatter.

After this PR, the formatter always outputs the contents of a formatted file (when used on a single file, and if the --check option is not specified).

If the `format: false` pragma is set then the source is echoed verbatim, without highlighting (because it's not possible to get the highlighting without the formatting).

This probably helps implementing the formatter in the vscode extension, see https://github.com/anoma/vscode-juvix/issues/98


janmasrovira (approved):

paulcadman:
@vrom911 would this help with the implementation of the highlighter in the vscode extension? If not we can defer merging this until after Prague.

vrom911:
I currently fixed the highlighter work in the extension, so no rush in this one 👌🏼

jonaprieto (approved):

PR #2181, opened 2023-06-06
Format returns 0 when file is not formatted

### LINKED ISSUE #2171: Format command should not exit with 1 when needs formatting
Currently, the command exit with code 1 when file is not formatted and the output of the command differs from the initial content of the module.

However, I would expect it to succeed even if file is not formatted. The purpose of the command is to format, and if it does so uit should not fail.

vrom911:
`format` command now returns code `0` most of the time.

It will return `1` when:
* some error occur, so can not format
* file is unformatted and `--check` option is used
* One or more files are not formatted in a Juvix project.

- Fixes #2171

### DIFF
--- a/app/Commands/Format.hs
+++ b/app/Commands/Format.hs
@@ -45,15 +45,16 @@ runCommand opts = do
                       "Use the --help option to display more usage information."
                     ]
                 return FormatResultFail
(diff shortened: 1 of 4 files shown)


paulcadman (dismissed):
🎉

paulcadman (approved):

PR #2182, opened 2023-06-06
Respect fixity in runtime printer

### LINKED ISSUE #2180: Respect constructor fixity when printing in the runtime
This would require storing the constructor fixities in the generated C code.

lukaszcz:
* Closes #2180

### DIFF
--- a/src/Juvix/Compiler/Backend/C/Translation/FromReg.hs
+++ b/src/Juvix/Compiler/Backend/C/Translation/FromReg.hs
@@ -11,6 +11,7 @@ import Juvix.Compiler.Backend.C.Translation.FromReg.Base
 import Juvix.Compiler.Reg.Data.InfoTable qualified as Reg
 import Juvix.Compiler.Reg.Extra qualified as Reg
 import Juvix.Compiler.Reg.Language qualified as Reg
+import Juvix.Data.Fixity qualified as Fixity
 import Juvix.Prelude
 
 fromReg :: Limits -> Reg.InfoTable -> MiniCResult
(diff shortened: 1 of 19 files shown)


src/Juvix/Compiler/Backend/C/Translation/FromReg.hs
```diff
@@ -39,9 +40,42 @@ fromReg lims tab =
...
+          Just fixity ->
+            ListInitializer
+              [ ExprInitializer $ getPrec (fixity ^. fixityPrecedence),
+                ExprInitializer $ ExpressionVar $ getAssoc (fixity ^. fixityArity)
+              ]
+          Nothing -> ExprInitializer $ macroVar "APP_FIXITY"
+
+        getPrec :: Precedence -> Expression
+        getPrec = \case
+          PrecMinusOmega -> macroVar "PREC_MINUS_OMEGA"
```
janmasrovira, line 69: let's define these literal strings separately

janmasrovira (commented):
This is great!

src/Juvix/Compiler/Backend/C/Translation/FromReg.hs
lukaszcz: Oh dear! There are several dozen strings in this file, used only in the Reg-to-C translation, and it's very unlikely they will ever change or be used anywhere else because they correspond to the macro/variable names in the runtime. If they do change (because of a change in the runtime), you need to change the surrounding translation logic anyway, so it's never going to be a simple replacement.

I don't see the point of wasting time on introducing dozens of lines of useless code of the form 
```haskell
strPrecMinusOmega :: String
strPrecMinusOmega = "PREC_MINUS_OMEGA"
```
that's never going to help with anything. The only advantage of this I can see is that one can search for all uses of the definition. But one can equally well use Ctrl+F with "PREC_MINUS_OMEGA", so not much of an advantage.

This separate string-literal-definition is a rule-of-thumb that in this particular case becomes clearly counterproductive.

What's the point? Let's just merge this.

janmasrovira (approved):

PR #2184, opened 2023-06-07
Use stderr for errors in the runtime

### LINKED ISSUE #1902: Make `trace` and `fail` write to `stderr`
Currently, `trace` and `fail` write to `stdout` (when compiled down to WASM/native). Since these are debugging functions, they should write to `stderr`. This requires fixing the runtime.

lukaszcz:
* Closes #1902

### DIFF
--- a/test/Runtime/Base.hs
+++ b/test/Runtime/Base.hs
@@ -24,6 +24,8 @@ clangCompile mkClangArgs inputFile outputFile execute step =
         execute outputFile'
     )
 
+-- The same as `P.readProcess` but instead of inheriting `stderr` redirects it
+-- to the child's `stdout`.
 readProcess :: FilePath -> [String] -> Text -> IO Text
 readProcess cmd args stdinText =
   withTempDir'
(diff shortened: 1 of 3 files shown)


test/Runtime/Base.hs
```diff
@@ -24,6 +24,30 @@ clangCompile mkClangArgs inputFile outputFile execute step =
         execute outputFile'
     )
 
+readProcess :: FilePath -> [String] -> Text -> IO Text
```
janmasrovira, line 27: Explain briefly what this function does

lukaszcz: The same as `P.readProcess` but instead of inheriting `stderr` redirects it to the child's `stdout`. I'll add a comment.

test/Runtime/Base.hs
```diff
@@ -24,6 +24,32 @@ clangCompile mkClangArgs inputFile outputFile execute step =
         execute outputFile'
     )
 
+-- The same as `P.readProcess` but instead of inheriting `stderr` redirects it
```
janmasrovira, line 27:
```suggestion
-- | The same as `P.readProcess` but instead of inheriting `stderr` redirects it
```

janmasrovira (approved):

janmasrovira (approved):

PR #2193, opened 2023-06-15
Check valid argument names in YAML

### LINKED ISSUE #2191: [BUG] Juvix to VampIR transpilation produces uncompilable VampIR code
**Describe the bug**
Transpiling [this code](https://gist.github.com/bazzilic/c0a88e62808d781b50c1d1d0651648eb) to VampIR produces an uncompilable code. VampIR compiler then fails with the following error:

``thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnableDetermineType { v: Variable { name: Some("output_note_2_value"), id: 95 } }', src/transform.rs:1234:75``

I think the culprit is this particular part of the generated VampIR code, which does not look like a valid VampIR expression:

image

**Desktop (please complete the following information):**

- OS: macOS Ventura 13.4
- Juvix version: 0.4.0-925d7cb
- Vamp-IR compiler version: 0.1.2 (commit 778d5b5)

**To Reproduce**

1. Download [this Juvix source code file](https://gist.github.com/bazzilic/c0a88e62808d781b50c1d1d0651648eb) 
2. Execute `juvix compile -t vampir ./bug.juvix` to produce `bug.pir`
3. Execute `vamp-ir halo2 compile --source bug.pir --output bug.halo2`
4. Vamp-IR compilation will fail

lukaszcz:
* Closes #2191

### DIFF
--- a/test/VampIR/Compilation/Base.hs
+++ b/test/VampIR/Compilation/Base.hs
@@ -23,9 +23,9 @@ vampirCompileErrorAssertion mainFile step = do
   entryPoint <- defaultEntryPointCwdIO mainFile
   r <- runIOEither entryPoint upToCore
   case r of
-    Left _ -> assertBool "" True
+    Left _ -> return ()
     Right res ->
       let tab = snd res ^. coreResultTable
        in case run $ runReader defaultCoreOptions $ runError @JuvixError $ toVampIR' tab of
-            Left _ -> assertBool "" True
+            Left _ -> return ()
             Right _ -> assertFailure "no error"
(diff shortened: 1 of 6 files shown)


test/VampIR/Compilation/Base.hs
```diff
@@ -13,3 +13,19 @@ vampirCompileAssertion paramsNum mainFile dataFile step = do
...
+vampirCompileErrorAssertion ::
+  Path Abs File ->
+  (String -> IO ()) ->
+  Assertion
+vampirCompileErrorAssertion mainFile step = do
+  step "Translate to JuvixCore"
+  entryPoint <- defaultEntryPointCwdIO mainFile
+  r <- runIOEither entryPoint upToCore
+  case r of
+    Left _ -> assertBool "" True
```
paulcadman, line 26: Suggestion: `Assertion` is just an alias for `IO ()`, so you can use `return ()` here instead of `assertBool "" True`

paulcadman (approved):

paulcadman (approved):

PR #2194, opened 2023-06-15
Refactor isValidChar

jonaprieto:
Move common functions to Prelude


lukaszcz (changes requested):
Let's name it more descriptively if we're making it global, e.g., `isValidIdentChar`.

A potential problem with this is that in the future valid VampIR names and valid C names might be different - then we'll need two functions again.

lukaszcz (approved):

PR #2195, opened 2023-06-16
Fix runtime C maybe benchmark example

paulcadman:
The constr_info_t struct has changed, so this example must be changed accordingly.

The benchmark builds are still broken because I missed this file in https://github.com/anoma/juvix/pull/2192

I've removed the unsupported wasm target from the `compile.sh` script in the benchmark directory to make it easier to spot errors.

### DIFF
--- a/tests/benchmark/compile.sh
+++ b/tests/benchmark/compile.sh
@@ -34,7 +34,6 @@ do
             cd juvix
             for f in *.juvix; do
                 execute "juvix compile -o `basename $f .juvix`.exe $f"
-                execute "juvix compile --target=wasm -o `basename $f .juvix`.wasm $f"
             done
             cd ..
         fi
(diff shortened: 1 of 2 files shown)


lukaszcz:
How do the benchmarks compile when doing the benchmarks? They don't use `compile.sh`? Now the WASM target name is `wasm32-wasi`.

lukaszcz (approved):

paulcadman:
> How do the benchmarks compile when doing the benchmarks? They don't use `compile.sh`? Now the WASM target name is `wasm32-wasi`.

The benchmarks don't use `compile.sh`, the compiler invocations are defined in https://github.com/anoma/juvix/blob/main/bench/Variants.hs. However there is no way of compiling all the benchmark examples without running the benchmark suite, which takes several hours. The `compile.sh` script is useful for sanity checking the examples.

PR #2204, opened 2023-06-19
Improve iterator formatting and syntax

### LINKED ISSUE #2202: Improve iterator formatting/syntax
The formatter should add a newline before the body:

```juvix
for (acc := 0) (x in xs)
  x + acc
```

However, I would also argue in favour of allowing optional braces (then the newline wouldn't be automatically inserted if things fit on one line):

```juvix
for (acc := 0) (x in xs) { x + acc }
```

The braces fit with the syntax of lambdas and the proposed optional braces for `case`. They may also be reasonable for multi-line bodies, because they clearly indicate where the body ends. Also, they are good for disambiguation in situations like:

```juvix
case M
| C1 := 
  for (acc := 0) (x in xs) {
    case N
    | D1 := ..
    | D2 := ..
  }
| C2 := ..
```

Though with the proposed optional braces in `case` one could use the braces in case instead.

lukaszcz:
* Closes #2202

### DIFF
--- a/src/Juvix/Compiler/Concrete/Print/Base.hs
+++ b/src/Juvix/Compiler/Concrete/Print/Base.hs
@@ -173,8 +173,12 @@ instance SingI s => PrettyPrint (Iterator s) where
         is' = parens . sepSemicolon <$> nonEmpty is
         rngs' = parens . sepSemicolon <$> nonEmpty rngs
         b = ppExpressionType _iteratorBody
+        b' =
+          if
+              | _iteratorBraces -> braces (oneLineOrNextNoIndent b)
+              | otherwise -> line <> b
     parensIf _iteratorParens $
-      nest (n <+?> is' <+?> rngs' <> oneLineOrNextNoIndent b)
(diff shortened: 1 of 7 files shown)


src/Juvix/Compiler/Concrete/Print/Base.hs
```diff
@@ -173,8 +173,12 @@ instance SingI s => PrettyPrint (Iterator s) where
         is' = parens . sepSemicolon <$> nonEmpty is
         rngs' = parens . sepSemicolon <$> nonEmpty rngs
         b = ppExpressionType _iteratorBody
+        b' =
+          if
+              | _iteratorBraces -> braces (oneLineOrNextNoIndent b)
+              | otherwise -> line <> b
```
janmasrovira, line 179:
```suggestion
        b'
            | _iteratorBraces = braces (oneLineOrNextNoIndent b)
            | otherwise = line <> b
```

janmasrovira (dismissed):

janmasrovira (approved):

PR #2205, opened 2023-06-19
Always print source of formatted file unless --check is specified

paulcadman:
This PR was already merged in https://github.com/anoma/juvix/pull/2173, but main was subsequently forced pushed as part of the 0.4.0 release and these changes were erased by mistake.

This PR changes the behaviour of the formatter when run on files that are already formatted. Previously the source of a file that was already formatted was not output by the formatter.

After this PR, the formatter always outputs the contents of a formatted file (when used on a single file, and if the --check option is not specified).

If the `format: false` pragma is set then the source is echoed verbatim, without highlighting (because it's not possible to get the highlighting without the formatting).

This probably helps implementing the formatter in the vscode extension, see https://github.com/anoma/vscode-juvix/issues/98


vrom911 (approved):
Thanks!

lukaszcz:
What happens when there is a syntax error?

paulcadman:
> What happens when there is a syntax error?

The corresponding parse error will be printed to stderr

lukaszcz:
> 

I hope the VSCode extension can handle that error and won't make the code disappear.

paulcadman:
> I hope the VSCode extension can handle that error and won't make the code disappear.

Good point! @vrom911 can you help with this?

PR #2207, opened 2023-06-19
Propogate formatter results when traversing subdirs

paulcadman:
Previously if a project looked like (where `Unformatted.juvix` contains unformatted code, and `Formatted.juvix` is fully formatted):

```
Dir
|- juvix.yaml
|- Unformatted.juvix
|- Subdir
   |- Formatted.juvix
```

and the user ran `juvix format Dir/` then the command would return exit code 0 instead of exit code 1. This is because only the result from formatting files within `Subdir` was used, the result from `Dir` was discarded.

This PR fixes this, the results of formatting all subdirectories in a project are combined appropriately.

### DIFF
--- a/src/Juvix/Formatter.hs
+++ b/src/Juvix/Formatter.hs
@@ -100,10 +100,10 @@ formatProject p = do
       [Path Rel File] ->
       FormatResult ->
       Sem r (FormatResult, Recurse Rel)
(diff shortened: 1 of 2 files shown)


vrom911 (approved):
Nice catch 👍🏼

PR #2212, opened 2023-06-20
Fix `isNegative` in the VampIR runtime

### LINKED ISSUE #2206: Fix `isNegative` in the generated VampIR code
Currently, it is incorrect. Use the implementation from https://anoma.github.io/VampIR-Book/section_4_3.html.

lukaszcz:
* Closes #2206


src/Juvix/Compiler/Backend/VampIR/Pretty/Options.hs
```diff
@@ -2,14 +2,17 @@ module Juvix.Compiler.Backend.VampIR.Pretty.Options where
...
+  { _optIntegerBits :: Int
+  }
 
 makeLenses ''Options
 
 defaultOptions :: Options
-defaultOptions = Options
+defaultOptions =
+  Options
+    { _optIntegerBits = 24
```
paulcadman, line 14: Where does the number `24` come from? Is it something that's determined by vamp-ir. Will it need to be user configurable at some point in the future?

lukaszcz: That's a good question: how many bits should we support? Should this be configurable by the user? Shouldn't we use different datatypes, because Int is supposed to be unbounded?

Efficiency of VampIR degenerates quickly with increasing number of bits, so I chose a middle ground for now, where you can count up to about eight million but VampIR compilation/proof/verification don't take ages (with 32 bits they would take roughly twice as long).

runtime/vampir/stdlib.pir
```diff
@@ -0,0 +1,47 @@
+def isBool x = (x * (x - 1) = 0);
```
paulcadman, line 1: I think a reference to this file needs to be added to the `extra-source-files` entry in the `package.yaml`:

https://github.com/anoma/juvix/blob/02e2f9f2dd093c504f132b832a0a1ef28b5c1f07/package.yaml#L22

This doesn't seem to be an issue for the stack build, but I think matters for `cabal install` and for other packaging purposes.

lukaszcz: done

paulcadman (approved):

PR #2213, opened 2023-06-20
Remove Subtree sections, all collapse all button

### LINKED ISSUE #2198: The generated documentation index looks poorly
image

Why do we have separate "Subtree" nodes in bold? They are in bold, but provide no information and only clutter the index and make it hard to read.

The actual parent nodes (Stdlib, Trait, System, ..) should be in bold and themselves collapsible (they should have the collapsing icon in front of them, not below them with an additional "Subtree" node).

Also, all nodes should be collapsed by default.

vrom911:
- Resolves #2198

Screenshot 2023-06-20 at 14 09 21

### DIFF
--- a/src/Juvix/Compiler/Backend/Html/Extra.hs
+++ b/src/Juvix/Compiler/Backend/Html/Extra.hs
@@ -51,6 +51,12 @@ jsLink js = do
       ! Attr.type_ "text/javascript"
     $ mempty
 
+closeAll :: Html
+closeAll =
+  script
+    ! Attr.type_ "text/javascript"
+    $ "function closeAll() { document.body.querySelectorAll('summary').forEach((e) => { document.body.querySelectorAll('details').forEach((e) => (e.hasAttribute('open')) ? e.removeAttribute('open') : '')})}"
+
 linuwialCss :: (Members '[Reader HtmlOptions] r) => Sem r Html
 linuwialCss = cssLink "linuwial.css"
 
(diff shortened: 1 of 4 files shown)


src/Juvix/Compiler/Backend/Html/Extra.hs
```diff
@@ -51,6 +51,12 @@ jsLink js = do
       ! Attr.type_ "text/javascript"
     $ mempty
 
+closeAll :: Html
+closeAll =
+  script
+    ! Attr.type_ "text/javascript"
+    $ "function closeAll() { document.body.querySelectorAll('summary').forEach((e) => { document.body.querySelectorAll('details').forEach((e) => (e.hasAttribute('open')) ? e.removeAttribute('open') : '')})}"
```
janmasrovira, line 58: Let's move this script to `assets/js`

janmasrovira, line 58: do you think we should have another button to expand everything? or maybe use the same button?
Either way, I think it would be nice to be able to do that

janmasrovira (commented):
I have one wording suggestion, but feel free to adapt it or just ignore it:
open all -> Expand all modules
close all -> Hide all modules

janmasrovira (approved):
Excellent

vrom911:
To summarise, this is the current version:

https://github.com/anoma/juvix/assets/8126674/032869d9-f8ef-4c1f-995d-5736675079d4

PR #2219, opened 2023-06-20
Remove abstract

### LINKED ISSUE #1690: The standard library definitions are not checked for termination
For example, adding the following definitions to Stdlib.Data.Nat doesn't make the compiler fail for programs using the library, even if they use the `div` function. If the standard library file is compiled directly, then termination checking fails for `div`.
```
sub : Nat → Nat → Nat;
sub zero _ := zero;
sub n zero := n;
sub (suc n) (suc m) := sub n m;

div : Nat → Nat → Nat;
div zero _ := zero;
div n m := suc (div (sub n m) m);
```

### LINKED ISSUE #2002: Remove `Abstract`
The `Abstract` language no longer has a reason to exist.
We have agreed to remove the `Abstract` language and move termination to `Internal`

### LINKED ISSUE #2224: Repl crashes when importing `Stdlib.Data.Int` after `Stdlib.Data.Nat`
To reproduce run `juvix repl` in a directory that is not part of a juvix project, or a project that has the stdlib as a dependency.
```
Stdlib.Prelude> import Stdlib.Data.Nat
Stdlib.Prelude> import Stdlib.Data.Int
impossible
  error, called at src/Juvix/Prelude/Base.hs:368:14 in juvix-0.4.0-7RQMqAwl1L1J48TJpqbZfT:Juvix.Prelude.Base
  impossible, called at src/Juvix/Compiler/Internal/Data/InfoTable.hs:173:43 in juvix-0.4.0-7RQMqAwl1L1J48TJpqbZfT:Juvix.Compiler.Internal.Data.InfoTable
Stdlib.Prelude>
```

The error happens during arity checking. `Nat` is missing from the Internal InfoTable. This is most likely related to how we handle imports in the translation from Abstract to Internal.

### LINKED ISSUE #2237: Add `dev internal reachability` command
This command should print the `DependencyInfo` record and should be useful when debugging reachability related bugs.

janmasrovira:
- Closes #2002 
- Closes #1690 
- Closes #2224
- Closes #2237


src/Juvix/Compiler/Internal/Translation/FromInternal.hs
```diff
@@ -86,23 +91,26 @@ typeCheckExpression ::
...
-    $ checkInclude i
+    . runTypesTableArtifacts
+    . runFunctionsTableArtifacts
+    . ignoreHighlightBuilder
+    . runBuiltinsArtifacts
+    . runNameIdGenArtifacts
+    . ignoreOutput @Example
+    . runReader table
+    . withEmptyVars
+    -- TODO Sotre cache in Artifacts and use it here
```
lukaszcz, line 111: typo

lukaszcz (approved):

PR #2227, opened 2023-06-26
Fix de Bruijn indices in LetRecs

### LINKED ISSUE #2226: The de Bruijn indices in the types in LetRecs are all messed up
I don't know how this even works. Probably only because we use ordinary Lets for non-recursive definitions and don't actually look into the types most of the time.

Just see the output of `from-concrete` for:
```
module letrec;

import Stdlib.Prelude open;

myfun : {A : Type} -> (A -> A) -> A -> A;
myfun {A} f a :=
      let
        go : Nat -> A -> A;
        go zero a := a;
        go (suc n) a := f (go n a);
      in
      go 5 a;

main : Nat;
main := myfun ((+) 1) 7;
```

The letrec binder is not accounted for in the lambdas of the value (the `go` function). That's a problem with the translation from Internal. Somehow, this compiles nonetheless.

Further, lambda-lifting messes up indices in the types. The types, in contrast to the values, are not in the scope of implicit binders for the values, but lambda-lifting seems to assume they are.

lukaszcz:
* Closes #2226 

For the program
```juvix
module letrec;

import Stdlib.Prelude open;

myfun : {A : Type} -> (A -> A) -> A -> A;
myfun {A} f a :=
      let
        go : Nat -> A -> A;
        go' : Nat -> A -> A;

        go zero a := a;
        go (suc n) a := f (go' n a);

        go' zero a := a;
        go' (suc n) a := f (go n a);
      in
      go 5 a;

main : Nat;
main := myfun ((+) 1) 7;
```
after translating to Core and lambda-lifting we got the incorrect indices in types of the let-bindings.
```
def myfun : Π A : Type, (A$0 → A$1) → A$1 → A$2 :=
  λ(A : Type)
    λ(f : A$0 → A$1)
      λ(a : A$1)
        let go' : Int → a$1 → a$2 := go'_9 A$2 f$1 in
        let go : Int → a$2 → a$3 := go_10 A$3 f$2 in
        go$0 5 a$2;
```
The indices have been corrected in this PR.

### DIFF
--- a/src/Juvix/Compiler/Core/Transformation/LambdaLetRecLifting.hs
+++ b/src/Juvix/Compiler/Core/Transformation/LambdaLetRecLifting.hs
@@ -159,7 +159,7 @@ lambdaLiftNode aboveBl top =
                   ]
           declareTopSyms
 
-          let -- TODO it can probably be simplified
+          let -- TODO it can probably be simplified, and it's wrong
               shiftHelper :: Node -> NonEmpty (Node, Binder) -> Node
               shiftHelper b = goShift 0
                 where
@@ -171,7 +171,7 @@ lambdaLiftNode aboveBl top =
                         | otherwise -> impossible
                       (y : ys) -> mkLet mempty bnd' (shift k x) (goShift (k + 1) (y :| ys))
                       where
-                        bnd' = over binderType (shift k . subsCalls) bnd
+                        bnd' = over binderType (shift k) bnd
           let res :: Node
               res = shiftHelper body' (nonEmpty' (zipExact letItems letRecBinders'))
           return (Recur res)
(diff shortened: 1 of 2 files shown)


lukaszcz:
> Is it possible to add a test that has been fixed by this?

It's really hard because I can't produce an example that actually makes things crash with the old code, because currently the indices in types are mostly ignored.

lukaszcz:
> > Is it possible to add a test that has been fixed by this?
> 
> It's really hard because I can't produce an example that actually makes things crash with the old code, because currently the indices in types are mostly ignored.

Any test looking at the output instead would be very unstable, and I don't think we should add it.

janmasrovira:
> > Is it possible to add a test that has been fixed by this?
> 
> It's really hard because I can't produce an example that actually makes things crash with the old code, because currently the indices in types are mostly ignored.

That's fine, then please, paste the output (before and after the fix) here as a comment

janmasrovira (approved):

PR #2231, opened 2023-06-27
Fold lets when the bound variable occurs at most once

lukaszcz:
For example, convert
```
let x := f a b c in
g x
```
to
```
g (f a b c)
```

### DIFF
--- a/src/Juvix/Compiler/Core/Info/FreeVarsInfo.hs
+++ b/src/Juvix/Compiler/Core/Info/FreeVarsInfo.hs
@@ -1,0 +1,49 @@
+module Juvix.Compiler.Core.Info.FreeVarsInfo where
+
+import Data.HashMap.Strict qualified as HashMap
+import Juvix.Compiler.Core.Extra
+import Juvix.Compiler.Core.Info qualified as Info
+import Juvix.Compiler.Core.Language
+
+newtype FreeVarsInfo = FreeVarsInfo
+  { -- map free variables to the number of their occurrences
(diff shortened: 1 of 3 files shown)


src/Juvix/Compiler/Core/Info/FreeVarsInfo.hs
```diff
@@ -0,0 +1,49 @@
...
+          fvi = FreeVarsInfo (HashMap.singleton _varIndex 1)
+      _ ->
+        modifyInfo (Info.insert fvi) node
+        where
+          fvi =
+            FreeVarsInfo $
+              foldr
+                ( \NodeChild {..} acc ->
+                    HashMap.unionWith (+) acc $
+                      HashMap.mapKeys (\idx -> idx - _childBindersNum) $
```
janmasrovira, line 37: Have you considered using `Map` instead of `HashMap`?
We are doing lots of `unionWith` and `mapKeys`. [`Map.unionWith`](https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Map-Strict.html#v:unionWith) is slightly more efficient than [`HashMap.unionWith`](https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Map-Strict.html#v:unionWith). We could also use [`Map.mapKeysMonotonic`](https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Map-Strict.html#v:mapKeysMonotonic).

janmasrovira (approved):

PR #2236, opened 2023-06-28
Local pragmas improvements

lukaszcz:
* Depends on #2234 
* Depends on #2231 
* Fixes the indices in `inline` and `specialize` for local lambda-lifted identifiers.
* Adds the `specialize-by` pragma which allows to specialize a local function by one of its free variables, or specialize arguments by name. For example:
```
funa : {A : Type} -> (A -> A) -> A -> A;
funa {A} f a :=
  let
    {-# specialize-by: [f] #-}
    go : Nat -> A;
    go zero := a;
    go (suc n) := f (go n);
  in go 10;
```
If `funa` is inlined, then `go` will be specialized by the actual argument substituted for `f`.


src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -352,7 +352,11 @@ translateCase translateIf dflt Case {..} = case _caseBranches of
 
 checkDepth :: Int -> Node -> Bool
 checkDepth 0 _ = False
-checkDepth d node = all (checkDepth (d - 1)) (childrenNodes node)
+checkDepth d node = case node of
+  NApp App {..} ->
+    checkDepth d _appLeft && checkDepth (d - 1) _appRight
```
janmasrovira, line 357: can you explain this to me?

lukaszcz: Previously e.g. `f a b c` had depth 4 because it's `((f a) b) c`. For the purposes of assessing how big terms are it's more reasonable to treat applications as flat.

janmasrovira (approved):

PR #2243, opened 2023-07-03
New syntax for function definitions

### LINKED ISSUE #2060: Syntax for mutually recursive types
The syntax for mutually recursive functions is straightforward: First we write the type signatures and then the clauses.

In the current syntax is not possible to define mutually recursive types without defining auxiliary type synonyms. For that, I propose a syntax enchancement copied from [Agda's syntax](https://agda.readthedocs.io/en/v2.6.3/language/mutual-recursion.html#mutual-recursion-forward-declaration). 

The enchancement consists in adding type definitions signatures, where the constructor list is not given.
E.g.

```agda
module LambdaCalculus;

type Lambda (V : Type) : Type; -- The `: Type` is optional

type App (V : Type) : Type;

type Expr (V : Type) :=
  | var : V -> Expr V
  | lam : Lambda V -> Expr V
  | app : App V -> Expr V;

-- here we give the constructors for `Lambda`
-- we need to check that signature matches the previous signature for `Lambda`
type Lambda (V : Type) :=
  | mkLambda : Expr V -> Lambda V;

type App (V : Type) :=
  | mkApp : Expr V -> Expr V -> App V;
```

Note that the previous example can be written with the help of type synonyms without the additional syntax.

```agda
module LambdaCalculus;

LambdaTy : Type -> Type;

AppTy : Type -> Type;

type Expr (V : Type) :=
  | var : V -> Expr V
  | lam : LambdaTy V -> Expr V
  | app : AppTy V -> Expr V;

type Lambda (V : Type) :=
  | mkLambda : Expr V -> Lambda V;

type App (V : Type) :=
  | mkApp : Expr V -> Expr V -> App V;

LambdaTy := Lambda;

AppTy := App;

```

### LINKED ISSUE #2189: Re-think function definitions
Our clause definition syntax is somewhat inconsistent. Sometimes we separate clauses with `|` (in lambdas and cases) and sometimes with `;` (at the top-level). Sometimes we use braces (in lambdas) and sometimes we don't (in cases and top-level definitions). Moreover, top-level clause definition syntax is inconsistent with the type definition syntax.

This is the root of the issue:
* #2051 

The lack of braces in `case` makes writing nested cases inconvenient:
```juvix
case a
| C x := (case f x
              | D := A
              | E := B)
| F := C;
```
The nested `case` needs to be put in parentheses to disambiguate.

Personally, I don't like the Haskell/Agda/Idris approach to function definitions. I don't like repeating the function name in each clause. Also, I wouldn't mind something more concise and closer to most other programming languages, where you can specify argument types next to arguments without bothering to write a separate signature.

Function definitions in Lean 4 come close to a syntax I would consider convenient and immediately understandable to someone not familiar with Haskell/Agda/Idris:
```lean
def twice (f : Nat -> Nat) (x : Nat) : Nat :=
  f (f x)

def compose (α β γ : Type) (g : β → γ) (f : α → β) (x : α) : γ :=
  g (f x)
```
Lean syntax naturally extends to multiple-clause pattern matching:
```lean
def length : List α → Nat
  | [] => 0
  | y :: ys => Nat.succ (length ys)

def drop : Nat → List α → List α
  | Nat.zero, xs => xs
  | _, [] => []
  | Nat.succ n, x :: xs => drop n xs
```

I find it ironic that while we're promoting the Haskell/Agda/Idris approach in Juvix, in our own Haskell code we're actively avoiding it. For example, we're trying to replace things like
```haskell
f :: List a -> a
f [] = A
f (x : xs) = B
```
with
```haskell
f :: List a -> a
f = \case
  [] -> A
  x : xs -> B
```

Proposals/ideas
============

1. Add optional braces to `case` and mandatory parentheses around the matched expression:
```juvix
case (f a b) {
| C x := A
| D y := B
}
```
Then the first `|` may be optional to make it more uniform with the lambdas and a new syntax for function clause definitions (below).

We need to add parentheses around the matched expression to avoid ambiguity with implicit arguments, but this seems natural. More precisely, we should be matching a single atom, so the following are still fine:
```juvix
case x
   A := ..
 | B := ..;

case x {
   A := ..
 | B := ..
};
```

2. Make braces optional in a lambda:
```juvix
\ 
| C x := A
| D y := B;
```
Then the lambda syntax is entirely analogous to the `case` syntax. Except that we don't require parentheses around top patterns in `case` because there is only one pattern matched -- this is very convenient but breaks the uniformity slightly.

3. Change top-level clause syntax to use `|` instead of repeating the function name and require all clauses for a function to be next to each other. It practically never happens anyway that the clauses of different functions are mixed (I'm not sure if we even allow it).
```juvix
func
  | x (C y) := A
  | Z (D y) := B;
```
Like in a lambda and a type definition, the first `|` is optional, which is handy when there is only one clause:
```juvix
func x y := A;
```
Now top-level clause definitions are analogous to cases, lambdas and type definitions. Note that having optional braces in `case` becomes more important with this change, because we now separate clauses with `|` and use `;` to end the entire definition:
```juvix
func
  | (C x) := case (f x) { A := .. | B := .. }
  | (D x) := case (g x) { A := .. | B := .. };
```
We could even allow optional braces for top-level clause definitions to make it completely consistent with `case` and lambda, but I don't see any practical use-case for it.
```juvix
func {
  (C x) := A
| (D x) := B
};
```

4. Allow to (optionally) merge the type signature with clause definitions.
```juvix
fact : Nat -> Nat
  | zero := 1
  | n@(suc n') := n * fact n';
```
With the type signature, the first `|` cannot be optional. We can still have a special case for a single clause with no arguments:
```juvix
id : Nat -> Nat := \ x := x;
```

6. Allow to move some arguments to the left of `:` when the signature is specified together with the clauses:
```juvix
id {A : Type} (x : A) : A := x;

compose {A B C : Type} (f : B -> C) (g : A -> B) (x : A) : C := f (g x);

map {A B : Type} (f : A -> B) : List A -> List B
  | nil := nil
  | (x :: xs) := f x :: map f xs;
```

With traits and the proposed universal quantification moved to the left:
```juvix
compose !{A B C} (f : B -> C) (g : A -> B) (x : A) : C := f (g x);

func !{A B} {{Ord A; Enum B}} (f : A -> B) : A -> B
  | ..
  | ..;

listOrd !{A} {{O : Ord A}} : Ord (List A) := MkOrd with {
  compare (lst1 : List A) (lst2 : List A) : Bool := ...;
};
```

Note that after all these changes signatures could still be specified separately from the clauses, but presumably this would be much less common (used mostly for mutual recursion):
```juvix
even : Nat -> Bool;
odd : Nat -> Bool;

even
  | zero := true
  | (suc n) := odd n;

odd
  | zero := false
  | (suc n) := even n;
```

janmasrovira:
- Closes #2060
- Closes #2189

- This pr adds support for the syntax described in #2189. It does not drop support for the old syntax.

It is possible to automatically translate juvix files to the new syntax by using the formatter with the `--new-function-syntax` flag. E.g.

```
juvix format --in-place --new-function-syntax
```
# Syntax changes
Type signatures follow this pattern:
```
f (a1 : Expr) .. (an : Expr) : Expr
```
where each `ai` is a non-empty list of symbols. Braces are used instead of parentheses when the argument is implicit.

Then, we have these variants:
1. Simple body. After the signature we have `:= Expr;`.
2. Clauses. The function signature is followed by a non-empty sequence of clauses. Each clause has the form:
```
| atomPat .. atomPat := Expr
```
# Mutual recursion
Now identifiers **do not need to be defined before they are used**, making it possible to define mutually recursive functions/types without any special syntax.
There are some exceptions to this. We cannot forward reference a symbol `f` in some statement `s` if between `s` and the definition of `f` there is one of the following statements:
1. Local module
2. Import statement
3. Open statement

I think it should be possible to drop the restriction for local modules and import statements


paulcadman:
In https://github.com/anoma/juvix-stdlib/blob/85751d3a7e5edd97a3d12dae197273731a2088cf/Stdlib/Data/Bool.juvix#L15 we now must write:

```
import Stdlib.Trait.Eq as Eq;
open Eq using {Eq};

module BoolTraits;
  Eq : Eq.Eq Bool :=
...
```

instead of:

```
import Stdlib.Trait.Eq as Eq;
open Eq using {Eq};

module BoolTraits;
  Eq : Eq Bool :=
...
```

(The latter now fails the termination checker).

The reasoning for allowing `Eq : Eq Bool` was that it was not possible refer to the lhs `Eq` in its type signature so there was no ambiguity.  Did you conclude that this was too confusing so it's better to qualify `Eq`, or was this change a side-effect of something else?

paulcadman:
If implementors of traits must use `Eq : Eq.Eq A := ...` then perhaps we should use the suffix `Trait` for trait definitions:

```
--- A trait defining equality
type EqTrait (A : Type) :=
  | mkEq : (A -> A -> Bool) -> EqTrait A;
```

Then implementors of traits would not have to do any additional imports other than the Prelude (as before) and could write the following (avoiding the stuttering):

```
Eq : EqTrait A := ...
```

What do you think?

janmasrovira:
> In https://github.com/anoma/juvix-stdlib/blob/85751d3a7e5edd97a3d12dae197273731a2088cf/Stdlib/Data/Bool.juvix#L15 we now must write:
> 
> ```
> import Stdlib.Trait.Eq as Eq;
> open Eq using {Eq};
> 
> module BoolTraits;
>   Eq : Eq.Eq Bool :=
> ...
> ```
> 
> instead of:
> 
> ```
> import Stdlib.Trait.Eq as Eq;
> open Eq using {Eq};
> 
> module BoolTraits;
>   Eq : Eq Bool :=
> ...
> ```
> 
> (The latter now fails the termination checker).
> 
> The reasoning for allowing `Eq : Eq Bool` was that it was not possible refer to the lhs `Eq` in its type signature so there was no ambiguity. Did you conclude that this was too confusing so it's better to qualify `Eq`, or was this change a side-effect of something else?

This was a side-effect of allowing forward references. Before the `Eq` on the rhs referred to the imported `Eq`. Now it refers to the `Eq` in the lhs, so it fails the termination checker.

janmasrovira:
> If implementors of traits must use `Eq : Eq.Eq A := ...` then perhaps we should use the suffix `Trait` for trait definitions:
> 
> ```
> --- A trait defining equality
> type EqTrait (A : Type) :=
>   | mkEq : (A -> A -> Bool) -> EqTrait A;
> ```
> 
> Then implementors of traits would not have to do any additional imports other than the Prelude (as before) and could write the following (avoiding the stuttering):
> 
> ```
> Eq : EqTrait A := ...
> ```
> 
> What do you think?

I agree that `Eq.Eq` looks bad. But I am not fully convinced by defining `type EqTrait (A : Type)`, since ultimately we want to use traits in many places and it will be tedious to repeat `xTrait` instead of `x` potentially multiple times in most of the type signatures. Imagine our Haskell codebase where every class constraint is suffixed by `Trait`.

One option to replace the existing
```
open Eq using {Eq};
```
By
```
open Eq using {Eq as EqTrait};
```
And then have 
```
Eq : EqTrait A := ...
```
As you suggested.

paulcadman:
The main issue I have with the `Eq.Eq` is that trait implementors must write (and discover that they must write) `import Stdlib.Trait.Eq as Eq;` instead of just importing `Stdib.Prelude` so `open Eq using {Eq as EqTrait};` wouldn't solve that. But I agree that the `Trait` suffix is not a good idea.

src/Juvix/Compiler/Concrete/Data/Scope.hs
```diff
@@ -105,3 +105,9 @@ emptyScope absPath =
       _scopeTopModules = mempty,
       _scopeLocalSymbols = mempty
     }
+
+newtype Reserved = Reserved
```
paulcadman, line 109: Is this type used anywhere?

janmasrovira:
> The main issue I have with the `Eq.Eq` is that trait implementors must write (and discover that they must write) `import Stdlib.Trait.Eq as Eq;` instead of just importing `Stdib.Prelude` so `open Eq using {Eq as EqTrait};` wouldn't solve that. But I agree that the `Trait` suffix is not a good idea.

That's true, but this is a temporary problem. In the not so distant future we will have trait instances, which will not bind any name, and so this won't be a problem anymore. So trait implementors (outside of the stdlib) will be able to write
```
import Stdlib.Prelude;

instance Eq Bool :=
    mkEq ...
```


In the meanwhile, another possible temporary patch would be to add a local module in the prelude:
```
module Trait;
  import Stdlib.Trait open public;
end;
```
And then the users can write:
```
import Stdlib.Prelude;

Eq : Trait.Eq Bool :=
    mkEq ...
```
It is maybe slightly better than the option I gave before, but it still does not fully solve the discoverability problem

paulcadman:
> And then the users can write:
> 
> ```
> import Stdlib.Prelude;
> 
> Eq : Trait.Eq Bool :=
>     mkEq ...
> ```
> 

This is very good - thanks! I'll raise it as an issue against stdlib once this PR is merged.

PR #2257, opened 2023-07-24
Separate modules namespace

### LINKED ISSUE #2110: Separate modules namespace
- Currently module names live in the same space as functions, types, constructors, etc.
This has the disadvantage that a module cannot export a type and a module with the same name.
This is sometimes useful. Consider the traits usecase (https://github.com/anoma/juvix-stdlib/pull/54).
Currently we write `ListTraits.Eq` to refer to the `Eq` instance for `List`. Ideally we would like to write `List.Eq` to refer to that instance, but this is impossible because the Prelude cannot export both the type `List` and the `List` module. In order to solve this, one option is to separate module names from the rest of the names. Syntax-wise, this would entail changing the syntax of `open` statements by requiring to prefix module names in `using {..}` and `hiding {..}` by the keyword `module`. E.g:
```
open Stdlib.Data.List using {List; module List}
``` 
Applying this change would only require us to add the `module` keyword in all `using/hiding {..}` that contain a module. The rest of the code should continue to work as before.

- Even though the usecase for traits is temporary (until we have trait inference/type classes), it is easy to see that this can be generally beneficial. Another interesting usecase is discussed in #2111.

janmasrovira:
- Closes #2110 

Now module names are separated from other symbols (types, functions, constructors, etc).
In practice this means that:
1. A module can now export a module symbol `S` and a non-module symbol `S` at the same time.
2. In `using` and `hiding` lists, we need to prefix module symbols with the `module` keyword. E.g. `open M using {module Aux}`. Other non-module symbols do not need a prefix.

### DIFF
--- a/app/Commands/Repl.hs
+++ b/app/Commands/Repl.hs
@@ -258,7 +258,8 @@ printDocumentation = replParseIdentifiers >=> printIdentifiers
             KNameLocal -> return Nothing
             KNameFunction -> getDocFunction n
             KNameConstructor -> getDocConstructor n
-            _ -> impossible
+            KNameLocalModule -> impossible
+            KNameTopModule -> impossible
           printDoc mdoc
           where
             printDoc :: Maybe (Concrete.Judoc 'Concrete.Scoped) -> Repl ()
@@ -313,7 +314,8 @@ printDefinition = replParseIdentifiers >=> printIdentifiers
                 KNameLocal -> return ()
                 KNameFunction -> printFunction n
                 KNameConstructor -> printConstructor n
-                _ -> impossible
+                KNameLocalModule -> impossible
+                KNameTopModule -> impossible
           where
             printLocation :: HasLoc s => s -> Repl ()
             printLocation def = do
--- a/app/Commands/Repl.hs
+++ b/app/Commands/Repl.hs
@@ -251,12 +251,14 @@ printDocumentation = replParseIdentifiers >=> printIdentifiers
 
         printIdentifier :: Concrete.ScopedIden -> Repl ()
         printIdentifier s = do
-          mdoc <- case s of
-            Concrete.ScopedAxiom a -> getDocAxiom (a ^. Scoped.nameId)
-            Concrete.ScopedInductive a -> getDocInductive (a ^. Scoped.nameId)
-            Concrete.ScopedVar {} -> return Nothing
-            Concrete.ScopedFunction f -> getDocFunction (f ^. Scoped.nameId)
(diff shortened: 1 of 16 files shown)


src/Juvix/Compiler/Internal/Translation/FromConcrete.hs
```diff
@@ -774,12 +774,15 @@ goExpression = \case
...
-      ScopedVar v -> Internal.IdenVar (goSymbol v)
-      ScopedFunction fun -> Internal.IdenFunction (goName fun)
-      ScopedConstructor c -> Internal.IdenConstructor (goName c)
+    goIden x = Internal.ExpressionIden $ case getNameKind x of
+      KNameAxiom {} -> Internal.IdenAxiom n'
+      KNameInductive {} -> Internal.IdenInductive n'
+      KNameLocal {} -> Internal.IdenVar n'
+      KNameFunction {} -> Internal.IdenFunction n'
+      KNameConstructor {} -> Internal.IdenConstructor n'
+      _ -> impossible
```
janmasrovira, line 783: list all cases

app/Commands/Repl.hs
```diff
@@ -251,12 +251,14 @@ printDocumentation = replParseIdentifiers >=> printIdentifiers
...
-            Concrete.ScopedFunction f -> getDocFunction (f ^. Scoped.nameId)
-            Concrete.ScopedConstructor c -> getDocConstructor (c ^. Scoped.nameId)
+          let n = s ^. Concrete.scopedIden . Scoped.nameId
+          mdoc <- case getNameKind s of
+            KNameAxiom -> getDocAxiom n
+            KNameInductive -> getDocInductive n
+            KNameLocal -> return Nothing
+            KNameFunction -> getDocFunction n
+            KNameConstructor -> getDocConstructor n
+            _ -> impossible
```
janmasrovira, line 261: list all cases

app/Commands/Repl.hs
```diff
@@ -303,13 +305,15 @@ printDefinition = replParseIdentifiers >=> printIdentifiers
...
-            Concrete.ScopedConstructor c -> printConstructor (c ^. Scoped.nameId)
+        printIdentifier s =
+          let n = s ^. Concrete.scopedIden . Scoped.nameId
+           in case getNameKind s of
+                KNameAxiom -> printAxiom n
+                KNameInductive -> printInductive n
+                KNameLocal -> return ()
+                KNameFunction -> printFunction n
+                KNameConstructor -> printConstructor n
+                _ -> impossible
```
janmasrovira, line 316: list all cases

paulcadman (approved):

PR #2260, opened 2023-07-26
Add field projections for records

### LINKED ISSUE #2258: Add field projections for records
Given a record type (a type with a single constructor that has been defined with record syntax), we should generate field projections.
E.g. if we have:
```
type Pair (A B : Type) := mkPair {
  fst : A;
 snd : B;
};
```
There should be the following functions in scope:
```
Pair.fst : {A B : Type} -> Pair A B -> A;
Pair.snd : {A B : Type} -> Pair A B -> B;
```

One already discussed way to implement this is to generate a local module that looks like this:
```
module Pair;
 fst {A B : Type} : Pair A B -> A
  | mkPair a b := a;

 snd : {A B : Type} : Pair A B -> B
  | mkPair a b := b;
end;
```
This module should be generated during scoping.

janmasrovira:
- Closes #2258 

# Overview
When we define a type with a single constructor and one ore more fields, a local module is generated with the same name as the inductive type. This module contains a projection for every field. Projections can be used as any other function.

E.g. If we have
```
type Pair (A B : Type) := mkPair {
 fst : A;
 snd : B;
};
```
Then we generate
```
module Pair;
 fst {A B : Type} : Pair A B -> A
  | (mkPair a b) := a;

 snd : {A B : Type} : Pair A B -> B
  | (mkPair a b) := b;
end;
```


app/Commands/Repl.hs
```diff
@@ -60,6 +60,7 @@ helpTxt =
   :multiline                Start a multi-line input. Submit with <Ctrl-D>
   :root                     Print the current project root
   :version                  Display the Juvix version
+  :dev        DEV CMD       Command reserved for debugging
```
paulcadman, line 63: As discussed consider moving this to dev repl

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -151,6 +156,12 @@ data Definition (s :: Stage)
   | DefinitionInductive (InductiveDef s)
   | DefinitionAxiom (AxiomDef s)
   | DefinitionTypeSignature (TypeSignature s)
+  | DefinitionProjectionDef (ProjectionDef s)
+
+_DefinitionInductive :: Traversal' (Definition s) (InductiveDef s)
```
paulcadman, line 161: Is this used by anything?

src/Juvix/Data/Effect/Fail.hs
```diff
@@ -15,6 +15,12 @@ runFail ::
 runFail = fmap (^? _Right) . runError @() . reinterpret (\Fail -> throw ())
 {-# INLINE runFail #-}
 
+ignoreFail ::
```
paulcadman, line 18: This doesn't seem to be used, but do you think it's useful to have it?

paulcadman (approved):

src/Juvix/Data/Effect/Fail.hs
janmasrovira: I am using it in #2263 so I think it is fine to keep it here

src/Juvix/Compiler/Concrete/Language.hs
janmasrovira: It was not being used. I removed it

app/Commands/Repl.hs
janmasrovira: this line will only be shown in the dev repl now.

paulcadman (approved):

PR #2264, opened 2023-07-28
zkLLVM integration

lukaszcz:
Work in progress

Current status:
* C runtime has been updated to support zkLLVM compilation.
* C code generation has been updated to support zkLLVM.
* Target `zkllvm` added. Compilation through C with zkLLVM clang to an *.ll file works.
* Running the zkLLVM assigner (assignment table and constraint generation) on the generated *.ll file works.
* Running the zkLLVM transpiler (proof generation) on the output of the assigner results in a segmentation fault in the transpiler. This is a bug in the zkLLVM transpiler which the zkLLVM team knows about and needs to fix.

Note: for CI to work it needs to use the zkLLVM version of clang.


jonaprieto:
Added to https://github.com/anoma/juvix/wiki/Catalogue-of-Deferred-Interesting-Pull-Requests

PR #2267, opened 2023-07-31
Disallow iterators with zero ranges

### LINKED ISSUE #2265: Make iterators have at least one range
- Currently iterators are allowed to have no ranges. Because of that, we had to introduce `AmbiguousIterator` in the parsing stage in #2250. We latter agreed that iterators with no ranges don't have any practical use, so we should force them to have at least one range and remove `AmbiguousIterator` from the code.

lukaszcz:
* Closes #2265

### DIFF
--- a/src/Juvix/Data/IteratorAttribs.hs
+++ b/src/Juvix/Data/IteratorAttribs.hs
@@ -19,7 +19,7 @@ instance FromJSON IteratorAttribs where
         _iteratorAttribsRangeNum <- keyMay "range" asIntegral
         unless
           (maybe True (> 0) _iteratorAttribsRangeNum)
-          (throwCustomError "the number of ranges must be positive")
+          (throwCustomError "the iterator must have at least one range")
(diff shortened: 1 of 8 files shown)


src/Juvix/Data/IteratorAttribs.hs
```diff
@@ -17,6 +17,9 @@ instance FromJSON IteratorAttribs where
         checkYamlKeys ["init", "range"]
         _iteratorAttribsInitNum <- keyMay "init" asIntegral
         _iteratorAttribsRangeNum <- keyMay "range" asIntegral
+        unless
+          (maybe True (> 0) _iteratorAttribsRangeNum)
+          (throwCustomError "the number of ranges must be positive")
```
janmasrovira, line 22:
```suggestion
          (throwCustomError "the iterator must have at least one range")
```

janmasrovira (approved):

PR #2270, opened 2023-08-03
User-friendly operator declaration syntax

### LINKED ISSUE #1964: User-friendly specification of operator priorities
Specifying operator priorities by numbers is, in my opinion, a rather daft idea. These numbers are hard to remember and they don't mean anything: what matters is the relative order of the operators (linearised if necessary).

Instead, the user should be able to specify priorities relative to priorities of already defined operators. Something like:
```
infixl +;
infixl * after +;
infix == before +;
infix > with ==;
```

Perhaps it should also be possible to specify operator groups to reduce unnecessary inter-module dependencies.
```
operator group additive;
operator group multiplicative after additive;
operator group comparison before additive;

infixl + with additive;
infixl * with multiplicative;
infix == with comparison;
infix > with comparison;
```

lukaszcz:
* Closes #1964 

Adds the possibility to define operator fixities. They live in a separate namespace. Standard library defines a few in `Stdlib.Data.Fixity`:
```

syntax fixity rapp {arity: binary, assoc: right};
syntax fixity lapp {arity: binary, assoc: left, same: rapp};
syntax fixity seq {arity: binary, assoc: left, above: [lapp]};

syntax fixity functor {arity: binary, assoc: right};

syntax fixity logical {arity: binary, assoc: right, above: [seq]};
syntax fixity comparison {arity: binary, assoc: none, above: [logical]};

syntax fixity pair {arity: binary, assoc: right};
syntax fixity cons {arity: binary, assoc: right, above: [pair]};

syntax fixity step {arity: binary, assoc: right};
syntax fixity range {arity: binary, assoc: right, above: [step]};

syntax fixity additive {arity: binary, assoc: left, above: [comparison, range, cons]};
syntax fixity multiplicative {arity: binary, assoc: left, above: [additive]};

syntax fixity composition {arity: binary, assoc: right, above: [multiplicative]};
```

The fixities are identifiers in a separate namespace (different from symbol and module namespaces). They can be exported/imported and then used in operator declarations:
```
import Stdlib.Data.Fixity open;

syntax operator && logical;
syntax operator || logical;
syntax operator + additive;
syntax operator * multiplicative;
```


jonaprieto:
I found instances of `HasLoc` for `FixitySyntaxDef` and `OperatorSyntaxDef`. However, in the editor, I got this:

Screenshot 2023-08-09 at 09 51 38

What is the issue?

jonaprieto:
Please solve the conflicts and fix the following issues for fixity:
- Prohibit the use of duplicate keys, , e.g. `syntax fixity sub {..., same: "seq", same: "ladd"};` 
- Prohibit the use of keys of the same kind, such as `syntax fixity sub {..., same: "seq", above: "ladd"};`

lukaszcz:
> * Prohibit the use of duplicate keys, , e.g. `syntax fixity sub {..., same: "seq", same: "ladd"};`

It turns out JSON/Yaml allow duplicate keys and the behaviour is to silently ignore one of them 🤦🏻

I'll try to fix that, but I'm not sure if that's even possible with the current library interface.

> * Prohibit the use of keys of the same kind, such as `syntax fixity sub {..., same: "seq", above: "ladd"};`

This will give an error at scoping, but maybe the error message is not informative enough, so I can add a check at parsing stage.

lukaszcz:
> > * Prohibit the use of duplicate keys, , e.g. `syntax fixity sub {..., same: "seq", same: "ladd"};`
> 
> It turns out JSON/Yaml allow duplicate keys and the behaviour is to silently ignore all but one of them 🤦🏻
> 
> I'll try to fix that, but I'm not sure if that's even possible with the current library interface.
> 

I give up on this one (for now). There just seems no way to make the Yaml parsing library fail with duplicate keys, and from the API it seems one can access only one key even in functions like `forEachInObject`. I tried, didn't work. We would need to use a different library, or give up on Yaml.

> > * Prohibit the use of keys of the same kind, such as `syntax fixity sub {..., same: "seq", above: "ladd"};`
> 
> This will give an error at scoping, but maybe the error message is not informative enough, so I can add a check at parsing stage.

Fixed that.

paulcadman:
> I give up on this one (for now). There just seems no way to make the Yaml parsing library fail with duplicate keys, and from the API it seems one can access only one key even in functions like `forEachInObject`. I tried to create a list of all keys, didn't work (duplicates already removed by the library). We would need to use a different library, or give up on Yaml.

The `yaml` library can emit warnings about duplicate keys that can be handled at decode time. Unfortunately the API is only easily available for the [decodeFileWithWarnings](https://hackage.haskell.org/package/yaml-0.11.11.2/docs/Data-Yaml.html#g:2) API, not the ByteString decoding API.

We'd need the library to implement something like the following, for our use case (adapted from the implementation of `decodeEither'`, which interestingly uses `unsafePerformIO`).

```
import Text.Libyaml qualified as Y

decodeEitherWarning :: FromJSON a => ByteString -> Either ParseException ([Warning], a)
decodeEitherWarning = either Left splitEither
              . unsafePerformIO
              . decodeHelper
              . Y.decode
    where
      splitEither :: ([Warning], Either String a) -> Either ParseException ([Warning], a)
      splitEither (_, Left s) = Left (AesonException s)
      splitEither (ws, Right x) = Right (ws, x)
```

lukaszcz:
> > I give up on this one (for now). There just seems no way to make the Yaml parsing library fail with duplicate keys, and from the API it seems one can access only one key even in functions like `forEachInObject`. I tried to create a list of all keys, didn't work (duplicates already removed by the library). We would need to use a different library, or give up on Yaml.
> 
> The `yaml` library can emit warnings about duplicate keys that can be handled at decode time. Unfortunately the API is only easily available for the [decodeFileWithWarnings](https://hackage.haskell.org/package/yaml-0.11.11.2/docs/Data-Yaml.html#g:2) API, not the ByteString decoding API.
> 
> We'd need the library to implement something like the following, for our use case (adapted from the implementation of `decodeEither'`, which interestingly uses `unsafePerformIO`).
> 
> ```
> import Text.Libyaml qualified as Y
> 
> decodeEitherWarning :: FromJSON a => ByteString -> Either ParseException ([Warning], a)
> decodeEitherWarning = either Left splitEither
>               . unsafePerformIO
>               . decodeHelper
>               . Y.decode
>     where
>       splitEither :: ([Warning], Either String a) -> Either ParseException ([Warning], a)
>       splitEither (_, Left s) = Left (AesonException s)
>       splitEither (ws, Right x) = Right (ws, x)
> ```

Thanks! I'll take a look at this again then, but maybe we can raise this as a separate issue? This affects all places we use Yaml in: pragmas, iterators, fixities. I didn't go deep enough to try reimplementing stuff from Aeson.BetterErrors with Text.Libyaml, but this should be explored.

jonaprieto (approved):

PR #2278, opened 2023-08-08
Improve iterator and named arguments parsing

lukaszcz:
Avoid excessive backtracking in iterator and named arguments parsing. This also improves error messages by committing to a parsing branch as early as possible.

### DIFF
--- a/tests/positive/NamedArguments.juvix
+++ b/tests/positive/NamedArguments.juvix
@@ -89,5 +89,5 @@ module FakeRecord;
     mkPair (snd := unit; fst := Type);
 
   pp2 : Pair (B := Unit; A := Type) :=
-    mkPair (fst := Type) unit;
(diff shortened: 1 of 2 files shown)


tests/positive/NamedArguments.juvix
```diff
@@ -89,5 +89,5 @@ module FakeRecord;
     mkPair (snd := unit; fst := Type);
 
   pp2 : Pair (B := Unit; A := Type) :=
-    mkPair (fst := Type) unit;
+    mkPair (fst := Type) (unit);
```
jonaprieto, line 92: Why this change?

lukaszcz: To catch more errors. When it begins with `(` it can still be a range, you see it's not only later.

jonaprieto (approved):
LGTM

PR #2290, opened 2023-08-10
Error on duplicate keys in YAML

### LINKED ISSUE #2285: Disallow duplicate keys in YAML
Because by default the YAML parser silently ignores duplicate keys, they are not detected in the YAML associated with pragmas, iterators and fixities. It is not easy to change this behaviour.

@paulcadman has the following suggestion:

The `yaml` library can emit warnings about duplicate keys that can be handled at decode time. Unfortunately the API is only easily available for the [decodeFileWithWarnings](https://hackage.haskell.org/package/yaml-0.11.11.2/docs/Data-Yaml.html#g:2) API, not the ByteString decoding API.

We'd need the library to implement something like the following, for our use case (adapted from the implementation of `decodeEither'`, which interestingly uses `unsafePerformIO`).

```
import Text.Libyaml qualified as Y

decodeEitherWarning :: FromJSON a => ByteString -> Either ParseException ([Warning], a)
decodeEitherWarning = either Left splitEither
              . unsafePerformIO
              . decodeHelper
              . Y.decode
    where
      splitEither :: ([Warning], Either String a) -> Either ParseException ([Warning], a)
      splitEither (_, Left s) = Left (AesonException s)
      splitEither (ws, Right x) = Right (ws, x)
```

* See discussion in PR #2270

lukaszcz:
* Closes #2285

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -49,6 +49,7 @@ dependencies:
   - gitrev == 1.3.*
   - hashable == 1.4.*
   - language-c == 0.9.*
+  - libyaml == 0.1.*
   - megaparsec == 9.2.*
   - microlens-platform == 0.4.*
   - parser-combinators == 1.3.*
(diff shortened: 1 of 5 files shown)


paulcadman:
Perhaps add a negative test for one of the users of the yaml decoder?

lukaszcz:
> Perhaps add a negative test for one of the users of the yaml decoder?

Done

paulcadman (approved):

PR #2291, opened 2023-08-10
Allow omitting `: Type` in implicit function argument definitions and in type definitions

### LINKED ISSUE #2281: Allow omitting the `Type` keyword in type and function definitions
The arguments to a type definition are currently always types, and even when they can be something else this is still a reasonable default which saves keystrokes. So we could write
```
type List A := nil | cons A (List A);
```
instead of the more verbose
```
type List (A : Type) := nil | cons A (List A);
```

In function definitions, we could allow omitting `Type` for implicit arguments. So
```
map {A B} (f : A -> B) : List A -> List B := ...
```
would be equivalent to
```
map {A B : Type} (f : A -> B) : List A -> List B := ...
```
Since we always infer implicit `Type` when omitted, the notation is unambiguous and no information is lost. It's just a shorter syntax for the most common case.

It is tedious to write this `: Type` everywhere when programming polymorphically, and it doesn't provide any new information.

lukaszcz:
* Closes #2281

### DIFF
--- a/src/Juvix/Compiler/Concrete/Language.hs
+++ b/src/Juvix/Compiler/Concrete/Language.hs
@@ -329,12 +329,30 @@ data IteratorSyntaxDef = IteratorSyntaxDef
 instance HasLoc IteratorSyntaxDef where
   getLoc IteratorSyntaxDef {..} = getLoc _iterSyntaxKw <> getLoc _iterSymbol
 
+data SigArgRhs (s :: Stage) = SigArgRhs
+  { _sigArgColon :: Irrelevant KeywordRef,
+    _sigArgType :: ExpressionType s
+  }
+
+deriving stock instance Show (SigArgRhs 'Parsed)
+
+deriving stock instance Show (SigArgRhs 'Scoped)
+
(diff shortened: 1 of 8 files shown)


janmasrovira:
I think it is possible for these type variables to be "Type -> Type", etc. And in the future they may have other more complex types. Moreover, we decided to force explicit types in signatures as a way to document the function/type. I think this change goes against that

lukaszcz:
> I think it is possible for these type variables to be "Type -> Type", etc. And in the future they may have other more complex types. Moreover, we decided to force explicit types in signatures as a way to document the function/type. I think this change goes against that

I think we discussed this some time ago, and I don't remember what the final conclusion was, if any. Maybe we can make
```
f {A B} : A -> B
```
always mean
```
f {A B : Type} : A -> B
```
i.e. always put implicit `Type` instead of `_`.

Then there is no ambiguity or loss of information - just shorter syntax for the most common case.

janmasrovira (approved):

jonaprieto:
⛴️

PR #2295, opened 2023-08-11
Allow wildcard arguments in new function definition syntax

### LINKED ISSUE #2288: Allow wildcards for variable names in short function definitions
It should be possible to write:
```
f (a _ : Bool) : Bool := a;
```
Currently, the wildcard is rejected.

lukaszcz:
* Closes #2288


janmasrovira:
Can you give some argument why someone would want wildcards on the left of the colon?
At the moment I can't find a good reason

lukaszcz:
> Can you give some argument why someone would want wildcards on the left of the colon?
> At the moment I can't find a good reason

Anytime you want to ignore an argument? E.g.
```
k2 {A : Type} (_ x : A) : A := x;
```
It's not very common, but you do want to ignore an argument sometimes, and then using `_` is more natural than being forced to use some invented name.

jonaprieto (approved):

PR #2305, opened 2023-08-22
Remove old function syntax

### LINKED ISSUE #2251: Drop support for old function syntax
- #2243 has been merged without dropping support for old syntax. At some point we should remove it.

Tasks:

1. Update negative tests to use the new syntax.
2. Remove the relevant types and code from language, parser and scoper.
3. Update documentation to reflect the syntax change.

### LINKED ISSUE #2252: Drop support for `open import` syntax
At some point we should drop support for the old `open import` syntax.

lukaszcz:
* Enables new function syntax in local let-declarations
* Closes #2251


janmasrovira:
would it be possible without too much annoyance to separate these three points in separate prs?

lukaszcz:
At this point it would cause a lot of annoyance to separate these. Because I discovered that new syntax in lets is not enabled after going half-way through removing the old syntax. So it's all mixed together in the same commits, which can't be easily broken up.

lukaszcz:
The "drop `open import` syntax" could be separated because it's just a few lines difference, and the basic data structure didn't change.

lukaszcz:
I can isolate "drop `open import` syntax" into a separate PR, but it would need to depend on this PR, because it's too cumbersome to isolate all changes of `open import` in *.juvix files. Should I do it then?

janmasrovira:
> I can isolate "drop `open import` syntax" into a separate PR, but it would need to depend on this PR, because it's too cumbersome to isolate all changes of `open import` in *.juvix files. Should I do it then?

Let's do that

janmasrovira (approved):

PR #2314, opened 2023-08-25
Fix record pattern formatting

### LINKED ISSUE #2279: Formatting record patterns messes up spacing
Formatting
```
mf : Pair (Pair Bool (List Nat)) (List Nat) → Bool
  | mkPair@{fst := mkPair@{fst; snd := nil}; snd := zero :: _} := fst
  | _ := false;
```
gives
```
mf : Pair (Pair Bool (List Nat)) (List Nat) → Bool
  | mkPair@{fst := mkPair@{fst; snd := nil};
  snd := zero :: _} := fst
  | _ := false;
```

lukaszcz:
* Closes #2279

### DIFF
--- a/tests/Compilation/positive/test060.juvix
+++ b/tests/Compilation/positive/test060.juvix
@@ -17,21 +17,20 @@ type Pair (A B : Type) :=
     };
 
 mf : Pair (Pair Bool (List Nat)) (List Nat) → Bool
-  | mkPair@{fst := mkPair@{fst; snd := nil}; snd := zero :: _} := fst
+  | mkPair@{fst := mkPair@{fst; snd := nil};
+           snd := zero :: _} := fst
   | _ := false;
 
 main : Triple Nat Nat Nat :=
   let
     p : Triple Nat Nat Nat := mkTriple 2 2 2;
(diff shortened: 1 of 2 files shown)


tests/Compilation/positive/test060.juvix
```diff
@@ -17,21 +17,20 @@ type Pair (A B : Type) :=
     };
 
 mf : Pair (Pair Bool (List Nat)) (List Nat) → Bool
-  | mkPair@{fst := mkPair@{fst; snd := nil}; snd := zero :: _} := fst
+  | mkPair@{fst := mkPair@{fst; snd := nil};
+           snd := zero :: _} := fst
```
janmasrovira, line 21: can we refacotr the `align (braces items)` so that the second field is aligned with the first field instead of the opening `{`?

lukaszcz: Now it's consistent with how it works with named arguments. I think that if we were to change this, we should change both in a separate issue.

janmasrovira (dismissed):

janmasrovira (approved):

PR #2318, opened 2023-08-25
Report termination errors after typechecking

### LINKED ISSUE #2293: Incomprehensible error message when the number of patterns doesn't match the type
Typechecking the program
```
module tyomit;

type List A := nil | cons A (List A);

map {A B} (f : A → B) : List A → List B
  | nil := nil
  | cons h t := cons (f h) (map f t);
```
results is
```
/Users/heliax/Documents/progs/juvix/tyomit.juvix:5:1-4: error:
The function map fails the termination checker.
```
The error should be that the number of patterns doesn't match the type.

Maybe we should just check e.g. at scoping whether all clauses have the same number of pattern. For now, we don't support clauses with different numbers of patterns anyway.

### LINKED ISSUE #2319: Expressions are not checked for termination in the repl
E.g. typing the following in the repl causes it to loop:
```
let x : Type := x in 1
```

janmasrovira:
- Closes #2293.
- Closes #2319 

I've added an effect for termination. It keeps track of which functions failed the termination checker, which is run just after translating to Internal. During typechecking, non-terminating functions are not normalized. After typechecking, if there is at least one function which failed the termination checker, an error is reported.
Additionally, we now properly check for termination of functions defined in a let expression in the repl.

### DIFF
--- a/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Checker.hs
+++ b/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Checker.hs
@@ -6,26 +6,27 @@ module Juvix.Compiler.Internal.Translation.FromInternal.Analysis.Termination.Che
 where
 
 import Data.HashMap.Internal.Strict qualified as HashMap
+import Data.HashSet qualified as HashSet
 import Juvix.Compiler.Internal.Data.InfoTable as Internal
 import Juvix.Compiler.Internal.Language as Internal
+import Juvix.Compiler.Internal.Translation.FromConcrete.Data.Context (NonTerminating (..))
 import Juvix.Compiler.Internal.Translation.FromInternal.Analysis.FunctionCall
 import Juvix.Compiler.Internal.Translation.FromInternal.Analysis.Termination.Data
 import Juvix.Compiler.Internal.Translation.FromInternal.Analysis.Termination.Error
 import Juvix.Compiler.Internal.Translation.FromInternal.Analysis.Termination.LexOrder
 import Juvix.Prelude
-import Data.HashSet qualified as HashSet
 
 -- | Returns the set of non-terminating functions.
 checkTermination ::
   InfoTable ->
   Module ->
-  HashSet Name
+  NonTerminating
 checkTermination infotable topModule = do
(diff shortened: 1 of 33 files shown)


src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Data/TerminationState.hs
```diff
@@ -0,0 +1,52 @@
...
+import Juvix.Compiler.Internal.Language
+import Juvix.Prelude
+
+data IsTerminating
+  = -- | Has been checked for termination.
+    TerminatingChecked
+  | -- | Has been checked for termination but failed.
+    TerminatingFailed
+  | -- | Has been marked as terminating in the source code.
+    TerminatingMarked
```
jonaprieto, line 21: Remove this.

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Checker.hs
```diff
@@ -1,45 +1,111 @@
...
+  FunctionTermination :: FunctionRef -> Termination m IsTerminating
+
+makeSem ''Termination
+
+functionIsTerminating :: (Members '[Termination] r) => FunctionRef -> Sem r Bool
+functionIsTerminating = fmap terminates . functionTermination
+  where
+    terminates :: IsTerminating -> Bool
+    terminates = \case
+      TerminatingMarked -> True
```
jonaprieto, line 36: Remove this as well.

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Data/TerminationState.hs
janmasrovira: done

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Checker.hs
janmasrovira: done

jonaprieto (approved):

PR #2321, opened 2023-08-29
Add css colors `ju-comment` and `ju-judoc`

janmasrovira:
This pr adds colors for comments and judoc for the html backend

### DIFF
--- a/assets/css/source-ayu-light.css
+++ b/assets/css/source-ayu-light.css
@@ -31,6 +31,14 @@ body {
     color: #787b80;
 }
 
+.ju-comment {
+    color: #ba919966;
+}
+
+.ju-comment {
+    color: #8b2252;
+}
+
 .ju-var {
     color: #5c6166;
 }
--- a/assets/css/source-ayu-light.css
+++ b/assets/css/source-ayu-light.css
@@ -35,7 +35,7 @@ body {
     color: #ba919966;
 }
 
-.ju-comment {
+.ju-judoc {
     color: #8b2252;
 }
 
--- a/assets/css/source-nord.css
+++ b/assets/css/source-nord.css
@@ -39,11 +39,19 @@ body {
     color: #a4b4d2
 }
 
+.ju-comment {
+    color: #83898d
+}
+
+.ju-judoc {
+    color: #8fbcbb
+}
+
 .ju-number {
     color: #d8dee9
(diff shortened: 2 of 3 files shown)


assets/css/source-ayu-light.css
```diff
@@ -31,6 +31,14 @@ body {
     color: #787b80;
 }
 
+.ju-comment {
+    color: #ba919966;
+}
```
jonaprieto, line 36: Keep just one.

jonaprieto:
Idea. Perhaps we can think about generating these color schemes automatically.
They are simple enough, and during the process, we can make sure the CSS
stylesheets define all the colors we need, one for each face.  

Also, we could offer some customisation, perhaps by exposing
some optional fields in the package Yaml to redefine the default
colors and other info, like the title or footer.
For example, we could do something like:

```yaml
# juvix.yaml
...
html:
  theme: dark
  colors:
  - comment: #c3c3c3
  - delimiter: #F3f3F3
  ...
...
```

assets/css/source-ayu-light.css
janmasrovira: solved

janmasrovira:
> Idea. Perhaps we can think about generating these color schemes automatically. They are simple enough, and during the process, we can make sure the CSS stylesheets define all the colors we need, one for each face.

We can explore that, as it would offer more safety.

> Also, we could offer some customisation, perhaps by exposing some optional fields in the package Yaml to redefine the default colors and other info, like the title or footer. For example, we could do something like:
> 
> ```yaml
> # juvix.yaml
> ...
> html:
>   theme: dark
>   colors:
>   - comment: #c3c3c3
>   - delimiter: #F3f3F3
>   ...
> ...
> ```

I'm not 100% convinced about using the `juvix.yaml` file for that. Maybe we can offer the user to pass a custom css file as an argument to the cli. The `juvix.yaml` file could then have an optional field that points to that custom css file.
Either way, let's discuss this in a separate issue.

jonaprieto (approved):

jonaprieto:
I'll open an issue to continue this discussion later.

> using the `juvix.yaml` file for that.

Why not? this usage example is a particular case, the way I see it, of providing specific parameters/input to the backend target for the build process.

PR #2332, opened 2023-09-01
New fixity/iterator syntax

### LINKED ISSUE #2329: Fixity yaml not visible enough in the html source using a dark theme
image

### LINKED ISSUE #2330: Syntax change proposal for fixity declarations
Currently fixity declarations use yaml syntax. E.g:
```
syntax fixity additive {arity: binary, assoc: left, above: [comparison, range, cons]};
```
The advantage is that we can use an existing familiar syntax and parser.
The problem is that there is no currently available parser that has location information for the parsed items. This makes it impossible to properly highlight and give go-to information to the parsed identifiers.

We could write a custom yaml parser for the subset that we support in the fixity declarations. But the `yaml` syntax does not really play well with juvix syntax.
1. We use `;` instead of `,` to separate items.
2.  We support `:` as a valid character. So `assoc:` is a valid juvix identifier.

We could implement a custom lexer, but that would make the implementation slightly more involved. Also, having special lexical rules in some specific context could be confusing.

I propose an alternative record-inspired syntax. E.g.
```
syntax fixity rapp := binary {assoc := right; above := [lapp; rapp]; below := [rapp] }
syntax fixity rapp := unary {assoc := postfix; below := [x; y] }
```
In this syntax we have two "constructors": `binary` and `unary`. The fixity properties are given in the same way we create records. Note that for lists we use the builtin list syntax `[a;b;c]`.

janmasrovira:
- Closes #2330
- Closes #2329 

This pr implements the syntax changes described in #2330. It drops support for the old yaml-based syntax.
Some valid examples:
```
syntax iterator for {init := 1; range := 1};

syntax fixity cons := binary {assoc := right};
syntax fixity cmp := binary;
syntax fixity cmp := binary {}; -- debatable whether we want to accept empty {} or not. I think we should
```
# Future work
This pr creates an asymmetry between iterators and operators definitions. Iterators definition do not require a constructor. We could add it to make it homogeneous, but it looks a bit redundant:
```
syntax iterator for := mkIterator {init := 1; range := 1};
```

We could consider merging iterator and fixity declarations with this alternative syntax.
```
syntax XXX for := iterator {init := 1; range := 1};

syntax XXX cons := binary {assoc := right};
```
where `XXX` is a common keyword. Suggestion by @lukaszcz XXX = declare


lukaszcz:
The syntactical asymmetry between `iterator` and `fixity` is justified because `fixity` declares a new idenfitier and assigns fixity to it, while `iterator` declares an already existing identifier to be an iterator.

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -287,11 +292,49 @@ deriving stock instance (Ord (SyntaxDef 'Parsed))
 
 deriving stock instance (Ord (SyntaxDef 'Scoped))
 
+data ParsedFixityFields (s :: Stage) = ParsedFixityFields
+  { _fixityFieldsAssoc :: Maybe BinaryAssoc,
+    _fixityFieldsPrecSame :: Maybe (SymbolType s),
+    _fixityFieldsPrecBelow :: Maybe [SymbolType s],
+    _fixityFieldsPrecAbove :: Maybe [SymbolType s],
+    _fixityFieldsBraces :: Irrelevant (KeywordRef, KeywordRef)
+  }
```
jonaprieto, line 301: I'd prefer to include this data in ParsedFixityInfo directly.

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -1687,10 +1726,36 @@ makeLenses ''ArgumentBlock
...
+fixityAssoc = to (^? fixityFields . _Just . fixityFieldsAssoc . _Just)
+
+fixityPrecSame :: SimpleGetter (ParsedFixityInfo s) (Maybe (SymbolType s))
+fixityPrecSame = to (^? fixityFields . _Just . fixityFieldsPrecSame . _Just)
+
+fixityPrecAbove :: SimpleGetter (ParsedFixityInfo s) (Maybe [SymbolType s])
+fixityPrecAbove = to (^? fixityFields . _Just . fixityFieldsPrecAbove . _Just)
+
+fixityPrecBelow :: SimpleGetter (ParsedFixityInfo s) (Maybe [SymbolType s])
+fixityPrecBelow = to (^? fixityFields . _Just . fixityFieldsPrecBelow . _Just)
```
jonaprieto, line 1743: They all look pretty similar except for the inner function fixityFields...
Curious if you factorise this somehow.

tests/positive/Termination/Data/Nat.juvix
```diff
@@ -1,7 +1,7 @@
 module Data.Nat;
 
-syntax fixity add {arity: binary, assoc: left};
-syntax fixity mul {arity: binary, assoc: left, above: [add]};
+syntax fixity add := binary {assoc := left};
+syntax fixity mul := binary {assoc := left; above := [add]};
```
jonaprieto, line 4: Not specific to this example, but two concerns here:
Shouldn't the formatter refine the above list? e.g:

```diff
-syntax fixity mul := binary {assoc := left; above := [add;X;add]};
+syntax fixity mul := binary {assoc := left; above := [add;X]};
```

In case there is field duplication entry, e.g.:

```
syntax fixity seq := binary {assoc := left; assoc:=right};
```
 The error should say that explicitly, instead of:

```
X.juvix:6:57:
  |
6 | syntax fixity seq := binary {assoc := left; assoc:=right};
  |                                                         ^
expecting above, below, or same
```

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -294,14 +297,21 @@ statement = P.label "<top level statement>" $ do
...
+          <|> StatementInductive
+            <$> inductiveDef Nothing
+          <|> StatementModule
+            <$> moduleDef
+          <|> StatementAxiom
+            <$> axiomDef Nothing
           <|> builtinStatement
-          <|> StatementFunctionDef <$> functionDefinition Nothing
+          <|> StatementFunctionDef
+            <$> functionDefinition Nothing
```
lukaszcz, line 314: Why these newlines here? I'd say they make things less readable.

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -604,21 +691,36 @@ recordUpdate = do
...
+      <|> AtomFunArrow
+        <$> kw kwRightArrow
+      <|> AtomHole
+        <$> hole
+      <|> AtomParens
+        <$> parens parseExpressionAtoms
+      <|> AtomBraces
+        <$> withLoc (braces parseExpressionAtoms)
+      <|> AtomRecordUpdate
+        <$> recordUpdate
```
lukaszcz, line 723: As above, why this newline change?

src/Juvix/Compiler/Concrete/Keywords.hs
```diff
@@ -87,15 +98,3 @@ allKeywords =
...
--- purposes
-nonKeywords :: [Keyword]
-nonKeywords =
-  [ kwAs,
-    kwEq,
-    kwFixity,
-    kwOperator,
-    kwAlias,
-    kwIterator
-  ]
```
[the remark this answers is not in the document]
janmasrovira: it was dead code, so I removed it. Should I restore it and delete it in another pr?

src/Juvix/Compiler/Concrete/Keywords.hs
```diff
@@ -37,31 +41,38 @@ import Juvix.Data.Keyword.All
...
 import Juvix.Prelude
 
 allKeywordStrings :: HashSet Text
-allKeywordStrings = keywordsStrings allKeywords
+allKeywordStrings = keywordsStrings reservedKeywords
 
-allKeywords :: [Keyword]
-allKeywords =
+reservedKeywords :: [Keyword]
+reservedKeywords =
```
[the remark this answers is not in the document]
janmasrovira: the new name better reflects the meaning of the definition. All keywords that are exported but not in this list are not reserved

src/Juvix/Compiler/Backend/C/Translation/FromReg.hs
```diff
@@ -73,10 +73,10 @@ fromReg lims tab =
...
         getAssoc :: OperatorArity -> Text
         getAssoc = \case
-          Fixity.Unary _ -> "assoc_none"
-          Fixity.Binary AssocNone -> "assoc_none"
-          Fixity.Binary AssocLeft -> "assoc_left"
-          Fixity.Binary AssocRight -> "assoc_right"
+          Fixity.OpUnary _ -> "assoc_none"
+          Fixity.OpBinary AssocNone -> "assoc_none"
+          Fixity.OpBinary AssocLeft -> "assoc_left"
+          Fixity.OpBinary AssocRight -> "assoc_right"
```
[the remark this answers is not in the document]
janmasrovira: The constructor names were conficting with those of `Arity` (defined in `Juvix.Data.FixityInfo`).

src/Juvix/Compiler/Concrete/Language.hs
```diff
@@ -269,6 +267,13 @@ deriving stock instance (Ord (AliasDef 'Parsed))
 
 deriving stock instance (Ord (AliasDef 'Scoped))
 
+data ParsedIteratorInfo = ParsedIteratorInfo
+  { _piteratorInfoInitNum :: Maybe (WithLoc Int),
+    _piteratorInfoRangeNum :: Maybe (WithLoc Int),
+    _piteratorInfoBraces :: Irrelevant (KeywordRef, KeywordRef)
+  }
```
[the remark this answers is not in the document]
janmasrovira: done

src/Juvix/Compiler/Concrete/Language.hs
janmasrovira: This encoding is better because it reflects in the AST that if we do not skip the fields, we at least have a reference to the brackets.
If we inlined everything in ParsedFixityInfo we would have
```
...
_fixityFieldsPrecAbove :: Maybe [SymbolType s],
_fixityFieldsBraces :: Irrelevant (Maybe (KeywordRef, KeywordRef))
...
```
So the types are not telling that `_fixityFieldsBraces == Nothing` implies `_fixityFieldsPrecAbove == Nothing`.

Also, I'd say that semantically it makes sense to have it separated

src/Juvix/Compiler/Concrete/Language.hs
janmasrovira: done

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
janmasrovira: that was unintentional. If I run the formatter that comes with lsp these lines get introduced. I've reverted it

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
janmasrovira: idem

tests/positive/Termination/Data/Nat.juvix
janmasrovira: The simplest way to parse this is by using [Applciative permutation](https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/Control-Applicative-Permutations.html) + megaparsec. That's what I'm currently doing, so I don't have fine control over error messages. If we really wanted custom errors, we would need to parse this with a more general parser and check for repeated fields. In my opinion, this is not worth the necessary effort and extra code.

jonaprieto (approved):

PR #2337, opened 2023-09-05
Add Internet effect

janmasrovira:
- This pr adds the `Internet` effect. The type of a function can inform whether that function requires internet connection. Moreover, now that we have the `--offline` flag (#2335), this effect will be part of the pipeline so that it can be accessed in any of the steps without having to pass some explicit argument such as `GitProcessMode`.


jonaprieto (approved):
LGTM

PR #2350, opened 2023-09-08
Fix instance import

lukaszcz:
Fixes a bug which prevented some instances from being imported from other modules.

### DIFF
--- a/src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
+++ b/src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
@@ -27,6 +27,10 @@ data InstanceInfo = InstanceInfo
     _instanceInfoResult :: Expression,
     _instanceInfoArgs :: [FunctionParameter]
   }
+  deriving stock (Eq)
+
+instance Hashable InstanceInfo where
+  hashWithSalt salt InstanceInfo {..} = hashWithSalt salt _instanceInfoResult
 
 -- | Maps trait names to available instances
 newtype InstanceTable = InstanceTable
@@ -40,7 +44,10 @@ makeLenses ''InstanceTable
 instance Semigroup InstanceTable where
   t1 <> t2 =
     InstanceTable $
-      HashMap.unionWith (++) (t1 ^. instanceTableMap) (t2 ^. instanceTableMap)
+      HashMap.unionWith combine (t1 ^. instanceTableMap) (t2 ^. instanceTableMap)
+    where
+      combine :: [InstanceInfo] -> [InstanceInfo] -> [InstanceInfo]
+      combine ii1 ii2 = nubHashable (ii1 ++ ii2)
 
 instance Monoid InstanceTable where
   mempty = InstanceTable mempty
--- a/src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
+++ b/src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
@@ -29,19 +29,33 @@ data InstanceInfo = InstanceInfo
   }
 
 -- | Maps trait names to available instances
-type InstanceTable = HashMap InductiveName [InstanceInfo]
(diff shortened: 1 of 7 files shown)


src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
```diff
@@ -29,20 +29,34 @@ data InstanceInfo = InstanceInfo
...
+  }
 
 makeLenses ''InstanceApp
 makeLenses ''InstanceInfo
+makeLenses ''InstanceTable
+
+instance Semigroup InstanceTable where
+  t1 <> t2 =
+    InstanceTable $
+      HashMap.unionWith (++) (t1 ^. instanceTableMap) (t2 ^. instanceTableMap)
```
janmasrovira, line 43: Should we handle deletion of duplicates or is it guaranteed that these two lists are disjoint?

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
```diff
@@ -29,20 +29,34 @@ data InstanceInfo = InstanceInfo
...
-type InstanceTable = HashMap InductiveName [InstanceInfo]
+newtype InstanceTable = InstanceTable
+  { _instanceTableMap :: HashMap InductiveName [InstanceInfo]
+  }
 
 makeLenses ''InstanceApp
 makeLenses ''InstanceInfo
+makeLenses ''InstanceTable
+
+instance Semigroup InstanceTable where
```
janmasrovira, line 40: You can derive semigroup and monoid using `deriving newtype (Monoid, Semigroup)`.
I am not saying you should, just a remainder

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
lukaszcz: I think they will be disjoint - this is invoked when we're combining things from different modules

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
lukaszcz: But the whole point that is we want different behaviour for Semigroup?

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
janmasrovira: If I open the same module twice I'm getting an ambiguity error.

`Main.juvix`
```
module Main;

import Default open;
import Default open;

x : Unit := default;
```

Default.jvuix
```
module Default;

type Unit := unit;

trait
type Default A := mkDefault { default : A };

instance
Unit-Default : Default Unit := mkDefault unit;

default {A : Type} {{x : Default A}} : A := case x
 | mkDefault a := a;
```

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
janmasrovira: ah yes! The semigroup instance for map just replaces the contents

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
lukaszcz: Ah, Ok, then I'll add filtering duplicates

lukaszcz: done

lukaszcz:
This requires more work. It seems also something is wrong with the dependencyBuilder, because with imports some instance identifiers are wrongly filtered out.

src/Juvix/Compiler/Internal/Data/InstanceInfo.hs
janmasrovira: did you push the changes? I can't see it in the diff

janmasrovira (approved):

PR #2351, opened 2023-09-08
Support more general inductive parameters in the type checker

janmasrovira:
This pr allows inductive type parameters to be any type. Until now, they had to be exactly `Type`. This allows us to define more general traits such as the `Monad` and `Functor`, as shown in the new test.
I have updated the aritychecker and the typechecker.

Pending work:
1. Update the positivity checker if necessary (@jonaprieto).
2. Update the necessary compilation steps in Core (@lukaszcz). 
3. Add compilation tests.

I think it is fine to merge the pending work as separate prs.


janmasrovira:
superseded by #2506

Issue #2246, opened 2023-07-07
Implement `Case` syntax changes


janmasrovira:
After 
- #2189

it was agreed that `case` expressions should have the form:

```
case atom {
 | pat := expr
 ..
 | pat := expr;
}
```

PR #2353, opened 2023-09-12
Add new case syntax

paulcadman:
## New case syntax

This PR adds a new syntax for case:

```
case <expression> of { 
  | <pattern> := <expression>
  | ...
}
```

The first pipe is optional.

## Formatter

The formatter:
* Removes the pipe if there's only one branch
* Adds a pipe before the first branch if there are multiple branches.

Additionally the formatter will convert the old syntax for case to the new syntax.

The old case syntax is still supported. This is so that we can migrate the standard library and other projects easily. After we have done this we can remove the old syntax.

## Implementation differences

The implementation differs from that specified in the issue. This is because in the proposed syntax, you can write something like:

```
case x { y := z }
```

The `x { y := z }` part looks like (to a human and to the Juvix parser) like a named application. Therefore we add an `of` keyword between the subject of the case and the braced list of branches. This also allows any expression to be the subject instead of just an expression atom as was originally proposed. 

- Closes https://github.com/anoma/juvix/issues/2246


src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -1808,6 +1808,28 @@ checkCaseBranch CaseBranch {..} = withLocalScope $ do
...
+    NewCaseBranch
+      { _newCaseBranchPattern = pattern',
+        _newCaseBranchExpression = expression',
+        ..
+      }
+  where
+    checkNotImplicit :: PatternArg -> Sem r ()
+    checkNotImplicit p =
+      when
+        (p ^. patternArgIsImplicit == Implicit)
```
janmasrovira, line 1830: I think we are missing `ImplicitInstance`

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -1808,6 +1808,28 @@ checkCaseBranch CaseBranch {..} = withLocalScope $ do
...
+        (throw (ErrCaseBranchImplicitPattern (CaseBranchImplicitPattern p)))
```
janmasrovira, line 1831: I see that we have no negative test for this error. However, I'm not sure if this error should exist, since this will be caught by the typechecker anyway. What do you think?

paulcadman: For me it depends on how good the type checker error is. 

Do you know how do you trigger this error?

janmasrovira: The error message for
```
   case b of {
      | {{true}} := false
```
is 
```
Expected an explicit pattern but found an implicit instance pattern: {{true}}
```

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
janmasrovira: :ok_hand:

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
paulcadman: I agree that this error should not exist as it duplicates the error thrown by the arity checker

janmasrovira (approved):

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
paulcadman: Raised the PR in https://github.com/anoma/juvix/pull/2357

paulcadman:
Let's merge https://github.com/anoma/juvix/pull/2357 and then fix this PR before merging

paulcadman:
https://github.com/anoma/juvix/pull/2357 is now merged, so we're ready to merge this one.

janmasrovira (approved):

PR #2369, opened 2023-09-14
Record creation syntax with function definitions

### LINKED ISSUE #2280: Record creation syntax analogous to record patterns
There is some asymmetry between record creation syntax and other record-related syntax (update, patterns). Only record creation uses parentheses instead of braces. This will look not-so-good when we have large record creations (e.g. with type class instances) spanning multiple lines - braces are better for enclosing large chunks of code.

I propose to add syntactic sugar to make record creation look the same as record pattern matching:
```
mkRecord@{
  field1 := value1;
  field2 := value2;
  field3 := value3;
}
```
would desugar to the same thing as
```
mkRecord (
  field1 := value1;
  field2 := value2;
  field3 := value3;
)
```

lukaszcz:
* Closes #2280 
* Record creation syntax uses normal function definition syntax like at the top-level or in lets.
* It is now allowed to omit the result type annotation in function definitions (the `: ResultType` part) with `_` inserted by default. This is allowed only for simple definitions of the form `x := value` in lets and record creation, but not at the top level.


src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -793,7 +794,7 @@ checkFunctionDef FunctionDef {..} = do
   sigDoc' <- mapM checkJudoc _signDoc
   (args', sigType', sigBody') <- withLocalScope $ do
     a' <- mapM checkArg _signArgs
-    t' <- checkParseExpressionAtoms _signRetType
+    t' <- maybe (return Nothing) (fmap Just . checkParseExpressionAtoms) _signRetType
```
janmasrovira, line 797:
```suggestion
    t' <- mapM checkParseExpressionAtoms _signRetType
```

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -2086,16 +2088,70 @@ checkExpressionAtom e = case e of
...
+        throw (ErrMissingFields (MissingFields (cname ^. nameConcrete) missingFields))
+      return
+        RecordCreation
+          { _recordCreationConstructor = constrIden,
+            _recordCreationFields = fields',
+            _recordCreationExtra = Irrelevant extra',
+            _recordCreationAtKw
+          }
+    Nothing ->
+      throw (ErrNotAConstructor (NotAConstructor (cname ^. nameConcrete)))
```
janmasrovira, line 2128: move the `Nothing` case before the `Just`. Otherwise it is a bit hard to read

src/Juvix/Compiler/Internal/Translation/FromConcrete.hs
```diff
@@ -761,12 +770,56 @@ goExpression = \case
...
+            _namedArgValue = Concrete.ExpressionIdentifier _fieldDefineIden
+          }
+      where
+        dummyAssignKw :: KeywordRef
+        dummyAssignKw =
+          KeywordRef
+            { _keywordRefKeyword = asciiKw ":=",
+              _keywordRefInterval = getLoc _fieldDefineFunDef,
+              _keywordRefUnicode = Ascii
+            }
```
janmasrovira, line 821: Use `Juvix.Compiler.Concrete.Gen.kw`.
E.g. `runReader (getLoc _fieldDefineFunDef) (Gen.kw Gen.kwAssign)`

lukaszcz: done

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
lukaszcz: done

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
lukaszcz: done

janmasrovira (approved):

PR #2382, opened 2023-09-22
Update standard library list functions to use traits

lukaszcz:


src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Traits/Resolver.hs
```diff
@@ -143,7 +143,7 @@ lookupInstance tab ty = do
     Just InstanceApp {..} ->
       lookupInstance' True tab _instanceAppHead _instanceAppArgs
     _ ->
-      throw (ErrNotATrait (NotATrait ty))
+      return []
```
paulcadman, line 146: was this committed by mistake?

lukaszcz: No, it's intentional. You will get an error earlier when checking instance types anyway, and this was causing some spurious errors with function parameters.

paulcadman (approved):

PR #2417, opened 2023-09-29
Improve the `specialise` pragma

### LINKED ISSUE #2401: Allow to specify arguments of the `specialise` pragma by name

### LINKED ISSUE #2416: Allow specialising instance arguments
So that we could write e.g.
```
{-# specialize: [1] #-}
merge {A} {{Ord A}} : List A → List A → List A
  | ...
```
and `merge` would get specialised by the instance argument of type `Ord A`. Currently, only explicit arguments can be specialised.

We should also be able to write
```
{-# specialize: true #-}
trait
type Natural N := mkNatural {
  + : N -> N -> N;
  * : N -> N -> N;
  fromNat : Nat -> N;
};
```
And then any instance argument of type `Natural N` would get specialised by default.

It should also be possible to indicate that only specific instances should be used for specialisation:
```
{-# specialize: true #-}
instance
naturalNatI : Natural Nat := ...
```

Once we have this, we can implement traits for numbers and use the associated operations without any loss in efficiency.

lukaszcz:
* Closes #2416 
* Closes #2401 

Work in progress


src/Juvix/Compiler/Core/Transformation/Optimize/Phase/Main.hs
```diff
@@ -17,14 +17,25 @@ optimize' CoreOptions {..} tab =
...
     $ tab
+  where
+    recs :: HashSet Symbol
+    recs = recursiveIdents tab
+
+    doInlining :: InfoTable -> InfoTable
+    doInlining tab' = inlining' _optInliningDepth recs' tab'
+      where
+        recs' =
+          if
```
janmasrovira, line 36: use guard

janmasrovira (approved):

PR #2433, opened 2023-10-05
Numeric, ordering and equality traits

lukaszcz:
* Adapts to https://github.com/anoma/juvix-stdlib/pull/86
* Adds a pass in `toEvalTransformations` to automatically inline all record projection functions, regardless of the optimization level. This is necessary to ensure that arithmetic operations and comparisons on `Nat` or `Int` are always represented directly with the corresponding built-in Core functions. This is generally highly desirable and required for the Geb target.
* Adds the `inline: always` pragma which indicates that a function should always be inlined during the mandatory inlining phase, regardless of optimization level.

### DIFF
--- a/src/Juvix/Compiler/Core/Evaluator.hs
+++ b/src/Juvix/Compiler/Core/Evaluator.hs
@@ -89,7 +89,7 @@ geval opts herr ctx env0 = eval' env0
           Closure env' (NLam (Lambda i' bi b)) ->
             let !v = eval' env r in evalBody i' bi env' v b
           lv
-            | opts ^. evalOptionsNormalize ->
+            | opts ^. evalOptionsNormalize || opts ^. evalOptionsNoFailure ->
                 let !v = eval' env r in goNormApp i lv v
             | otherwise ->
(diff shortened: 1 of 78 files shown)


src/Juvix/Compiler/Core/Evaluator.hs
```diff
@@ -106,7 +106,7 @@ geval opts herr ctx env0 = eval' env0
           NCtr (Constr _ tag args) ->
             branch n env args tag def bs
           v'
-            | opts ^. evalOptionsNormalize ->
+            | opts ^. evalOptionsNormalize || opts ^. evalOptionsNoFailure ->
```
paulcadman, line 109: question out of curiosity: What is the purpose of `opts ^. evalOptionsNoFailure` - is it required by a particular backend?

paulcadman (approved):

src/Juvix/Compiler/Core/Evaluator.hs
paulcadman: Thanks!

Issue #2336, opened 2023-09-01
Proposal: Use a Juvix source file to define a package instead of YAML


paulcadman:
This issue is for discussion of the idea to replace the current juvix.yaml package file format with a Juvix source file.

Some advantages of using Juvix instead of YAML:

1. Users could take advantage of Juvix tooling when managing configuration (e.g when debugging missing/misspelled fields)
2. Typed configuration (user would import a module with the types / records used in configuration)
3. Line / column information when parsing (this is useful when reporting dependency errors and is lacking in the Haskell yaml parsers)

# Prior art

Some languages use language source code to define builds/packages already:

Lean4 / lake
https://github.com/leanprover/lake/blob/a2ced44b7e5e30c2a6b84b420e1bbdd8d6d8e079/examples/targets/lakefile.lean

Swift / swift package manager
https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html

Kotlin / gradle
https://docs.gradle.org/current/userguide/kotlin_dsl.html

lukaszcz:
I'm in favour of the idea. Would make it easier to write (in Juvix) tools manipulating Juvix packages and to do meta-programming with package information later (no need for YAML parser in Juvix).

janmasrovira:
:+1:

jonaprieto:
Perhaps something like this:

```
-- Package.juvix

module MyPackage;

import App;

examplePackage := mkPackage {
  name := "MyPackage";
  version := [ 0 ; 3 ; 1 ];
  main := App.main;
  dependencies: [
    ( "stdlib" , [ 0 ; 1 ; 0 ]);
    ("juvix-con , [ 0 ; 0 ; 1 ]);
};

```

paulcadman:
Example package / dependency type. Version String would be parsed when evaluated.

```
module Package;

type Dependency :=
  | pathDependency {path : String}
  | gitDependency {
      url : String;
      name : String;
      ref : String
    };

type Package :=
  mkPackage {
    name : String;
    version : String;
    dependencies : List Dependency;
    main : Maybe String;
    buildDir : Maybe String
  };

package : Package :=
  mkPackage@{
    name := "quickcheck";
    version := "0.7.1";
    main := just "Example.juvix";
    buildDir := nothing;
    dependencies :=
      [ pathDependency (
        path := ".juvix-build/stdlib/"
        )
      ; gitDependency (
          url := "https://github.com/anoma/juvix-stdlib";
          name := "stdlib";
[...]

PR #2434, opened 2023-10-05
Use a Juvix source file to define a package

paulcadman:
Depends on:
* ~~https://github.com/anoma/juvix/pull/2459~~
* https://github.com/anoma/juvix/pull/2462

This PR is part of a series implementing:
* https://github.com/anoma/juvix/issues/2336

This PR adds the package file loading function, including a file evaluation effect. It integrates this with the existing `readPackage` function and adds tests / smoke tests.

## Package.juvix format

Instead of `juvix.yaml` (which is still supported currently) users can now place a `Package.juvix` file in the root of their project. The simplest `Package.juvix` file you can write is:

```
module Package;

import PackageDescription open;

```

The [PackageDescription](https://github.com/anoma/juvix/blob/35b2f618f093895f32929ac0f2c0affcdab8d627/include/package/PackageDescription.juvix) module defines the `Package` type. Users can use "go-to definition" in their IDE from the Package file to see the documentation and definitions.

Users may also import `Stdlib.Prelude` in their Package file. This is loaded from the global project. No other module imports are supported.

Notes:
* If a directory contains both `Package.juvix` and `juvix.yaml` then `Package.juvix` is used in preference.

## Default stdlib dependency

The `Dependency` type has a constructor called `defaultStdlib`. This means that any project can use the compiler builtin standard library dependency. With `juvix.yaml` this dependency is only available when the `dependencies` field is unspecified.

```
module Package;

import PackageDescription open;

package : Package := defaultPackage { dependencies := [defaultStdlib] };
```

## Validation

As well as the standard type checking validation that the Juvix compiler provides additional validation is made on the file.

* The Package module must contain the identifier `package` and it must have type `Package` that's obtained from the global `PackageDescription` module.
* Every dependency specified in the Package.juvix must be unique.

* Closes https://github.com/anoma/juvix/issues/2336

### Package with name and version

```
module Package;

import PackageDescription open;

package : Package :=
  defaultPackage {name := "a-package";
                  version := mkVersion 0 1 0};
```

### Package with GitHub dependency

```
module Package;

import PackageDescription open;

package : Package :=
  defaultPackage {name := "a-package";
                  version := mkVersion 0 1 0;
                  dependencies := [defaultStdlib;
                                   github (org := "anoma";
                                           repo := "juvix-containers";
                                           ref := "v0.7.1")]};
```

## Package with main and buildDir fields

```
module Package;

import Stdlib.Prelude open;
import PackageDescription open;

package : Package :=
  defaultPackage {name := "a-package";
                  version := mkVersion 0 1 0;
                  dependencies := [defaultStdlib;
                                   github (org := "anoma";
                                           repo := "juvix-containers";
                                           ref := "v0.7.1")];
                  buildDir := just "/tmp/build";
                  main := just "HelloWorld.juvix"
                };

```


janmasrovira (approved):
:motor_boat:

PR #2442, opened 2023-10-10
Allow `syntax operator` in the rhs of a record declaration

### LINKED ISSUE #2431: Allow operator definitions for record projections
This doesn't currently work.

```
syntax operator + additive;
syntax operator * multiplicative;

trait
type Natural A :=
  mkNatural {
    + : A -> A -> A;
    * : A -> A -> A;
  };
```

The problem seems to be that `+` and `-` are implicitly defined in a local module `Natural`. Perhaps we should make local modules inherit the syntax declarations from the enclosing scope?

janmasrovira:
- Closes #2431.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
+++ b/src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
@@ -954,9 +954,12 @@ checkInductiveDef InductiveDef {..} = do
         checkRecordStatements :: [RecordStatement 'Parsed] -> Sem r [RecordStatement 'Scoped]
         checkRecordStatements = \case
           [] -> return []
-          RecordStatementField f : fs -> do
-            f' <- checkField f
-            (RecordStatementField f' :) <$> checkRecordStatements fs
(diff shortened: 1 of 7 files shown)


src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -944,27 +944,35 @@ checkInductiveDef InductiveDef {..} = do
...
+              (RecordStatementOperator d :) <$> checkRecordStatements fs
+            RecordStatementField d -> do
+              d' <- checkField d
+              (RecordStatementField d' :) <$> checkRecordStatements fs
+
+        checkField :: RecordField 'Parsed -> Sem r (RecordField 'Scoped)
+        checkField RecordField {..} = do
+          type' <- checkParseExpressionAtoms _fieldType
+          -- TODO local scope??
+          withLocalScope $ do
```
lukaszcz, line 968: I don't understand the comment. We have `withLocalScope` here. Either the comment should be made more clear or removed

lukaszcz (approved):

PR #2444, opened 2023-10-13
Instance coercions

### LINKED ISSUE #2426: Trait inheritance
It should be possible to inherit traits, so that their instances are automatically and implicitly coerced to the ancestor traits when needed. 

A separate but closely related issue is to be able to declare record fields as "implicit instance" with double braces, so that when creating an instance the parent instances can be automatically inferred. This can already be achieved by specifying the constructor type signature in full
```
mkIntegral : {{natural : Natural N}} -> (- : N -> N -> N) -> (fromInt : Int -> N) -> Integral N
```
but one cannot yet use double braces with the record declaration syntax.

I provisionally propose the following syntax.

```
trait
type Natural N := mkNatural {
  + : N -> N -> N;
  * : N -> N -> N;
  fromNat : Nat -> N;
};

trait
type Integral N := mkIntegral {
  inherit {{natural : Natural N}};
  - : N -> N -> N;
  fromInt : Int -> N;
};

instance
naturalNatI : Natural Nat := mkNatural@{
  + := (Nat.+);
  * := (Nat.*);
  fromNat := id;
};

instance
naturalIntI : Natural Int := mkNatural@{
  + := (Int.+);
  * := (Int.*);
  fromNat := Int.ofNat;
};

instance
integralIntI : Integral Int := mkIntegral@{
  -- `natural := naturalIntI` is inferred here because of the double braces 
  -- in the declaration of the `natural` field
  - := (Int.-);
  fromInt := id;
};

calculate {A} {{Integral A}} (x y : A) : A := x + y - 7;
-- in `x + y` the instance `Integral.natural integralIntI = naturalIntI` is automatically inferred, 
-- because we have an implicit coercion `Integral.natural` from `Integral` to `Natural` 
-- declared with `inherit`
```

We could also opt for a more general coercion mechanism. Then perhaps the `coercion` keyword would be more appropriate than `inherit`. Perhaps it should then go inside the double braces as a keyword associated with the field/function?
```
trait
type Integral N := mkIntegral {
  {{coercion natural : Natural N}};
  - : N -> N -> N;
  fromInt : Int -> N;
};
```

lukaszcz:
* Closes #2426 

Work in progress

### DIFF
--- a/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Checker.hs
+++ b/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Checker.hs
@@ -197,6 +197,7 @@ checkFunctionDef FunctionDef {..} = do
           _funDefName,
           _funDefTerminating,
           _funDefInstance,
+          _funDefCoercion,
           _funDefBuiltin,
           _funDefPragmas
         }
@@ -242,7 +243,7 @@ checkInstanceType FunctionDef {..} = case mi of
     tab <- ask
     unless (isTrait tab _instanceInfoInductive) $
       throw (ErrTargetNotATrait (TargetNotATrait _funDefType))
-    is <- map fst <$> subsumingInstances (tab ^. infoInstances) ii
+    is <- subsumingInstances (tab ^. infoInstances) ii
     unless (null is) $
       throw (ErrSubsumedInstance (SubsumedInstance ii is (getLoc _funDefName)))
     let metaVars = HashSet.fromList $ mapMaybe (^. paramName) _instanceInfoArgs
--- a/src/Juvix/Compiler/Concrete/Keywords.hs
+++ b/src/Juvix/Compiler/Concrete/Keywords.hs
@@ -34,6 +34,7 @@ import Juvix.Data.Keyword.All
     kwBracketR,
     kwBuiltin,
     kwCase,
+    kwCoercion,
     kwColon,
     kwEnd,
     kwEq,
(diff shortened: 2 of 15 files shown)


src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Checker.hs
```diff
@@ -36,6 +37,15 @@ registerNameIdType uid ty = do
...
   modify (set (highlightTypes . at uid) (Just ty))
 
+checkTable ::
+  (Members '[Reader InfoTable, Error TypeCheckerError] r) =>
+  Sem r ()
+checkTable = do
+  tab <- ask
+  let s = toList $ cyclicCoercions (tab ^. infoCoercions)
+  unless (null s) $
+    throw (ErrCoercionCycles (CoercionCycles (nonEmpty' s)))
```
paulcadman, line 47:
```suggestion
  whenJust (nonEmpty s)
    $ throw
    . ErrCoercionCycles
    . CoercionCycles
```

paulcadman (dismissed):

tests/negative/Internal/LoopingCoercion.juvix
```diff
@@ -0,0 +1,21 @@
...
+
+trait
+type T2 A := mkT2 {pp : A -> A};
+
+coercion
+fromT1toT2 {A} {{T1 A}} : T2 A := mkT2@{
+  pp := T1.pp
+};
+
+coercion
```
paulcadman, line 16: I think the negative tests need to. be updated to use the new keyword and then we can merge this.

lukaszcz: Yes, thanks, I forgot about them

paulcadman (approved):

PR #2446, opened 2023-10-17
Dependent default values

### LINKED ISSUE #2445: Allow default values to refer to previous arguments
e- Currently default values can only refer to names available at the top scope (i.e. the scope where the function/constructor being defined is).
Sometimes it can be useful to refer to previos arguments. For instance consider the example given in #2427:
```
trait
type Ord A := mkOrd {
  cmp : A -> A -> Ordering
  < (x y : A) : Bool := case cmp x y of {
    | LT := true
    | _ := false
  };
  ..
};
```
In the default definition of `<` we need to refer to `cmp`.

## Implementation
We briefly discussed online one possible way to implement this: If we have:
```
f {a : Nat := 0} {b : Nat := a + 1}  {c : Nat := a + b} := ...
```
Then if we find `f` we should generate the following:
```
let a := 0
     b := a + 1
     c := a + b
in f a b c;
```

Note that default arguments happens in `FromConcrete.NamedArguments` and in the aritychecker.
So if we have `f {b := 1}`, then the NamedArguments desugarer will generate
```
let a := _
     b := 1
in f {a} {b}
```
And the arity checker will introduce the final argument:
```
let a := _
     b := 1
in 
  let 
    c := a + b
   in  f {a} {b} {c}
```

janmasrovira:
- Closes #2445


janmasrovira:
It is possible to enter an infinite loop when inserting default arguments.

E.g.
```
fun {a : Nat := fun} : Nat := a * a;

main : Nat := fun;
```
In order to prevent this problem, we need dependency information for default arguments before translating to Internal. Because I didn't want to delay this pr further, so I'll work on a patch separately.

janmasrovira:
> It is possible to enter an infinite loop when inserting default arguments.
> 
> E.g.
> 
> ```
> fun {a : Nat := fun} : Nat := a * a;
> 
> main : Nat := fun;
> ```
> 
> In order to prevent this problem, we need dependency information for default arguments before translating to Internal. I didn't want to delay this pr further, so I'll work on a patch separately. Also, this is a problem with default arguments in general, not only for dependent default values.

this has been fixed in [e886aa2](https://github.com/anoma/juvix/pull/2446/commits/e886aa219e1b603e34cef11f9e2c8fc1520b0173)

src/Juvix/Compiler/Internal/Extra/Clonable.hs
```diff
@@ -0,0 +1,260 @@
...
+    underBinder fun $ \fun' -> do
+      body' <- freshNameIds _funDefBody
+      defaultSig' <- freshNameIds _funDefArgsInfo
+      return
+        FunctionDef
+          { _funDefName = fun' ^. funDefName,
+            _funDefType = ty',
+            _funDefBody = body',
+            _funDefArgsInfo = defaultSig',
+            ..
```
lukaszcz, line 259: I think the `..` is dangerous here. One needs to remember that all `Expression` fields need to be refreshed here. It's easy to forget when one adds a new field, and with `..` the compiler won't give an error but silently use the original non-refreshed field.

src/Juvix/Compiler/Internal/Extra/Clonable.hs
```diff
@@ -0,0 +1,260 @@
...
+            _letExpression = e'
+          }
+
+instance Clonable SimpleBinder where
+  freshNameIds SimpleBinder {..} = do
+    ty' <- freshNameIds _sbinderType
+    return
+      SimpleBinder
+        { _sbinderType = ty',
+          ..
```
lukaszcz, line 188: Potentially dangerous `..`

src/Juvix/Compiler/Internal/Extra/Clonable.hs
```diff
@@ -0,0 +1,260 @@
...
+
+instance Clonable Application where
+  freshNameIds Application {..} = do
+    l' <- freshNameIds _appLeft
+    r' <- freshNameIds _appRight
+    return
+      Application
+        { _appLeft = l',
+          _appRight = r',
+          ..
```
lukaszcz, line 45: Potentially dangerous `..`

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Data/Inference.hs
```diff
@@ -166,7 +214,7 @@ strongNormalize' = go
     goApp (Application l r i) = do
       l' <- go l
       case l' of
-        ExpressionSimpleLambda (SimpleLambda lamVar _ lamBody) -> do
+        ExpressionSimpleLambda (SimpleLambda (SimpleBinder lamVar _) lamBody) -> do
           go (substitutionE (HashMap.singleton lamVar r) lamBody)
```
lukaszcz, line 218: I'm not sure if this is correct. What if `r` contains some binders, e.g., lets or lambdas? They could be duplicated by the substitution, which would violate the assumption that name ids are unique for different binders. I think either there should be an explicit assumption here that `r` contains no binders (which should be guaranteed / checked somehow) or the bound variables in `r` should be refreshed.

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Data/Inference.hs
```diff
@@ -213,7 +261,7 @@ weakNormalize' = go
...
```
lukaszcz, line 265: As above. I think this is not entirely correct.

lukaszcz: BTW, I can't find any place in the pipeline where `SimpleLambda` is created. Isn't this dead code which should be removed?

lukaszcz (changes requested):
Address the comments. I think `SimpleLambda` can be removed, but not completely sure.

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Data/Inference.hs
janmasrovira: > BTW, I can't find any place in the pipeline where `SimpleLambda` is created. Isn't this dead code which should be removed?

It is used in `Inference.functionDefEval`. 
`SimpleLambda` represents the subset of lambdas that we can normalize during inference (basically type synonyms).

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Data/Inference.hs
janmasrovira: Your are right, I've had this on the back on my mind for some time, but since it wasn't causing problems I didn't change it. Probably because SimpleLamba is only used for type synonyms, and we haven't used nested type synonyms applied to arguments with binders.
However, this pr does not change the behaviour of this, I merely adopted the syntax to use the new SimpleBinder constructor.

src/Juvix/Compiler/Internal/Extra/Clonable.hs
janmasrovira: wildcard removed

src/Juvix/Compiler/Internal/Extra/Clonable.hs
janmasrovira: wildcard removed

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/TypeChecking/Data/Inference.hs
lukaszcz: Okay, let's do it in a separate PR

lukaszcz (approved):

PR #2450, opened 2023-10-17
Constant folding

### LINKED ISSUE #2154: Constant folding
Functions whose all arguments are values can simply be evaluated with the Core evaluator and the calls replaced with the obtained values. This optimisation would affect all backends.

Perhaps we might want to restrict this to non-recursive functions, or those whose termination has been proven and not just declared with `terminating`.

lukaszcz:
* Closes #2154 
* Evaluates closed applications with value arguments when the result type is zero-order. For example, `3 + 4` is evaluated to `7`, and `id 3` is evaluated to `3`,  but `id id` is not evaluated because the target type is not zero-order (it's a function type).


src/Juvix/Compiler/Core/Data/IdentDependencyInfo.hs
```diff
@@ -53,3 +56,45 @@ createSymbolDependencyInfo tab = createDependencyInfo graph startVertices
...
 recursiveIdents = nodesOnCycles . createCallGraph
+
+-- | identifiers from which some recursive identifier can be reached
+recursiveIdentsClosure :: InfoTable -> HashSet Symbol
+recursiveIdentsClosure tab =
+  -- unfortunately, there is no Graph library function which would allow to
+  -- compute this in linear time; hence, we implement this directly
+  run $
+    evalState (mempty :: HashSet Symbol) $
+      foldM (dfs mempty) mempty (HashMap.keys graph)
```
janmasrovira, line 67: the formatted code looks nicer if we write vertical chains following the pattern `f . g . h $ x` instead of `f $ g $ h $ x`

```suggestion
  run .
    evalState (mempty :: HashSet Symbol) $
      foldM (dfs mempty) mempty (HashMap.keys graph)
```

janmasrovira (approved):

PR #2456, opened 2023-10-20
Show import with public keyword in the html doc

### LINKED ISSUE #2455: Public imports not showing in html documentation
The bug was introduced in 9d9631aa26bc1b121a3b0fa23691318916d604ec

janmasrovira:
- Closes #2455


jonaprieto:
Let's add one smoke test as in here: https://github.com/anoma/juvix/blob/4dd80e71adc5eac68b528e23f4cbd639caa39307/tests/smoke/Commands/html.smoke.yaml#L17-L28

Instead of checking for `<!DOCTYPE HTML>`, put something sensible related to this fix.

jonaprieto (approved):

PR #2458, opened 2023-10-20
Refactor EntryPoint, Package, Pipeline, Root packages 

paulcadman:
Depends on:
*  https://github.com/anoma/juvix/pull/2451

This PR is part of a series implementing:
* https://github.com/anoma/juvix/issues/2336

In attempt to make the main PR:
* https://github.com/anoma/juvix/pull/2434
easier to review.

Changes in this PR refactor EntryPoint, Package, Pipeline, Root, splitting out the types into Base modules so that they can be imported and used in subsequent PRs with fewer dependencies to avoid package cycles.

### DIFF
--- a/src/Juvix/Compiler/Pipeline/Root/Base.hs
+++ b/src/Juvix/Compiler/Pipeline/Root/Base.hs
@@ -1,0 +1,15 @@
+module Juvix.Compiler.Pipeline.Root.Base where
+
+import Juvix.Compiler.Pipeline.Package.Base
+import Juvix.Prelude
+
+data Roots = Roots
+  { _rootsRootDir :: Path Abs Dir,
+    _rootsPackage :: Package,
+    _rootsPackageGlobal :: Bool,
+    _rootsBuildDir :: Path Abs Dir,
+    _rootsInvokeDir :: Path Abs Dir
+  }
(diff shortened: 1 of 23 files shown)


src/Juvix/Compiler/Pipeline/Root/Base.hs
```diff
@@ -0,0 +1,15 @@
+module Juvix.Compiler.Pipeline.Root.Base where
+
+import Juvix.Compiler.Pipeline.Package.Base
+import Juvix.Prelude
+
+data Roots = Roots
```
jonaprieto, line 6: Is it better to call this one simply `Root`?

paulcadman: Yes I think this is a better name. But would you mind if we did not change this until the other PRs are merged? It will cause many conflicts.

jonaprieto (approved):

src/Juvix/Compiler/Pipeline/Root/Base.hs
jonaprieto: Yes, I approved it because of the same reason.

jonaprieto (approved):

Issue #2430, opened 2023-10-05
`juvix init` Panics if System Language Uses Cyrillic Alphabet


agureev:
`juvix init` command panics producing output

`juvix: <stdout>: commitAndReleaseBuffer: invalid argument (cannot encode character '\10024')`

if the system language uses Cyrillics instead of English seemingly unable to process the glitter/stars unicode.

Problem got solved by changing system language to English producing the expected output. It is probably worth mentioning this issue somewhere, e.g. after the `doctor` command.

The system info is:
- Endeavor OS
- Juvix Version 0.5.2

paulcadman:
We should use `setLocaleEncoding utf8` before any console output.

https://hackage.haskell.org/package/base-4.18.1.0/docs/GHC-IO-Encoding.html#v:setLocaleEncoding

paulcadman:
Serokell made a Haskell library to try to set a utf8 compatible locale for Haskell applications: https://serokell.io/blog/haskell-with-utf8

I cannot reproduce the error in the bug so I can't check to see if it's fixed.

@agureev if possible, could you compile juvix using the [set-output-locale-with-utf8](https://github.com/anoma/juvix/tree/set-output-locale-with-utf8) branch and see if it fixes the issue (without changing the system language)?

agureev:
Currently the error is still reproducible on commit `63e795f5` and it still seems to do with language settings. Here is the locale setup as well as the error message (which, as you may note, is the same as before, seemingly having to do with processing the glitter unicode):

```
LANG=ru_RU.UTF-8
LC_CTYPE="ru_RU.UTF-8"
LC_NUMERIC=en_CA.UTF-8
LC_TIME=en_CA.UTF-8
LC_COLLATE="ru_RU.UTF-8"
LC_MONETARY=en_CA.UTF-8
LC_MESSAGES="ru_RU.UTF-8"
LC_PAPER=en_CA.UTF-8
LC_NAME=en_CA.UTF-8
LC_ADDRESS=en_CA.UTF-8
LC_TELEPHONE=en_CA.UTF-8
LC_MEASUREMENT=en_CA.UTF-8
LC_IDENTIFICATION=en_CA.UTF-8
LC_ALL=
```

Running `juvix init` produces 

`juvix: <stdout>: commitAndReleaseBuffer: invalid argument (cannot encode character '\10024')`

agureev:
Here's a direct test using `export`:

```
$ juvix init
[...]

PR #2463, opened 2023-10-23
Use with-utf8 to manage setting a compatible terminal locale

paulcadman:
See: https://serokell.io/blog/haskell-with-utf8 for motivation for this library.

This is an attempt to fix:
* https://github.com/anoma/juvix/issues/2430


paulcadman:
This doesn't fix https://github.com/anoma/juvix/issues/2430 - so we'll close this for now.

PR #2465, opened 2023-10-23
Update clang formatting

jonaprieto:
Using

```
clang-format --version
Homebrew clang-format version 17.0.2
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
       - id: check-yaml
       - id: check-added-large-files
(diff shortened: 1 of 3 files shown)


jonaprieto:
> Updating the config seems to have fixed it

I didn't notice you push, sorry. Btw, I don't know if you noticed that the pre-commit updates to v17.0.3. I was trying to update with homebrew but it's throwing errors.

paulcadman:
> > Updating the config seems to have fixed it
> 
> I didn't notice you push, sorry. Btw, I don't know if you noticed that the pre-commit updates to v17.0.3. I was trying to update with homebrew but it's throwing errors.

I was able to update the homebrew formula to v17.0.3 however this may not be the one that you actually use:

There's a copy of clang-format bundled with llvm:

```
$ which clang-format
/opt/homebrew/opt/llvm@16/bin/clang-format
```

so if you have homebrew llvm on your PATH, you'll use the copy of `clang-format` from llvm rather than the one installed from the homebrew formula.

paulcadman (approved):

PR #2467, opened 2023-10-24
Add FileExt type

jonaprieto:
This PR introduces FileExt type, and consequently, one can generalise methods and matches based on the file extension; for example, `parseInputJuvixAsmFile` is now an app. `parseInputFile FileExtJuvixAsm`

### DIFF
--- a/src/Juvix/Data/FileExt.hs
+++ b/src/Juvix/Data/FileExt.hs
@@ -13,10 +13,15 @@ data FileExt
   | FileExtJuvixGeb
   | FileExtJuvixCore
   | FileExtJuvixAsm
+  | FileExtVampIR
+  | FileExtVampIRParams
+  | FileExtPlonk
+  | FileExtHalo
   | FileExtLisp
-  | FileC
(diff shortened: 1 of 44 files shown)


src/Juvix/Data/FileExt.hs
```diff
@@ -0,0 +1,106 @@
...
+
+-- | File extensions Juvix interacts with.
+data FileExt
+  = FileExtJuvix
+  | FileExtJuvixMarkdown
+  | FileExtJuvixGeb
+  | FileExtJuvixCore
+  | FileExtJuvixAsm
+  | FileExtLisp
+  | FileC
```
paulcadman, line 17: Rename `FileC` -> `FileExtC` ?

janmasrovira (approved):

src/Juvix/Data/FileExt.hs
jonaprieto: Done

paulcadman (approved):

Issue #2472, opened 2023-10-25
Juvix source files are decoded using the system locale instead of UTF8


paulcadman:
Thanks to @agureev for finding this issue.

## Description of Issue

The following session was recorded on Artem's machine using Arch linux
```
$ export LANG=ru_RU.koi8r
$ locale
LANG=ru_RU.koi8r
LC_CTYPE="ru_RU.koi8r"
LC_NUMERIC="ru_RU.koi8r"
LC_TIME="ru_RU.koi8r"
LC_COLLATE="ru_RU.koi8r"
LC_MONETARY="ru_RU.koi8r"
LC_MESSAGES="ru_RU.koi8r"
LC_PAPER="ru_RU.koi8r"
LC_NAME="ru_RU.koi8r"
LC_ADDRESS="ru_RU.koi8r"
LC_TELEPHONE="ru_RU.koi8r"
LC_MEASUREMENT="ru_RU.koi8r"
LC_IDENTIFICATION="ru_RU.koi8r"
LC_ALL=
$ juvix repl
Juvix REPL version 0.5.2-830b3be: https://juvix.org/. Run :help for help
/home/artgur/.config/juvix/0.5.2/global-project/.juvix-build/stdlib/Stdlib/Data/Bool/Base.juvix:34:20: error:
/home/artgur/.config/juvix/0.5.2/global-project/.juvix-build/stdlib/Stdlib/Data/Bool/Base.juvix:34:20:
   |
34 | if : {A : Type} → Bool → A → A → A
   |                    ^
expecting ->
```

The file `Base.juvix` should be written to / read from disk using the UTF-8 encoding. However the non-UTF8 system locale (koi8r) is being used to read the file meaning that `→` is not recognised by Juvix.

## Cause

We are using API from https://hackage.haskell.org/package/text-2.1/docs/Data-Text-IO.html#v:readFile to read the contents of files. This API has the following warning:

> Beware that this function (similarly to [readFile](https://hackage.haskell.org/package/base-4.16.3.0/docs/Prelude.html#v:readFile)) is locale-dependent. Unexpected system locale may cause your application to read corrupted data or throw runtime exceptions about "invalid argument (invalid byte sequence)" or "invalid argument (invalid character)". This is also slow, because GHC first converts an entire input to UTF-32, which is afterwards converted to UTF-8.

## Solution

[...]

PR #2473, opened 2023-10-25
Use system locale independent readFile and writeFile APIs from with-utf8

paulcadman:
The problem with readFile and writeFile from text [Data.Text.IO](https://hackage.haskell.org/package/text-2.0.2/docs/Data-Text-IO.html) is that they use the system locale to determine the text encoding format.

Our assumption is that all Juvix source files are UTF-8 encoded.

I cannot reproduce the issue with using the old APIs on my machine, it can be reproduced on Arch linux. I'm not sure how to write a specific test for this.

* Closes https://github.com/anoma/juvix/issues/2472

### DIFF
--- a/app/Commands/Dev/Asm/Compile.hs
+++ b/app/Commands/Dev/Asm/Compile.hs
@@ -19,7 +19,7 @@ runCommand opts = do
           { _entryPointTarget = tgt,
             _entryPointDebug = opts ^. compileDebug
           }
-  s <- embed (readFile (toFilePath file))
+  s <- readFile (toFilePath file)
   case Asm.runParser (toFilePath file) s of
     Left err -> exitJuvixError (JuvixError err)
     Right tab -> do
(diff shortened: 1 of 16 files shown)


janmasrovira:
For the record, I've reproduced the bug (`juvix init` and `juvix repl`)  on my linux machine with `/etc/local.conf`:
```
LANG=en_GB.ISO-8859-1
LC_ADDRESS=en_GB.ISO-8859-1
LC_IDENTIFICATION=en_GB.ISO-8859-1
LC_MEASUREMENT=en_GB.ISO-8859-1
LC_MONETARY=en_GB.ISO-8859-1
LC_NAME=en_GB.ISO-8859-1
LC_NUMERIC=en_GB.ISO-8859-1
LC_PAPER=en_GB.ISO-8859-1
LC_TELEPHONE=en_GB.ISO-8859-1
LC_TIME=en_GB.ISO-8859-1
```

janmasrovira (approved):

Issue #2452, opened 2023-10-19
Versioning for the Package.juvix format


paulcadman:
Following on from:
* https://github.com/anoma/juvix/issues/2336

## Contract

The Package.juvix format is defined by the `Package` type in the `PackageDescription` module and the contract of the `Package.juvix` file (namely, that `Package.juvix` contains a top level identifier `package` with the type `Package`).

## Requirement

We need:
1. The flexibility to change this contract
2. To support projects that use older versions of the contract (backwards compatibility). i.e we don't want to break existing projects if we change the `Package` type.

We can achieve this by versioning the `Package.juvix` contract and supporting older versions of the contract in new compilers. The version specification needs to be independent from the contract itself. For example the version cannot be part of the `Package` type because we need to choose the version of the `Package` type to use before running the pipeline. 

## Proposal

@janmasrovira had the very good idea to version the PackageDescription module paths. i.e we would package modules with names `PackageDescription.V1`, `PackageDescription.V2` and make them all available to the `Package` module a loading time.

The Package validation would check that the `package` identifier in the `Package` module comes from one of the `PackageDescription` modules and there would be a separate backend translation for each version.

PR #2478, opened 2023-10-27
Version the PackageDescription module

paulcadman:
This PR adds a version number to the module name of the `PackageDescription` module. This allows us to change the Package file format in a backwards compatible way in the future.

Now the simplest Package.juvix file looks like:

```
module Package;

import PackageDescription.V1 open;

```

## Adding new versions

The idea is that new versions of the PackageDescription format will be
added as files of the form:
include/package/PackageDescription/Vx.juvix

The correspondence between a version of the PackageDescription file and the package type name is recorded in [`packageDescriptionTypes`](https://github.com/anoma/juvix/blob/9ba869d836f9f4a6bf678bee65838594fbd8613a/src/Juvix/Compiler/Pipeline/Package/Loader.hs#L15).

The `package` identifier type must come from **one** of the versions
of the PackageDescription module.

* Closes https://github.com/anoma/juvix/issues/2452

### DIFF
--- a/src/Juvix/Compiler/Pipeline/Package/Loader/EvalEff.hs
+++ b/src/Juvix/Compiler/Pipeline/Package/Loader/EvalEff.hs
@@ -13,7 +13,7 @@ makeLenses ''TypeSpec
 data EvalEff m a where
   Eval' :: Node -> EvalEff m Value
   LookupIdentifier :: Text -> EvalEff m Node
-  -- | Assert that the Node has a type given by one of the ;TypeSpec;s
+  -- | Assert that the Node has a type given by one of the 'TypeSpec's
   AssertNodeType :: (Foldable f) => Node -> f TypeSpec -> EvalEff m ()
 
 makeSem ''EvalEff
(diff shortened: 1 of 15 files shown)


src/Juvix/Compiler/Pipeline/Package/Loader/EvalEff.hs
```diff
@@ -13,7 +13,8 @@ makeLenses ''TypeSpec
 data EvalEff m a where
   Eval' :: Node -> EvalEff m Value
   LookupIdentifier :: Text -> EvalEff m Node
-  AssertNodeType :: Node -> TypeSpec -> EvalEff m ()
+  -- | Assert that the Node has a type given by one of the ;TypeSpec;s
```
jonaprieto, line 16: `;TypeSpec;`?

jonaprieto (commented):
LGTM

src/Juvix/Compiler/Pipeline/Package/Loader/EvalEff.hs
paulcadman: Fixed this - thanks

jonaprieto (approved):

PR #2480, opened 2023-10-30
Rename Roots type to Root

paulcadman:
This was suggested by @jonaprieto in https://github.com/anoma/juvix/pull/2458#discussion_r1368476371 - but we deferred it until the Package file PR sequence was merged.


jonaprieto (approved):
Good, this looks much better

PR #2482, opened 2023-10-30
Use `DisjointSet` to unify variables

janmasrovira:


jonaprieto:
Added to https://github.com/anoma/juvix/wiki/Catalogue-of-Deferred-Interesting-Pull-Requests

PR #2483, opened 2023-10-30
Update juvix init to generate Package.juvix instead of juvix.yaml

paulcadman:
`juvix init` now generates a `Package.juvix` file in the current directory instead of a `juvix.yaml` file. It uses the prompts from the user to fill in the name and version options.

### Validity check

After the file is generated, the Juvix project in the current directory is loaded to check that the generated file is valid.

### Version support

Each version of the PackageDescription module must have a corresponding [PackageDescriptionType](https://github.com/anoma/juvix/blob/c39c27000c99e6053dce83a34fca7c9f66b6c605/src/Juvix/Compiler/Pipeline/Package/Loader/Versions.hs#L12) which defines:

* The path relative to `include/package` that the file exists at (e.g `PackageDescription/V1.juvix`)
* The name of the Package type in the file (e.g `Package`)
* A function that translates a Package type into a Concrete function definition that represents the code that is generated in the `Package.juvix`
* A function that returns a Bool that determines if the generated function definition depends on the standard library. (i.e that the standard library needs to be imported by Package.juvix).

The last point is necessary because in the V1 spec, the `SemVer` type uses `Maybe` to encode the release and meta components of the of the version:

```
package : Package :=
  defaultPackage
    {name := "dd";
     version := mkVersion 1 2 3 {just "prerel"} {just "meta"}};
```

So if the user specifies a release or meta component in their version - then the standard library needs to be imported to provide the `just` constructor. If the user only specifies major, minor and patch components then the standard library does not need to be imported.

* Closes https://github.com/anoma/juvix/issues/2485

### DIFF
--- a/app/Commands/Init.hs
+++ b/app/Commands/Init.hs
@@ -30,6 +30,7 @@ init = do
   pkg <- getPackage
   say ("creating " <> pack (toFilePath packageFilePath))
   embed (Utf8.writeFile @IO (toFilePath packageFilePath) (renderPackage pkg))
+  checkPackage
   say "you are all set"
   where
     renderPackage :: Package -> Text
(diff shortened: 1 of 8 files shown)


paulcadman:
> Why did you mark it as 0.5.4?

I'm not sure when 0.5.3 cutoff is/was - we can put it in 0.5.3 if you like.

janmasrovira (approved):

PR #2490, opened 2023-11-01
Cache .hie directory in CI

paulcadman:
We use the options in `package.yaml`.

```
         - -fhide-source-paths
         - -fwrite-ide-info -hiedir=.hie
```
If a previously available .hie directory is missing then GHC will rebuild the whole project with reason: `[HIE file is missing]`. So we need to cache it to take advantage of incremental builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: main/.hie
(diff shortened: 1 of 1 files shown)


paulcadman:
Compare Stack Setup in

* https://github.com/anoma/juvix/actions/runs/6718948000/job/18259629007
* https://github.com/anoma/juvix/actions/runs/6719337660/job/18260772473

This change only saves 1.5 minutes from the build in total.

jonaprieto:
> This change only saves 1.5 minutes from the build in total. This is better than nothing I suppose!

Thanks!

jonaprieto (approved):

Issue #2487, opened 2023-10-31
Remove `juvix.yaml` support


paulcadman:
After:

* https://github.com/anoma/juvix/issues/2336
* https://github.com/anoma/juvix/issues/2485

We should remove support for the `juvix.yaml` package file.

* Remove the {To, From}JSON instances for Package and other utilities for reading/writing the `juvix.yaml` file.
* Add a utility (e.g `juvix dev migrate-package`) which translates an existing `juvix.yaml` files to an equivalent `Package.juvix` file.

PR #2503, opened 2023-11-07
Migrate all Juvix projects from juvix.yaml to Package.juvix in the repository

paulcadman:
This PR:

* Modifies entry point `_entryPointBuildDir` to use the `BuildDir` type instead of `SomeBase Dir`. This allows delayed resolution of the default build directory which was useful for the Package -> Concrete translation point below.
* Modifies `juvix dev root` to render the current package as a Package.juvix file.
* Modifies the Package -> Concrete translation to recognise default arguments. So, for example, an empty `juvix.yaml` file will be translated into the following (instead of the `name`, `version`, and `dependencies` arguments being populated).

        
        module Package;

        import Stdlib.Prelude open;
        import PackageDescription.V1 open;

* Adds a temporary command (removed when juvix.yaml support is removed) `juvix dev migrate-juvix-yaml` that translates `juvix.yaml` into an equivalent `Package.juvix` in the current project.
* Adds a temporary script `migrate-juvix-yaml.sh` (removed when juvix.yaml support is removed) which can be run in the project to translate all Juvix projects in the repository.
* Actually translate all of the `juvix.yaml` files to `Package.juvix` using the script.

* Part of https://github.com/anoma/juvix/issues/2487

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -63,7 +63,7 @@ runAppIO args@RunAppIOArgs {..} =
     AskRoot -> return _runAppIOArgsRoot
     AskInvokeDir -> return invDir
     AskPkgDir -> return (_runAppIOArgsRoot ^. rootRootDir)
-    AskBuildDir -> return (_runAppIOArgsRoot ^. rootBuildDir)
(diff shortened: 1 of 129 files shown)


jonaprieto (approved):
LGTM

Issue #2476, opened 2023-10-27
Fix variable refreshing in substitution


lukaszcz:
In `Inference.weakNormalize'` and `Inference.strongNormalize'` in the case for `SimpleLambda` we should be refreshing the bound variables in `r`.

lukaszcz:
This actually should be changed in general for substitution, which is just wrong currently.

PR #2504, opened 2023-11-08
Fix `clone` in `subsHoles`

lukaszcz:
Fix `clone` (it was in the wrong place) introduced in https://github.com/anoma/juvix/pull/2501

### DIFF
--- a/src/Juvix/Compiler/Internal/Extra/Base.hs
+++ b/src/Juvix/Compiler/Internal/Extra/Base.hs
@@ -159,8 +159,8 @@ subsHoles s = leafExpressions helper
   where
     helper :: Expression -> Sem r Expression
     helper e = case e of
-      ExpressionHole h -> maybe (clone e) return (s ^. at h)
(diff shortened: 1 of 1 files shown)


janmasrovira (dismissed):

jonaprieto:
Isn't a test to check this fix something?

lukaszcz:
> Isn't a test to check this fix something?

This fixes a mistake in #2476 

It's hard to make a test other than one which would go through the term and check whether all bound variables are unique (and it would be more time-consuming to write this test than fix the problem). Because for now we don't have a concrete example of a program where violating this invariant actually breaks things.

jonaprieto (approved):

jonaprieto:
> > Isn't a test to check this fix something?
> 
> This fixes a mistake in #2501
> 
> It's hard to make a test other than one which would go through the term and check whether all bound variables are unique (and it would be more time-consuming to write this test than fix the problem). Because for now we don't have a concrete example of a program where violating this invariant actually breaks things.

Okay

PR #2481, opened 2023-10-30
Interleave arity and typechecking

janmasrovira:
- Closes #2362 

This pr implements a new typechecking algorithm. This algorithm can be activated using the global flag `--new-typechecker`. This flag will only take effect on the compilation pipeline but not the repl.

The main difference between the new and old algorithm is that the new one inserts holes during typechecking. Thus, it does not require the arity checker pass.

The new algorithm does not yet implement default arguments. The plan is to make the change in the following steps:
1. Merge this pr.
2. Merge #2506.
3. Implement default arguments for the new algorithm.
4. Remove the arity checker and the old algorithm.

PR #2506, opened 2023-11-08
General inductive parameters

janmasrovira:
- Depends on #2481 

This pr allows inductive type parameters to be any type. Until now, they had to be exactly `Type`. This allows us to define more general traits such as the `Monad` and `Functor`, as shown in the new test.
This is only supported under the temporary `--new-typechecker` flag.

Pending work:

Update the positivity checker if necessary (@jonaprieto).
Update the necessary compilation steps in Core (@lukaszcz).
Add compilation tests.


lukaszcz (approved):

jonaprieto:
Shortly, we can merge this PR since it does not replace the default typechecker. However, we need to address the positivity checker separately. I added two examples below to make it clear that there is an issue, although it is not straightforward to me at the moment how to solve it. 

 

```
module NegativeEvil;

type Evil (f : Type -> Type) :=
  | magic : f (Evil f) -> Evil f;
```

First, the `Evil` data type is not strictly positive but it passes our checker, so we have an issue. Second, ignoring the positivity condition over the Evil definition for a moment, consider the rest of this program.  

```
type Empty := mkEmpty {}; 

toEmpty (A : Type) : Type := A -> Empty;

panic : toEmpty (Evil toEmpty) 
  | x@(magic h) := h x ;

god : Empty := panic (magic panic); 
```

The above produces the following error, 
```
The expression panic has type:
  Evil λ! {A := A → Empty} → Empty
but is expected to have type:
  _ (Evil _)
```

The message is a bit difficult to understand, at least to me, so we must fix it too.
Also, this should typecheck ignoring the positivity condition.

Similarly as with `Evil`, one could derive a term of `Empty` using for example, other types like:

```
type Free (f : Type -> Type) (A : Type)
 : Type :=
| Leaf : A -> Free f A
| Branch : f (Free f A) -> Free f A;
```

rokopt:
> @rokopt could get benefit of being able to typecheck this `Free` data type, isn't it? your `BT A` could be expressed as `Free (\ {T := Prod T T}) A`, or similar I'd say.

Yes, that's right!

PR #2507, opened 2023-11-09
Run test suite in parallel

paulcadman:
## Overview

This PR makes the compiler pipeline thread-safe so that the test suite can be run in parallel.

This is achieved by:
* Removing use of `{get, set, with}CurrentDir` functions.
* Adding locking around shared file resources like the the global-project and internal build directory.

NB: **Locking is disabled for the main compiler target**, as it is single threaded they are not required.

## Run test suite in parallel

To run the test suite in parallel you must add `--ta '+RTS -N -RTS'` to your stack test arguments. For example:

```
stack test --fast --ta '+RTS -N -RTS'
```

The `-N` instructs the Haskell runtime to choose the number of threads to use based on how many processors there are on your machine. You can use `-Nn` to see the number of threads to `n`.

These flags are already [set in the Makefile](https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/Makefile#L26) when you or CI uses `stack test`.

## Locking

The Haskell package [filelock](https://hackage.haskell.org/package/filelock) is used for locking. File locks are used instead of MVars because Juvix code does not control when new threads are created, they are created by the test suite. This means that MVars created by Juvix code will have no effect, because they are created independently on each test-suite thread. Additionally the resources we're locking live on the filesystem and so can be conveniently tagged by path.

### FileLock

The filelock library is wrapped in a FileLock effect:

https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Data/Effect/FileLock/Base.hs#L6-L8

There is an [IO interpreter](https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Data/Effect/FileLock/IO.hs#L8) that uses filelock and an [no-op interpreter](https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Data/Effect/FileLock/Permissive.hs#L7) that just runs actions unconditionally.

### TaggedLock

To make the file locks simpler to use a TaggedLock effect is introduced:

https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Data/Effect/TaggedLock/Base.hs#L5-L11

And convenience function:

https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Data/Effect/TaggedLock.hs#L28

This allows an action to be locked, tagged by a directory that may or may not exist. For example in the following code, an action is performed on a directory `root` that may delete the directory before repopulating the files. So the lockfile cannot be stored in the `root` itself.

https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Extra/Files.hs#L55-L60

## Pipeline

As noted above, we only use locking in the test suite. The main app target pipeline is single threaded and so locking is unnecessary. So the interpretation of locks is parameterised so that locking can be disabled https://github.com/anoma/juvix/blob/e6dca22cfdcff936add5e7134f9c6f20416504a5/src/Juvix/Compiler/Pipeline/Run.hs#L64


src/Juvix/Compiler/Pipeline.hs
```diff
@@ -37,11 +37,12 @@ import Juvix.Compiler.Reg.Data.InfoTable qualified as Reg
 import Juvix.Compiler.Reg.Translation.FromAsm qualified as Reg
 import Juvix.Data.Effect.Git
 import Juvix.Data.Effect.Process
+import Juvix.Data.Effect.TaggedLock
 import Juvix.Prelude
 
-type PipelineEff = '[PathResolver, EvalFileEff, Error PackageLoaderError, Error DependencyError, GitClone, Error GitProcessError, Process, Log, Reader EntryPoint, Files, NameIdGen, Builtins, Error JuvixError, HighlightBuilder, Internet, Embed IO]
+type PipelineEff = '[PathResolver, EvalFileEff, Error PackageLoaderError, Error DependencyError, GitClone, Error GitProcessError, Process, Log, TaggedLock, Reader EntryPoint, Files, NameIdGen, Builtins, Error JuvixError, HighlightBuilder, Internet, Embed IO, Resource, Final IO]
```
janmasrovira, line 43: Is it necessary to have both `Embed IO` and `Final IO`?

paulcadman: I think so, because of the way we interpret the effects at the top level:

https://github.com/anoma/juvix/blob/c4d349473db82eacb938e8d2e4f456417a4e9361/src/Juvix/Compiler/Pipeline/Run.hs#L57-L75

The IO affects like `runFilesIO` depend on `Embed IO` but `resourceToIOFinal` depends on `Final IO`. We also have `embedToFinal @IO` in the stack of interpreters but for some reason I don't understand this does not mean we can remove the `Embed IO` effect (without doing something else), we get a type error:

<details>

<summary>Type error output, expand for details</summary>

```
src/Juvix/Compiler/Pipeline/Run.hs:75:7: error:
    • Couldn't match type ‘Resource’ with ‘Embed IO’
      Expected: Sem PipelineEff a
                -> Sem
                     '[Juvix.Compiler.Pipeline.Package.Loader.EvalEff.EvalFileEff,
                       Error PackageLoaderError, Error DependencyError,
                       Scoped CloneArgs Git, Error GitProcessError, Process, Log,
                       TaggedLock, Reader EntryPoint, Files, NameIdGen, Builtins,
                       Error JuvixError, HighlightBuilder, Internet, Embed IO, Resource,
                       Final IO]
                     (ResolverState, a)
        Actual: Sem
                  '[PathResolver,
                    Juvix.Compiler.Pipeline.Package.Loader.EvalEff.EvalFileEff,
                    Error PackageLoaderError, Error DependencyError, GitClone,
                    Error GitProcessError, Process, Log, TaggedLock, Reader EntryPoint,
                    Files, NameIdGen, Builtins, Error JuvixError, HighlightBuilder,
                    Internet, Resource, Final IO]
                  a
                -> Sem
                     '[Juvix.Compiler.Pipeline.Package.Loader.EvalEff.EvalFileEff,
                       Error PackageLoaderError, Error DependencyError, GitClone,
                       Error GitProcessError, Process, Log, TaggedLock, Reader EntryPoint,
                       Files, NameIdGen, Builtins, Error JuvixError, HighlightBuilder,
                       Internet, Resource, Final IO]
                     (ResolverState, a)
    • In the second argument of ‘(.)’, namely ‘runPathResolver'’
      In the second argument of ‘(.)’, namely
        ‘runEvalFileEffIO . runPathResolver'’
      In the second argument of ‘(.)’, namely
        ‘mapError (JuvixError @PackageLoaderError)
           . runEvalFileEffIO . runPathResolver'’
   |
75 |     . runPathResolver'
   |       ^^^^^^^^^^^^^^^^
```

</details>

janmasrovira: I think we could get rid of the `Embed IO` if we refactored the functions that have a constraint like `Member (Embed IO) r` with `MonadIO r`. However, I don't think there is a point of doing that in this pr, so it's fine as it is

janmasrovira (approved):

Issue #2517, opened 2023-11-16
Typechecking infers incorrect types


lukaszcz:
Input:
```
module ty1;

import Stdlib.Prelude open;

trait
type Functor (f : Type -> Type) :=
  mkFunctor {
    fmap : {A B : Type} -> (A -> B) -> f A -> f B
  };

instance
listFunctorI : Functor List :=
  mkFunctor@{
    fmap := map;
  };
```
Output after running `uvix dev internal typecheck ty1.juvix --print-result --new-typechecker`:
```
import Stdlib.Prelude

mutual {
  listFunctorI : Functor List
    := let 
      not-mutual fmap : (f : A → B) → List A → List B
        := map {A} {B}
     in let 
      not-mutual f : Type → Type := List
      not-mutual fmap : {A : Type} → {B : Type} → (A → B) → f A → f B
        := λ {| {_ω1276} {_ω1277} := fmap}
     in mkFunctor {f} fmap

  type Functor (f : Type → Type) =
    mkFunctor : (fmap : {A : Type} → {B : Type} → (A → B) → f A → f B) → Functor f
}

mutual fmap : {f : Type → Type} → {{Functor f}} → {A : Type} → {B : Type} → (A → B) → f A → f B
  := λ : {f : Type → Type} → {{Functor f}} → {A : Type} → {B : Type} → (A → B) → f A → f B {| {_ω1282} ({{(mkFunctor a : Functor _ω1282)}}) {_ω1283} {_ω1284} := a {A} {B}}
```

In the fist `fmap`, the type variables `A` and `B` are not bound, which later makes the translation from Internal to Core fail.

lukaszcz:
This is actually a bug with the new named arguments syntax. Just before having type constructor parameters for inductive types we never used type variables in constructors.

lukaszcz:
Well, the bug is in the typechecker, both the new and old one, the named arguments syntax only causes it to appear.

Running `juvix dev internal pretty ty1.juvix --new-typechecker` results in:
```
mutual {
  listFunctorI : Functor List
    := let 
      not-mutual fmap : _ := map
     in let 
      not-mutual f : Type → Type := _
      not-mutual fmap : {A : Type} → {B : Type} → (A → B) → f A → f B
        := fmap
     in mkFunctor {f} fmap

  type Functor (f : Type → Type) =
    mkFunctor : (fmap : {A : Type} → {B : Type} → (A → B) → f A → f B) → Functor f
}
```
[...]

Issue #2247, opened 2023-07-07
Bug in typechecking


janmasrovira:
The following
```
module Lambda;

type Box (A : Type) :=
  | box : A → Box A;

x : Box ((A : Type) → A → A);
x := box λ {A a := a};
```
gets typechecked into:
```
ype Box (A : Type) =
  box : A → Box A

x : Box ((A : Type) → A → A)
x := box {Type → errorHere@A → A} λ : Type → A → A { A a := a}
```
The error is that that the `A` next to `errorHere` is out of scope

paulcadman:
@janmasrovira please raise a separate issue for adding an error message.

PR #2524, opened 2023-11-20
Improve inference for `--new-typechecker`

janmasrovira:
This pr applies a number of fixes to the new typechecker.
The fixes implemented are:
1.  When guessing the arity of the body, we properly use the type information of the variables in the patterns.
2. When generating wildcards, we name them properly so that they align with the name in the type signature.
3. When compiling named applications, we inline all clauses of the form `fun : _ := body`. This is a workaround to https://github.com/anoma/juvix/issues/2247 and https://github.com/anoma/juvix/issues/2517
4. I've had to ignore test027 (Church numerals). While the typechecker passes and one can see that the types are correct, there is a lambda where its clauses have different number of patterns. Our goal is to support that in the near future (https://github.com/anoma/juvix/issues/1706). This is the conflicting lambda:
    ```
    mutual num : Nat → Num
      := λ : Nat → Num {| (zero : Nat) := czero
      | ((suc n : Nat)) {A} := csuc (num n) {A}}
    ```
5.  I've added non-trivial a compilation test involving monad transformers.


paulcadman (approved):
I also tested compilation of juvix-containers etc. with the new typechecker.

janmasrovira:
> I also tested compilation of juvix-containers etc. with the new typechecker.

thanks!

PR #2543, opened 2023-11-28
Fix: remove ErrArity from Typechecking errors

jonaprieto:


janmasrovira:
this is already done in #2532

jonaprieto:
> this is already done in #2532
All right. Anyway, we should be more careful when reviewing PRs to
not see this kind of PRs.

PR #2546, opened 2023-12-01
Update to the latest juvix-stdlib

paulcadman:
This PR updates the juvix-stdlib submodule ref to the current juvix-stdlib/main.

NB: The Markdown test is not stable after changes to the stdlib - the ids (deriving from internal name ids?) will change and so the expected file must be regenerated.


jonaprieto:
> NB: The Markdown test is not stable after changes to the stdlib - the ids (deriving from internal name ids?) will change and so the expected file must be regenerated.

Yes, that's unfortunate. It's hard to check these IDs are the correct ones, so it makes sense, perhaps, just for testing to have another flag to ignore them.

jonaprieto (approved):

PR #2548, opened 2023-12-05
Read `Package` on demand and cache it

janmasrovira:
This patch dramatically increases the efficiency of `juvix dev root`, which was unnecessarily parsing all dependencies included in the `Package.juvix` file. Other commands that do not require the `Package` will also be faster.

It also refactors some functions so that the `TaggedLock` effect is run globally.

I've added `singletons-base` as a dependency so we can have `++` on the type level. We've tried to define a type family ourselves but inference was not working properly.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -77,6 +77,7 @@ dependencies:
   - process == 1.6.*
   - safe == 0.3.*
   - singletons == 3.0.*
+  - singletons-base == 3.1.*
   - singletons-th == 3.1.*
   - Stream == 0.4.*
   - string-interpolate == 0.3.*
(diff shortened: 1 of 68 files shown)


package.yaml
```diff
@@ -77,6 +77,7 @@ dependencies:
   - process == 1.6.*
   - safe == 0.3.*
   - singletons == 3.0.*
+  - singletons-base == 3.1.*
```
paulcadman, line 80: Is this import now required? Was it just added for the type level list append experiments?

paulcadman (approved):

Issue #2559, opened 2023-12-10
Translation from JuvixAsm to Nock


lukaszcz:
* Implement translation from JuvixAsm to Nock. 

  One can use the argument on the left of `*` in the Nock evaluation function as a stack. The Nock stack would need to contain (from bottom to top): pointer to stdlib, pointer to JuvixAsm compiled functions, the function arguments (JuvixAsm argument area), the value stack and the temporary stack mixed together. Each function can have a separate Nock stack - on function call a new stack is created which initially contains the new arguments and a pointer to compiled functions and stdlib at the end (bottom). In the translation, one needs to keep track of the current value and temporary stack heights and adjust the references to temporaries, arguments and function symbols accordingly. JuvixAsm constructors can be represented by Nock terms with the tag in the head, e.g., `cons 1 (cons 2 nil)` would be `[cons 1 [cons 2 nil]]` with unique numbers used for the tags `cons` and `nil`.
* Test the compilation pipeline to Nock with the Nock evaluator.
* Depends on #2557 
* Depends on #2555

paulcadman:
Implemented in:
* https://github.com/anoma/juvix/pull/2570
* https://github.com/anoma/juvix/pull/2586
* https://github.com/anoma/juvix/pull/2591

Issue #2558, opened 2023-12-10
Nock serialization


lukaszcz:
* Implement serialization of our Nock representation to actual Nockma format.
* Depends on #2557

paulcadman:
Fixed by:

* https://github.com/anoma/juvix/pull/2564

PR #2570, opened 2023-12-29
Nockma compile

janmasrovira:
This PR is a snapshot of the current work on the JuvixAsm -> Nockma translation. The compilation of Juvix programs to Nockma now works so we decided to raise this PR now to avoid it getting too large.

## Juvix -> Nockma compilation

You can compile a frontend Juvix file to Nockma as follows:

example.juvix
```
module example;

import Stdlib.Prelude open;

fib : Nat → Nat → Nat → Nat
  | zero x1 _ := x1
  | (suc n) x1 x2 := fib n x2 (x1 + x2);

fibonacci (n : Nat) : Nat := fib n 0 1;

sumList (xs : List Nat) : Nat :=
  for (acc := 0) (x in xs)
    acc + x;

main : Nat := fibonacci 9 + sumList [1; 2; 3; 4];
```

```
$ juvix compile -t nockma example.juvix
```

This will generate a file `example.nockma` which can be run using the nockma evaluator:

```
$ juvix dev nockma eval example.nockma
```

Alternatively you can compile JuvixAsm to Nockma:

```
$ juvix dev asm compile -t nockma example.jva
```

## Tests

We compile an evaluate the JuvixAsm tests in https://github.com/anoma/juvix/blob/cb3659e08e552ee9ca40860077e39a4070cf3303/test/Nockma/Compile/Asm/Positive.hs

We currently skip some because either:
1. They are too slow to run in the current evaluator (due to arithmetic operations using the unjetted nock code from the anoma nock stdlib).
2. They trace data types like lists and booleans which are represented differently by the asm interpreter and the nock interpreter
3. They operate on raw negative numbers, nock only supports raw natural numbers

## Next steps

On top of this PR we will work on improving the evaluator so that we can enable the slow compilation tests.


lukaszcz:
I refactored the Nockma pipeline to fit more with how other pipelines are handles, moving the JuvixAsm transformations to the function `toNockma` in `Juvix.Compiler.Asm.Pipeline`, and adjusting `Juvix.Compiler.Pipeline` and other places accordingly.

I also removed the `--nockma-debug` option, because there already is a `--debug` option. To generate `trace` pseudo-Nockma instructions, use `--debug` or `-g`.

paulcadman:
> I refactored the Nockma pipeline to fit more with how other pipelines are handled, moving the JuvixAsm transformations to the function `toNockma` in `Juvix.Compiler.Asm.Pipeline`, and adjusting `Juvix.Compiler.Pipeline` and other places accordingly.
> 
> I also removed the `--nockma-debug` option, because there already is a `--debug` option. To generate `trace` pseudo-Nockma instructions, use `--debug` or `-g`.

Thanks!

lukaszcz:
I think ultimately the `--nockma-pretty` option should also be removed in favour of having a separate "pseudo-nockma" target which supports readable names, debugging instructions, generally some inessential but convenient sugar over the actual Nock format. This target would then be serialized to actual Nockma, omitting the traces and translating names to numbers:
* #2558.

app/Commands/Dev/Core/Repl.hs
```diff
@@ -26,7 +26,7 @@ parseText = Core.runParser replPath defaultModuleId
 
 runRepl :: forall r. (Members '[Embed IO, App] r) => CoreReplOptions -> Core.InfoTable -> Sem r ()
 runRepl opts tab = do
-  embed (putStr "> ")
+  putStr "> "
```
lukaszcz, line 29: I don't understand what happened here and why is it in this PR? We don't need `embed` anymore? This PR changes that all over the place - it should be separate PR doing this refactor.

app/Commands/Dev/Geb/Infer.hs
```diff
@@ -27,7 +27,6 @@ runCommand opts = do
             Geb.ppOut
               opts
               (tyMorph ^. Geb.typedMorphismObject)
-          embed $ putStrLn ""
```
lukaszcz, line 30: Why removing newline here (not only embed)?

app/Commands/Dev/Geb/Repl.hs
```diff
@@ -114,25 +114,25 @@ checkTypedMorphism gebMorphism = Repline.dontCrash $ do
     Right _ -> printError (error "Checking only works on typed Geb morphisms.")
 
 runReplCommand :: String -> Repl ()
-runReplCommand input =
+runReplCommand input_ =
```
lukaszcz, line 117: Analogously with the `input` that somehow appeared in `Prelude` or somewhere in this PR and requires adjusting variable names all over the place. This is an unrelated (?) refactor.

scripts/nockma-stdlib-parser.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Usage:
+# chmod +x nockma-stdlib-parser.sh
+# ./nockma-stdlib-parser < anoma.hoon
+
+# Use grep to find lines matching the pattern and awk to format the output
+# Reads from stdin
+grep -oP '\+\+  \K\w+\s+::\s+\+\d+' | awk '{gsub("::  \\+", ""); print}'
```
lukaszcz, line 8: What does this script do? There should be a comment.

src/Juvix/Compiler/Asm/Transformation/TempHeight.hs
```diff
@@ -0,0 +1,76 @@
...
+import Juvix.Compiler.Asm.Data.Stack
+import Juvix.Compiler.Asm.Transformation.Base
+
+computeFunctionTempHeight ::
+  (Member (Error AsmError) r) =>
+  InfoTable ->
+  FunctionInfo ->
+  Sem r FunctionInfo
+computeFunctionTempHeight tab fi = do
+  ps :: [Command] <- recurseFun sig fi
```
lukaszcz, line 12: This can be done with the simplified recursor `recurseS`. You don't need full memory information & validity checking here -- just stack height info. I'll change it.

src/Juvix/Compiler/Backend/Geb/Translation/FromSource.hs
```diff
@@ -25,23 +25,23 @@ fromSource ::
   Path Abs File ->
   Text ->
   Sem r Geb.Expression
-fromSource fileName input =
-  case runParser fileName input of
+fromSource fileName input' =
+  case runParser fileName input' of
```
lukaszcz, line 29: These `input` name changes really should be in a separate PR, if possible. It's too many of them,

src/Juvix/Compiler/Backend/Html/Translation/FromTyped.hs
```diff
@@ -288,7 +288,7 @@ goTopModule cs m = do
   htmlOpts <- ask @HtmlOptions
   runReader (htmlOpts {_htmlOptionsKind = HtmlDoc}) $ do
     fpath <- moduleDocPath m
-    Prelude.embed (putStrLn ("Writing " <> pack (toFilePath fpath)))
+    putStrLn ("Writing " <> pack (toFilePath fpath))
```
lukaszcz, line 291: Same with removing `embed`

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
```diff
@@ -82,8 +80,8 @@ fromSource e = mapError (JuvixError @ParserError) $ do
       Path Abs File ->
       Sem r (Module 'Parsed 'ModuleTop)
     goFile fileName = do
-      input <- getFileContents fileName
-      mp <- runModuleParser fileName input
+      input_ <- getFileContents fileName
+      mp <- runModuleParser fileName input_
```
lukaszcz, line 84: Unrelated (?) name changes.

src/Juvix/Compiler/Core/Translation/FromSource.hs
```diff
@@ -25,16 +25,16 @@ import Text.Megaparsec qualified as P
 -- | Note: only new symbols and tags that are not in the InfoTable already will be
 -- generated during parsing
 runParser :: Path Abs File -> ModuleId -> InfoTable -> Text -> Either MegaparsecError (InfoTable, Maybe Node)
-runParser fileName mid tab input =
+runParser fileName mid tab input_ =
```
lukaszcz, line 28: Name change

src/Juvix/Compiler/Nockma/Evaluator.hs
```diff
@@ -44,14 +45,15 @@ subTermT = go
 subTerm :: (Member (Error NockEvalError) r) => Term a -> Path -> Sem r (Term a)
 subTerm term pos = do
   case term ^? subTermT pos of
-    Nothing -> throw InvalidPath
+    -- Nothing -> throw (InvalidPath "subterm")
```
lukaszcz, line 48: Remove comment

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+type Offset = Natural
+
+data CompilerFunction = CompilerFunction
+  { _compilerFunctionName :: FunctionId,
+    _compilerFunctionArity :: Natural,
+    _compilerFunction :: Sem '[Compiler, Reader CompilerCtx] ()
+  }
+
+data StackId
+  = CurrentFunction
```
lukaszcz, line 58: What do you need to store the current function for? "Current function" is never referenced in JuvixAsm - functions are always referenced by name.

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+  | ValueStack
+  | TempStack
+  | AuxStack
```
lukaszcz, line 61: What's an `AuxStack`, what do you store there?

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+    output (pushOnStack FrameStack frame)
+
+  -- 3.
+  --  i. Take a copy of the function from the function library
+  --  ii. Replace its argument area with the arguments from the value stack
+  --  iii. Push this copy to the current function stack
+
+  -- Setup function to call with its arguments
+  -- given n, we compute [R..R] of length n
+  if
```
lukaszcz, line 704: Oh dear, you're emulating a call stack on top of a Nock VM! No wonder it's slow. You should not be handling the call stack explicitly -- Nock VM already does this and you should use it.

A call should just change the Nock subject (transferring the arguments, etc) and evaluate the compiled function call with the new subject. The result of the evaluation should just push to the top of the value stack in the calling function.

The `ret` instruction is then just identity -- returns the top of the value stack. I perhaps should've made it more clear that in JuvixAsm `ret` can only appear in tail positions (like `tcall`).

For example,
```
push arg1
push arg2
call f;
P;
```
should be translated to Nock similar to:
```
[Push arg1*
  [Push arg2*
    [Push
        ["replace subject with new one, with 2 arguments from top of the stack"; fetch & eval f]
        ["remove the 2nd & 3rd value from top of the value stack (the arguments) - that's one indexing and one replace in Nock"; P*]
    ]
  ]
]
```
where `X*` is the translation of `X`, and `Push` is the nock "push" combinator 8.

In fact, almost always the call in JuvixAsm will be like above with the pushes corresponding to the number of arguments. Then we can compile this to slightly better:
```
[Push
  [Push arg1*
    [Push arg2*
      ["replace subject with new one, with 2 arguments from top of the stack"; fetch & eval f]
    ]
  ]
  P*
]
```
(we could in fact always recreate the original call tree and avoid removing the arguments after call return).

It also seems merging the stacks as I initially intended (#2559) might lead to a simpler nock code for "replace subject with new one" (but it's a bit more difficult, because you need to keep track of the value/temporary stack heights).

It matters a lot here whether you generate 1 or 10 instructions for a common operation. This translates to roughly an order of magnitude difference in the running time of generated programs.

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+  Maybe (Sem r ()) ->
+  [(Asm.Tag, Sem r ())] ->
+  Sem r ()
+caseCmd defaultBranch = \case
+  [] -> sequence_ defaultBranch
+  (tag, b) : bs -> do
+    -- push the constructor tag at the top
+    push (OpAddress # topOfStack ValueStack ++ constructorPath ConstructorTag)
+    push (constructorTagToTerm tag)
+    testEq
```
lukaszcz, line 839: Why do you need to push the stack twice to test for equality, just to immediately replace the pushed values with the equality result? We have `[= X Y]` in nock, so just use `[= [@ S] [K TagValue]]`. That would make testing equality several times faster (or more, because slicing/replacement might not be so cheap in Nock VM).

We can't write this the way one writes Haskell, hoping that it'll somehow get optimized later. There's nothing that can optimize this -- we're generating the code and it's all common basic operations.

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+    branch b (caseCmd defaultBranch bs)
```
lukaszcz, line 840: Similarly with this branch. It doesn't need to take the value from the stack, because we know what it's branching on -- we can just do something like `[If [= [@ S] [K TagValue]] B1 B2]`.

src/Juvix/Compiler/Nockma/Translation/FromSource/Base.hs
```diff
@@ -36,7 +35,7 @@ runParserProgram :: FilePath -> Text -> Either MegaparsecError (N.Program Natura
 runParserProgram = runParserFor program
 
 runParserFor :: Parser a -> FilePath -> Text -> Either MegaparsecError a
-runParserFor p f input = case P.runParser (spaceConsumer >> p <* eof) f input of
+runParserFor p f input' = case P.runParser (spaceConsumer >> p <* eof) f input' of
```
lukaszcz, line 38: Name change

test/BackendGeb/Eval/Base.hs
```diff
@@ -63,8 +63,8 @@ gebEvalAssertion' _mainFile expectedFile step gebMorphism = do
 gebEvalErrorAssertion :: Path Abs File -> (String -> IO ()) -> Assertion
 gebEvalErrorAssertion mainFile step = do
   step "Parse"
-  input <- readFile (toFilePath mainFile)
-  case Geb.runParser mainFile input of
+  input' <- readFile (toFilePath mainFile)
+  case Geb.runParser mainFile input' of
```
lukaszcz, line 67: Name change

test/BackendGeb/FromCore/Base.hs
```diff
@@ -17,9 +17,9 @@ coreToGebTranslationAssertion ::
   Assertion
 coreToGebTranslationAssertion root mainFile expectedFile step = do
   step "Parse Juvix Core file"
-  input <- readFile . toFilePath $ mainFile
+  input' <- readFile . toFilePath $ mainFile
   entryPoint <- set entryPointTarget TargetGeb <$> testDefaultEntryPointIO root mainFile
-  case Core.runParserMain mainFile defaultModuleId mempty input of
+  case Core.runParserMain mainFile defaultModuleId mempty input' of
```
lukaszcz, line 22: name changes

test/Scope/Positive.hs
```diff
@@ -50,7 +50,7 @@ testDescr PosTest {..} = helper renderCodeNew
                 entryPoint <- testDefaultEntryPointIO tRoot file'
 
                 let evalHelper :: Text -> Sem (PipelineEff PipelineAppEffects) a -> IO (PipelineResult a)
-                    evalHelper input m = snd <$> testRunIO entryPoint {_entryPointStdin = Just input} m
+                    evalHelper input' m = snd <$> testRunIO entryPoint {_entryPointStdin = Just input'} m
```
lukaszcz, line 53: name change

lukaszcz (changes requested):
Move the unrelated changes to a separate refactoring PR (`embed` removal, `input` name change).

src/Juvix/Compiler/Nockma/Translation/FromAsm.hs
```diff
@@ -0,0 +1,1048 @@
...
+  deriving stock (Enum, Bounded, Eq, Show)
+
+-- | A closure has the following structure:
+-- [code totalArgsNum argsNum args], where
+-- 1. code is code to run when fully applied.
+-- 2. totalArgsNum is the number of arguments that the function
+--     which created the closure expects.
+-- 3. argsNum is the number of arguments that have been applied to the closure.
+-- 4. args is the list of args that have been applied.
+--    The length of the list should be argsNum.
```
lukaszcz, line 74: To enable fast append (with one replacement operation) we might want to store the path to the end of supplied arguments list

lukaszcz: Or we can store a list of argument sequences. So we have (in pseudo-Haskell data notation)
```
data Closure = Closure {
  code :: Nock,
  -- stored args
  args :: NonEmpty (PathToEnd, NonEmpty Value),
  -- number of expected arguments (to be still supplied to the closure - when it reaches 0 we do the call),
  argsNum :: Int
}
```
The extending the closure args is just list cons (we know the number of supplied arguments, so we can precompute PathToEnd at compile time):
```
args' = (PathToEndOfSuppliedArgs, suppliedArgs) : args
```
At the end transferring the closure args to function args requires going through the list recursively and appending:
```
flatten ((p, args) : as) = replace p args (flatten as)
```
I think we can't get around doing the appending at least once.

Closure extensions are rare, however, so we might want to have a data structure for the common special case:
```
data FlatClosure = FlatClosure {
  code :: Nock,
  -- path to end of stored args, for closure extension
  pathToEnd :: Path,
  -- stored args
  args :: NonEmpty Value,
  -- number of expected arguments (to be still supplied to the closure - when it reaches 0 we do the call),
  argsNum :: Int
}
```

app/Commands/Dev/Core/Repl.hs
paulcadman: Done

app/Commands/Dev/Geb/Infer.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

app/Commands/Dev/Geb/Repl.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

scripts/nockma-stdlib-parser.sh
paulcadman: Added a comment to the script

src/Juvix/Compiler/Backend/Geb/Translation/FromSource.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

src/Juvix/Compiler/Backend/Html/Translation/FromTyped.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

src/Juvix/Compiler/Concrete/Translation/FromSource.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

src/Juvix/Compiler/Core/Translation/FromSource.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

src/Juvix/Compiler/Nockma/Translation/FromSource/Base.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

test/BackendGeb/Eval/Base.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

test/BackendGeb/FromCore/Base.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

test/Scope/Positive.hs
paulcadman: This change was moved to https://github.com/anoma/juvix/pull/2582

src/Juvix/Compiler/Nockma/Evaluator.hs
paulcadman: Done

lukaszcz (approved):
Okay, there are still a few unrelated refactors, but let's merge it

PR #2580, opened 2024-01-11
Fast nockma eval

janmasrovira:
Adds annotations to cells to indicate that it is a call to the stdlib and might be evaluated faster in the Haskell evaluator.

The syntax for stdlib calls is as follows:
```
[stdlib@add args@<args-term> <left-term> <right-term>]
```
where `add` is the name of the function being called, `<args-term>` is a nockma term that points to the position of the arguments, and `<left-term>` and `<right-term>` are the actual components of the cell.

### DIFF
--- a/test/Nockma/Compile/Positive.hs
+++ b/test/Nockma/Compile/Positive.hs
@@ -232,6 +232,18 @@ tests =
       pushNat 33333
       pushNat 22222
       add,
+    Test "mul big" (eqStack ValueStack [nock| [1111088889 nil] |]) $ do
+      pushNat 33333
+      pushNat 33333
+      mul,
+    Test "sub big" (eqStack ValueStack [nock| [66666 nil] |]) $ do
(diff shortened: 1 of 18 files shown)


test/Nockma/Compile/Positive.hs
```diff
@@ -167,7 +172,7 @@ eqStack st = eqSubStack st []
 unfoldTerm :: Term Natural -> NonEmpty (Term Natural)
 unfoldTerm t = case t of
   TermAtom {} -> t :| []
-  TermCell Cell {..} -> _cellLeft NonEmpty.<| unfoldTerm _cellRight
+  TermCell (Cell l r) -> l NonEmpty.<| unfoldTerm r
```
lukaszcz, line 175: Why this change? Doesn't it go against our general style (of using named fields)?

lukaszcz (approved):

PR #2586, opened 2024-01-22
Support  `MemRepTuple` in the Nockma backend

janmasrovira:
We can represent Anoma types like [resource](https://github.com/anoma/anoma/blob/e18e50e3c380b2867da6cb81f427334eb5c0b2d9/hoon/resource-machine.hoon#L7) as Juvix records.

The Nockma encoding of types uses Nockma 'tuples' where each component of the tuple holds a value of a field. So for Juvix->Anoma integration it is convenient to compile values of record types as Nockma tuples.

We already have the concept of representing constructors of inductive types that have only one non-zero-field constructor in the compiler, see [`MemRepTuple`](https://github.com/anoma/juvix/blob/1147e1fce131f5b423fd558e5636e6aaf23120ac/src/Juvix/Compiler/Tree/Language/Rep.hs#L13).

In this PR, as part of the Nockma step, we mark constructors that satisfy the requirements of the `MemRepTuple` translation as such. Then we use a tuple encoding for those constructors.

### DIFF
--- a/src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
+++ b/src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
@@ -46,6 +46,7 @@ data ConstructorInfo = ConstructorInfo
     _constructorTag :: Tag,
     _constructorType :: Type,
     _constructorArgNames :: [Maybe Text],
+    _constructorArgsNum :: Int,
     _constructorFixity :: Maybe Fixity
   }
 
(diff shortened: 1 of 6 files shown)


src/Juvix/Compiler/Core/Data/Stripped/InfoTable.hs
```diff
@@ -46,6 +46,7 @@ data ConstructorInfo = ConstructorInfo
     _constructorTag :: Tag,
     _constructorType :: Type,
     _constructorArgNames :: [Maybe Text],
+    _constructorArgsNum :: Int,
```
lukaszcz, line 49: `_constructorArgsNum` is always equal to `length (typeArgs _constructorType)`, I guess? There should be a comment that this field is only a "cache" for this value.

lukaszcz (approved):

PR #2583, opened 2024-01-16
JuvixTree parser and pretty printer

lukaszcz:
This PR implements:
* JuvixTree parser.
* JuvixTree pretty printer.
* `juvix dev tree read file.jvt` command which reads and pretty prints a JuvixTree file.
* The `tree` target in the `compile` command.
* Removal of `StackRef` in JuvixAsm. This makes JuvixAsm more consistent with JuvixTree and simplifies the data structures. `StackRef` is not needed for compilation from Core.

Tests for the parser will appear in a separate PR, when I implement an automatic translation of JuvixAsm to JuvixTree files.

PR #2590, opened 2024-01-23
Add the `juvix dev tree compile` command

lukaszcz:
* Adds the `juvix dev tree compile` CLI command.
* Depends on #2589 
* Depends on #2587 
* Depends on #2583

### DIFF
--- a/app/Commands/Dev/Tree/Compile/Base.hs
+++ b/app/Commands/Dev/Tree/Compile/Base.hs
@@ -1,0 +1,93 @@
+module Commands.Dev.Tree.Compile.Base where
+
+import Commands.Base
+import Commands.Dev.Tree.Compile.Options
+import Commands.Extra.Compile qualified as Compile
+import Juvix.Compiler.Asm.Pretty qualified as Asm
+import Juvix.Compiler.Backend qualified as Backend
+import Juvix.Compiler.Backend.C qualified as C
+import Juvix.Compiler.Nockma.Pretty qualified as Nockma
+import Juvix.Compiler.Tree.Data.InfoTable qualified as Tree
+
+data PipelineArg = PipelineArg
+  { _pipelineArgOptions :: CompileOptions,
+    _pipelineArgFile :: Path Abs File,
+    _pipelineArgTable :: Tree.InfoTable
+  }
+
+getEntry :: (Members '[Embed IO, App, TaggedLock] r) => PipelineArg -> Sem r EntryPoint
+getEntry PipelineArg {..} = do
+  ep <- getEntryPoint (AppPath (preFileFromAbs _pipelineArgFile) True)
+  return $
+    ep
+      { _entryPointTarget = getTarget (_pipelineArgOptions ^. compileTarget),
+        _entryPointDebug = _pipelineArgOptions ^. compileDebug,
+        _entryPointUnsafe = _pipelineArgOptions ^. compileUnsafe,
+        _entryPointOptimizationLevel = fromMaybe defaultOptLevel (_pipelineArgOptions ^. compileOptimizationLevel),
+        _entryPointInliningDepth = _pipelineArgOptions ^. compileInliningDepth
+      }
+  where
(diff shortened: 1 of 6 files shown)


paulcadman (dismissed):

app/Commands/Dev/Tree/Compile/Base.hs
```diff
@@ -0,0 +1,93 @@
...
+runCPipeline ::
+  forall r.
+  (Members '[Embed IO, App, TaggedLock] r) =>
+  PipelineArg ->
+  Sem r ()
+runCPipeline pa@PipelineArg {..} = do
+  entryPoint <- getEntry pa
+  C.MiniCResult {..} <- getRight (run (runReader entryPoint (runError (treeToMiniC _pipelineArgTable :: Sem '[Error JuvixError, Reader EntryPoint] C.MiniCResult))))
+  cFile <- inputCFile _pipelineArgFile
+  embed @IO (writeFile (toFilePath cFile) _resultCCode)
```
janmasrovira, line 55: we recently added `writeFileEnsureLn` to the prelude. Should we use that normally? I'd say yes

app/Commands/Dev/Tree/Compile/Base.hs
```diff
@@ -0,0 +1,93 @@
...
+      | otherwise = defaultOptimizationLevel
+
+runCPipeline ::
+  forall r.
+  (Members '[Embed IO, App, TaggedLock] r) =>
+  PipelineArg ->
+  Sem r ()
+runCPipeline pa@PipelineArg {..} = do
+  entryPoint <- getEntry pa
+  C.MiniCResult {..} <- getRight (run (runReader entryPoint (runError (treeToMiniC _pipelineArgTable :: Sem '[Error JuvixError, Reader EntryPoint] C.MiniCResult))))
```
janmasrovira, line 53: this line would look nicer with the `f . g $ x` pattern

janmasrovira, line 53: this line would look nicer with the `f . g $ x` shape.
Also, we usually put type annotations next to the run function. Something like
```suggestion
  C.MiniCResult {..} <- getRight 
        . run 
        . runReader entryPoint 
        . runError @JuvixError
        $ treeToMiniC _pipelineArgTable
```

app/Commands/Dev/Tree/Compile/Options.hs
```diff
@@ -0,0 +1,23 @@
...
+  )
+where
+
+import Commands.Extra.Compile.Options
+import CommonOptions
+import Data.List.NonEmpty qualified as NonEmpty
+
+treeSupportedTargets :: NonEmpty CompileTarget
+treeSupportedTargets =
+  NonEmpty.fromList
```
janmasrovira, line 13: we have `nonEmpty'` in the prelude to avoid having to import `Data.List.NonEmpty` that often

app/Commands/Dev/Tree/Compile/Base.hs
lukaszcz: I will leave this for future refactor. There are 3 places just in this single file with `writeFile` used this way, and I think it's better if we change all of them at once, to be consistent.

app/Commands/Dev/Tree/Compile/Base.hs
lukaszcz: done

app/Commands/Dev/Tree/Compile/Options.hs
lukaszcz: done

app/Commands/Dev/Tree/Compile/Base.hs
lukaszcz: This was copied and modified from `Commands/Dev/Core/Compile/Base.hs`, which perhaps should also be changed in a future refactor.

janmasrovira (approved):

PR #2604, opened 2024-01-27
Use `writeFileEnsureLn` in place of `writeFile`

janmasrovira:
Now we guarantee that whenever we write a file there is a newline character at the end, which is a [Unix convention](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205).

The juvix prelude now exports `writeFileEnsureLn` and it no longer exports `writeFile`. If at some point we need the behaviour of `writeFile` I'd suggest that we export it renamed as `writeFileVerbatim` or something similar.

### DIFF
--- a/src/Juvix/Prelude/Base.hs
+++ b/src/Juvix/Prelude/Base.hs
@@ -150,7 +150,8 @@ import Data.Text qualified as Text
 import Data.Text.Encoding
 import Data.Text.IO hiding (appendFile, putStr, putStrLn, readFile, writeFile)
 import Data.Text.IO qualified as Text
-import Data.Text.IO.Utf8
+import Data.Text.IO.Utf8 hiding (writeFile)
+import Data.Text.IO.Utf8 qualified as Utf8
 import Data.Traversable
 import Data.Tuple.Extra hiding (both)
 import Data.Type.Equality (type (~))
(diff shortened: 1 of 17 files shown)


src/Juvix/Prelude/Base.hs
```diff
@@ -589,11 +590,14 @@ runInputInfinite s =
...
-  let t' = case Text.unsnoc t of
-        Nothing -> t
-        Just (_, y) -> case y of
-          '\n' -> t
-          _ -> Text.snoc t '\n'
-   in writeFile p t'
+ensureLn :: Text -> Text
+ensureLn t =
+  case Text.unsnoc t of
+    Nothing -> t
```
paulcadman, line 596: Something I didn't think about at the time - should we add a newline if the text is empty?

janmasrovira: I don't think so. See [definition of Text File](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_403).

> A file that contains characters organized into zero or more lines. [..]

It is allowed for a text file to have no lines, then the newline character should not be added.

paulcadman: That's right! Thanks

paulcadman (approved):

Issue #2606, opened 2024-01-29
Translation from JuvixTree to Nockma


paulcadman:
JuvixTree is a new intermediate language in the compiler between JuvixCore and JuvixAsm (see https://github.com/anoma/juvix/pull/2581). JuvixTree has been designed to make it simpler to write an efficient compilation to Nockma.

The work for this issue is to replace the JuvixAsm -> Nockma compilation with JuvixTree -> Nockma.

PR #2614, opened 2024-01-31
Translation from `Juvix Tree` to `Nockma`

janmasrovira:
This PR replaces the JuvixAsm -> Nockma translation with a JuvixTree -> Nockma translation.

We can now enable some of the JuvixTree tests that did not work with JuvixAsm->Nockma because they were too slow. 

## Notes

We have changed [test031: temp stack with branching](https://github.com/anoma/juvix/blob/22ee87f0e7b82d44bc467014c4c1d26b9f4c2ce2/tests/Tree/positive/test031.jvt) to avoid using negative numbers (because negative integers are not supported in Nockma). 

Three tree tests trace/output lists. Lists are serialised differently by the asm and nockma pretty printers so they cannot share a single test output file. We have created separate nockma output files for these tests (see eg. [test028.nockma.out](https://github.com/anoma/juvix/blob/22ee87f0e7b82d44bc467014c4c1d26b9f4c2ce2/tests/Tree/positive/out/test028.nockma.out)).

* Closes https://github.com/anoma/juvix/issues/2606


lukaszcz (approved):
Well done! Now the JuvixTree tests can be compiled to Nockma and evaluated in reasonable time. Closures are still very slow, and maybe we'll need to think harder about them in the future, but for now it's fine.

We should still do in separate PRs:
* Tests for the Nockma backend that go all the way from Juvix input, not only from the intermediate JuvixTree language. The native/WASM compilation pipeline tests could be modified for this purpose.
* Make the tests run the external Nockma interpreter in addition to our internal one.

PR #2616, opened 2024-02-01
JuvixTree validation

lukaszcz:
* Validation (type checking) of JuvixTree. Similar to JuvixAsm validation, will help with debugging.
* Depends on #2608

### DIFF
--- a/src/Juvix/Compiler/Tree/Transformation/Validate.hs
+++ b/src/Juvix/Compiler/Tree/Transformation/Validate.hs
@@ -46,8 +46,8 @@ inferType tab funInfo = goInfer mempty
         checkBinop ty1' ty2' rty = do
           ty1 <- goInfer bl _nodeBinopArg1
           ty2 <- goInfer bl _nodeBinopArg2
-          _ <- unifyTypes' loc tab ty1 ty1'
-          _ <- unifyTypes' loc tab ty2 ty2'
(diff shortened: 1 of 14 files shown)


src/Juvix/Compiler/Tree/Transformation/Validate.hs
```diff
@@ -0,0 +1,267 @@
...
+        goInfer bl _nodeBinopArg2
+      where
+        loc = _nodeBinopInfo ^. nodeInfoLocation
+
+        checkBinop :: Type -> Type -> Type -> Sem r Type
+        checkBinop ty1' ty2' rty = do
+          ty1 <- goInfer bl _nodeBinopArg1
+          ty2 <- goInfer bl _nodeBinopArg2
+          _ <- unifyTypes' loc tab ty1 ty1'
+          _ <- unifyTypes' loc tab ty2 ty2'
```
janmasrovira, line 50: I'd suggest to either use `void` or define `unifyTypes'_`

lukaszcz: Thanks, I forgot how `void` is called in Haskell and somehow the web search didn't turn it up

lukaszcz: done

janmasrovira (approved):

PR #2648, opened 2024-02-13
Fix `writeFileEnsureLn`

janmasrovira:
- This fixes a regression introduced in #2624, where the newline was not being inserted


paulcadman:
We tried to fix this but there's some bad interaction when wrapping IO actions in our TaggedLocks. Hopefully this will be fixed when moving from polysemy to effectful. Closing this for now.

PR #2650, opened 2024-02-14
JuvixReg transformation: initialize variables assigned in other branches

### LINKED ISSUE #2576: Insert assignments in JuvixReg to guarantee continuous memory use
Cairo requires memory to be used continuously, e.g., one cannot assign to memory at address 100 and leave memory at address 99 uninitialized at the end of the program (this is checked by the Cairo runner). Hence, we cannot leave any temporary variables in JuvixReg uninitialized. We need to implement a transformation on JuvixReg that initializes variables not assigned in an if/case branch, but assigned in other branches (variables assigned in only one branch will appear e.g. as a result of transforming JuvixReg into SSA form).

lukaszcz:
* Closes #2576 
* Adds a JuvixReg transformation `InitBranchVars` which inserts assignments to initialize variables assigned in other branches. Assumes the input is in SSA form (which is preserved).
* Adds tests for the `InitBranchVars` transformation.
* Depends on #2647


src/Juvix/Prelude/Base.hs
```diff
@@ -365,6 +365,11 @@ zip4Exact [] [] [] [] = []
 zip4Exact (x1 : t1) (x2 : t2) (x3 : t3) (x4 : t4) = (x1, x2, x3, x4) : zip4Exact t1 t2 t3 t4
 zip4Exact _ _ _ _ = error "zip4Exact"
 
+allEqual :: (Eq a) => [a] -> Bool
```
paulcadman, line 368: This function exists in the `extra` library (which Juvix depends on), so we could use/export this instead.

https://hackage.haskell.org/package/extra-1.7.14/docs/Data-List-Extra.html#v:allSame

lukaszcz: Done

paulcadman (approved):

PR #2652, opened 2024-02-14
Support compilation to Anoma compatible functions

paulcadman:
This PR adds a `anoma` target to the `juvix compile`. This target compiles a Juvix `main` function to a Nockma/Anoma "function". Unlike the native, wasm, and nockma targets the main function may have any type signature.

## Anoma calling convention

Anoma calls functions by evaluating the formula `[call L replace [RL args] @ S]` against a subject equal to the function. Here `args` is a Nockma term that evaulates to a tuple of arguments that should be passed to the function.

The anoma target compiles the `main` function to Nockma in the same way as the nockma target. The main function is then [wrapped](https://github.com/anoma/juvix/blob/9a658465aebe687565ec4394f965983bb66295a1/src/Juvix/Compiler/Nockma/Translation/FromTree.hs#L627) to make it compatible with the Anoma calling convention.

## Testing

The anoma calling convention is [unit tested](https://github.com/anoma/juvix/blob/9a658465aebe687565ec4394f965983bb66295a1/test/Nockma/Eval/Positive.hs#L117) and [smoke tested](https://github.com/anoma/juvix/blob/9a658465aebe687565ec4394f965983bb66295a1/tests/smoke/Commands/compile.smoke.yaml#L159).

## Nockma REPL

To facilitate testing the Nockma REPL can now load a nockma file as an initial subject.


src/Juvix/Compiler/Core/Data/TransformationId.hs
```diff
@@ -69,7 +69,10 @@ toVampIRTransformations = [CombineInfoTables, FilterUnreachable, CheckVampIR, Le
 
 toStrippedTransformations :: [TransformationId]
 toStrippedTransformations =
-  [CombineInfoTables, FilterUnreachable, CheckExec, LambdaLetRecLifting, TopEtaExpand, OptPhaseExec, MoveApps, RemoveTypeArgs]
+  [CombineInfoTables, FilterUnreachable, LambdaLetRecLifting, TopEtaExpand, OptPhaseExec, MoveApps, RemoveTypeArgs]
```
lukaszcz, line 72: The order was significant here -- "illegal" things (for `CheckExec` - axioms) may get filtered out or brought in from imported modules. Plus, there should be similar checks for Anoma/Nockma and later Cairo. I'll make a commit to this PR to refactor the transformation pipeline to allow for different kinds of checks.

paulcadman: I see - we moved the CheckExec to a toExec transformation (https://github.com/anoma/juvix/blob/8f7a6426955d2fc893c12dc2ac58c0ef5ca322ba/src/Juvix/Compiler/Pipeline.hs#L178) to allow Anoma main functions to have arbitrary signatures. But in the case of Anoma we want to check that axioms are not present so I think your refactor is a good idea, thanks.

test/Anoma/Compilation/Positive.hs
```diff
@@ -0,0 +1,522 @@
...
+  where
+    mkTestIO :: IO Test
+    mkTestIO = do
+      _testProgramSubject <- withRootCopy compileMain
+      let _testProgramFormula = anomaCall args
+          _testEvalOptions = defaultEvalOptions
+      return Test {..}
+
+    withRootCopy :: (Prelude.Path Abs Dir -> IO a) -> IO a
+    withRootCopy action = withSystemTempDir "test" $ \tmpRootDir -> do
```
lukaszcz, line 28: The stored standard library is not shared between different tests, which defeats per-module compilation and makes the tests run 3-4 times slower than they otherwise would.

tests/Anoma/Compilation/positive/test007.juvix
```diff
@@ -0,0 +1,27 @@
...
+
+map' {A B} (f : A → B) : List A → List B :=
+  \ {
+    | nil := nil
+    | (h :: t) := f h :: map' f t
+  };
+
+terminating
+map'' {A B} {{Partial}} (f : A → B) (x : List A) : List B :=
+  if (null x) nil (f (phead x) :: map'' f (tail x));
```
lukaszcz, line 15: Compiling this to Nockma relies on Core-level optimizations. The function `phead` can internally call a `fail` function which takes a string as argument. If `Partial` is not specialised and this internal call to `fail` is not inlined to directly call the builtin, you'll be translating strings and get a crash.

lukaszcz:
I added checks for strings in the Anoma pipeline, which makes the 4 tests indirectly depending on them fail. They compiled before because you had Core-level optimizations turned on, which inlined the functions taking strings, but we shouldn't rely on this.

src/Juvix/Compiler/Core/Data/TransformationId.hs
lukaszcz: I commited a refactor.

paulcadman:
> I added checks for strings in the Anoma pipeline, which makes the 4 tests indirectly depending on them fail. They compiled before because you had Core-level optimizations turned on, which inlined the functions taking strings, but we shouldn't rely on this.

Should these tests be removed then?

test/Anoma/Compilation/Positive.hs
paulcadman: Running the tests concurrently in the same directory triggers the leaking IO bypassing the TaggedLock that we've seen before. It's possible that using Effectful will fix this. An alternative would be to synchronously build the standard library somehow before running the test suite and copy that to each temporary location.

lukaszcz:
> > I added checks for strings in the Anoma pipeline, which makes the 4 tests indirectly depending on them fail. They compiled before because you had Core-level optimizations turned on, which inlined the functions taking strings, but we shouldn't rely on this.
> 
> Should these tests be removed then?

I don't know, maybe they can be fixed, e.g., by removing functions depending on `Partial` (test 007 definitely can be). I didn't look too much into them, leaving this decision to you.

test/Anoma/Compilation/Positive.hs
lukaszcz: Somehow, it works for the Juvix->Native/Wasm compilation tests (in `tests/Compilation/positive`). What's different?

I remember I've been dealing specifically with this issue when implementing per-module-compilation, and ultimately used file-level locking which solved the problem.

lukaszcz: I don't remember what's happening exactly for the main exec pipeline compilation tests, but the stored standard library modules seem to be reused judging by decrease in runtime after first test.

tests/Anoma/Compilation/positive/test007.juvix
paulcadman: We'd like to keep support for Strings so we can intercept messages from builtin fail: https://github.com/anoma/juvix/issues/2665

We've re-enabled support for PrimString in the Anoma target and refactored the failing tests to use `failwith` (i.e builtin fail) directly instead of via the Partial trait.

lukaszcz: I don't completely understand the rationale behind this replacement -- if you have strings in the Nockma backend you can use `Partial`, if you don't you can't use `failwith` either.

test/Anoma/Compilation/Positive.hs
paulcadman: The anoma compilation tests are using the same functions to run the pipeline as the main exec compilation tests, yet they fail for the leaking IO / tagged lock issue. TaggedLock in general is too unreliable and I'm thinking about a plan to remove it. I propose to leave the anoma compilation tests as they are (using separate directories) for now until we work out how we will replace tagged lock.

tests/Anoma/Compilation/positive/test007.juvix
lukaszcz: > but we can support builtin fail because it's compiled directly to `OpFail` in Tree

I think that's not an entirely correct assumption, because of eta-expansions in the pipeline. So lambda-abstractions with string argument type might be added on top of the builtin, and the string argument would be applied to a lambda-function not to the builtin directly. Otherwise my checks for strings in `checkBuiltins'` wouldn't fail with only `failwith` because `OpFail` is already handled specially there. It might work because of subsequent optimisation which inlines the eta-expanded functions.

paulcadman: I see. I'll replace the PrimString check and modify the tests to avoid using Partial/failwith.

paulcadman: We've now modified the tests to avoid using Partial/failwith and restored the Strings check in the anoma backend.

src/Juvix/Compiler/Pipeline.hs
```diff
@@ -119,21 +119,33 @@ upToGeb ::
 upToGeb spec =
   upToStoredCore >>= \Core.CoreResult {..} -> storedCoreToGeb spec _coreResultModule
 
+upToAnoma ::
+  (Members '[HighlightBuilder, Reader Parser.ParserResult, Reader EntryPoint, Reader Store.ModuleTable, Files, NameIdGen, Error JuvixError, GitClone, PathResolver] r) =>
+  Sem r (Nockma.Term Natural)
+upToAnoma = upToStoredCore >>= \Core.CoreResult {..} -> Nockma.TermCell <$> coreToAnoma _coreResultModule
```
lukaszcz, line 125: It should be `storedCoreToAnoma` (using `storedCoreToTree`) instead of `coreToAnoma`. Otherwise, we're doing the `Core.toStored` transformations twice.

lukaszcz (changes requested):
Change [src/Juvix/Compiler/Pipeline.hs](https://github.com/anoma/juvix/pull/2652/files/ac795111bc167bd0698fd5dc71f3ba1293ca6da5#diff-a6521e11f7a18d421e424b4b31087e5008aa2ea6bad1246f3a5fb01eeb1df28c) as discussed in the comment.

src/Juvix/Compiler/Pipeline.hs
paulcadman: Thank you. Done

lukaszcz (approved):

Issue #2562, opened 2023-12-10
Translation from JuvixReg to Cairo ASM


lukaszcz:
* Translate from JuvixReg in SSA form to Cairo ASM. Since Juvix is purely functional, this fits well with the Cairo read-only memory model. We never need to modify or free memory - we allocate it once and assign the necessary fields. Allocate a new stack frame for each function call, which stores the arguments and local variables.
* Test the compilation pipeline to Cairo ASM with the Cairo ASM interpreter.
* Depends on #2561 
* Depends on #2555

PR #2659, opened 2024-02-16
Add field element type

### LINKED ISSUE #2571: Add field element type
We should have a field element type for VampIR and Cairo.

Currently, we have an integer type which is treated inconsistently: unbounded integer in the Core, 24bit integers for VampIR, 31bit integers without overflow check for the WASM/native backend. It should always be an unbounded integer type, and VampIR should instead use the field element type.

lukaszcz:
* Closes #2571
* It is reasonable to finish this PR before tackling #2562, because the field element type is the primary data type in Cairo. 
* Depends on #2653

Checklist
---------

- [x] Add field type and operations to intermediate representations (JuvixCore, JuvixTree, JuvixAsm, JuvixReg).
- [x] Add CLI option to choose field size.
- [x] Add frontend field builtins.
- [x] Automatic conversion of integer literals to field elements.
- [x] Juvix standard library support for fields.
- [x] Check if field size matches when loading a stored module.
- [x] Update the Cairo Assembly (CASM) interpreter to use the field type instead of integer type.
- [x] Add field type to VampIR backend.
- [x] Tests


package.yaml
```diff
@@ -63,6 +63,7 @@ dependencies:
   - file-embed == 0.0.*
   - filelock == 0.1.*
   - filepath == 1.4.*
+  - finite-field == 0.10.*
```
paulcadman, line 66: Do we need to depend on this package as we have a source copy of the required module in our codebase?

stack.yaml
```diff
@@ -4,6 +4,7 @@ resolver: nightly-2024-02-06
 extra-deps:
   - git: https://github.com/Vekhir/aeson-better-errors.git
     commit: 1ec49ab7d1472046b680b5a64ae2930515b47714
+  - finite-field-0.10.0@sha256:f676a7abf829f27fee189036219c0326dce68a435bcdddc26319afff4008d8f4,2164
```
paulcadman, line 7: Do we need to depend on this package as we have a source copy of the required module in our codebase?

package.yaml
lukaszcz: That module depends on another module from the package which you didn't include. It doesn't compile without it.

janmasrovira (commented):
Since we are only using `finite-field` for the `FiniteField` typeclass, I think it makes sense to also copy that definition into our codebase and drop the dependency. I've done that in #2673

janmasrovira (approved):

PR #2663, opened 2024-02-20
Replace `polysemy` by `effectful`

janmasrovira:


src/Juvix/Compiler/Pipeline/Loader/PathResolver.hs
```diff
@@ -401,33 +401,74 @@ expectedPath' m = do
...
+                | otherwise = Nothing
+              env' :: ResolverEnv
+              env' =
+                ResolverEnv
+                  { _envRoot = root',
+                    _envLockfileInfo = Nothing,
+                    _envSingleFile
+                  }
+
+          let runner ::
```
arybczak, line 453: It looks to me like you're unnecessarily jumping through tons of hoops here.

You constructed `st'` and `env'`, so IIUC simply calling `local (const env') $ bracket (put st') (put st) $ unlift y` instead of, well, all of this here should do the trick.

janmasrovira: thank you! we've simplified the code following your advice

arybczak:
I was curious about this PR because this is a nice example of effectful vs polysemy, but I'm **very** surprised at the second benchmark results.

I think something's not right with the translation and I was curious enough that I tried to reproduce this locally and I can, but when I tried to build the project with profiling to see what's going on, GHC fails to compile the repline/haskeline module because you do some weird things there and apparently GHC gets confused.

janmasrovira:
> I was curious about this PR because this is a nice example of effectful vs polysemy, but I'm **very** surprised at the second benchmark results.
> 
> I suspect something's not right with the translation and I was curious enough that I tried to reproduce this locally and I can, but when I tried to build the project with profiling to see what's going on, GHC fails to compile the repline/haskeline module because you do some weird things there and apparently GHC gets confused 😕

We were also surprised with that benchmark and we are investigating the cause. I'll share the findings as soon as we find them. 
Ufortunately, I can't help you with the profiling problem. We are able to compile with profiling flags in linux but not in mac.

PR #2690, opened 2024-03-18
Anoma transaction fix

paulcadman:
When we first implemented the Nockma backend we wrongly assumed that the only entry point for Juvix compiled Nockma modules would be the main function. Using this assumption we could add a setup step in the main function that put the Anoma stdlib and compiled functions from the Juvix module in a static place in the Nockma subject. References to the Anoma stdlib and functions in the module could then be resolved statically.

However, one of the use cases for Juvix -> Nockma compilation is for Anoma to run logic functions that are fields of a transaction. So the user writes a Juvix program with main function that returns a transaction. The __result__ of the main function is passed to Anoma. When Anoma calls the logic function on a field of the transaction, the setup part of the main function is not run so the subject is not in the required state. In fact, the logic function is not even callable by Anoma because non-main functions in the Juvix module use a calling convention that assumes the subject has a particular shape.

This PR solves the problem by making all functions in the Juvix module use the Anoma calling convention. The issue of setup is solved by requiring an extra Nock term, called the Juvix environment, to be injected into any Nock term arising from the compiled Juvix module at path RRL, whenever it is called. This is similar to evaluation of calls in a particular context in hoon https://anoma.github.io/anoma/calling.html#evaluating-calls-in-the-anoma-context

We also add some extra tracing. You can now annotate output cells with a tag that will be displayed in the output.

We added unittests for `listToTuple`, `appendRights` helper functions.

### DIFF
--- a/test/Nockma/Eval/Positive.hs
+++ b/test/Nockma/Eval/Positive.hs
@@ -3,12 +3,12 @@ module Nockma.Eval.Positive where
 import Base hiding (Path, testName)
 import Data.HashMap.Strict qualified as HashMap
 import Juvix.Compiler.Core.Language.Base (defaultSymbol)
+import Juvix.Compiler.Nockma.Anoma
 import Juvix.Compiler.Nockma.Evaluator
 import Juvix.Compiler.Nockma.Language
 import Juvix.Compiler.Nockma.Pretty
 import Juvix.Compiler.Nockma.Translation.FromSource.QQ
 import Juvix.Compiler.Nockma.Translation.FromTree
-import Nockma.Base
 
 type Check = Sem '[Reader [Term Natural], Reader (Term Natural), EmbedIO]
 
--- a/test/Nockma/Eval/Positive.hs
+++ b/test/Nockma/Eval/Positive.hs
@@ -146,9 +146,8 @@ anomaCallingConventionTests =
 
 juvixCallingConventionTests :: [Test]
 juvixCallingConventionTests =
-  [True]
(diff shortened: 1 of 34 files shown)


test/Nockma/Eval/Positive.hs
```diff
@@ -23,14 +23,18 @@ makeLenses ''Test
 
 mkNockmaAssertion :: Test -> Assertion
 mkNockmaAssertion Test {..} = do
+  -- putStrLn (ppTrace _testProgramFormula)
+  -- writeFileEnsureLn
+  --   $(mkAbsFile "/home/jan/projects/juvix-effectful/out.nockma")
+  --   (ppPrint _testProgramSubject <> "\n\n" <> ppPrint _testProgramFormula)
```
janmasrovira, line 29:
```suggestion
```

test/Nockma/Eval/Positive.hs
```diff
@@ -23,14 +23,18 @@ makeLenses ''Test
...
   let (traces, evalResult) =
         run
           . runReader _testEvalOptions
           . runOutputList @(Term Natural)
           . runError @(ErrNockNatural Natural)
           . runError @(NockEvalError Natural)
-          $ eval _testProgramSubject _testProgramFormula
-
+          $ do
+            eval _testProgramSubject _testProgramFormula
```
janmasrovira, line 37:
```suggestion
          $ eval _testProgramSubject _testProgramFormula
```

janmasrovira (dismissed):

janmasrovira (approved):

paulcadman:
Closing in favour of https://github.com/anoma/juvix/pull/2693

PR #2698, opened 2024-03-21
Check modification time when loading stored modules

### LINKED ISSUE #2680: Improve per-module compilation by checking modification time
Currently, when checking whether a module should be recompiled we always compute the hash of the source and compare with the stored hash. We could also store the modification time, assume the source hasn't changed if the modification time didn't change and compute the source hash only if it did change. This would speed up stored module loading.

lukaszcz:
* Closes #2680

### DIFF
--- a/src/Juvix/Compiler/Pipeline/Driver.hs
+++ b/src/Juvix/Compiler/Pipeline/Driver.hs
@@ -10,6 +10,7 @@ where
 
 import Data.HashMap.Strict qualified as HashMap
 import Data.List.NonEmpty qualified as NonEmpty
+import Data.Time
 import Juvix.Compiler.Concrete (ImportCycle (ImportCycle), ScoperError (ErrImportCycle))
 import Juvix.Compiler.Concrete.Data.Highlight
 import Juvix.Compiler.Concrete.Language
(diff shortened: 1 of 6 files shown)


janmasrovira:
I've done a rudimentary benchmark to check whether getting the last modified time was cheaper than computing the hash. It turns out that computing the hash is much faster.

The benchmark consists in replicating the juvix stdlib 50 times and then compare the time it takes to hash every juvix file as opposed to get the last modification time.

```bash
#!/usr/bin/env bash

ORIGINAL_DIR="juvix-stdlib"

rm -rf benchtmp
mkdir benchtmp

for i in {1..50}; do
    NEW_DIR="benchtmp/${ORIGINAL_DIR}-${i}"
    cp -rf "$ORIGINAL_DIR" "$NEW_DIR"
done

hyperfine --warmup 2 \
    --command-name "hash-256" \
    'find benchtmp -type f -name "*.juvix" -print0 | xargs -0 sha256sum' \
    --command-name "last-modified-time" \
    'find benchtmp -type f -name "*.juvix" -exec stat --format="%y %n" {} \;'
```

The result is:
```
Benchmark 1: hash-256
  Time (mean ± σ):      56.4 ms ±   0.5 ms    [User: 17.2 ms, System: 40.5 ms]
  Range (min … max):    55.4 ms …  57.6 ms    51 runs

Benchmark 2: last-modified-time
  Time (mean ± σ):      1.636 s ±  0.013 s    [User: 1.216 s, System: 0.419 s]
  Range (min … max):    1.614 s …  1.656 s    10 runs

Summary
  hash-256 ran
   29.01 ± 0.33 times faster than last-modified-time
```

janmasrovira:
I'd say the benchmark with cache is more relevant since that will be the most common.
For reference, my results are:
```
Benchmark 1: hash-256
  Time (mean ± σ):      59.8 ms ±   1.3 ms    [User: 17.5 ms, System: 43.8 ms]
  Range (min … max):    57.7 ms …  64.4 ms    48 runs

Benchmark 2: last-modified-time
  Time (mean ± σ):      47.4 ms ±   1.3 ms    [User: 12.9 ms, System: 36.1 ms]
  Range (min … max):    45.5 ms …  53.0 ms    60 runs

Summary
  last-modified-time ran
    1.26 ± 0.04 times faster than hash-256
```

PR #2701, opened 2024-03-26
Move the Nockma standard library to a separate file

### LINKED ISSUE #2688: Move the Nockma standard library to a separate file in `runtime/src/nockma/stdlib.nockma`
There are two reasons:
1. Compiling `Juvix.Compiler.Nockma.Stdlib` takes several seconds, and this file is recompiled almost every time. This slows down development - one needs to wait several seconds after small changes to files unrelated to the Nockma backend.
2. The Nockma standard library is a part of the runtime - it is included in the generated code. Hence, it should not be under GPL unless we want the generated programs to be obligatorily GPL-licensed (see https://github.com/anoma/juvix/issues/2553). Moving all runtime files to a single directory makes it easier to re-license them separately under a more permissive license.

lukaszcz:
* Closes #2688


janmasrovira (approved):

paulcadman:
Please add stdlib.nockma to extra files in package.yaml https://github.com/anoma/juvix/blob/ee2f8aefbc3b1926f36c05da639a20f0b27cae59/package.yaml#L31

paulcadman (approved):

PR #2708, opened 2024-04-02
Alternative instances for `Eq`, `Hashable` for `StorageKey`

janmasrovira:
These instances make explicit what we want to take into account when testing for equality and hashing. With the previous instances, the `StdlibCall` in `CellInfo` would still be used for hashing, which is undesired.

### DIFF
--- a/src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
+++ b/src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
@@ -24,14 +24,28 @@ stripTags = \case
           $ c
       )
 
-instance (Eq a) => Eq (StorageKey a) where
-  s1 == s2 = stripTags (s1 ^. storageKeyTerm) == stripTags (s2 ^. storageKeyTerm)
+instance (NockmaEq a) => NockmaEq (StorageKey a) where
+  nockmaEq (StorageKey s1) (StorageKey s2) = nockmaEq s1 s2
 
-instance (Hashable a) => Hashable (StorageKey a) where
-  hashWithSalt s k = hashWithSalt s (stripTags (k ^. storageKeyTerm))
+instance (NockmaEq a) => Eq (StorageKey a) where
+  (==) = nockmaEq
+
+instance (NockmaEq a, Hashable a) => Hashable (StorageKey a) where
+  hashWithSalt salt k = goTerm salt (k ^. storageKeyTerm)
+    where
+      goCell :: Int -> Cell a -> Int
+      goCell s c = hashWithSalt s (c ^. cellLeft, c ^. cellRight)
+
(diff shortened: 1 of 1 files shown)


src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
```diff
@@ -24,14 +24,28 @@ stripTags = \case
...
 
-instance (Hashable a) => Hashable (StorageKey a) where
-  hashWithSalt s k = hashWithSalt s (stripTags (k ^. storageKeyTerm))
+instance (NockmaEq a) => Eq (StorageKey a) where
+  (==) = nockmaEq
+
+instance (NockmaEq a, Hashable a) => Hashable (StorageKey a) where
+  hashWithSalt salt k = goTerm salt (k ^. storageKeyTerm)
+    where
+      goCell :: Int -> Cell a -> Int
```
paulcadman, line 36: Does the salt need to be passed in as an argument to each function? Or can the `salt` argument be used directly instead?

src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
```diff
@@ -24,14 +24,28 @@ stripTags = \case
...
+      goAtom :: Int -> Atom a -> Int
+      goAtom s a = hashWithSalt s (a ^. atom)
+
+      goTerm :: Int -> Term a -> Int
+      goTerm s = \case
+        TermAtom a -> goAtom s a
+        TermCell c -> goCell s c
 
 instance (PrettyCode a, NockNatural a) => PrettyCode (StorageKey a) where
   ppCode k = ppCode (stripTags (k ^. storageKeyTerm))
```
paulcadman, line 48: I guess `stripTags` should be removed and the PrettyCode instance updated to remove the things that are not used in the hashing to avoid confusion when printing out keys?

janmasrovira: Yes, I think we should do that

src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
janmasrovira: The top `salt` can be used since the `go` functions are not recursive

janmasrovira: done

src/Juvix/Compiler/Nockma/Evaluator/Storage.hs
janmasrovira: done

paulcadman: 🌟 thank you!

PR #2718, opened 2024-04-10
Support for Cairo builtins

### LINKED ISSUE #2683: Support Cairo builtins
* Cairo builtins include range checks (to implement integer arithmetic) and cryptographic primitives. The first version of our Cairo backend doesn't support any builtins -- only Cairo Assembly instructions. Supporting builtins requires adapting the calling convention and code generation, because communication with the builtins is done via memory, pointers to which need to be passed around.
* Depends on #2563

lukaszcz:
This PR implements generic support for Cairo VM builtins. The calling convention in the generated CASM code is changed to allow for passing around the builtin pointers. Appropriate builtin initialization and finalization code is added. Support for specific builtins (e.g. Poseidon hash, range check, Elliptic Curve operation) still needs to be implemented in separate PRs.

* Closes #2683


src/Juvix/Compiler/Casm/Translation/FromCairo.hs
```diff
@@ -5,7 +5,7 @@ import Juvix.Compiler.Casm.Data.Result
 import Juvix.Compiler.Casm.Language
 
 fromCairo :: [Cairo.Element] -> Result
-fromCairo elems0 = Result mempty (go 0 [] elems0)
+fromCairo elems0 = Result mempty (go 0 [] elems0) mempty
```
janmasrovira, line 8: since there are two `mempty`, it would be more readable with record syntax

paulcadman (approved):

PR #2729, opened 2024-04-16
Support parallel module compilation in justfile

paulcadman:
Thanks to @janmasrovira for figuring out that the stack `--ghc-options=-j` flag enables [parallel module compilation in GHC](https://downloads.haskell.org/ghc/latest/docs/users_guide/using.html#using-ghc-make).

This PR adds support for this in the project justfile.

You can configure the argument to `-j` using the `numParallelJobs` option, for example:

```
just numParallelJobs=24 build
+ stack build --fast -j24 --ghc-options=-j24
```

If `numParallelJobs` is not set then `-j` is passed with no arguments in `--ghc-options` (this is equivalent to passing the number of cpus of the machine.) and is passed with the number of cpus of the machine for the stack `-j` option (the stack `-j` option requires an argument). 

The `numParallelJobs` option also sets the argument to the [stack `-j` option](https://docs.haskellstack.org/en/stable/global_flags/#-jobs-or-j-option).

To disable build parallelism set the `disableParallel` flag:

```
just disableParallel=yes build
+ stack build --fast
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,4 +1,4 @@
-# set to non-empty string to disable parallel tests
+# set to non-empty string to disable parallel builds and tests
 #
 # e.g:
 #   just disableParallel=yes test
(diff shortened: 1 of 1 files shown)


janmasrovira (approved):
:fast_forward: :fast_forward: :fast_forward:

PR #2730, opened 2024-04-16
Use the justfile for CI builds

paulcadman:
This PR changes the CI build to use the justfile instead of the Makefile to run builds and tests. CI builds now take advantage of parallel module builds from https://github.com/anoma/juvix/pull/2729.

In order support this the runtime build target in the justfile now supports `runtimeCcArg` and `runtimeLibtoolArg` so that the `CC` and `LIBTOOL` Makefile argument can be set. This is required for the macOS build.

In addition this PR upgrades the stack setup step action. Previously the stack build flags included `--fast` which meant the whole project was rebuilt in the `test` step, this has also been fixed.

Overall this speeds up the CI:
* Linux now takes 30mins (from 40mins)
* macOS now takes 60mins (from 80mins)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ env:
   VAMPIRVERSION: v0.1.3
   CAIRO_VM_VERSION: ae06ba04f3b6864546b6baeeebf1b0735ddccb5d
   JUST_ARGS: enableOptimized=yes runtimeCcArg=$CC runtimeLibtoolArg=$LIBTOOL
+  STACK_BUILD_ARGS: --pedantic -j4 --ghc-options=-j
 
 jobs:
   pre-commit:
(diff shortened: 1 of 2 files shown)


paulcadman:
> > In addition this PR upgrades the stack setup step action. Previously the stack build flags included `--fast` which meant the whole project was rebuilt in the `test` step, this has also been fixed.
> 
> `--fast` is actually slow in this case, because the tests running time dwarfs compilation time and `--fast` turns off optimizations. I wonder how much of the improvement is due to running a faster test executable and how much due to parallelization of compilation.
> 
> Besides, we should test the release executable in the CI (compiled with optimizations).

This PR makes sure that the project is built only once. Previously we built the project twice, the first build (with the `--fast` flag) was never used.

Previously:
* We first build with the `--fast` flag as part of the Stack setup step, and then rebuilt the project without the `--fast` flag in the test step 

Now:
* We build the project without the `--fast` flag in the Stack setup step, so the project does not get rebuilt in the test step. In particular the tests are run with the optimized build.

janmasrovira (approved):

PR #2734, opened 2024-04-17
Support `extract-module-statements` attribute in juvix code blocks

paulcadman:
This PR adds support for the `extract-module-statements` attribute for Juvix code blocks:

So if you write something like the following block in a Juvix markdown file:

````
```juvix extract-module-statements
module Foo;
   type T := t;
end;
```
````

The statement `type T := t;` from the body of the module is rendered in the output. The `module Foo;` , and `end;` lines are not rendered in the output. 

A block with the `extract-module-statements` must contain a single local module statement and nothing else. An error is reported if this is not the case.

The `extract-module-statements` attribute also takes an optional argument. It sets the number of statements from the module body to drop from the output.

In the following example, the output will contain the single line `a : T  := t;`.

````
```juvix extract-module-statements 1
module Foo;
   type T := t;
   a : T  := t;
end;
```
````

### CODE REVIEW

File: src/Juvix/Compiler/Backend/Markdown/Data/MkJuvixBlockOptions.hs [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+lexeme = L.lexeme spaceConsumer
+
+symbol :: Text -> Parser ()
+symbol = void . L.symbol spaceConsumer
+
+decimal :: Parser Int
+decimal = lexeme L.decimal
+
+parseOptions :: Parser MkJuvixBlockOptions
+parseOptions = spaceConsumer >> parseHide <|> parseExtractModule <|> parseShow
```
janmasrovira, line 47: I think with do notation is easier to read. We should remove the other `eof` occurences in this module and have it only here.
```suggestion
parseOptions = do
  spaceConsumer
  opts <- parseHide 
    <|> parseExtractModule 
    <|> parseShow
  eof
  return opts
```

src/Juvix/Compiler/Backend/Markdown/Data/MkJuvixBlockOptions.hs
```diff
@@ -0,0 +1,61 @@
...
+parseHide :: Parser MkJuvixBlockOptions
+parseHide = symbol optionHide $> MkJuvixBlockOptionsHide <* lexeme eof
+
+parseExtractModule :: Parser MkJuvixBlockOptions
+parseExtractModule = do
+  symbol optionExtractModuleStatements
+  dropNum <- fromMaybe 0 <$> optional decimal
+  let res =
+        MkJuvixBlockOptionsExtractModule (JuvixBlockOptionsExtractModule dropNum)
+  return res <* lexeme eof
```
janmasrovira, line 61: remove `lexeme eof`
```suggestion
  return (MkJuvixBlockOptionsExtractModule (JuvixBlockOptionsExtractModule dropNum))
```

src/Juvix/Compiler/Backend/Markdown/Data/MkJuvixBlockOptions.hs
```diff
@@ -0,0 +1,61 @@
...
+decimal = lexeme L.decimal
+
+parseOptions :: Parser MkJuvixBlockOptions
+parseOptions = spaceConsumer >> parseHide <|> parseExtractModule <|> parseShow
+
+parseShow :: Parser MkJuvixBlockOptions
+parseShow = eof $> MkJuvixBlockOptionsShow
+
+parseHide :: Parser MkJuvixBlockOptions
+parseHide = symbol optionHide $> MkJuvixBlockOptionsHide <* lexeme eof
```
janmasrovira, line 53: remove `lexeme eof`
```suggestion
parseHide :: Parser MkJuvixBlockOptions
parseHide = symbol optionHide $> MkJuvixBlockOptionsHide
```

src/Juvix/Compiler/Backend/Markdown/Data/MkJuvixBlockOptions.hs
```diff
@@ -0,0 +1,61 @@
...
+symbol = void . L.symbol spaceConsumer
+
+decimal :: Parser Int
+decimal = lexeme L.decimal
+
+parseOptions :: Parser MkJuvixBlockOptions
+parseOptions = spaceConsumer >> parseHide <|> parseExtractModule <|> parseShow
+
+parseShow :: Parser MkJuvixBlockOptions
+parseShow = eof $> MkJuvixBlockOptionsShow
```
janmasrovira, line 50: remove `eof`
```suggestion
parseShow :: Parser MkJuvixBlockOptions
parseShow = return MkJuvixBlockOptionsShow
```

src/Juvix/Compiler/Backend/Markdown/Data/MkJuvixBlockOptions.hs
paulcadman: Thanks 🌟

janmasrovira (approved):

PR #2752, opened 2024-04-22
Juvix to Isabelle/HOL translation

### LINKED ISSUE #2689: Translate Juvix types to Isabelle/HOL
This is needed to be able to write the type specs in Juvix.

lukaszcz:
* Closes #2689 
* Adds the command `juvix isabelle program.juvix` which translates a given file to an Isabelle/HOL theory.
* Currently, only a single module is translated.
* By default translates types and function signatures. Translating function signatures can be disabled with the `--only-types` option.

Blocked by:
- https://github.com/anoma/juvix/issues/2763


jonaprieto:
> How do we test HTML and Markdown generation?

For Markdown, we have tests in `test/BackendMarkdown/` and smoke tests. For Html, as I as I remember, we only have smoke tests. 

> any ideas of how this could be tested?

Let's install Isabelle in the CI, and test with it that the generation produces valid theories.

janmasrovira (dismissed):

jonaprieto (changes requested):
Wait for me, I need to test with some examples from the specs, before merging this.

src/Juvix/Compiler/Backend/Isabelle/Pretty/Base.hs
```diff
@@ -0,0 +1,157 @@
...
+
+instance PrettyCode RecordField where
+  ppCode RecordField {..} = do
+    n <- ppCode _recordFieldName
+    ty <- ppCodeQuoted _recordFieldType
+    return $ n <+> "::" <+> ty
+
+instance PrettyCode Theory where
+  ppCode Theory {..} = do
+    n <- ppCode _theoryName
```
jonaprieto, line 118: The theory's name must correspond to the file's name. Currently, it's printing the qualified name of the Juvix module.

jonaprieto: That won't work either. Use the unqualified name of the module as the name of the theory.

jonaprieto: Indeed. But we can generate the Isabelle theories following the module's folder structure of the Juvix modules. We do this for HTML output.

jonaprieto (approved):

PR #2821, opened 2024-06-08
Disable macOS CI

lukaszcz:
It's failing because of rate limits exceeded when installing RISC0 VM. We should disable the macOS CI until we figure out how to reliably get around this problem.


janmasrovira (approved):

paulcadman:
Can you follow this PR up by caching the RISC0 VM install?

PR #2838, opened 2024-06-18
Update ormolu to 0.7.4.0

janmasrovira:
This pr updates the version of ormolu to 0.7.4.0. Moreover, we drop the `mrkkrp/ormolu-action` github action. Instead, we invoke the ormolu version that is in the stackage snapshot selected in the `stack.yaml` file.

One can locally install the new version thus:
```
cd juvix
stack install ormolu
```
The output of `ormolu --version` should be 
```
ormolu 0.7.4.0
using ghc-lib-parser 9.8.2.20240223
```

# Questionable changes
image

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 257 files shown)


janmasrovira:
not planned for any time soon

PR #2863, opened 2024-06-27
Improve performance of formatting a project

janmasrovira:
Currently formatting a project is equivalent to running `juvix format` on each individual file. Hence, the performance is quadratic wrt the number of modules in the project. This pr fixes that and we now we only process each module once.

# Benchmark (1236% faster :rocket:)
Checking the standard library
```
hyperfine --warmup 1 'juvix format --check' 'juvix-main format --check'
Benchmark 1: juvix format --check
  Time (mean ± σ):     450.6 ms ±  33.7 ms    [User: 707.2 ms, System: 178.7 ms]
  Range (min … max):   396.0 ms … 497.0 ms    10 runs

Benchmark 2: juvix-main format --check
  Time (mean ± σ):      6.019 s ±  0.267 s    [User: 9.333 s, System: 1.512 s]
  Range (min … max):    5.598 s …  6.524 s    10 runs

Summary
  juvix format --check ran
   13.36 ± 1.16 times faster than juvix-main format --check
```

# Other changes:
1. The `EntryPoint` field `entryPointModulePath` is now optional.
2. I've introduced a new type `TopModulePathKey` which is analogous to `TopModulePath` but wihout location information. It is used in hashmap keys where the location in the key is never used. This is useful as we can now get a `TopModulePathKey` from a `Path Rel File`.
3. I've refactored the `_formatInput` field in `FormatOptions` so that it doesn't need to be a special case anymore.
4. I've introduced a new effect `Forcing` that allows to individually force fields of a record type with a convenient syntax.
5. I've refactored some of the constraints in scoping so that they only require `Reader Package` instead of  `Reader EntryPoint`.
6. I've introduced a new type family so that local modules are no longer required to have `ModuleId` from their type. Before, they were assigned one, but it was never used.

# Future work:
1. For project-wise formatting, the compilation is done in parallel, but the formatting is still done sequentially. That should be improved.

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -298,9 +298,11 @@ runPipelineEntry entry p = runPipelineOptions $ do
   r <- runIOEither entry (inject p) >>= fromRightJuvixError
   return (snd r ^. pipelineResult)
 
-runPipelineSetup :: (Members '[App, EmbedIO, Reader PipelineOptions, TaggedLock] r) => Sem (PipelineEff' r) a -> Sem r a
--- runPipelineSetup p = ignoreProgressLog $ do -- TODO restore
-runPipelineSetup p = appRunProgressLog $ do
+runPipelineSetup ::
+  (Members '[App, EmbedIO, Reader PipelineOptions, TaggedLock] r) =>
+  Sem (PipelineEff' r) a ->
+  Sem r a
+runPipelineSetup p = ignoreProgressLog $ do
   args <- askArgs
   entry <- getEntryPointStdin' args
   r <- runIOEitherPipeline entry (inject p) >>= fromRightJuvixError
--- a/app/App.hs
+++ b/app/App.hs
@@ -32,7 +32,9 @@ data App :: Effect where
   AskGlobalOptions :: App m GlobalOptions
   FromAppPathFile :: AppPath File -> App m (Path Abs File)
   GetMainAppFile :: Maybe (AppPath File) -> App m (AppPath File)
+  GetMainAppFileMaybe :: Maybe (AppPath File) -> App m (Maybe (AppPath File))
   GetMainFile :: Maybe (AppPath File) -> App m (Path Abs File)
+  GetMainFileMaybe :: Maybe (AppPath File) -> App m (Maybe (Path Abs File))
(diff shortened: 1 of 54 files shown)


app/App.hs
```diff
@@ -32,7 +32,9 @@ data App :: Effect where
   AskGlobalOptions :: App m GlobalOptions
   FromAppPathFile :: AppPath File -> App m (Path Abs File)
   GetMainAppFile :: Maybe (AppPath File) -> App m (AppPath File)
+  GetMainAppFileMaybe :: Maybe (AppPath File) -> App m (Maybe (AppPath File))
   GetMainFile :: Maybe (AppPath File) -> App m (Path Abs File)
+  GetMainFileMaybe :: Maybe (AppPath File) -> App m (Maybe (Path Abs File))
```
paulcadman, line 37: Q: Do `GetMain{File, AppFile}` need to be defined on the effect as they can be derived from `GetMain{File, AppFile}Maybe?

src/Juvix/Data/Effect/Forcing.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Juvix.Prelude.Base
+
+data Forcing (a :: GHCType) :: Effect where
+  ForcesField :: (NFData b) => Lens' a b -> Forcing a m ()
+  Forces :: Lens' a b -> Sem '[Forcing b] () -> Forcing a m ()
+
+makeSem ''Forcing
+
+forcing :: a -> Sem '[Forcing a] () -> a
```
paulcadman, line 13: What does forcing do? Perhaps write a docstring to explain?

janmasrovira: it is a special case of `evalForcing` for `r = []`. I added this because I think it will be common enough to use the `Forcing` effect on its own.

paulcadman: What does the `Forcing` effect do?

janmasrovira: I've added documentation. Please let me know if it needs more explanation

app/App.hs
janmasrovira: I'll improve the app effect in a separate pr

PR #2874, opened 2024-07-02
Reduce Internal boilerplate

janmasrovira:
This pr modifies the `HasExpressions` class to allow traversing direct subexpressions rather than only leaf expressions. I've added the `lens` library as a dependency to have access to generic traversals defined in [Control.Lens.Plated](https://hackage.haskell.org/package/lens-5.3.2/docs/Control-Lens-Plated.html).


janmasrovira:
> This PR switches to a different lens library and uses more general library functions to implement some traversals over program representation. I think once it's done we should benchmark it to have an idea how this affects performance.

afaik the implementation is equivalent so performance should be the same. It's just that `lens` is a bigger library.
I've run the benchmark on the stdlib and it seems to indicate that.
```
hyperfine --warmup 1 --prepare 'juvix clean' 'juvix typecheck index.juvix' 'juvix-main typecheck index.juvix' Benchmark 1: juvix typecheck index.juvix
  Time (mean ± σ):     436.9 ms ±   7.3 ms    [User: 874.8 ms, System: 103.0 ms]
  Range (min … max):   427.7 ms … 447.0 ms    10 runs

Benchmark 2: juvix-main typecheck index.juvix
  Time (mean ± σ):     438.6 ms ±   7.0 ms    [User: 883.1 ms, System: 88.4 ms]
  Range (min … max):   426.2 ms … 447.9 ms    10 runs

Summary
  juvix typecheck index.juvix ran
    1.00 ± 0.02 times faster than juvix-main typecheck index.juvix
```

I included the `lens` library because it has the functions that I needed and it was easier for me this way. Then we have two options, we use `lens` from now on, or we copy to our codebase the (relatively small) fragment that we need for this pr. That's something that we can discuss on a call next week.

paulcadman:
> 
> I included the `lens` library because it has the functions that I needed and it was easier for me this way. Then we have two options, we use `lens` from now on, or we copy to our codebase the (relatively small) fragment that we need for this pr. That's something that we can discuss on a call next week.

There's [microlens-pro](https://hackage.haskell.org/package/microlens-pro) that provides Prisms / Isos if that's what you need. (this is not an endorsement).

paulcadman (approved):

PR #2875, opened 2024-07-03
Pragmas for record fields

### LINKED ISSUE #2872: It should be possible to have pragmas inside record declarations
The following
```
type Pr (A B : Type) := mkPr {
     {-# inline: true #-}
     pfst : A;
     psnd : B
};
```
currently gives an error
```
  |
6 |      {-# inline: true #-}
  |      ^
unexpected '{'
expecting <identifier>, builtin, syntax, or }
```
Instead, the pragma should be accepted and associated with the generated record projection.

This would be useful to have for the Isabelle translation, to be able to indicate that certain functions in a trait should be translated to specific Isabelle operators.

lukaszcz:
* Closes #2872

### DIFF
--- a/tests/positive/Format.juvix
+++ b/tests/positive/Format.juvix
@@ -354,6 +354,17 @@ module OperatorRecord;
     in n + m * m;
 end;
 
+module RecordFieldPragmas;
+  type Pr (A B : Type) :=
+    mkPr {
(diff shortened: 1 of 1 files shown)


paulcadman (approved):
It's good we can also use judoc on record fields 👍

PR #2876, opened 2024-07-08
Migrate old named application syntax

### LINKED ISSUE #2668: Migrate old named application syntax
Migrate existing code and remove support for the old named application syntax (https://github.com/anoma/juvix/pull/2250) the new syntax was added in https://github.com/anoma/juvix/pull/2494

And raise a PR to update documentation.

janmasrovira:
- Closes #2668

This pr migrates the old named application syntax to the new one. In order to migrate a juvix file to the new syntax it suffices to run the formatter.
After the next release, we should completely remove the support for the old syntax.

## Other changes
I've improved Scope negative tests. Previously, when a negative test failed, you could only see the title of the test and the message "Incorrect Error", as well as the Haskell file and line where the test is defined.
This is extremely incovenient because you have to go to the haskell test file, go to the line where the error is defined, look at the name of the file and then visit that file. Moreover, you need to manually run the scoper on that file to see the error that was returned. 
I've fixed that and it now shows all relevant information. Example:
image
I've implemented this only using the `Generic` instance for the `ScoperError` type, so doing something similar for the rest of negative tests should be straightforward.


tests/positive/Format.juvix
```diff
@@ -297,13 +297,22 @@ module Traits;
   type Show A := mkShow {show : A → String};
 
   instance
-  showStringI : Show String := mkShow (show := id);
+  showStringI : Show String :=
```
paulcadman, line 300: Does it make sense to you why the juvix formatter would break lines on these code blocks given that we increased the ribbon width?

janmasrovira: The explanation is that after `:=` we break if the body does not fit in a line. The formatter _always_ inserts newlines in the new named application syntax. I.e. you'll never see `mkShow@{show := id}`.
I think there are 3 sensible ways to format the named application syntax.
1. [what we currently do] Always break.
2. Break iff there are more than 1 argument.
3. Break only if it does not fit in a line.

I'm fine with both 1 and 2
If we went, with 2 my expectation is that there would not be a linebreak after `:=`.

paulcadman: I see - I'm also happy with what we currently do. Thanks for the explanation.

paulcadman (approved):

PR #2883, opened 2024-07-10
Make `juvix format` line width 100 with ribbon width 100

paulcadman:
This PR increases the ribbon width of `juvix format` from 60 to 100 characters.

Reasons for the compromise to a fixed 100 chars ribbon width:

* It is clear that the ribbon width of 60 characters was too small.
* A ribbon width of 100 is an acceptable compromise between formatting code for display and editing code in multiple buffers on the same screen.
* We would like to avoid making the formatter configurable so that Juvix code has a consistent look and to save future Juvix users from discussions about formatting. Maxim: "juvix format's style is no one's favourite, yet juvix format is everyone's favourite" (thanks go fmt).

## Definition of ribbon width from the [docs](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html)

> The page has a certain maximum width, which the layouter tries to not exceed, by inserting line breaks where possible. The functions given in this module make it fairly straightforward to specify where, and under what circumstances, such a line break may be inserted by the layouter, for example via the [sep](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:sep) function.
> 
> There is also the concept of ribbon width. The ribbon is the part of a line that is printed, i.e. the line length without the leading indentation. The layouters take a ribbon fraction argument, which specifies how much of a line should be filled before trying to break it up. A ribbon width of 0.5 in a document of width 80 will result in the layouter to try to not exceed 0.5*80 = 40 (ignoring current indentation depth).

Examples from [`anoma-app-patterns:/Token/Transaction.juvix`](https://github.com/anoma/anoma-app-patterns/blob/8d7e892de33159b4b7c59e2cf11282ba49853e71/Token/Transaction.juvix). NB: The file in the repo is unformatted so will not match the layout in the left column below.

| Current (line width 150, ribbon width 60) | This PR (line width 100, ribbon width 100) |
| --- | --- |
| Screenshot 2024-07-10 at 12 22 46 | Screenshot 2024-07-10 at 14 41 33 |
| Screenshot 2024-07-10 at 12 23 10 | Screenshot 2024-07-10 at 14 42 01 |
| Screenshot 2024-07-10 at 12 23 21 | Screenshot 2024-07-10 at 14 42 14 |
| Screenshot 2024-07-10 at 12 23 34 | Screenshot 2024-07-10 at 14 42 29 |
| Screenshot 2024-07-10 at 12 23 50 | Screenshot 2024-07-10 at 14 42 40 |
| Screenshot 2024-07-10 at 12 24 13 | Screenshot 2024-07-10 at 14 42 57 |

### DIFF
--- a/tests/positive/Format.juvix
+++ b/tests/positive/Format.juvix
@@ -16,11 +16,7 @@ Bool; true; false; mkShow; module Show};
 terminating
 -- Comment between terminating and type sig
 go : Nat → Nat → Nat
-  | n s :=
-    ite
-      (s < n)
-      (go (sub n 1) s)
-      (go n (sub s n) + go (sub n 1) s);
+  | n s := ite (s < n) (go (sub n 1) s) (go n (sub s n) + go (sub n 1) s);
 
 module {- local module -}
 M;
@@ -65,19 +61,7 @@ t5 : String := "a" M., "b" M., "c", "d";
 t2 : String := "a", "b", "c", "d";
 
 -- comma chain does not fit in a line
-t3 : String :=
-  "a"
-    , "b"
-    , "c"
-    , "d"
-    , "e"
-    , "f"
-    , "g"
-    , "h"
-    , "i"
-    , "j"
-    , "k"
-    , "1234";
+t3 : String := "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "1234";
 
 -- escaping in String literals
 e1 : String := "\"\n";
@@ -121,14 +105,9 @@ t : String :=
     +l1 "Hellooooooooo"
     +l1 "Hellooooooooo"
         +l6 "Hellooooooooo"
-        +l6 ("Hellooooooooo"
-          +r6 "Hellooooooooo"
-          +r6 "Hellooooooooo")
+        +l6 ("Hellooooooooo" +r6 "Hellooooooooo" +r6 "Hellooooooooo")
         +l6 "Hellooooooooo"
-        +l6 "Hellooooooooo"
-          +l7 "Hellooooooooo"
-          +l7 "Hellooooooooo"
-          +l7 "Hellooooooooo"
+        +l6 "Hellooooooooo" +l7 "Hellooooooooo" +l7 "Hellooooooooo" +l7 "Hellooooooooo"
       , "hi"
       , "hi";
 
@@ -296,8 +275,7 @@ module Traits;
   showStringI : Show String := mkShow (show := id);
 
   instance
-  showBoolI : Show Bool :=
-    mkShow (show := λ {x := ite x "true" "false"});
+  showBoolI : Show Bool := mkShow (show := λ {x := ite x "true" "false"});
 
   instance
   showNatI : Show Nat := mkShow (show := natToString);
@@ -309,16 +287,14 @@ module Traits;
   g : {A : Type} → {{Show A}} → Nat := 5;
 
   instance
-  showListI {A} {{Show A}} : Show (List A) :=
-    mkShow (show := showList);
+  showListI {A} {{Show A}} : Show (List A) := mkShow (show := showList);
 
   showMaybe {A} {{Show A}} : Maybe A → String
     | (just x) := "just (" ++str Show.show x ++str ")"
     | nothing := "nothing";
 
   instance
-  showMaybeI {A} {{Show A}} : Show (Maybe A) :=
-    mkShow (show := showMaybe);
+  showMaybeI {A} {{Show A}} : Show (Maybe A) := mkShow (show := showMaybe);
 
   f {A} {{Show A}} : A → String
     | x := Show.show x;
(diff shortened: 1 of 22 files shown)


heueristik:
> Setting the ribbon width to 100 is too large because it's common for text editors to have buffer width around 80 characters.

Two questions:
Which editors wouldn't be able to display code with >80 chars properly?
Will this encourage people to write very short, non-descriptive variable names in production code? This could lead to more errors.

heueristik:
> Maxim: "juvix format's style is no one's favourite, yet juvix format is everyone's favourite" (thanks go fmt)

It is also worth noting that `go fmt` has no line length limit:

> Go has no line length limit. Don’t worry about overflowing a punched card. If a line feels too long, wrap it and indent with an extra tab.

https://go.dev/doc/effective_go#formatting

paulcadman:
> I'd say not enforcing a line length limit is less strict than having an option to enforce different line length limits. It allows people to choose the line length arbitrarily and differently within even a single project.

I think we can agree on this. Q: can we implement this using prettyprinter, if so what's the effort required? Perhaps @janmasrovira knows?

paulcadman:
> For me personally, ribbon width of 100 chars would still be palatable, but with anything longer than this used consistently I'd need to give up on two panes side by side.

Thanks - I'll update this PR to line width 100 and ribbon width 100 (i.e `100 1.0`).

paulcadman:
> > Maxim: "juvix format's style is no one's favourite, yet juvix format is everyone's favourite" (thanks go fmt)
> 
> It is also worth noting that `go fmt` has no line length limit:
> 
> > Go has no line length limit. Don’t worry about overflowing a punched card. If a line feels too long, wrap it and indent with an extra tab.
> 
> https://go.dev/doc/effective_go#formatting

For clarity I didn't make this comment to endorse all of gofmts design choices, just the maxim.

janmasrovira:
I'm fine with either choice. Let's try it for a while, if we feel like it can be improved, we can always change it.

I'd like to add that we don't respect the user newlines by design. The goal of our formatter is indeed to make code look readable, but also (and not less important) to remove as much formatting responsability from the user.

janmasrovira (dismissed):

tests/positive/Format.juvix
```diff
@@ -65,19 +61,7 @@ t5 : String := "a" M., "b" M., "c", "d";
 t2 : String := "a", "b", "c", "d";
 
 -- comma chain does not fit in a line
```
janmasrovira, line 63: The list needs to be made longer so that we force linebreaks

tests/positive/Format.juvix
```diff
@@ -138,31 +117,15 @@ g : (_ : Type) -> Nat
...
-    -> Nat
+  : {A : Type} -> List A -> List A -> List A -> List A -> List A -> List A -> List A -> Nat
   | _ _ _ _ _ _ _ := 1;
 
 axiom undefined : {A : Type} -> A;
 
 -- 200 in the body is indented with respect to the start of the chain
 -- (at {A : Type})
 exampleFunction2
-  : {A : Type}
```
janmasrovira, line 157: idem

tests/positive/Format.juvix
```diff
@@ -369,32 +329,16 @@ longLongLongArg : Int := 0;
 
 longLongLongListArg : List Int := [];
 
-l1 : List Int :=
```
janmasrovira, line 372: idem

tests/positive/Format.juvix
```diff
@@ -369,32 +329,16 @@ longLongLongArg : Int := 0;
...
-  ; longLongLongArg
-  ; longLongLongArg
-  ; longLongLongArg
-  ; longLongLongArg
-  ];
+l1 : List Int := [1; 2; longLongLongArg; longLongLongArg; longLongLongArg; longLongLongArg];
 
 l2 : List Int := [1; 2; 3];
 
-l3 : List (List Int) :=
```
janmasrovira, line 383: idem

janmasrovira (changes requested):
I now realize that some changes are needed in the Format.juvix test.
Some lines were written with the intention to force the formatter to introduce a newline.

paulcadman:
> I now realize that some changes are needed in the Format.juvix test. Some lines were written with the intention to force the formatter to introduce a newline.

Thanks - I'll fix these now

tests/positive/Format.juvix
paulcadman: done

tests/positive/Format.juvix
paulcadman: done

tests/positive/Format.juvix
paulcadman: done

tests/positive/Format.juvix
paulcadman: done

janmasrovira (approved):

PR #2890, opened 2024-07-12
Puns for named application

janmasrovira:
Since it is common to want to assign a named argument a variable of the same name, we add special syntax for it. E.g.
```
f (fieldA : A) (fieldB : B) : S :=
  mkS@{
    fieldC := fieldA; -- normal named argument
    fieldA;  -- pun
    fieldB   -- pun
  };
```


paulcadman (approved):
⭐

PR #2900, opened 2024-07-15
Use ormolu from stackage in the CI

janmasrovira:
1. Adds the command `just format check`, which checks that all Haskell files are formatted.
2. In CI, we use install ormolu from stackage and run it. This will facilitate consistency between CI and local setups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - name: Install hpack
+        run: sudo apt install -y haskell-hpack
+
+      - name: Generate juvix.cabal
+        run: hpack
+
       - uses: mrkkrp/ormolu-action@v11
         with:
           version: 0.5.3.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: extractions/setup-just@v2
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -43,20 +43,35 @@ jobs:
...
-            --ghc-opt -XDerivingStrategies
-            --ghc-opt -XImportQualifiedPost
-            --ghc-opt -XMultiParamTypeClasses
-            --ghc-opt -XPatternSynonyms
-            --ghc-opt -XStandaloneDeriving
-            --ghc-opt -XTemplateHaskell
-            --ghc-opt -XUnicodeSyntax
-            --ghc-opt -XBangPatterns
+          path: |
+            ~/.stack
```
paulcadman, line 63: This works fine but did you consider the following?

1. Cache the path `~/.local/bin/ormolu` with key (e.g) `${{ runner.os }}-ormolu-${{ env.STACK_RESOLVER }}`
2. Guard the install ormolu step with `if: steps.stack-cache.outputs.cache-hit != 'true'` (`stack-cache` here refers to the `id` of the `Cache stack` step, see [example cache workflow](https://github.com/actions/cache?tab=readme-ov-file#example-cache-workflow))

This way the `stack install` step will not run if the `ormolu` binary already exists. This would avoid the hackage cache download triggered by `stack install` on each ormolu run.

janmasrovira: I've applied your suggestion. Thanks!

paulcadman:
Before merging this, let's check that the cache works after https://github.com/anoma/juvix/pull/2868 is merged and we rebase.

paulcadman (approved):

PR #2908, opened 2024-07-18
Logger

janmasrovira:
1. Adds the `--log-level LOG_LEVEL` flag  to the CLI. This flag can be given `error`, `warn`, `info`, `progress`, `debug` as argument to filter the logged messages.
2. Removes the `--only-errors` flag.
3. Adds the `--ide-end-error-char CHAR`, which receives a character as an argument, which is appended to the end of error messages. This is handy to facilitate parsing of errors messages from the ide. This functionality was previously embeded in the old `--only-errors` flag.

### DIFF
--- a/app/App.hs
+++ b/app/App.hs
@@ -75,11 +75,9 @@ reAppIO args@RunAppIOArgs {..} =
     GetMainFile m -> getMainFile' m
     GetMainFileMaybe m -> getMainFileMaybe' m
     FromAppPathDir p -> liftIO (prepathToAbsDir invDir (p ^. pathPath))
-    RenderStdOut t
-      | _runAppIOArgsGlobalOptions ^. globalOnlyErrors -> return ()
-      | otherwise -> do
-          sup <- liftIO (Ansi.hSupportsANSIColor stdout)
-          renderIO (not (_runAppIOArgsGlobalOptions ^. globalNoColors) && sup) t
+    RenderStdOut t -> do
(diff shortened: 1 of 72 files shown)


app/App.hs
```diff
@@ -293,23 +282,29 @@ runPipelineOptions m = do
...
   Sem (PipelineEff' r) a ->
   Sem r a
-runPipelineSetup p = ignoreProgressLog $ do
+runPipelineSetup p = do
   args <- askArgs
   entry <- getEntryPointStdin' args
   r <- runIOEitherPipeline entry (inject p) >>= fromRightJuvixError
   return (snd r)
 
+-- say :: (Member App r) => Text -> Sem r ()
```
paulcadman, line 300: Can be removed? (I can't delete using GitHub suggestion because there's no way to remove the extra newline)

paulcadman (commented):
suggestion: Add a smoke test for `--ide-end-error-char`

janmasrovira:
> Please open an issue in vscode-juvix once you merge this or another kind of PRs that remove/change the CLI that the vscode extension may depend on.

- I've created an issue: https://github.com/anoma/vscode-juvix/issues/133

app/App.hs
janmasrovira: I forgot to remove that. It is now fixed

janmasrovira:
> suggestion: Add a smoke test for `--ide-end-error-char`

I've done that in 99e67c90462c0fe823b921d22f5a94b149a332e2

paulcadman (approved):

PR #2917, opened 2024-07-26
Export Juvix source code to latex

janmasrovira:
This pr adds two new commands related to latex.
1. `juvix dev latex getJuvixSty`. This command prints the contents of the `juvix.sty` latex package to stdout. It has no options and the expected usage is `juvix dev latex getJuvixSty > juvix.sty`.
2. `juvix dev latex export`. Expects a .juvix file as an argument and outputs to stdout the highlighted module in latex format. Optional flags `--from LINE`, `--to LINE` to output only the specified line range. There is a `--mode` flag to choose how you want the output. 
   ```
   • standalone: Output a ready to compile LaTeX file
   • wrap: Wrap the code in a Verbatim environment
   • raw: Output only the code (default: standalone)
   ```
4. As shown in the standalone output, one is allowed to choose the theme when importing the juvix package like this: `\usepackage[theme = latte]{juvix}`. Available themes are `latte`, `frappe`, `macchiato`, `mocha`.

Examples:
To generate the following output I ran these commands:
```
cd examples/milestones/HelloWorld
mkdir latex
juvix dev latex getJuvixSty > latex/juvix.sty
juvix dev latex export HelloWorld.juvix > latex/main.tex
cd latex
xelatex main.tex
open main.pdf
```
1. with `\usepackage[theme = latte]{juvix}`:
image
1. with `\usepackage[theme = frappe]{juvix}`:
image
1. with `\usepackage[theme = macchiato]{juvix}`:
image
1. with `\usepackage[theme = mocha]{juvix}`:
image


src/Juvix/Extra/Paths/Base.hs
```diff
@@ -11,6 +11,9 @@ assetsDirQ = FE.makeRelativeToProject "assets" >>= FE.embedDir
 cssDirQ :: Q Exp
 cssDirQ = FE.makeRelativeToProject "assets/css" >>= FE.embedDir
 
+juvixStyQ :: Q Exp
+juvixStyQ = FE.makeRelativeToProject "latex/juvix.sty" >>= FE.embedFileRelative
```
paulcadman, line 15: Suggestion: Use `include/latex/juvix.sty` for the file. The `include/` directory in the repository is intended for extra things to be included in the binary.

paulcadman, line 15: Do you need to use both `FE.makeRelativeToProject` and `FE.embedFileRelative` ? `FE.embedFileRelative` will work with the relative path?

app/Commands/Dev/Latex/Export.hs
```diff
@@ -0,0 +1,59 @@
...
+          $ moduleToLatex c m
+  renderStdOut $
+    case _exportMode of
+      ExportStandalone -> standalone ltx
+      ExportRaw -> ltx
+      ExportWrap -> verb ltx
+
+verb :: Text -> Text
+verb code =
+  Text.unlines
```
paulcadman, line 32: Minor: It might be clearer to use the string interpolation quasi-quote for these text blocks:

https://github.com/anoma/juvix/blob/bd3b7f140104668e72134bf9e9adb138d82baf09/app/Commands/Repl.hs#L53

https://www.stackage.org/haddock/nightly-2024-08-02/string-interpolate-0.3.4.0/Data-String-Interpolate.html

src/Juvix/Extra/Paths/Base.hs
janmasrovira: done

janmasrovira: done

paulcadman (dismissed):

app/Commands/Dev/Latex/Export.hs
janmasrovira: done

PR #2928, opened 2024-07-30
Fix name signature bug and extend test for instance fields

### LINKED ISSUE #2923: Add tests for PR #2916
* Add tests for #2916 
* See my comment on the PR page

janmasrovira:
- Closes #2923 

This pr fixes a bug where all fields were assigned to be explicit arguments in the NameSignature Builder. A single line change was enough to fix it.
```diff
-           RecordStatementField RecordField {..} -> addSymbol @s Explicit Nothing _fieldName _fieldType
+           RecordStatementField RecordField {..} -> addSymbol @s (fromIsImplicitField _fieldIsImplicit) Nothing _fieldName _fieldType
```

I've also added a compilation test for instance fields.


test/Compilation/Positive.hs
```diff
@@ -455,5 +455,10 @@ tests =
         "Test076: Builtin Maybe"
         $(mkRelDir ".")
         $(mkRelFile "test076.juvix")
-        $(mkRelFile "out/test076.out")
+        $(mkRelFile "out/test076.out"),
+      posTestEval
```
paulcadman, line 459: Can this be `posTest` instead of `posTestEval` so we test compilation post-core?

janmasrovira: my understanding is that after core it doesn't make a difference wether a field is an instance field or not. Should I still change it to `posTest`?

paulcadman: We can keep it like this if you like. The reason to change it is to add more test coverage to the post-core pipeline.

paulcadman (approved):

test/Compilation/Positive.hs
janmasrovira: I've made the change. I think it is an overall relevant test.

paulcadman (approved):

PR #2937, opened 2024-08-06
Add do notation

### LINKED ISSUE #2355: Add do-notation
- Depends on #2354.

We would support three kinds of statements (bind, let, expression), and we could borrow the syntax from haskell.
```
do {
  <var> <- <expr>; -- bind
  let <fundef>; -- pure let
  <expr>; -- monadic expression
};
```

In the future we can consider some extensions:
1. Allow patterns of single constructor types on the left of `<-`.
2. @paulcadman mentioned [Idris !-notation](https://idris2.readthedocs.io/en/latest/tutorial/interfaces.html#notation).
3. Do notation for `Applicative`.

janmasrovira:
- Closes #2355
- Depends on #2943 

Example:
```
minusOne : Nat -> Maybe Nat
  | zero := nothing
  | (suc n) := just n;


minusThree (n : Nat) : Maybe Nat :=
  do {
    x1 <- minusOne n;
    x2 <- minusOne x1;
    let
      x2' : Nat := x2;
    in
    x3 <- minusOne x2';
    pure x3;
  };
```


paulcadman (approved):
🌯

PR #2973, opened 2024-08-27
Don't fold lets with `fail`, `trace` or `>->` in the body

lukaszcz:
This is to avoid optimizing e.g.
```
let x := A in
trace "after A" >-> x
```
into 
```
trace "after A" >-> A
```
which is annoying when debugging.

### DIFF
--- a/src/Juvix/Compiler/Core/Extra/Utils.hs
+++ b/src/Juvix/Compiler/Core/Extra/Utils.hs
@@ -183,7 +183,44 @@ containsDebugOperations = ufold (\x xs -> x || or xs) isDebugOp
   where
     isDebugOp :: Node -> Bool
     isDebugOp = \case
-      NBlt (BuiltinApp {..}) -> _builtinAppOp `elem` [OpTrace, OpFail, OpSeq]
+      NBlt BuiltinApp {..} ->
+        case _builtinAppOp of
+          OpTrace -> True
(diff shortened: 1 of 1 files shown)


src/Juvix/Compiler/Core/Extra/Utils.hs
```diff
@@ -177,6 +177,15 @@ isFalseConstr = \case
   NCtr Constr {..} | _constrTag == BuiltinTag TagFalse -> True
   _ -> False
 
+-- | Check if the node contains `trace`, `fail` or `seq` (`>->`).
+containsDebugOperations :: Node -> Bool
+containsDebugOperations = ufold (\x xs -> x || or xs) isDebugOp
+  where
+    isDebugOp :: Node -> Bool
+    isDebugOp = \case
+      NBlt (BuiltinApp {..}) -> _builtinAppOp `elem` [OpTrace, OpFail, OpSeq]
```
janmasrovira, line 186:
```suggestion
      NBlt BuiltinApp {..} -> _builtinAppOp `elem` [OpTrace, OpFail, OpSeq]
```

janmasrovira (commented):
I'd prefer if we had a function `isDebugOp : BuiltinOp -> Bool` with an explicit pattern match instead of an ad hoc list

janmasrovira (approved):

PR #2974, opened 2024-08-27
Isabelle/HOL translation: comments

### LINKED ISSUE #2962: Include comments in the Isabelle/HOL translation
The PR

* #2958 

includes Judoc comments in the Isabelle/HOL translation.

Including all comments is more difficult, because in general they are not directly associated with nodes in the syntax tree and are forgotten when translating from Concrete.

lukaszcz:
* Closes #2962 
* Depends on #2963 
* In Isabelle/HOL comments cannot appear in internal syntax. All comments inside a Juvix definition are moved outside: to before the definition or before the earliest function clause.


janmasrovira (commented):
Did you consider `ExactPrint`?

janmasrovira:
The idea is that every time you print something you attach a location to that. Then the effect takes care of automatically inserting the comments in the correct place.

If you want to put comments only in certain places, it could be done by providing the location in those places and omiting it everywhere else.

janmasrovira (approved):

PR #2975, opened 2024-08-27
Improve performance of anomaEncode / anomaDecode in the Core evaluator

paulcadman:
This PR:

* Adds a new implementation of {decode, encode}ByteString functions, used by anomaEncode and anomaDecode in the Core evaluator
* Adds property tests for roundtripping and benchmarks for the new functions.

The old implementation used [bitvec](https://hackage.haskell.org/package/bitvec) to manipulate the ByteString. This was far too slow. The new implementation uses bit operations directly on the input integer and ByteArray. 

It's now possible to run [anoma-app-patterns:`Tests/Swap.juvix`](https://github.com/anoma/anoma-app-patterns/blob/feature/tests/Tests/Swap.juvix) to completion. 

For encoding, if the size of the output integer exceeds 64 bits (and therefore a BigInt must be used) then the new implementation has quadratic time complexity in the number of input bytes if an implementation of `ByteString -> Integer` is used as follows:

```
byteStringToIntegerLE :: ByteString -> Integer
byteStringToIntegerLE = BS.foldr (\b acc -> acc `shiftL` 8 .|. fromIntegral b) 0
```

```
byteStringToInteger' :: ByteString -> Integer
byteStringToInteger' = BS.foldl' (\acc b -> acc `shiftL` 8 .|. fromIntegral b) 0

```

I think this is because `shiftL` is expensive for large Integers. To mitigate this I'm splitting the input ByteString into 1024 byte chunks and processing each separately. Using this we get 100x speed up at ~0.25Mb input over the non-chunked approach and linear time-complexity thereafter. 

## Benchmarks

The benchmarks for encoding and decoding 250000 bytes:

```
 ByteString Encoding to/from integer
      encode bytes to integer:   OK
        59.1 ms ± 5.3 ms
      decode bytes from integer: OK
        338  ms ±  16 ms
```

The previous implementation would never complete for this input.

Benchmarks for encoding and decoding 2 * 250000 bytes:

```
    ByteString Encoding to/from integer
      encode bytes to integer:   OK
        121  ms ± 8.3 ms
      decode bytes from integer: OK
        651  ms ±  27 ms
```

Benchmarks for encoding and decoding 4 * 250000 bytes:

```
    ByteString Encoding to/from integer
      encode bytes to integer:   OK
        249  ms ±  17 ms
      decode bytes from integer: OK
        1.317 s ±  16 ms
```

### DIFF
--- a/bench2/Benchmark/Nockma.hs
+++ b/bench2/Benchmark/Nockma.hs
@@ -1,11 +1,13 @@
 module Benchmark.Nockma where
 
 import Benchmark.Nockma.Encoding qualified as NockmaEncoding
+import Benchmark.Nockma.Encoding.ByteString qualified as NockmaByteString
 import Test.Tasty.Bench
 
 bm :: Benchmark
 bm =
   bgroup
     "Nockma"
-    [ NockmaEncoding.bm
+    [ NockmaEncoding.bm,
+      NockmaByteString.bm
     ]
--- a/bench2/Benchmark/Nockma/Encoding/ByteString.hs
+++ b/bench2/Benchmark/Nockma/Encoding/ByteString.hs
@@ -1,0 +1,19 @@
+module Benchmark.Nockma.Encoding.ByteString where
+
+import Juvix.Compiler.Nockma.Encoding.ByteString qualified as Encoding
+import Juvix.Prelude.Base
+import System.Random
+import Test.Tasty.Bench
+
+randomBytes :: Int -> IO ByteString
+randomBytes numBytes = do
+  sg <- getStdGen
+  return (fst (genByteString numBytes sg))
+
+bm :: Benchmark
+bm =
+  bgroup
+    "ByteString Encoding to/from integer"
+    [ env (randomBytes 250000) (\bs -> bench "encode bytes to integer" (nf Encoding.encodeByteString bs)),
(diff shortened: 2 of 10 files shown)


paulcadman:
> Can't we solve this just by eliminating the use of shifts on Integers? They are the culprits here, but just bit-testing should be constant time. I changed two functions in `Nockma.Encoding.Base` to:

Thank you, this is a better approach!

I also looked at the low-level library functions `integerToMutableByteArray` et al. They seem quite complicated to use as they involve doing memory allocations and pointer manipulation.

paulcadman:
> > Can't we solve this just by eliminating the use of shifts on Integers? They are the culprits here, but just bit-testing should be constant time. I changed two functions in `Nockma.Encoding.Base` to:
> 
> Thank you, this is a better approach!

@lukaszcz  I've reverted to the bitvec implementation using your suggestion for decoding. For me the `juvix eval Tests/Swap.juvix` test is still too slow without the chunking of the bytestring when encoding to an Integer (1m45s without chunking vs 20s with chunking). So I have kept the chunking bytestring -> integer function for encoding.

paulcadman:
> Or maybe we could just use `Data.Serialize`? The serialization/deserialization implementation for `Integer` should be linear there.

anomaEncode 'serializes' an arbitrary ByteString to an Integer. So I'm not sure we can use Data.Serialize because `serialize : Integer -> ByteString` is not surjective?

PR #2984, opened 2024-08-30
Compilation of side conditions in pattern matches

### LINKED ISSUE #2804: Side conditions in pattern matching
Allow side conditions in pattern matches:
```
| pat1 pat2 pat3 if condition := body
```
For example, allow to write
```
case lst of
| [x] if F x := B1
| [x] := B2
| _ := B3
```
instead of
```
case lst of
| [x] := 
  if
    | F x := B1
    | else := B2
| _ := B3
```

Syntax proposal:

```
case lst of
| [x] 
     | if F x := A
     | if G x := B
     | else := C (else branch is optional)
| _ := D;
```

lukaszcz:
* Closes #2804 
* Requires #3003
* Front-end syntax for side conditions was implemented in #2852. This PR implements compilation of side conditions.
* Adds side-conditions to `Match` nodes in Core. Updates Core parsing, printing and the evaluator.
* Only side-conditions without an `else` branch are allowed in Core. If there is an `else` branch, the side conditions are translated in `fromInternal` into nested ifs. Because with `else` the conditions are exhaustive, there are no implications for pattern exhaustiveness checking.
* Adjusts the "wildcard row" case in the pattern matching compilation algorithm to take into account the side conditions.


janmasrovira:
As discussed, let's first merge the removal of `naive-match-to-case`.

janmasrovira (approved):

PR #2989, opened 2024-09-02
Improve instance termination checking

janmasrovira:


janmasrovira:
the approach taken in this pr does not work. Some refactors might still be worth applying, but closing for now

PR #3011, opened 2024-09-09
Fix location of scoped modulePathName

### LINKED ISSUE #2737: Formatting of imports breaks lines unnecessarily
The import
```
import Stdlib.Cairo.Ec as Ec;
```
gets formatted into
```
import
Stdlib.Cairo.Ec as Ec;
```
I don't understand the reason. It's merely 30 characters long - the line shouldn't be broken up.

janmasrovira:
Closes #2737.

This issues caused the formatter to sometimes insert unwanted line breaks.


janmasrovira:
@lukaszcz I added a test

PR #3024, opened 2024-09-11
Fix JuvixTree parsing and pretty printing

lukaszcz:
Recent changes to the compiler left JuvixTree parsing and pretty printing not in sync with each other.


janmasrovira:
We should have tests to check that pretty output can be parsed back

janmasrovira (approved):

Issue #3037, opened 2024-09-13
linux static release build fails


paulcadman:
After the update to GHC 9.10.1 in:

* https://github.com/anoma/juvix/pull/2991

We updated the linux static build workflow that's used to build the linux binary release.

* https://github.com/anoma/juvix/pull/3034

However the linux static build fails:

https://github.com/anoma/juvix/actions/runs/10852415273

The dependency `text` fails with:

```
text                         > In file included from /usr/include/c++/13.2.1/cstring:42,
text                         >                  from simdutf/simdutf.h:7,
text                         >                  /tmp/stack-f4da676a69f752ea/text-2.1.1/from simdutf/simdutf.cpp:4:0: error: 
text                         > 
text                         > /usr/include/fortify/string.h: In function 'size_t simdutf::scalar::{anonymous}::utf16_to_utf8::convert_valid(const char16_t*, size_t, char*)':
text                         > /usr/include/fortify/string.h:40:27: error:
text                         >      error: inlining failed in call to 'always_inline' 'void* memcpy(void*, const void*, size_t)': target specific option mismatch
text                         >        40 | _FORTIFY_FN(memcpy) void *memcpy(void * _FORTIFY_POS0 __od,
text                         >           |                           ^~~~~~
text                         >    |
text                         > 40 | _FORTIFY_FN(memcpy) void *memcpy(void * _FORTIFY_POS0 __od,
text                         >    |                           ^
text                         > 
text                         > /tmp/stack-f4da676a69f752ea/text-2.1.1/simdutf/simdutf.cpp:8578:15: error:
text                         >      note: called from here
text                         >      8578 |       ::memcpy(&v, data + pos, sizeof(uint64_t));
text                         >           |       ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
text                         >      |
text                         > 8578 |       ::memcpy(&v, data + pos, sizeof(uint64_t));
text                         >      |               ^
text                         > 
text                         > `g++' failed in phase `C++ Compiler'. (Exit code: 1)
```

Perhaps the version of gcc shipped with the alpine container we use for the static build is not compatible? I'll need to investigate.

* https://quay.io/repository/benz0li/ghc-musl

## Impact

We cannot make static release binaries for linux until this is fixed.

PR #3052, opened 2024-09-19
Revert GHC 9.10.1 update

paulcadman:
We cannot build linux static binaries with GHC 9.10.1:

* https://github.com/anoma/juvix/issues/3037

This PR reverts the GHC update to unblock Juvix releases / nightly releases. We can try the update again when stackage nightly updates to GHC 9.10.1.


paulcadman:
The CI was failing because the agent ran out of space. The CI cache contained both GHC 9.8.2 and  GHC 9.10.1 build artefacts.

I've purged the cache to remove the GHC 9.10.1 build artefacts.

I've also added the commit from https://github.com/anoma/juvix/pull/2991 to free up space on the CI agent by removing software that we don't use.

PR #3102, opened 2024-10-14
Compile-time configuration

### LINKED ISSUE #3077: Make WASM support optional
It should be possible to compile Juvix without WASM support. We don't need it anymore, and it seems to be the biggest potential stumbling block when compiling Juvix, especially on Mac where the default clang doesn't support WASM.

### LINKED ISSUE #3100: Make Rust compiler optional
We currently require it to compile the Rust runtime. It should be optional, same as WASM support (c.f. issue #3077).

lukaszcz:
* Closes #3077 
* Closes #3100 
* Adds a compilation-time configuration script that creates a `config/config.json` file which is then read by the `Makefile`/`justfile` and embedded into the Juvix binary.

### DIFF
--- a/config/configure.sh
+++ b/config/configure.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-# This script should be run from the root of the project
-
 if [ ! -d "config" ]; then
     printf "This script should be run from the root of the project.\n" 1>&2
     exit 1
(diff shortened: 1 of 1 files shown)


config/configure.sh
```diff
@@ -0,0 +1,30 @@
...
+    WASM="false"
+fi
+
+if $CARGO --version >/dev/null 2>&1; then
+    RUST="true"
+else
+    RUST="false"
+fi
+
+printf "{\n  \"wasm\": $WASM,\n  \"rust\": $RUST,\n  \"clang\": \"$CC\",\n  \"cargo\": \"$CARGO\"\n}\n" > config/config.json
```
paulcadman, line 30: Suggestion: use `cat` instead of `printf` to generate the json file.

```suggestion
cat <<EOF > config/config.json
{
  "wasm": $WASM,
  "rust": $RUST,
  "clang": "$CC",
  "cargo": "$CARGO"
}
EOF
```

paulcadman (approved):

PR #3105, opened 2024-10-16
Fix closure representation in the Nock backend

### LINKED ISSUE #3042: Nockma backend: reduce the number of things stored in closures
I believe it's not necessary to store all this stuff in closures, or some rarely used stuff can be moved to one "extra" field, the pointer to which would be just passed around most of the time.

The point is that at each function call you need to create a list with all these things. Hence, the length of this list has a significant impact on the overall running time.

In a purely functional programming language, function calls are the main way of transferring control ("loops" are implemented with function calls) and functions themselves are typically small. This means that the total number of operations performed is proportional (with a relatively small constant) to the number of function calls. Hence, any constant factor inefficiency in function call implementation translates to a comparable constant factor slowdown of the entire program. Function calls need to be as efficient as possible.

### LINKED ISSUE #3083: Nockma backend: Adjust closure implementation to match the Nock calling convention
We should use the `curry` Nock stdlib function to implement partial application, or otherwise implement partial application without storing the partially applied arguments in the arguments tuple. When called by Nock from outside, the arguments tuple is just replaced with the remaining arguments, so we can't represent a partially applied function by storing the partially applied arguments there.

lukaszcz:
* Closes #3083 
* Closes #3042 

The representation of closures is changed to make it more efficient and compatible with the Nock calling convention.


src/Juvix/Compiler/Nockma/Language.hs
```diff
@@ -460,6 +468,12 @@ opAddress txt p = TermCell (txt @ OpAddress #. p)
 opQuote :: (IsNock x) => Text -> x -> Term Natural
 opQuote txt p = TermCell (txt @ OpQuote #. p)
 
+opTrace :: Term Natural -> Term Natural
+opTrace val = OpHint # (nockHintAtom NockHintPuts # val) # val
+
+opTrace' :: Term Natural -> Term Natural -> Term Natural
```
paulcadman, line 474: Is this used anywhere?

src/Juvix/Compiler/Nockma/Translation/FromTree.hs
```diff
@@ -1092,26 +1070,25 @@ callFunWithArgs fun args = do
...
+
+curryClosure :: Term Natural -> [Term Natural] -> Term Natural -> Sem r (Term Natural)
+curryClosure f args newArity = do
+  let args' = (foldTerms (nonEmpty' $ map (\x -> (OpQuote # OpQuote) # x) args <> [OpQuote # OpAddress # closurePath ArgsTuple]))
+  return . makeClosure $ \case
+    FunCode -> (OpQuote # OpCall) # (OpQuote # closurePath FunCode) # (OpQuote # OpReplace) # ((OpQuote # closurePath ArgsTuple) # args') # (OpQuote # OpQuote) # f
+    ArgsTuple -> OpQuote # nockNilTagged "argsTuple" -- We assume the arguments tuple is never accessed before being replaced by the caller.
+    ClosureRemainingArgsNum -> newArity
+    FunctionsLibrary -> OpQuote # functionsLibraryPlaceHolder
+    StandardLibrary -> OpQuote # stdlibPlaceHolder
```
paulcadman, line 1091: These fields don't need to be addresses to a the functions library and stdlib in the function wrapper? Is this because the function code is not evaluated in this context? Perhaps it's a good idea to add a comment?

paulcadman (commented):
Is it possible to add additional tests that use currying that would have failed with the previous approach?

src/Juvix/Compiler/Nockma/Translation/FromTree.hs
```diff
@@ -987,45 +961,43 @@ runCompilerWith _opts constrs moduleFuns mainFun =
...
 
+    -- This term is not quoted and cannot be evaluated.
     funcsLib :: Term Natural
     funcsLib = Str.theFunctionsLibrary @ makeList compiledFuns
       where
         compiledFuns :: [Term Natural]
         compiledFuns =
-          (OpQuote # (666 :: Natural)) -- TODO we have this unused term so that indices match. Remove it and adjust as needed
+          (nockNilTagged "anomaGetPlaceholder")
+            : (nockNilTagged "mainFunctionPlaceholder")
```
paulcadman, line 971: Was the `666` placeholder previously for the `main` function that is added to `allFuns` but not `libFuns`? 

The `libFunShift` is required because `anomaGet` is now added to the functions library?

paulcadman: I'd just like to clarify why `libFunShift` is `1` but 2 extra functions are being added to the front of the functions library. We should comment this while we remember it.

paulcadman (approved):

PR #3122, opened 2024-10-23
Formatter: add braces when the iterator body is not enclosed

### LINKED ISSUE #3091: Enforce braces around iterator bodies
Currently, the braces are optional. I think we should make them mandatory to make the syntax more consistent.

For example:
```
for (acc := t) (x in d) {
    deleteWith p x acc
}
```
Now we also allow:
```
for (acc := t) (x in d)
  deleteWith p x acc
```

lukaszcz:
* Closes #3091
* Formatter adds braces when the body is not enclosed in braces or parentheses. Braces-enclosed body is always printed as a block on a new line:
```
for (acc := 0) (x in lst) {
  x + acc
}
```
* If the body is enclosed in ordinary parentheses, then they are preserved and the iterator is printed on a single line, if possible:
```
for (acc := 0) (x in lst) (x + acc)
```
This is sometimes useful when you want iterator application as an argument to something.

### DIFF
--- a/tests/positive/RecordIterator.juvix
+++ b/tests/positive/RecordIterator.juvix
@@ -19,4 +19,7 @@ foldl
   (g : B -> elem -> B)
   (ini : B)
   (ls : container)
-  : B := for (acc := ini) (x in ls) {g acc x};
+  : B :=
+  for (acc := ini) (x in ls) {
+    g acc x
+  };


janmasrovira:
It might be confusing to support both parentheses and braces with the same meaning. Should we always use parentheses?

janmasrovira (approved):

PR #3128, opened 2024-10-25
Run Nockma in an Anoma node

janmasrovira:
# Changes
1. Adds a new command `juvix dev anoma node`. This command runs the anoma node.
2. Adds a flag `--anoma-dir` to `juvix dev nockma run`. When given, it must point to the anoma clone. Then, it will run the nockma code in the anoma node and report the result (with no traces).

# Prerequisites
1. An anoma clone at some specific commit. 
   ```
   git clone git@github.com:anoma/anoma.git
   cd anoma
   git checkout 98e3660b91cd55f1d9424dcff9420425ae98f5f8
   
   # build anoma
   mix deps.get
   mix escript.install hex protobuf
   mix compile

   # build the client
   mix do --app anoma_client escript.build
   ```
2. The `mix` command (elixir).
3. The [`grpcurl`](https://github.com/fullstorydev/grpcurl) command. To install a single binary in `~/.local/bin` you can run:
   ```
   curl -sSL "https://github.com/fullstorydev/grpcurl/releases/download/v1.9.1/grpcurl_1.9.1_linux_x86_64.tar.gz" |  tar -xz -C ~/.local/bin --no-wildcards grpcurl
   ```

# Testing
I've not included any test. It can be tested locally like this:
```
cd juvix/tests/Anoma/Compilation/positive
juvix compile anoma test001.juvix
echo 20 > args.debug.nockma
juvix dev nockma run --anoma-dir ~/projects/anoma test001.nockma --args args.debug.nockma

### DIFF
--- a/src/Anoma/Effect/Base.hs
+++ b/src/Anoma/Effect/Base.hs
@@ -80,12 +80,12 @@ withSpawnAnomaClient body = do
           { std_out = CreatePipe
           }
 
-withSapwnAnomaNode ::
+withSpawnAnomaNode ::
   forall r a.
   (Members '[EmbedIO, Logger, Error SimpleError, Process, Reader AnomaPath] r) =>
   (Int -> Handle -> ProcessHandle -> Sem r a) ->
   Sem r a
-withSapwnAnomaNode body = withSystemTempFile "start.exs" $ \fp tmpHandle -> do
+withSpawnAnomaNode body = withSystemTempFile "start.exs" $ \fp tmpHandle -> do
   liftIO (B.hPutStr tmpHandle anomaStartExs)
   hClose tmpHandle
   cproc <- cprocess (toFilePath fp)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Anoma/Effect/Base.hs [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+        )
+          { std_out = CreatePipe
+          }
+
+withSapwnAnomaNode ::
+  forall r a.
+  (Members '[EmbedIO, Logger, Error SimpleError, Process, Reader AnomaPath] r) =>
+  (Int -> Handle -> ProcessHandle -> Sem r a) ->
+  Sem r a
+withSapwnAnomaNode body = withSystemTempFile "start.exs" $ \fp tmpHandle -> do
```
paulcadman, line 88: minor: typo `withSpawnAnomaNode`

src/Anoma/Effect/Base.hs
```diff
@@ -0,0 +1,150 @@
...
+  liftIO (B.hPutStr tmpHandle anomaStartExs)
+  hClose tmpHandle
+  curDir <- getCurrentDir
+  asks (^. anomaPath) >>= setCurrentDir
```
paulcadman, line 92: suggestion: Use `anomaPath` directly to make an absolute path for the process working directory. 

Use of `setCurrentDir` will cause issues if this function is ever used concurrently in the future so we should try to avoid using it in general.

src/Anoma/Effect/Base.hs
janmasrovira: fixed

src/Anoma/Effect/Base.hs
janmasrovira: fixed in c8872e0542b109621f0dfbc19ccb4cd96fad7f9f

paulcadman: The calls to `setCurrentDir` are still there. Can they be removed now the cwd is being set on CreateProcess?

janmasrovira: I forgot to remove those :disappointed: it is fixed now

PR #3135, opened 2024-10-30
Add `dev nockma  encode` command

janmasrovira:
- New command `juvix dev nockma encode --help`
```
Usage: juvix dev nockma encode --to ENCODING --from ENCODING

  Encode and decode nockma terms

Available options:
  --to ENCODING            Choose the source encoding.
                           • base64: Jam and Base 64 encoding
                           • bytes: Jam encoding
                           • debug: Nockma code with annotations
                           • text: Nockma code without annotations
  --from ENCODING          Choose the target encoding.
                           • base64: Jam and Base 64 encoding
                           • bytes: Jam encoding
                           • debug: Nockma code with annotations
                           • text: Nockma code without annotations
```

### DIFF
--- a/src/Anoma/Effect/RunNockma.hs
+++ b/src/Anoma/Effect/RunNockma.hs
@@ -6,12 +6,14 @@ where
 
 import Anoma.Effect.Base
 import Anoma.Rpc.RunNock
-import Juvix.Compiler.Nockma.Encoding (decodeCue64, encodeJam64)
+import Data.ByteString.Base64 qualified as Base64
+import Juvix.Compiler.Nockma.Encoding
+import Juvix.Compiler.Nockma.Language
 import Juvix.Compiler.Nockma.Language qualified as Nockma
+import Juvix.Data.CodeAnn
 import Juvix.Prelude
 import Juvix.Prelude.Aeson (Value)
 import Juvix.Prelude.Aeson qualified as Aeson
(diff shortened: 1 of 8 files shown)


src/Anoma/Effect/RunNockma.hs
```diff
@@ -60,5 +55,5 @@ runNockma prog inputs = do
   res :: Response <- anomaRpc runNockGrpcUrl (Aeson.toJSON msg) >>= fromJSON
   logVerbose (mkAnsiText ("Response Payload:\n" <> Aeson.jsonEncodeToPrettyText res))
   case res of
-    ResponseProof x -> decodeJam64 x
+    ResponseProof x -> decodeCue64 x
```
paulcadman, line 58: Q: `decodeCue64` is correct here because the value of the 'proof' field is jammed - so we cue it to see the nock term?

janmasrovira: it's just a matter of renaming `decodeJam64` -> `decodeCue64`. They are the same function. I could leave it to `decodeJam64` if that's more intuitive. What do you think?

janmasrovira: I removed `decodeJam64` in af0dd16827799614ea0e30b911bec87a86475664 because it was a duplicate of `decodeCue64`

paulcadman (approved):

PR #3159, opened 2024-11-07
Don't write log messages to stdout

lukaszcz:
The "Cloning (...)" message was written to stdout which messed up markdown generation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,6 @@ repos:
     rev: v4.5.0
     hooks:
       - id: check-yaml
-        args: ["--disable", "duplicate-keys"]
       - id: check-added-large-files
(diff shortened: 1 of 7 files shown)


janmasrovira (commented):
I don't think we should suppress git messages

paulcadman:
I think we should have logs for when things are being cloned in the normal use-case. Perhaps we could configure the output based on the current log level instead to cover use cases where you don't want them printed?

janmasrovira:
Ideally we would redirect the git messages to the Logger effect, probably on log level info

janmasrovira:
I mentioned the Logger effect, which prints to stderr, so that it integrates well with the existing cli flags

janmasrovira (approved):

PR #3168, opened 2024-11-12
HTML generation: make the light theme lighter

### LINKED ISSUE #3141: Why is the background in generated HTML docs grey?
I would expect it to be ordinary white. Grey is less readable because it's closer to the color of the text.

image

lukaszcz:
* Closes #3141 
* Adds the `latte-light` theme with lighter background and makes it the default. This is a bit subjective, but in my opinion the light theme should not have a background darker than the browser window pane. It should be close to white.

### DIFF
--- a/app/Commands/Html/Options.hs
+++ b/app/Commands/Html/Options.hs
@@ -47,7 +47,7 @@ parseHtml = do
       (eitherReader parseTheme)
       ( long "theme"
           <> metavar "THEME"
-          <> value Latte
+          <> value LatteLight
           <> showDefault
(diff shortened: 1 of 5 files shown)


paulcadman:
Could you add before and after screenshots in the PR?

janmasrovira:
I'm reluctant to change the values of `ctp-base` and `ctp-mantle` because these are part of the catppuccin palette

janmasrovira:
one option is to create a new theme `latte-light`, which is `latte` plus the changes that you proposed. This way we keep the original palettes untouched

paulcadman (approved):

PR #3169, opened 2024-11-13
Detect termination for nested local definitions

### LINKED ISSUE #3147: Termination checking doesn't work with local lets
Typechecking
```
module termRec;

import Stdlib.Prelude open;

type MyList A :=
  | myNil
  | myCons@{
      elem : A;
      next : MyList A;
    };

go {A B} (f : A -> B) : MyList A -> MyList B
  | myNil := myNil
  | myCons@{elem; next} := myCons@{elem := f elem; next := go f next};
```
gives the error
```
/Users/heliax/Documents/progs/juvix/termRec.juvix:9-12:7-3: error:
The following functions fail the termination checker:
• next
• go
```
But the following type-checks:
```
go {A B} (f : A -> B) : MyList A -> MyList B
  | myNil := myNil
  | myCons@{elem; next} := myCons (f elem) (go f next);
```

Termination checking needs to be adapted to deal with record creation/update syntax (or whatever it is desugared to in Internal).

lukaszcz:
* Closes #3147 

When we call a function that is currently being defined (there may be several such due to nested local definitions), we add a reflexive edge in the call map instead of adding an edge from the most nested definition. For example, for

```juvix
go {A B} (f : A -> B) : List A -> List B
  | nil := nil
  | (elem :: next) :=
    let var1 := f elem;
        var2 := go f next;
    in var1 :: var2;
```

we add an edge from `go` to the recursive call `go f next`, instead of adding an edge from `var2` to `go f next` as before.

This makes the above type-check.

The following still doesn't type-check, because `next'` is not a subpattern of the clause pattern of `go`. But this is a less pressing problem.

```juvix
go {A B} (f : A -> B) : List A -> List B
  | nil := nil
  | (elem :: next) :=
    let var1 := f elem;
        var2 (next' : List A) : List B := go f next';
    in myCons var1 (var2 next);
```

### DIFF
--- a/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/FunctionCall.hs
+++ b/src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/FunctionCall.hs
@@ -59,12 +59,12 @@ viewCall = \case
 
             Consider:
             ```
-            isEven (x : Nat) : Nat :=
+            isEven (x : Nat) : Bool :=
               let
-                isEven' : Nat -> Nat
+                isEven' : Nat -> Bool
                   | zero := true
                   | (suc n) := isOdd' n;
-                isOdd' : Nat -> Nat
+                isOdd' : Nat -> Bool
                   | zero := false
                   | (suc n) := isEven' n;
                in isEven' x;


src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/FunctionCall.hs
```diff
@@ -33,19 +37,25 @@ viewCall = \case
...
               Just x' -> do
-                s <- asks (elemIndex x' . (^. sizeEqual))
+                s <- asks (elemIndex x' . (^. sizeEqual) . findSizeInfo)
                 return $ case s of
                   Nothing -> CallRow Nothing
                   Just s' -> CallRow (Just (s', REq))
               Nothing -> return (CallRow Nothing)
+          findSizeInfo :: SizeInfoMap -> SizeInfo
+          findSizeInfo infos =
+            fromMaybe (maybe emptySizeInfo snd . headMay $ infos ^. sizeInfoMap)
```
janmasrovira, line 55: I don't understand this line. Why is it ok to return the head even if the function name does not match?

src/Juvix/Compiler/Internal/Translation/FromInternal/Analysis/Termination/Checker.hs
```diff
@@ -100,48 +102,48 @@ checkTerminationShallow' topModule = do
           | Nothing <- order -> TerminatingFailed
 
 scanModule ::
-  (Members '[State CallMap] r) =>
+  (Members '[State CallMap, Reader SizeInfoMap] r) =>
```
janmasrovira, line 105: why add the Reader here? The goal should be to add the `Reader SizeInfoMap` only to places where it can be nonempty

janmasrovira (approved):

PR #3170, opened 2024-11-13
Use tagged version of ghc-musl in static linux build

paulcadman:
The issue with alpine ncurses packaging has been fixed upstream and the ghc-musl containers have been rebuilt. Therefore we can resume using the tagged releases of the ghc-musl container.

https://github.com/benz0li/ghc-musl/issues/10

Thanks @benz0li for help with diagnosing and fixing this issue.

### DIFF
--- a/.github/workflows/linux-static-binary.yaml
+++ b/.github/workflows/linux-static-binary.yaml
@@ -15,7 +15,7 @@ jobs:
   build:
     name: Build static binary
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


benz0li:
P.S.: `quay.io/benz0li/ghc-musl:9.8.2` is now frozen – i.e. not going to be rebuilt[^1].

[^1]: Exceptions are made if an image contains a serious bug.

> Whenever a new version of GHC is released, the previous version's docker image is rebuilt once again and then frozen.

– https://github.com/benz0li/ghc-musl?tab=security-ov-file#vulnerabilities-in-prior-versions

PR #3175, opened 2024-11-14
Fix nock compilation and builtin evaluation of sha256 stdilb

paulcadman:
This PR:

1. Fixes the compilation of the sha256 builtin anoma lib call

A sha256 hash is 32 bytes long, not 64 bytes. This number is used when constructing the ByteArray representation (i.e `[length payload]` cell) of the output of Anoma stdlib sha256 call. The Anoma stdlib sha256 call just returns the atom payload.

2. Fixes the evaluation of the sha256 stdlib call

Previously we were converting the sha256 hash bytestring to Base16 format. This is convenient when displaying the ByteString hash. However the Anoma nock interpreter outputs the raw bytes so we must change the builtin evaluator to match this behaviour.

After these fixes we can re-enable the test084 anoma compilation test.


janmasrovira (dismissed):

paulcadman:
I forgot to update the other sha256 tests, this is now done.

janmasrovira (approved):

PR #3181, opened 2024-11-20
Allow pattern-matching in variable definitions

### LINKED ISSUE #3118: Allow pattern-matching in lets (and other definitions) for records and other single-constructor types
We should allow:
```
let (x, y) := f a
in
```
and
```
let record@{field1; field2} := f a
in
```
This is convenient syntactic sugar which easily translates to pattern matching. We should allow it only for single-constructor types so there is no issue with non-exhaustive patterns.

For consistency, we can allow it everywhere where a function definition is expected.

lukaszcz:
* Closes #3118 
* Allows patterns on the left-hand sides of definitions. To avoid clashes with constructor names, the patterns need to be enclosed in parentheses or have the form `n@...`. A standalone symbol `x` in a definition `x := ...` is always interpreted as an identifier even if there exists a constructor `x`.
* Left-hand side patterns are desugared in the translation from Concrete to Internal. If the inductive type has more than one constructor, non-exhaustive pattern matching error will later be given by the pattern matching compiler.

### DIFF
--- a/src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
+++ b/src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
@@ -1220,7 +1220,6 @@ checkFunctionDef fdef@FunctionDef {..} = do
   let def =
         FunctionDef
           { _signName = sigName',
-            -- _signPattern = sigPattern',
             _signDoc = sigDoc',
             _signBody = sigBody',
             _signTypeSig = sig',


janmasrovira:
reviewing...

src/Juvix/Compiler/Concrete/Translation/FromParsed/Analysis/Scoping.hs
```diff
@@ -1142,19 +1197,36 @@ checkFunctionDef fdef@FunctionDef {..} = do
...
+      p' <- runReader PatternNamesKindFunctions (checkParsePatternAtom p)
+      return
+        FunctionDefNameScoped
+          { _functionDefName = name',
+            _functionDefNamePattern = Just p'
+          }
   let def =
         FunctionDef
           { _signName = sigName',
+            -- _signPattern = sigPattern',
```
janmasrovira, line 1223: this comment leaked from my pr

janmasrovira (approved):

PR #3182, opened 2024-11-20
Set std_in to NoStream on the Anoma client process

paulcadman:
This PR sets `std_in` to `NoStream` on the Anoma client process.

https://hackage.haskell.org/package/process-1.6.25.0/docs/System-Process.html#t:StdStream

We do not read from the Anoma client process stdin so we do not need to open a handle to stdin.

This fixes an issue where the Anoma client stdin does not get cleaned up correctly on exit which causes subsequent input to the terminal to be corrupted.


janmasrovira (approved):
good catch

PR #3187, opened 2024-11-22
Add `juvix dev anoma prove`

paulcadman:
> ⚠️ Based on https://github.com/anoma/juvix/pull/3183

Add `juvix dev anoma prove`

This is similar to `juvix dev nockma run`.

The difference is that the mamaged client is used and it outputs the result as a jammed file.


app/Commands/Dev/Nockma/Run/Anoma.hs
```diff
@@ -14,25 +14,10 @@ makeLenses ''RunCommandArgs
...
-        renderStdOutLn (ppPrint tr)
-      renderStdOutLn (annotate AnnImportant "Result:")
-      renderStdOutLn (ppPrint (res ^. runNockmaResult))
+  res <- runNock (runArgs ^. runCommandProgramFile) (runArgs ^. runCommandArgsFile)
+  let traces = res ^. runNockmaTraces
+  renderStdOutLn (annotate AnnImportant $ "Traces (" <> show (length traces) <> "):")
+  forM_ traces $ \tr ->
+    renderStdOutLn (ppPrint tr)
+  renderStdOutLn (annotate AnnImportant "Result:")
+  renderStdOutLn (ppPrint (res ^. runNockmaResult))
```
janmasrovira, line 23: We should change it here as well
```suggestion
    renderStdOutLn (ppOutDefault tr)
  renderStdOutLn (annotate AnnImportant "Result:")
  renderStdOutLn (ppOutDefault (res ^. runNockmaResult))
```

app/Commands/Dev/Anoma/Prove.hs
```diff
@@ -0,0 +1,27 @@
...
+import Commands.Dev.Anoma.Base
+import Commands.Dev.Anoma.Prove.Options
+import Juvix.Compiler.Nockma.Encoding.Jam qualified as Encoding
+import Juvix.Compiler.Nockma.Pretty hiding (Path)
+
+runCommand :: forall r. (Members (Anoma ': Error SimpleError ': AppEffects) r) => ProveOptions -> Sem r ()
+runCommand opts = do
+  res <- runNock (opts ^. proveFile) (opts ^. proveArgs)
+  let traces = res ^. runNockmaTraces
+  forM_ traces (renderStdOutLn . ppPrint)
```
janmasrovira, line 15: We should be using ppOutDefault instead of ppPrint so that we see colors in the output
```suggestion
  forM_ traces (renderStdOutLn . ppOutDefault)
```

PR #3189, opened 2024-11-22
Improve generated code for derived instances

### LINKED ISSUE #3186: Improve the inlining optimization to handle derived instances well
Currently, derived instances which are recursive are inefficient, because the recursive call is done via the generic `Eq.eq` instead of calling the function being defined directly. Thus a recursive call is compiled to a case on the instance of `Eq` to fetch the function to be called. This instance should be inlined and the case eliminated. It currently isn't unless the pragma `inline: case` is present. This situation should be detected automatically.

lukaszcz:
* Closes #3186 
* Changes the generated code for derived Eq instances to define a local recursive function with a let instead of a lambda. This is necessary, because otherwise an indirection is generated on each recursive call -- instance resolution inserts the instance identifier, which then is matched on and the function for the recursive call extracted instead of being called directly. This cannot be (easily) optimized in Core because, to avoid looping in the compiler, in general one cannot do inlining of recursive functions.
* Minor improvements to the inlining optimization and the main optimization phase: do inlining before specialization and more liberal rules for heuristic case value inlining.

### DIFF
--- a/src/Juvix/Compiler/Core/Data/IdentDependencyInfo.hs
+++ b/src/Juvix/Compiler/Core/Data/IdentDependencyInfo.hs
@@ -97,11 +97,20 @@ recursiveIdentsClosure tab =
         chlds = fromJust $ HashMap.lookup sym graph
 
 -- | Complement of recursiveIdentsClosure
+nonRecursiveReachableIdents' :: InfoTable -> HashSet Symbol
+nonRecursiveReachableIdents' tab =
+  HashSet.difference
+    (HashSet.fromList (HashMap.keys (tab ^. infoIdentifiers)))
+    (recursiveIdentsClosure tab)
+
+nonRecursiveReachableIdents :: Module -> HashSet Symbol
+nonRecursiveReachableIdents = nonRecursiveReachableIdents' . computeCombinedInfoTable
+
 nonRecursiveIdents' :: InfoTable -> HashSet Symbol
 nonRecursiveIdents' tab =
   HashSet.difference
     (HashSet.fromList (HashMap.keys (tab ^. infoIdentifiers)))
-    (recursiveIdentsClosure tab)
+    (recursiveIdents' tab)
 
 nonRecursiveIdents :: Module -> HashSet Symbol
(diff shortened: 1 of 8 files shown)


janmasrovira:
I've used this code to benchmark the potential gains of replacing the lambda with a recursive function defined in a let. I've used the main branch (f916c0a95aa9996b77ebac0d264ff3b1d33970e2) to benchmark this, so the changes in the inlining phase do not affect this.
```
module main;

import Stdlib.Prelude open;

type UNat :=
  | uzero
  | usuc UNat;

instance
eq-lambda : Eq UNat :=
  mkEq
    λ{
      uzero uzero := true
      | (usuc n) (usuc m) := Eq.eq n m
      | _ _ := false
    };

eq-let : Eq UNat :=
  let
    eq : UNat -> UNat -> Bool
      | uzero uzero := true
      | (usuc n) (usuc m) := eq n m
      | _ _ := false;
  in mkEq eq;

r : Nat -> UNat
  | zero := uzero
  | (suc n) := usuc (r n);

instance
fromnat : FromNatural UNat := mkFromNatural r;

n : UNat := fromNat 10000000;

main1 : Bool := Eq.eq {{eq-lambda}} n n;

main2 : Bool := Eq.eq {{eq-let}} n n;

main : Bool := ...;
```

I've used the native backend.
```
juvix compile native main.juvix --output lambda  # with main = main1
juvix compile native main.juvix --output let     # with main = main2
```

I've run ` hyperfine --warmup 1 './let' './lambda'`:
```
Benchmark 1: ./let
  Time (mean ± σ):     256.5 ms ±  11.3 ms    [User: 46.1 ms, System: 208.0 ms]
  Range (min … max):   245.5 ms … 282.7 ms    10 runs

Benchmark 2: ./lambda
  Time (mean ± σ):     368.6 ms ±   6.1 ms    [User: 71.6 ms, System: 293.1 ms]
  Range (min … max):   360.3 ms … 378.7 ms    10 runs

Summary
  ./let ran
    1.44 ± 0.07 times faster than ./lambda
```

PR #3193, opened 2024-11-26
Add `juvix dev anoma add-transaction` command

paulcadman:


paulcadman:
This is failing because due to Anoma node compilation error, perhaps due to the use of an old version of the protobuf compiler. Node engineering is looking into this.

janmasrovira (approved):

PR #3213, opened 2024-12-02
Fix Anoma in the CI

lukaszcz:
Makes `ANOMA_VERSION` point to the latest commit in the `lukasz/juvix-integration-tracking` branch. We need to manually keep this branch up-to-date, but controlling the branch solves the problem with disappearing commits (e.g., after force-push).


paulcadman:
Thanks - so we need to merge `artem/juvix-node-integration-v0.28` into `lukasz/juvix-integration-tracking` when we want to update with the latest testnet changes.

paulcadman (approved):

PR #3215, opened 2024-12-02
Insert hash to packages without pre-release annotation

### LINKED ISSUE #2985: Package name & version are not set properly in moduleIds
In the `Compiler/Pipeline/Driver.hs` file, when processing the imports we should change the `_entryPointPackage` field of the `EntryPoint` to the package of the imported module. We don't do that, which results in wrong package information being associated with the `moduleId`. This can lead to a name clash when there are two modules with the same name in two different packages, or in two different versions of the same package used as dependencies in the project.

The PR

* #2977 

fixes this for the recursive pipelines (for HTML and Isabelle/HOL generation) by modifying `processRecursiveUpTo`. The solution is a hack, because it reads the global package from disk every time a module in this package is processed. The problem is that  for the global package, `Package` is not available in `PackageInfo`. The extra disk reads is why I didn't implement this solution in general. The `Package` should be read only once and information about the `Package`, not only `PackageLike`, should be available in `PackageInfo`.

### LINKED ISSUE #3154: Juvix Compiler Fails building it's own stdlib tests
**Describe the bug**
```bash
[6 of 6] Compiling Test
juvix: Maybe.fromJust: Nothing
  error, called at libraries/base/Data/Maybe.hs:150:21 in base:Data.Maybe
  fromJust, called at src/Juvix/Compiler/Core/Data/Module.hs:71:30 in juvix-0.6.6-63HcRSLJPON7XBeV31D9oO:Juvix.Compiler.Core.Data.Module
make[1]: *** [Makefile:6: build/Test] 錯誤 1
make[1]: 離開目錄「/home/taichi/Documents/Work/Repo/anoma-repos/juvix-stdlib/test」
make: *** [Makefile:21: test] 錯誤 2
```

**Desktop (please complete the following information):**

- OS: [e.g. iOS] Arch Linux & ubunut
- Juvix Version [e.g. 22] Juvix version 0.6.6-0961d87

**To Reproduce**

1. Go to: juvix-stdlib [0080b1183ab55e5180e69bfc3987e4cd6edbc230]
2. run: make test

Also see https://github.com/anoma/juvix-stdlib/actions/runs/11717091275/job/32636224379

for CI failure

**Expected behavior**
Haskell not to freakout

**Screenshots**

**Additional context**

janmasrovira:
- Fixes #2985 
- Fixes #3154


src/Juvix/Data/SHA256.hs
```diff
@@ -4,13 +4,52 @@ import Crypto.Hash.SHA256 qualified as SHA256
...
+runSHA256Builder m = fmap
+  ( first
+      ( hashToText
+          . SHA256.finalize
+      )
+  )
+  $ reinterpret' m (runState SHA256.init)
+  $ \case
+    BuilderDigestFiles f -> do
+      fs <- mapM readFileBS' (toList f)
```
paulcadman, line 43: There may be an issue with file system path ordering, but some of my assumptions might be faulty:

(I'm assuming that the digest will change if the order of the input files change.)

For the packageId we require that the digest of the package files doesn't change between runs of juvix etc. However we use the file system to return the list of files and calls like this are not guaranteed to be deterministic. (The returned order of paths may change if the metadata of the files change for example).

So maybe we should add an API `digestSortedFiles` which sorts the input first - or just modify `digestFiles` to do the sorting. 

Apologies if you're sorting the input files somewhere, I couldn't find it.

janmasrovira: that's a very good point. I'll fix  that

janmasrovira: fixed in 37f30042532561b29082f60065932cb393172eec

paulcadman (approved):

PR #3224, opened 2024-12-04
Global option `--no-check`

lukaszcz:
The `--no-check` option disables the backend-specific sanity check in Core (the transformations in `Check/*`, which check the type of main, use of builtins, etc). This is useful for debugging.


janmasrovira:
I think it'd be better to prefix these kind of flags with `dev`, e.g. `dev-no-check` or `--dev-no-sanity-check`

janmasrovira (approved):

PR #3264, opened 2024-12-27
Fix flatparse location

janmasrovira:
- :warning: Depends on #3263 

The flatparse line and column numbers are 0-based, whereas we use 1-based numbers. This pr fixes that.


janmasrovira:
subsumed by #3265

PR #3270, opened 2025-01-07
Optimize single-argument unknown function invocations with known non-function target types

### LINKED ISSUE #2935: Optimize single-argument unknown function invocations with known non-function target types
* This is a special case of the arity-specializing optimization for calls to unknown functions (see #1545), which is easier to implement and doesn't need arity analysis with a data-flow algorithm.
* When an unknown function `f` is called and it has type `T1 -> T2` where `T2` is a concrete non-function type (i.e. not function and not dynamic `*`), then we know that `f` must be a closure expecting exactly one argument (because zero-argument closures are not allowed, and it cannot expect more because of the type). Then we can convert the general call with `ccall` or `apply_1` into a direct closure call with one argument. 
* This optimization should be performed as a transformation on JuvixTree, replacing `ccall(f, a)` with `call(f, a)` when the result of `f` is known not to be a function.

lukaszcz:
* Closes #2935 
* Implements the `ConvertUnaryCalls` optimization which replaces generic calls to unknown unary functions with known non-function target types by direct closure calls
* Adds an optimization phase in JuvixTree
* Adds options to the JuvixTree pipeline
* Fixes a bug in Rust code generation for direct closure calls

### DIFF
--- a/src/Juvix/Compiler/Tree/Transformation/Optimize/ConvertUnaryCalls.hs
+++ b/src/Juvix/Compiler/Tree/Transformation/Optimize/ConvertUnaryCalls.hs
@@ -6,7 +6,7 @@ import Juvix.Compiler.Tree.Extra.Recursors
 import Juvix.Compiler.Tree.Extra.Type
 import Juvix.Compiler.Tree.Transformation.Base
 
--- Replaces generic calls (with CallClosures) to unknown unary functions with
+-- | Replaces generic calls (with CallClosures) to unknown unary functions with
 -- known non-function target types by direct closure calls (with Call)
 convertUnaryCalls :: InfoTable -> InfoTable
 convertUnaryCalls tab = mapT convert tab


src/Juvix/Compiler/Tree/Transformation/Optimize/ConvertUnaryCalls.hs
```diff
@@ -0,0 +1,51 @@
+module Juvix.Compiler.Tree.Transformation.Optimize.ConvertUnaryCalls where
```
janmasrovira, line 1: add a brief description of the optimization

test/Main.hs
```diff
@@ -28,9 +28,9 @@ import Typecheck qualified
 
 slowTests :: IO TestTree
 slowTests =
-  sequentialTestGroup
+  testGroup
     "Juvix slow tests"
-    AllFinish
+    -- AllFinish
```
janmasrovira, line 33: probably remove this line

janmasrovira (approved):

PR #3313, opened 2025-02-03
Support ByteArray and List arguments to anoma prove CLI

paulcadman:
This PR adds support for two additional argument types fro the prove CLI:

```
  --arg ARG_TYPE:ARG       An argument to the program:
                           • bytearray:FILE where FILE is a file containing bytes that represent a ByteArray
                           • list:FILE where FILE is a file containing a newline delimited list of base64 encoded jammed nouns that represent a Nock list of jammed nouns
```

I tested these arguments with the following programs:

First a program that accepts an ed25519 signing key, this returns `64`.

```
module Kudos.ReadSigningKey;

import Stdlib.Prelude open;
import Applib open;

main (signingKey : ByteArray) : Nat := size signingKey;
```

Secondly a program that accepts a list of encoded atoms, decodes them and takes the sum.

```
module Kudos.TestList;

import Stdlib.Prelude open;
import Applib open;

main (xsEncoded : List (Encoded Nat)) : Nat :=
  let
    xs := map Encode.decode xsEncoded;
  in for (acc := 0) (x in xs) {
    acc + x
  };
```

I tested this with the input containing the jammed atoms representing 1, 1 and 2:

```
DA==
DA==
SA==
```

### DIFF
--- a/src/Anoma/Effect/RunNockma.hs
+++ b/src/Anoma/Effect/RunNockma.hs
@@ -6,6 +6,7 @@ where
 
 import Anoma.Effect.Base
 import Anoma.Rpc.RunNock
+import Data.ByteString.Base64 qualified as Base64
 import Juvix.Compiler.Nockma.Encoding
 import Juvix.Compiler.Nockma.Language qualified as Nockma
 import Juvix.Data.CodeAnn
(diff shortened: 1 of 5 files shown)


src/Anoma/Effect/RunNockma.hs
```diff
@@ -18,7 +19,7 @@ data RunNockmaArg
   = -- | An argument that must be jammed before it is sent
     RunNockmaArgTerm (Nockma.Term Natural)
   | -- | An argument that is already jammed and must not be jammed again before it is sent
-    RunNockmaArgJammed (Nockma.Atom Natural)
+    RunNockmaArgJammed ByteString
```
janmasrovira, line 22: make much more sense now

janmasrovira (approved):

PR #3324, opened 2025-02-10
Modular compilation in JuvixTree

lukaszcz:
* Adds the `Pipeline.Modular` module which implements modularization in the backend pipeline (currently only Core.Stripped and Tree). The backend modularization framework is separate from the existing modularization framework up to Core. The intermediate representation of each module is saved on disk, but currently the process is not parallelized (it's probably not worth it -- the processing after Core is relatively cheap).
* The function `runIOEitherModular` in `Pipeline.Modular.Run` can be used to run the pipeline from source using both modularization frameworks (up to Core and backend).
* Currently, the backend modularization framework is used only for `dev compile tree` and the compilation pipeline tests (source to native/WASM).


janmasrovira (commented):
review in progress

janmasrovira (approved):

PR #3335, opened 2025-02-14
Fix `main` detection

lukaszcz:
The `main` function is now detected by name only in the main compiled file. Any `main` functions in imported modules are not registered as `main` in Core (this caused compiler crashes because unreachable definitions in modules with `main` were filtered out).


janmasrovira (approved):

janmasrovira:
I've checked locally that this fixes https://github.com/anoma/juvix/pull/3335

PR #3346, opened 2025-02-28
Expose Cairo range check builtin

### LINKED ISSUE #3333: Add range check builtin in Cairo
Needed by https://github.com/anoma/anoma/issues/1674

lukaszcz:
* Closes #3333 
* Adds the `rangeCheck` builtin: `rangeCheck x bound` succeeds and returns true if `0 <= x <= bound`, assuming `0 <= bound < 2^128`.
* `rangeCheck x bound` is an assertion in Cairo: the program fails if `x > bound`, so `rangeCheck` cannot be used to branch on whether `0 <= x <= bound`.

### DIFF
--- a/test/Base.hs
+++ b/test/Base.hs
@@ -230,6 +230,20 @@ readProcessCwd' menv mcwd cmd args stdinText =
         return r
     )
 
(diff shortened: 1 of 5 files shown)


janmasrovira (commented):
would it be possible to produce errors through Sem and add a negative test?

janmasrovira (approved):

PR #3379, opened 2025-03-31
Noun-like type representation in Nockma

janmasrovira:
Adds a new type representation for types that can be directly encoded with either an atom or a cell without a tag.
E.g.
Consider this inductive type:
```
type Term :=
  | atom Nat
  | cell@{
      left : Term;
      right : Term;
    };
```
And terms:
```
x : Term := Term.cell (Term.atom 30) (Term.cell (Term.atom 40) (Term.atom 80));
y : Term := Term.atom 20
```
Then x and y will be represented unambiguously as
```
x ~ [30 40 80]
y ~ 20
```

### DIFF
--- a/src/Juvix/Compiler/Nockma/Translation/FromTree.hs
+++ b/src/Juvix/Compiler/Nockma/Translation/FromTree.hs
@@ -374,25 +374,25 @@ supportsNounNockmaRep tab ci = fmap NockmaMemRepNoun . run . runFail $ do
       return NockmaMemRepAtom
     2 -> return NockmaMemRepCell
     _ -> impossible
-    where
-      -- Returns True if all elements of some type are representable with an
-      -- Atom. There may be false negatives. In that case, a less optimal
-      -- representation might be chosen, but it shouldn't effect correctness.
-      typeRepresentedAsAtom :: Tree.Type -> Bool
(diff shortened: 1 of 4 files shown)


src/Juvix/Compiler/Nockma/Translation/FromTree.hs
```diff
@@ -469,38 +517,44 @@ compile = \case
...
+                    NockmaMemRepList constr -> case constr of
+                      NockmaMemRepListConstrNil -> impossible
+                      NockmaMemRepListConstrCons ->
+                        indexTuple
+                          IndexTupleArgs
+                            { _indexTupleArgsLength = 2,
+                              _indexTupleArgsIndex = argIx
+                            }
+                    NockmaMemRepNoun n -> case n of
+                      NockmaMemRepAtom -> []
```
paulcadman, line 541: Perhaps use `emptyPath` from `Juvix.Compiler.Nockma.Language.Path` instead of `[]`?

paulcadman (dismissed):

src/Juvix/Compiler/Nockma/Translation/FromTree.hs
janmasrovira: done

paulcadman (approved):

PR #3391, opened 2025-04-11
Anoma HTTP JSON client communication

lukaszcz:
* Adapts our Haskell code for communication with Anoma to the HTTP JSON client.
* The Anoma node tests can no longer be run in parallel, because the client port is fixed. If we don't run them sequentially, starting some clients in the tests will fail because another client is already running in parallel with the same port.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ env:
   RISC0_VM_VERSION: v1.0.1
   # This is the top commit hash in the branch lukasz/juvix-ci-stable
   # of the anoma repository.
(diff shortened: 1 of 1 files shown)


janmasrovira:
> The Anoma node tests can no longer be run in parallel, because the client port is fixed. If we don't run them sequentially, starting some clients in the tests will fail because another client is already running in parallel with the same port.

I think this should be addressed somehow

janmasrovira:
the test suite in anoma-apps will need to be fixed. We run the tests in parallel there
https://github.com/anoma/anoma-apps/blob/f9a013bfd06e3ccd038a008cc396b2d165481247/run-tests.sh

janmasrovira (approved):

PR #3402, opened 2025-04-25
Update anoma

janmasrovira:


janmasrovira:
reminder that we will need to update the applib accordingly

PR #3403, opened 2025-04-25
Additional Anoma compilation CLI options

lukaszcz:
* Adds the `--text` option which enables generation of an unannotated textual representation of Nock code
* Adds the `--hoon` option which enables generation of a hoon file that can be read by `urbit eval`
* Adds the `--no-anomalib` option which disables bundling the Anoma standard library


src/Juvix/Prelude/Base.hs
```diff
@@ -47,3 +47,16 @@ findRepeated = map (head . snd . fst) . findRepeatedOn id
 
 allDifferent :: forall a. (Ord a) => [a] -> Bool
 allDifferent = null . findRepeated
+
+-- Function to group elements in fixed-size chunks
+groupEvery :: Int -> [a] -> [[a]]
+groupEvery _ [] = []
+groupEvery n xs = take n xs : groupEvery n (drop n xs)
```
janmasrovira, line 54: use [splitAt](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:splitAt)

janmasrovira: did you push the changes?

janmasrovira (approved):

